PR #301, opened 2026-03-28
feat: Handoffs

amol-anand:
A new SLICC handoff flow so local coding agents like Codex/Claude/Cursor can hand work off directly into the SLICC browser agent instead of stopping at “please test this in the browser.” The problem I'm trying to solve is that last-mile validation often needs a real logged-in browser session, open tabs, extension context, and human-confirmed continuation, which local agents don’t always have.

At a high level, I’ve built:
- a new handoff relay flow in SLICC where a handoff can be posted to the worker and opened as a relay page
- extension support to receive that handoff, show an Accept/Dismiss prompt, and inject it into the current cone on approval
- pending handoff handling in the UI, including extension badge support and a Chat tab count indicator
- a reusable slicc-handoff skill package + helper script so agents can understand phrases like “handoff to slicc” and create the right payload automatically

Next step on my side is end-to-end worker testing, and for that I need access to the Cloudflare account so I can add the R2 bucket bindings needed for the handoff storage flow.

### DIFF
--- a/.agents/skills/slicc-handoff/SKILL.md
+++ b/.agents/skills/slicc-handoff/SKILL.md
@@ -1,0 +1,51 @@
+---
+name: slicc-handoff
+description: Create a generic SLICC handoff when the user wants work moved to the SLICC agent, continued in SLICC, or tested in SLICC. Prefer the repo wrapper command and keep the payload concise, browser-aware, and self-contained.
+---
+
+# SLICC Handoff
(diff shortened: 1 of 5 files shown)


amol-anand:
Closing this one.. I have another one which avoids the need for R2 but still allows for seamless handoffs to the browser agent from any agent trigger surface

PR #316, opened 2026-03-31
feat: add GitHub OAuth provider

catalan-adobe:
## Summary

- Add GitHub OAuth provider (`packages/webapp/providers/github.ts`) with authorization code grant flow, git token bridging, and user profile display
- Add `POST /github/token` and `POST /github/revoke` routes to the tray hub Cloudflare Worker for server-side code exchange and token revocation
- Add `github-token-changed` event listener in `GitCommands` to invalidate cached git tokens when the OAuth provider writes a new token

## How it works

GitHub OAuth App (authorization code grant) with server-side code exchange via the tray hub worker. The provider follows the same `ProviderConfig` pattern as the Adobe provider but is simpler — no LLM streaming, no model discovery, no token renewal.

On login, the OAuth token is written to `/workspace/.git/github-token` in the global VFS so isomorphic-git uses it for push/clone/fetch. A `github-token-changed` event invalidates the `GitCommands` cache so the new token is picked up immediately.

## Manual setup required

1. Create a GitHub OAuth App at `github.com/settings/developers`
2. Copy client ID into `packages/webapp/providers/github-config.json`
3. Set worker secret: `npx wrangler secret put GITHUB_CLIENT_SECRET`

## Test plan

- [ ] Worker routes: 7 new tests (token exchange, error passthrough, CORS, 405, revoke, revoke CORS, revoke 405)
- [ ] Provider: 6 new tests for `extractCodeFromUrl` (query params, single param, missing, empty, extension URL)
- [ ] Git: `resetTokenCache` static method callable without error
- [ ] Manual: full OAuth flow in CLI mode after OAuth App is created
- [ ] Manual: full OAuth flow in extension mode

### DIFF
--- a/docs/superpowers/plans/2026-03-30-github-oauth-provider.md
+++ b/docs/superpowers/plans/2026-03-30-github-oauth-provider.md
@@ -1,0 +1,900 @@
+# GitHub OAuth Provider Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a GitHub OAuth provider to Slicc that enables OAuth-based login, automatic git token bridging, and GitHub API access.
+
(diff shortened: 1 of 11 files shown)


trieloff:
Hey @catalan-adobe — heads up that #491 takes a different approach to the same problem. Instead of GitHub-specific routes (`POST /github/token`, `POST /github/revoke`), it introduces a generic OAuth token broker with two routes (`POST /oauth/token`, `POST /oauth/revoke`) driven by a provider registry. Adding a new provider is config-only — no new routes or handler functions.

Your PR's GitHub provider work (git token bridge, user profile fetch, `github-token-changed` event) was a great reference — #491 reuses those patterns but wires them through the generic broker and adds GitHub Models LLM streaming.

If #491 merges first, #316 would be superseded. Let me know if you have any concerns or want to discuss the approach.

PR #378, opened 2026-04-09
feat: add X-Session-Id header to all LLM API requests

catalan-adobe:
## Summary

- Wraps pi-ai `streamSimple` to inject `X-Session-Id` HTTP header on every LLM request, regardless of provider (Anthropic, OpenAI, Google, Mistral, Bedrock, etc.)
- Session ID encodes the cone→scoop relationship: cone uses its JID directly, scoops use `coneJid/scoopFolder`
- Also sets `sessionId` on the pi-agent-core Agent for providers that use it for caching (e.g., OpenAI's `prompt_cache_key`)

## Test plan

- [x] Typecheck passes (`npx tsc --noEmit`)
- [x] All 52 scoop-context tests pass
- [ ] Verify `X-Session-Id` header appears in outgoing LLM requests via browser DevTools Network tab
- [ ] Verify scoops show `coneJid/folder` format in the header


catalan-adobe:
@trieloff, any concern adding this header to every LLM provider calls?
First idea is to help us monitor usage of the Adobe LLM provider.

trieloff:
Let me have a quick check if this can be used to de-anonymize users, which we should prevent.

packages/webapp/src/scoops/scoop-context.ts
```diff
@@ -110,14 +110,18 @@ export class ScoopContext {
...
     this.scoop = scoop;
     this.callbacks = callbacks;
     this.fs = fs;
     this.sessionStore = sessionStore ?? null;
     this.skillsFs = skillsFs ?? null;
-    this.sessionId = scoop.jid;
+    // Session ID encodes the cone→scoop relationship:
+    //   cone:  "cone_17126…"
+    //   scoop: "cone_17126…/my-task-scoop"
+    this.sessionId = scoop.isCone ? scoop.jid : `${coneJid ?? 'unknown'}/${scoop.folder}`;
```
trieloff: This looks like it's able to leak info. Better use a hash of the scoop name (hashed with jid, so it's auto-salted)

trieloff (changes requested):
- use crypto.randomUUID, as the millisecond timestamp, together with the provider's own logs would aid deannonymization
- better scope this only to the adobe provider
- don't leak scoop names
- reset the session id at once a day

packages/webapp/src/scoops/scoop-context.ts
catalan-adobe: Done in 929c59a9. Applied the hashed-folder approach — scoop IDs are now `<uuid>/<hash(folder, uuid)>` where the daily-rotating UUID serves as the salt, so the same folder name under two cones produces two different hashes. Helper lives in `packages/webapp/src/scoops/llm-session-id.ts`.

catalan-adobe:
@trieloff thanks for the privacy review — all four items addressed in 929c59a9.

**1. `crypto.randomUUID` instead of millisecond timestamp**
Each cone gets a v4 UUID from `crypto.randomUUID()`, cached per-cone in `localStorage` with a UTC date stamp. The `Date.now()`-derived JID no longer leaves the browser.

**2. Scoped to the Adobe provider only**
The `streamSimple` wrapper checks `model.provider === 'adobe'` at call time. Anthropic, OpenAI, Google, Mistral, Bedrock, etc. now receive no `X-Session-Id` header at all. Checking per-call (not per-init) means a mid-session model switch behaves correctly.

**3. No scoop folder leakage**
Scoop IDs are `<dailyUuid>/<hash(folder, dailyUuid)>`. SHA-256 truncated to 16 hex chars; the UUID doubles as the salt so the same folder under two cones produces two different hashes. The folder string never reaches Adobe.

**4. Daily rotation**
The stored entry has the shape `{uuid, date: "YYYY-MM-DD"}` (UTC). Any `ScoopContext.init()` on a new day mints a fresh UUID; scoop suffixes rotate with it because the hash salt changes. Within a session, the ID is stable (captured in the streamFn closure).

Also reverted `this.sessionId` to `scoop.jid` so `SessionStore` (IndexedDB) keeps restoring saved conversations across day boundaries. The persistence key and the outgoing Adobe ID are now cleanly separated concerns.

**Verification**
- New helper: `packages/webapp/src/scoops/llm-session-id.ts`
- New tests: `packages/webapp/tests/scoops/llm-session-id.test.ts` (7 tests covering UUID shape, same-day stability, cross-day rotation, folder-name opacity, salt effectiveness)
- Full suite: 2410 passed, typecheck and both webapp + chrome-extension builds clean.

trieloff (approved):

PR #386, opened 2026-04-10
fix: abort in-flight scoop work on drop_scoop

catalan-adobe:
## Summary

- `drop_scoop` now aborts the in-flight LLM agent, clears queued follow-up prompts, and suppresses all post-disposal callbacks so dropped scoops stop consuming tokens, executing tools, and writing files
- Two-layer defense: Layer 1 (`ScoopContext.disposed` flag) aborts at the source; Layer 2 (`scoops.has(jid)` guard in orchestrator) catches any callback that leaks through
- Cleans up the `scoopResponseBuffer` during scoop unregistration to prevent stale text accumulation

## Problem

Calling `drop_scoop` removed a scoop from the registry but did not terminate the underlying LLM API call. The scoop continued streaming, executing shell commands, writing files, and opening browser tabs — all invisible since `list_scoops` no longer showed it. Completion messages from dropped scoops arrived unexpectedly at the cone, and full token cost was incurred with no way to stop it.

Root cause: `ScoopContext.dispose()` nulled the agent reference without calling `agent.abort()`. The existing `stop()` method already had the correct abort logic, but `dispose()` (called by the drop flow) bypassed it entirely.

### `packages/webapp/src/scoops/scoop-context.ts`
- Add `disposed` flag field, set as first operation in `dispose()`
- Call `agent.clearAllQueues()` and `agent.abort()` before nulling references in `dispose()`
- Guard `setStatus()` — early return when disposed, preventing the `prompt()` finally block from routing stale completion messages
- Guard `prompt()` catch block — suppress error callbacks from aborted agents
- Guard `handleAgentEvent()` — suppress all agent event callbacks (streaming deltas, tool execution, recovery) after disposal

### `packages/webapp/src/scoops/orchestrator.ts`
- Add `scoops.has(jid)` guard to all four `contextCallbacks` closures (`onResponse`, `onResponseDone`, `onError`, `onStatusChange`) as defense-in-depth
- Add `scoopResponseBuffer.delete(jid)` to `unregisterScoop()` cleanup

### `packages/webapp/tests/scoops/scoop-context.test.ts`
- Test: `dispose()` calls `abort()` and `clearAllQueues()` before nulling agent
- Test: status callbacks suppressed after disposal (in-flight prompt finally block)
- Test: error callbacks suppressed from aborted prompt
- Test: all agent event callbacks suppressed after disposal (streaming, tool start/end, turn end, agent end)

## Test plan

- [x] All 56 scoop-context tests pass (4 new)
- [x] All 394 scoops tests pass
- [x] Full test suite: 2410 passed, 0 failures
- [x] Typecheck clean (browser + CLI + worker)
- [x] Prettier formatted
- [x] Production build succeeds
- [x] Chrome extension build succeeds
- [x] Manual verification: extension loaded, scoops drop cleanly


trieloff:
Good find.

PR #394, opened 2026-04-13
fix(webapp): harden jsh and bsh script discovery

trieloff:
## Summary
- add a shared `ScriptCatalog` for `.jsh` and `.bsh` discovery, invalidated by `FsWatcher`
- route `WasmShell`, `which`, standalone/offscreen startup, and `BshWatchdog` through the shared catalog instead of duplicate registries
- harden mount + watcher behavior so mounted and nested script trees stay discoverable and cache invalidation covers mount-backed mutations
- add regression coverage for mount shadowing, nested mounts, watcher invalidation, shell-owned catalogs, and related package test stability

## Verification
- `npm run test -w @slicc/webapp`
- `npm run build -w @slicc/webapp`
- `npm run typecheck -w @slicc/webapp` *(still fails on pre-existing unrelated errors in `packages/webapp/src/providers/built-in/bedrock-camp.ts:455` and `packages/webapp/src/shell/wasm-shell.ts:497`)*


packages/vfs-root/workspace/skills/skill-authoring/SKILL.md
```diff
@@ -55,9 +55,11 @@ Type `commands` in the terminal for the full list. Key commands:
 Virtual filesystem stored in IndexedDB, survives tab closes and refreshes. Mount local directories:
 
 ```
-mount /workspace/myproject
+mount /mnt/myproject
 ```
 
+Use an empty mount point. Mounting over existing files is blocked so built-in skills and scripts stay discoverable.
```
trieloff:
```suggestion
Use an empty mount point. Mounting over existing files is blocked so built-in skills and scripts stay discoverable. `ln -s` the mounted files into the place where you need them. 
```

PR #412, opened 2026-04-15
fix: disable automated extension publish to cws

catalan-adobe:
@trieloff, @karlpauls, proposing to disable extension automated publish (at least for the time of Summit) to ensure we have a working version for our labs.


trieloff:
ok

PR #426, opened 2026-04-17
fix(git): route isomorphic-git through VirtualFS so mounts work

### LINKED ISSUE #421: bug: git commands fail on mounted directories (HEAD not found)
## Summary

`git` commands fail when run inside a mounted directory (mounted via the `mount` command / File System Access API). The error manifests as `HEAD not found` even though `.git/HEAD` is visibly present in the OS folder.

## Root Cause

`GitCommands` passes raw LightningFS directly to isomorphic-git:

\`\`\`ts
// packages/webapp/src/git/git-commands.ts
this.lfs = options.fs.getLightningFS();
await git.clone({ fs: this.lfs, dir: cwd, ... });
\`\`\`

`VirtualFS.readFile()` is mount-aware — it calls `findMount()` and routes through the File System Access API handle for mounted paths. But `getLightningFS()` returns raw LightningFS which only sees IndexedDB and is completely unaware of mount points. So isomorphic-git reads `.git/HEAD` through LightningFS, finds nothing, and throws.

## Affected Scenario

| Scenario | Works? |
|---|---|
| Repo cloned via `git clone` into `/workspace` (stored in LightningFS/IndexedDB) | ✅ |
| Repo inside a **mounted** directory (File System Access API via `mount` command) | ❌ |

## Fix

Pass isomorphic-git a custom `fs` adapter backed by `VirtualFS` (which is mount-aware) instead of raw LightningFS. The adapter needs to implement the [isomorphic-git fs interface](https://isomorphic-git.org/docs/en/plugin_fs) using `VirtualFS` methods (`readFile`, `writeFile`, `readdir`, `mkdir`, `stat`, `lstat`, `rm`, etc.).

## Relevant Files

- `packages/webapp/src/git/git-commands.ts` — uses `getLightningFS()`, needs a VirtualFS-backed adapter
- `packages/webapp/src/fs/virtual-fs.ts` — `getLightningFS()` (line 92), `findMount()` (line 254), `readFile()` (line 321)

trieloff:
## Summary

Fixes #421 — `git` commands fail on mounted directories with "HEAD not found", because `GitCommands` passed raw LightningFS to isomorphic-git instead of a mount-aware FS.

- **New `vfs-fs-adapter.ts`** — builds a `PromiseFsClient` over `VirtualFS`. Mounted paths route through `VirtualFS` (File System Access API); LFS paths stay on the raw LightningFS fast path so `statusMatrix` filemode bytes remain byte-identical to the pre-adapter behavior.
- **`git-commands.ts`** — one-line constructor change (`this.lfs = createIsomorphicGitFs(options.fs).promises`). All 58 `git.*` call sites untouched.
- **`virtual-fs.ts`** — fixes a latent mount-write bug surfaced while testing: `new Uint8Array(content.buffer)` dropped `byteOffset`/`byteLength`, so pooled `Buffer` views (e.g. from iso-git's `Buffer.concat` when writing `.git/index`) wrote the whole backing pool. Now preserves the view bounds.
- **Tests** — new `git-commands-mount.test.ts` does an end-to-end `init → add → commit → log` on a mounted repo, plus a "status doesn't say 'HEAD not found'" guard and an LFS-only regression check. `fsa-test-helpers.ts` gained `createWritable()` and `removeEntry()` on the mock handles so full round-trip tests are possible.

## Test plan

- [x] `npx vitest run -w packages/webapp` — 2274/2274 pass
- [x] `npx prettier --check` on all touched files — clean
- [ ] Manual: mount a real folder via the `mount` command, run `git status` / `git log` inside it

### DIFF
--- a/packages/webapp/src/fs/virtual-fs.ts
+++ b/packages/webapp/src/fs/virtual-fs.ts
@@ -532,14 +532,15 @@ export class VirtualFS {
       try {
         const fh = await VirtualFS.fsaGetFile(mount.handle, mount.relParts, true);
         const writable = await fh.createWritable();
+        // Preserve byteOffset/byteLength: pooled Buffer instances share a
+        // backing ArrayBuffer with other allocations, so `content.buffer`
+        // alone would write the whole pool.
         const data =
           typeof content === 'string'
             ? new TextEncoder().encode(content)
-            : new Uint8Array(
-                content instanceof Uint8Array
-                  ? (content.buffer as ArrayBuffer)
-                  : (content as ArrayBuffer)
-              );
+            : content instanceof Uint8Array
+              ? new Uint8Array(content.buffer, content.byteOffset, content.byteLength)
+              : new Uint8Array(content as ArrayBuffer);
         await writable.write(data as unknown as FileSystemWriteChunkType);
         await writable.close();
       } catch (err) {
--- a/packages/webapp/src/fs/virtual-fs.ts
+++ b/packages/webapp/src/fs/virtual-fs.ts
@@ -408,6 +408,18 @@ export class VirtualFS {
   }
 
   /**
+   * Check whether an absolute path is under any active mount point.
+   * Non-allocating (iterates the mount map directly) — safe to call on a hot
+   * path such as the per-operation check in the isomorphic-git fs adapter.
+   */
+  isPathUnderMount(path: string): boolean {
+    for (const mountPath of this.mountPoints.keys()) {
+      if (path === mountPath || path.startsWith(mountPath + '/')) return true;
+    }
+    return false;
+  }
+
+  /**
    * Find the mount point that owns `path`.
    * Returns the handle and the path segments relative to the mount root,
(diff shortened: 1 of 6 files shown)


trieloff:
Dropping a note from the cone side: I'm sliccy, running inside SLICC, and I can confirm this patch is already paying off. The user mounted /Users/trieloff/Developer/slicc at /mnt/slicc via the File System Access API, and from that mount I was able to run git log, git branch -a, and git log claude-1 successfully — the exact operations that used to fail with "HEAD not found".

For reference, the very first thing I tried was git log --oneline from inside /mnt/slicc/.yolo/claude-1/. That one still fails, because the .git file there is a worktree pointer to a host path (/Users/trieloff/Developer/slicc/.git/worktrees/claude-1) that isn't inside any mount — so git can't resolve the gitdir. Stepping up one level to the real .git directory worked perfectly. Not a bug in this PR, but worth noting that worktree checkouts need the parent repo (or the worktree's gitdir) mounted too.

🍦 Self-licking confirmed.

PR #430, opened 2026-04-17
feat(shell): add `agent` supplemental command with AgentBridge + bash allowlist

trieloff:
## Summary

Introduces the `agent` shell command and the infrastructure to support spawning Pi-based sub-agents from inside scoops, alongside a set of hardening fixes across the scoop/restricted-FS/agent-bridge stack.

Major additions:
- **`agent` supplemental command + AgentBridge**: `feat(shell): add \`agent\` supplemental command scaffolding`, `feat(scoops): implement AgentBridge for agent shell command`, `feat(scoops): publish AgentBridge on globalThis.__slicc_agent in CLI + extension bootstrap`, plus multiple bridge fixes (registry visibility, parent modelId inheritance, ephemeral-scoop registration, cwd validation, side-panel proxy publication, mid-stream delta persistence, chronologically-last assistant output).
- **Bash allowlist rewritten on just-bash AST**: `refactor(bash-allowlist): rewrite on just-bash AST`, plus follow-ups closing subshell-in-double-quotes, bare-`&`/newline, arithmetic-expansion, and nested `ParameterExpansion`/`BraceExpansion` bypasses. Vendored patched `just-bash` exposing its AST parser to the browser; added a tsc-backed type surface.
- **RestrictedFS hardening**: closed sync and async lstat ancestor-symlink escapes (VAL-FS-019 parity), implemented the `VfsAdapter` sync+realpath contract.
- **Orchestrator fix**: remove scoop from registry before firing terminal `onStatusChange` callback.
- **Factory/skills docs**: new `upstream-integration-worker` skill and numerous scrutiny/user-testing round notes in `.factory/`.

53 commits, 443 files, +27k/-768.

## Test plan

- [ ] `npx prettier --check .`
- [ ] `npm run typecheck`
- [ ] `npm run test`
- [ ] `npm run build`
- [ ] `npm run build -w @slicc/chrome-extension`
- [ ] Manual: spawn a scoop via `agent` from the CLI float; confirm it shows in `list_scoops` and the UI
- [ ] Manual: same flow in the Chrome extension float (side-panel proxy path)
- [ ] Manual: verify bash-allowlist rejects the previously-bypassing forms (subshell-in-dq, bare `&`/newline, arithmetic braced expansion)

## Maintenance

The local source-of-truth for this PR is the `droid-2` branch. `origin/agent-cli` is kept as a cleaned mirror via a detached worktree at `../slicc-agent-cli`. Stripped from the mirror to keep the diff reviewable:

[...]

PR #433, opened 2026-04-18
feat(shell): per-scoop command allow-list via just-bash commands filter

trieloff:
## Summary

Adds an optional `allowedCommands` to `WasmShellOptions` (and surfaces it on `ScoopConfig`) so each scoop can restrict which shell commands are executable. Leans on just-bash's native `commands:` filter — no AST walker, no vendored types.

- **Default (omitted / `['*']`):** every built-in, custom, and `.jsh` command is available. Existing behavior unchanged.
- **Restricted:** only listed names are registered on the underlying `Bash` instance. Anything else fails with exit code 127 (`command not found`) at dispatch — including inside pipelines, subshells, and command substitution, because just-bash resolves every simple command through the same registry.
- Applies uniformly to:
  - just-bash built-ins (`getCommandNames()` + `getNetworkCommandNames()`)
  - slicc custom commands (`git`, `mount`, skills, supplemental)
  - `.jsh` auto-discovered scripts

4 files, +176/-8:

- `packages/webapp/src/shell/wasm-shell.ts` — the filter plus `isCommandAllowed()` gate on jsh registration/fallback
- `packages/webapp/src/scoops/types.ts` — new `ScoopConfig.allowedCommands`
- `packages/webapp/src/scoops/scoop-context.ts` — thread the config field through when constructing the scoop's `WasmShell`
- `packages/webapp/tests/shell/wasm-shell.test.ts` — 9 new tests covering default / wildcard / empty / pipeline / substitution / custom / `.jsh` / `/usr/bin`

Motivation: the existing allow-list infrastructure on `droid-2`/PR #430 parses every bash invocation into an AST and walks every node looking for disallowed substitutions (~600 LOC + vendored just-bash internal type surface). That work is redundant with just-bash's `BashOptions.commands` primitive, which filters the command registry directly. Disallowed names in `$(...)`, `<(...)`, subshells, eval, etc. all resolve through the same registry and therefore fail the same way.

## Test plan

- [x] `npm run test` — 2624 pass / 2 skipped (including 9 new `WasmShell command allow-list` cases)
- [x] `npx prettier --check` on all changed files
- [x] `npm run typecheck` — no new errors (pre-existing unrelated `lucide` type resolution error on `main` untouched)
- [ ] `npm run build`

### DIFF
--- a/packages/webapp/src/scoops/scoop-context.ts
+++ b/packages/webapp/src/scoops/scoop-context.ts
@@ -161,6 +161,7 @@ export class ScoopContext {
         env: Object.keys(secretEnv).length > 0 ? secretEnv : undefined,
         browserAPI: browser,
         jshDiscoveryFs: this.skillsFs ? effectiveSkillsFs : undefined,
+        allowedCommands: this.scoop.config?.allowedCommands,
       });
       log.info('WasmShell initialized', { folder: this.scoop.folder });
       const skills = await loadSkills(effectiveSkillsFs, this.skillsDir);
--- a/packages/webapp/src/scoops/types.ts
+++ b/packages/webapp/src/scoops/types.ts
@@ -77,6 +77,13 @@ export interface ScoopConfig {
    * they always use an unrestricted filesystem.
    */
   writablePaths?: readonly string[];
+  /**
+   * Shell command allow-list. When omitted (or when it contains `'*'`), every
+   * built-in, custom, and `.jsh` command is available — the default. Otherwise
+   * only commands whose names appear in the list can execute inside this
+   * scoop's shell, including through pipelines and substitution.
+   */
+  allowedCommands?: readonly string[];
 }
 
 /** Message from any channel */
--- a/packages/webapp/src/shell/wasm-shell.ts
+++ b/packages/webapp/src/shell/wasm-shell.ts
@@ -404,6 +404,31 @@ export class WasmShell {
       customCommands,
     });
 
+    // Network-command post-registration cleanup (Codex P1 on #433).
+    //
+    // just-bash's `BashOptions.commands` filter controls only the non-network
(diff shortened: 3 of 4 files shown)


trieloff:
Pushed [0337402](https://github.com/ai-ecoverse/slicc/commit/0337402) (on top of a rebase onto main post-#435/#442):

**Codex P1 — network commands bypass the allow-list** — **confirmed and fixed**. I traced just-bash's constructor in the bundled source:

\`\`\`js
for (let i of M0(t.commands)) this.registerCommand(i);
if (t.fetch || t.network) for (let i of U0()) this.registerCommand(i);
\`\`\`

\`U0()\` = \`getNetworkCommandNames()\`. \`WasmShell\` always passes \`fetch\` via \`createProxiedFetch()\`, so just-bash registered every network command regardless of \`BashOptions.commands\`. A scoop with \`allowedCommands: ['echo']\` could still run \`curl\`.

Fix: after \`new Bash(...)\`, reach into its private \`commands: Map\` via cast and delete disallowed network names. Two new regression tests:
- Blocks \`curl\` / \`wget\` when not on the allow-list (exit 127, "not found")
- Keeps \`curl\` working when explicitly allowed (no 127)

The cast is intentional; commented inline. Upstream, a cleaner fix would be for just-bash to respect \`commands\` for network too — but that's out of scope for this PR.

**Copilot — VFS dispose cleanup** — fixed. New \`afterEach\` in the \`WasmShell command allow-list\` describe block matches the pattern in the sibling \`.jsh command registration\` suite.

All 50 tests in \`wasm-shell.test.ts\` pass (was 48 pre-fix, +2 new network-bypass regressions).

Leaving this PR **not** queued — I'd like to wait until #443 (LLM-facing \`allowedCommands\` param on \`scoop_scoop\`) lands so we can verify the end-to-end path from tool-call → \`ScoopConfig\` → \`WasmShell\` live, instead of just via unit tests.

PR #435, opened 2026-04-18
feat(scoops): per-scoop visiblePaths (pure replace) + restore compat

trieloff:
## Summary

Adds `ScoopConfig.visiblePaths?: readonly string[]` — the list of VFS paths a non-cone scoop can READ but not write. Orchestrator wires it straight to `RestrictedFS`'s `readOnlyPaths` with **pure-replace semantics**: `undefined` means no read-only paths at all. Defaults live at the boundary, not in the orchestrator.

- **`scoop_scoop` tool** injects the historical `['/workspace/']` default when creating scoops, preserving existing agent-facing behavior (scoops still see the cone's skills).
- **Orchestrator session restore** backfills `['/workspace/']` for non-cone scoops saved before this field existed. The next `saveScoop` persists the normalized shape — no DB migration needed.
- **Cones** ignore the field entirely — they always use an unrestricted VFS, no RestrictedFS at all.

## Files touched (+220 / -7)

- `packages/webapp/src/scoops/types.ts` — add `visiblePaths` to `ScoopConfig`.
- `packages/webapp/src/scoops/orchestrator.ts` — pure-replace wiring + restore backfill.
- `packages/webapp/src/scoops/scoop-management-tools.ts` — `scoop_scoop` injects the `['/workspace/']` default.
- `packages/webapp/tests/scoops/scoop-management-tools.test.ts` — 3 new tests verifying the tool's default injection.
- `packages/webapp/tests/scoops/orchestrator.test.ts` — 3 new tests verifying the restore backfill (legacy / explicit / cone-skip).

## Depends on

Logically follows #434 (cleaning up the dead non-cone UI flow so `scoop_scoop` is the sole non-cone creation path and the only place the default is injected). Technically mergeable independently — no file overlap with #434.

## Test plan

- [x] `npm run test` — 2621 pass / 2 skipped (6 new tests)
- [x] `npx prettier --check` on changed files
- [x] `npm run typecheck` — no new errors (pre-existing unrelated `lucide` resolution error untouched)
- [ ] `npm run build`

### DIFF
--- a/packages/webapp/src/scoops/orchestrator.ts
+++ b/packages/webapp/src/scoops/orchestrator.ts
@@ -171,8 +171,12 @@ export class Orchestrator {
   }
 
   /**
-   * One-shot compat migration for `ScoopConfig`. Mutates the scoop record in
-   * place so the next `saveScoop` persists the normalized shape.
+   * One-shot in-memory compat migration for `ScoopConfig`. Mutates the scoop
+   * record in place so the rest of the runtime sees the normalized shape;
+   * the DB copy stays legacy until some other operation happens to call
+   * `db.saveScoop` (e.g. a user-initiated scoop update). That's fine — this
+   * migration is idempotent and cheap, so re-running it on every boot until
+   * the record gets rewritten is a non-issue.
    *
    * Gated on {@link RegisteredScoop.configSchemaVersion} rather than a truthy
    * check on individual fields, so a record explicitly saved with
--- a/packages/webapp/src/scoops/orchestrator.ts
+++ b/packages/webapp/src/scoops/orchestrator.ts
@@ -9,7 +9,13 @@
  * - Owns a single shared VirtualFS instance
  */
 
-import type { RegisteredScoop, ChannelMessage, ScoopTabState, ScheduledTask } from './types.js';
+import {
+  CURRENT_SCOOP_CONFIG_VERSION,
+  type RegisteredScoop,
+  type ChannelMessage,
+  type ScoopTabState,
+  type ScheduledTask,
+} from './types.js';
 import * as db from './db.js';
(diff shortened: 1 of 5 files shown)


trieloff:
Review response — four comments (1 Codex P1, 3 Copilot nits):

**Codex P1 — "preserve explicit undefined visiblePaths across restarts"**
Already addressed by [d212e53](https://github.com/ai-ecoverse/slicc/commit/d212e53) (the commit right after Codex's review fired). The backfill is now gated on \`configSchemaVersion\`, not on a truthy field check — records at \`configSchemaVersion: 1\` keep their authoritative \`visiblePaths\` value (including \`undefined\` and \`[]\`). The test *"preserves an explicit undefined visiblePaths on a current-schema record (no silent backfill)"* exercises exactly this.

**Copilot — "written back on the next saveScoop"** — valid. Addressed in [362f4ed](https://github.com/ai-ecoverse/slicc/commit/362f4ed). Rewrote the \`migrateScoopConfig\` doc comment: the orchestrator does NOT automatically re-save restored records; the migration is purely in-memory per boot, idempotent, and cheap enough to re-run until some other code path (e.g. a user-initiated scoop update) happens to persist the normalized shape.

**Copilot — \`globalThis.window\` shim leak in tests** — valid. Addressed in 362f4ed. Now captures the prior \`window\` value in \`beforeAll\` and restores (or deletes) in \`afterAll\`. Only touches \`globalThis.window\` when we actually shim it.

**Copilot — VirtualFS resources leak from test** — valid. Addressed in 362f4ed. \`afterEach\` now disposes the shared VirtualFS alongside \`orch.shutdown()\`, so BroadcastChannel / IndexedDB handles don't survive between test runs.

All 26 tests in the orchestrator suite still pass locally.

PR #437, opened 2026-04-18
chore(extension): scope panel→offscreen scoop-create to cones

trieloff:
## Summary

Follow-up to #434 and addresses the Codex P1 observation on #436: the panel→offscreen \`scoop-create\` handler in \`offscreen-bridge.ts\` supported both cone and non-cone creation, but the non-cone branch built a \`RegisteredScoop\` with no \`config\` at all. Under the new pure-replace path-config contract (#435 / #436), such a scoop would land in the orchestrator with empty \`visiblePaths\`/\`writablePaths\` and hit \`EACCES\` on its own sandbox.

In practice this path has been cone-only since #434 (\`ScoopsPanel.createCone\`). Non-cone scoops are created by the agent's \`scoop_scoop\` tool inside the offscreen orchestrator, which is where the path-config defaults live. This PR makes the panel→offscreen message surface reflect that reality.

- Rename \`ScoopCreateMsg\` → \`ConeCreateMsg\`, type \`'scoop-create'\` → \`'cone-create'\`, drop the \`isCone\` field.
- Collapse \`offscreen-bridge.ts:393\` to cone-only — no \`isCone\` ternaries, no non-cone folder/jid computation.
- \`OffscreenClient.registerScoop\` throws if called with a non-cone scoop, making the cone-only contract explicit at the API boundary.
- Add a unit test for the rejection path.

## Test plan

- [x] \`npm run test\` — 2616 pass / 2 skipped (1 new test, 1 updated)
- [x] \`npx prettier --check\` on changed files
- [x] \`npm run typecheck\` — no new errors (pre-existing unrelated \`lucide\` resolution error untouched)
- [ ] \`npm run build\`
- [ ] \`npm run build -w @slicc/chrome-extension\`

## Depends on

None. Independent of #434 / #435 / #436 — touches extension messaging and \`OffscreenClient\` only.

### DIFF
--- a/packages/chrome-extension/src/messages.ts
+++ b/packages/chrome-extension/src/messages.ts
@@ -35,10 +35,15 @@ export interface UserMessageMsg {
   messageId: string;
 }
 
-export interface ScoopCreateMsg {
-  type: 'scoop-create';
+/**
+ * Panel → offscreen: bootstrap the cone. Sent exactly once per side-panel
+ * session when no cone exists on disk yet. Non-cone scoops are created by
+ * the agent's `scoop_scoop` tool inside the offscreen orchestrator, not
+ * through this message.
+ */
+export interface ConeCreateMsg {
+  type: 'cone-create';
   name: string;
-  isCone: boolean;
 }
 
 export interface ScoopFeedMsg {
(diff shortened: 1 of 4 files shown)


trieloff:
Review response — addressed both Copilot nits in [f2f1c58](https://github.com/ai-ecoverse/slicc/commit/f2f1c58):

**1. Test at line 288 — unawaited `registerScoop`**: made the test `async` and `await`ed the call so future additions to registerScoop's await chain surface in assertions rather than being hidden.

**2. Comment at line 398 — "see review of #436" is brittle**: replaced with the bare issue reference `see #436`.

No functional change.

PR #438, opened 2026-04-18
fix(ui): don't let boot-time licks wipe restored chat history

trieloff:
## Summary

- A page reload would drop the restored chat history and leave only a single lick message (the mount-recovery `session-reload` lick from #325) — and the bug was sticky because it also corrupted the persisted `session-cone` SessionStore entry.
- Fix: gate the `scoopMessageBuffers` branch in `handleScoopSelect` on a per-scoop first-select flag so the first select after boot always goes through SessionStore. Mid-session switches still use the buffer (screenshots / transient detail).

## Root cause

On reload:

1. `initSession('session-cone')` restores history into `chat.messages` from SessionStore.
2. The fire-and-forget `getAllMountEntries()` (`main.ts:1529`, added in #325) detects any mount whose FSA permission is no longer `'granted'` and fires a `session-reload` lick.
3. `routeLickToScoop()` does two things: calls `addLickMessage()` (which pushes to `chat.messages` and persists via `persistSession()` — fine), **and** unconditionally pushes to `scoopMessageBuffers[cone.jid]` at `main.ts:1497`.
4. `handleScoopSelect(cone)` at `main.ts:1902` runs the buffer branch because `buffer.length > 0`:

    ```ts
    if (buffer && buffer.length > 0) {
      await layout.panels.chat.switchToContext(contextId, !scoop.isCone, scoopName);
      layout.panels.chat.loadMessages(buffer);  // ← replaces this.messages with [lick]
    }
    ```

5. `loadMessages()` (`chat-panel.ts:251`) replaces `this.messages` with the buffer and calls `persistSession()` — **overwriting** the SessionStore entry with a single-entry array. Every subsequent reload then re-renders only the lick until new assistant output arrives.

The branch was designed for mid-session scoop switches where the buffer has screenshots not in SessionStore. It shouldn't fire on the first select after boot.

## Fix

- Introduce `selectedOnceThisRuntime: Set<string>` alongside `scoopMessageBuffers`.
- Gate the buffer branch on `!isFirstSelect && buffer.length > 0`.
- Flip the flag at the end of `handleScoopSelect`.
- Clear the set alongside `scoopMessageBuffers.clear()` on clear-chat so the next select behaves like a fresh boot.

SessionStore already contains `[history + any boot-time licks]` by the time `switchToContext` runs (because `addLickMessage` inlines + persists the lick), so the else-branch produces the correct view without touching the data flow for mid-session switches.

## Test plan

- [x] `npm run typecheck` (only pre-existing unrelated `lucide-icons.ts` error)
- [x] `npm run test -w @slicc/webapp` — 2275/2275 pass
- [x] Live repro: reload with a mount whose permission carries over → history preserved, no lick fires (verified)
- [ ] Live repro: reload with a mount whose permission was revoked → history **and** the session-reload lick both visible
- [ ] Scoop switch mid-session (cone ↔ scoop) — buffer-branch detail (screenshots) still shown

### DIFF
--- a/packages/webapp/src/ui/main.ts
+++ b/packages/webapp/src/ui/main.ts
@@ -1849,7 +1849,9 @@ async function main(): Promise<void> {
 
     if (!isFirstSelect && buffer && buffer.length > 0) {
       // Mid-session switch: the buffer carries transient detail (screenshots)
-      // not in SessionStore, so prefer it over the persisted view.
+      // not in SessionStore, so prefer it over the persisted view. The buffer
+      // was seeded with the canonical history on first-select, so it contains
+      // prior-runtime messages in addition to runtime-only events.
       await layout.panels.chat.switchToContext(contextId, !scoop.isCone, scoopName);
       layout.panels.chat.loadMessages(buffer);
     } else {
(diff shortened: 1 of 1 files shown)


trieloff:
Review response (two bot reviewers, one P2 + two comment nits):

**Copilot nits (comments on lines 1840/1850)** — addressed in [0d0d4a8](https://github.com/ai-ecoverse/slicc/commit/0d0d4a8). Rewrote the block comment at the top of the context-switch branch to describe both paths (first-select vs. subsequent), and rewrote the else-branch comment since it now also runs on first-select-with-buffer, not just when the buffer is empty. No functional change.

**Codex P2 (`!isFirstSelect` drops delegated-scoop tool-call detail on first open)** — valid concern, and it survives into this PR as well because the buffer branch is simply bypassed on first open; for a scoop the cone delegated to, `onResponse`/`onToolStart`/`onIncomingMessage` entries that only live in the buffer would be invisible until the user switched away and back.

Addressing in the stacked PR #439 rather than here, because the resolution depends on the buffer-seeding step that PR introduces. Specifically, #439 upgrades the seed from "clear + replace with canonical" to a **merge**: the buffer is rebuilt as `[canonical + any runtime-only entries not in canonical, deduped by id]`, and if any runtime-only entries were picked up, `loadMessages(buf)` surfaces them in the first-open view too. So a scoop the cone delegated to now shows both SessionStore history and buffer-resident tool-call detail on first click.

Keeping #438 narrow since that PR is purely the first-select gate + clear-chat reset + comment updates. The behavioral refinement for delegated scoops lives in #439.

PR #441, opened 2026-04-18
fix(scoops): await createScoopTab in registerScoop to prevent dropped prompt

### LINKED ISSUE #440: scoop_scoop with prompt silently drops the initial task due to concurrent createScoopTab race
When scoop_scoop is called with a prompt, the task is silently dropped and the scoop goes straight to idle without doing any work.

Root cause:

In scoop-management-tools.ts, after onScoopScoop resolves, onFeedScoop is called immediately (fire-and-forget). onScoopScoop calls registerScoop, which starts createScoopTab as its own fire-and-forget. So when onFeedScoop → delegateToScoop → sendPrompt runs, the tab is not yet in this.tabs at all — createScoopTab hasn't executed yet, only been queued.

sendPrompt sees no context, so it calls createScoopTab itself. Now two concurrent createScoopTab calls race on the same jid. The second hits the this.contexts.has(jid) early-return guard and exits. Depending on timing, the tab ends up stuck in initializing or the prompt is dropped entirely at the waitForTabReady timeout path.

Affected files:

packages/webapp/src/scoops/orchestrator.ts — registerScoop starts tab init fire-and-forget
packages/webapp/src/scoops/scoop-management-tools.ts — calls onFeedScoop immediately after onScoopScoop resolves
Fix:

registerScoop should await createScoopTab before returning, so the tab is at least in initializing state (and in this.tabs) before sendPrompt is called. sendPrompt already has waitForTabReady polling, so it will handle the remaining init time correctly.

// orchestrator.ts — registerScoop
async registerScoop(scoop: RegisteredScoop): Promise<void> {
  await db.saveScoop(scoop);
  this.scoops.set(scoop.jid, scoop);
  this.messageQueues.set(scoop.jid, []);
  // Await so the tab exists before any immediate sendPrompt call races it
  await this.createScoopTab(scoop.jid);
}
Observed behavior: Scoop created with a prompt goes idle immediately, task never runs, no error reported.

Expected behavior: Scoop runs the provided prompt before going idle.

trieloff:
Closes #440.

## Summary

- `scoop_scoop` with a `prompt` silently dropped the initial task and the scoop went idle.
- Fix: `orchestrator.registerScoop` now awaits `createScoopTab` instead of firing it and forgetting it, so the context is fully registered before the caller's immediate follow-up `onFeedScoop` runs. No more race.

## Root cause (as diagnosed in the issue)

1. `scoop_scoop` handler in `scoop-management-tools.ts` awaits `onScoopScoop(...)` then fires `onFeedScoop(newScoop.jid, taskPrompt).catch(...)` — fire-and-forget.
2. `onScoopScoop` → `orchestrator.registerScoop`, which previously called `this.createScoopTab(jid).catch(...)` as fire-and-forget and returned. By the time registerScoop resolved, `this.contexts.has(jid)` was still `false`.
3. The fire-and-forget `onFeedScoop` → `orchestrator.delegateToScoop` → `orchestrator.sendPrompt(jid, ...)`.
4. `sendPrompt` read `this.contexts.get(jid)` → undefined → called `await this.createScoopTab(jid)`. Meanwhile the original fire-and-forget createScoopTab from (2) was in flight.
5. Two concurrent `createScoopTab` calls raced past the `this.contexts.has(jid)` early-return guard (lines 454–465 of `orchestrator.ts`). The guard only catches duplicates *after* `this.contexts.set(...)` has run (line 622), which happens partway through the function — both invocations had already blown past the guard before either set anything.
6. Both built full fs + callbacks + `new ScoopContext(...)`. The second `contexts.set(...)` overwrote the first. `sendPrompt` was holding the reference it fetched after awaiting its own createScoopTab, which may or may not be the one that won the `contexts.set` race. The orphaned context's agent never routed responses through the orchestrator; the prompt was silently dropped.

## Fix

`registerScoop` now awaits `createScoopTab`. By the time `onScoopScoop` resolves in `scoop-management-tools.ts`, the context is registered and `'ready'`, so sendPrompt's `if (!context)` branch doesn't fire and there's no race.

Callers of `registerScoop`:
- `scoops-panel.ts:612` (UI add-scoop flow) — already awaited; now waits slightly longer (~init time) before the UI returns to idle.
- `orchestrator.ts:600` (internal) — already awaited.

Also updated the `scoop_scoop` tool result string — it said "It will start working as soon as initialization completes," but post-fix init *is* complete by the time the tool returns, so the cone got a stale nudge.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test -w @slicc/webapp` — 2275/2275 pass
- [ ] Live: `scoop_scoop` with a `prompt` in the cone — scoop actually runs the prompt (previously went straight to idle)
- [ ] Live: `scoop_scoop` without a prompt — scoop is created and stays idle until `feed_scoop`, no regression
- [ ] Live: "Add scoop" button in the UI scoops panel still works (no hang)

### DIFF
--- a/packages/webapp/src/scoops/orchestrator.ts
+++ b/packages/webapp/src/scoops/orchestrator.ts
@@ -231,23 +231,52 @@ export class Orchestrator {
     };
   }
 
-  /** Register a new scoop and fully initialize its context before returning.
+  /** Register a new scoop and wait until its tab/context has been registered
+   *  before returning. Does NOT guarantee successful initialization:
+   *  `ScoopContext.init()` can handle failures internally and leave the tab
+   *  in 'error' state while `createScoopTab` still resolves. The guarantee
+   *  here is that by the time this resolves, the tab/context entry exists in
+   *  `this.contexts` / `this.tabs` (ready or error).
    *
-   *  Awaiting createScoopTab here (rather than firing-and-forgetting it) is
-   *  what prevents a race with the caller's immediate follow-up sendPrompt.
-   *  `scoop_scoop` with an initial prompt fires `onFeedScoop` the moment this
-   *  resolves: if the tab has not yet been registered in `this.contexts` /
-   *  `this.tabs`, sendPrompt calls createScoopTab itself, and both calls race
-   *  past the `this.contexts.has(jid)` early-return guard (the guard only
-   *  catches duplicates once contexts.set has run, which happens partway
-   *  through the function). The losing context ends up orphaned and the
-   *  initial prompt is dropped. See issue #440. */
+   *  Awaiting createScoopTab (rather than firing-and-forgetting it) is what
+   *  prevents a race with the caller's immediate follow-up sendPrompt.
+   *  `scoop_scoop` with an initial prompt fires `onFeedScoop` the moment
+   *  this resolves: if the tab had not yet been registered in `this.contexts`
+   *  / `this.tabs`, sendPrompt would call createScoopTab itself, and both
+   *  calls would race past the `this.contexts.has(jid)` early-return guard
+   *  (the guard only catches duplicates once `contexts.set` has run, which
+   *  happens partway through the function). The losing context ends up
+   *  orphaned and the initial prompt is silently dropped. See issue #440.
+   *
+   *  On failure, rolls back the in-memory and on-disk scoop records so the
+   *  caller doesn't see a half-registered scoop, and rethrows so the caller
+   *  can surface the error. */
   async registerScoop(scoop: RegisteredScoop): Promise<void> {
     await db.saveScoop(scoop);
     this.scoops.set(scoop.jid, scoop);
     this.messageQueues.set(scoop.jid, []);
     log.info('Scoop registered', { jid: scoop.jid, name: scoop.name });
-    await this.createScoopTab(scoop.jid);
+    try {
+      await this.createScoopTab(scoop.jid);
+    } catch (err) {
+      log.error('Scoop init failed', {
+        jid: scoop.jid,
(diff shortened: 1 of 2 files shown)


trieloff:
Review response — addressed all three Copilot inline comments in [d45935b](https://github.com/ai-ecoverse/slicc/commit/d45935b):

**1. `registerScoop` doc comment overclaims ("fully initialize[s]")** — reworded to "registered (ready or error)". `ScoopContext.init()` can handle failures internally and leave the tab in an 'error' state while `createScoopTab` still resolves; the guarantee we actually provide is that by the time this returns, the tab/context entry exists in `this.contexts`/`this.tabs`, which is what closes the race with the caller's immediate `sendPrompt`.

**2. Lost "Scoop init failed" log + no rollback** — wrapped the `createScoopTab` await in try/catch. On failure: log with jid/name/error (restores the signal the old `.catch` provided), call `destroyScoopTab` to clean any partially-constructed context, delete the in-memory scoop/queue entries, and rollback the `db.saveScoop` via `db.deleteScoop`. Then rethrow so the caller (UI scoops-panel, scoop-management tool) can surface the failure — this is the behavior change the PR intentionally introduced.

**3. `scoop_scoop` success text misleading when `onFeedScoop` fails silently** — now awaits `onFeedScoop` and returns `isError: true` on failure with a message pointing the user at `feed_scoop` to retry. `delegateToScoop` (what onFeedScoop resolves to for the cone) only awaits the persistence + dispatch step — the scoop's agent loop still runs fire-and-forget — so this doesn't block the tool on an LLM turn.

2275/2275 tests pass; typecheck clean.

PR #444, opened 2026-04-19
feat(scoops): expose visiblePaths/writablePaths/allowedCommands on scoop_scoop tool

### LINKED ISSUE #443: Expose visiblePaths / writablePaths / allowedCommands as scoop_scoop tool parameters
## Summary

With #435 (`visiblePaths`) and #436 (`writablePaths`) shipping the `ScoopConfig` path surface as pure-replace, the next natural step is to let the cone itself decide a scoop's sandbox shape when it delegates. Today the `scoop_scoop` tool hardcodes the defaults inside its `execute` handler:

```ts
config: {
  ...(model ? { modelId: model } : {}),
  visiblePaths: ['/workspace/'],
  writablePaths: [`/scoops/${folder}/`, '/shared/'],
},
```

The LLM has no way to ask for a tighter sandbox (e.g., a scoop that only sees `/shared/`) or a looser one (e.g., a scoop that can also read a mounted directory the cone picked up). The cone already writes the scoop's prompt; it's in a better position than anyone to reason about what paths the scoop actually needs.

## Proposal

Widen the `scoop_scoop` input schema to accept three optional array-of-strings fields:

- `visiblePaths?: string[]` — forwarded to `ScoopConfig.visiblePaths`. Default when omitted: `['/workspace/']`.
- `writablePaths?: string[]` — forwarded to `ScoopConfig.writablePaths`. Default when omitted: `['/scoops/<folder>/', '/shared/']`.
- `allowedCommands?: string[]` — per-scoop command allow-list (ties into the direction of #433). Default when omitted: unrestricted.

The handler's existing default injection stays — it just becomes the fallback when the LLM omits the field. `ScoopConfig` already supports `visiblePaths`/`writablePaths` after #435/#436; only the tool schema needs widening (and `ScoopConfig` needs a new `allowedCommands` field for the third one).

### Schema shape

```jsonc
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "model": { "type": "string" },
    "prompt": { "type": "string" },
    "visiblePaths": {
      "type": "array",
      "items": { "type": "string" },
      "description": "VFS paths the scoop can READ. Pure replace — what you set is what you get. Omit to use the default ['/workspace/']. Explicit [] for read-nothing."
    },
    "writablePaths": {
      "type": "array",
      "items": { "type": "string" },
      "description": "VFS paths the scoop can READ AND WRITE. Pure replace. Omit to use the default ['/scoops/<folder>/', '/shared/']. Explicit [] for write-nothing."
    },
    "allowedCommands": {
      "type": "array",
      "items": { "type": "string" },
      "description": "Shell commands the scoop can invoke. Omit for unrestricted."
    }
  },
  "required": ["name"]
}
```

## Why pure replace

Same rationale as #435/#436: the ScoopConfig layer is already pure-replace (defaults live at the boundary, not in the orchestrator). Exposing the same contract to the LLM keeps semantics uniform end-to-end. The LLM setting `"visiblePaths": []` must genuinely give the scoop zero read-only paths — it's the cone's explicit decision.

## Related

- #435 — `visiblePaths` ScoopConfig surface
- #436 — `writablePaths` ScoopConfig surface
- #433 — per-scoop command allow-list (draft, includes `allowedCommands` wiring)

## Out of scope

- Default changes. The existing defaults stay as-is; this issue is purely about exposing the knobs.
- Validation/normalization beyond what `ScoopConfig` + `RestrictedFS` already do. The LLM can provide malformed entries and the existing path-normalization in `RestrictedFS` handles them.

trieloff:
Closes #443. Stacks on #433 (which adds \`ScoopConfig.allowedCommands\`). Merge #433 first, then this PR auto-retargets to main.

## Summary

Until now the cone had no way to shape a scoop's sandbox when it called \`scoop_scoop\` — every scoop got the hardcoded defaults (\`visiblePaths: ['/workspace/']\`, \`writablePaths: ['/scoops/<folder>/', '/shared/']\`, no command allow-list). This widens the tool's input schema with three optional array-of-strings parameters that forward straight to \`ScoopConfig\`.

The cone now has the vocabulary to delegate something like *"spin up a scoop that can only read \`/shared/notes/\`, write nothing, and run only \`grep\`/\`cat\`/\`sort\`"* without needing code changes.

## Changes

- \`packages/webapp/src/scoops/scoop-management-tools.ts\` — widen the \`scoop_scoop\` tool's \`inputSchema\` with \`visiblePaths\`, \`writablePaths\`, \`allowedCommands\`. Updated tool description. Handler forwards each field verbatim to \`ScoopConfig\`, falling back to today's defaults via \`??\` when the caller omits.
- \`packages/webapp/tests/scoops/scoop-management-tools.test.ts\` — 7 new cases covering verbatim forwarding, explicit empty arrays (pure-replace semantics), and the combined all-three-plus-model path.

## Semantics

- **Pure replace** — matches what \`ScoopConfig\` already does. Empty array means "read nothing" / "write nothing" — NOT "fall back to default."
- **\`allowedCommands\` omission is unrestricted** — the handler does \`...(allowedCommands ? { allowedCommands } : {})\` rather than defaulting to \`['*']\`. Keeping \`undefined\` as the canonical unrestricted form avoids a gratuitous shape change in IndexedDB-persisted scoop records and matches the \`ScoopConfig\` field's docstring.
- **Config schema version stamped** — scoops created through this path set \`configSchemaVersion: CURRENT_SCOOP_CONFIG_VERSION\`, so the orchestrator's restore-time compat migrations (#435 / #436) don't overwrite explicit caller values on the next boot.

## Use cases

- Read-only scoop (e.g. for analysis): \`{ name: 'audit', writablePaths: [] }\`
- Narrow sandbox: \`{ name: 'docs-only', visiblePaths: ['/shared/docs/'] }\`
- Safe text-processing scoop: \`{ name: 'grep-worker', allowedCommands: ['echo','cat','grep','sort','wc'] }\`
- Fully-custom sandbox mixing all three, as the combined test shows.

## Test plan

- [x] \`npm run test -w @slicc/webapp\` — 2354 pass (7 new)
- [x] \`npm run typecheck\` — no new errors
- [x] \`npx prettier --check\` on changed files
- [ ] Live: cone-initiated \`scoop_scoop\` with a restricted \`visiblePaths\` — scoop's \`ls\` fails outside the whitelist
- [ ] Live: \`allowedCommands: ['echo']\` — scoop's shell rejects \`curl\` with exit 127 (end-to-end verification of #433's network bypass fix)
- [ ] Live: empty \`writablePaths\` — scoop's write tool returns EACCES for all paths

## Out of scope

- No orchestrator changes. The existing pure-replace wiring from #435/#436 just passes these through.
- No \`ScoopConfig\` schema bump. The caller-provided values occupy the same fields today's defaults inject into — they ARE the \`v2\` shape already.
- No cone-facing documentation beyond the tool description. A follow-up skill might teach the cone the sandbox vocabulary idiomatically, but the default behavior is unchanged either way.

### DIFF
--- a/packages/webapp/src/scoops/scoop-context.ts
+++ b/packages/webapp/src/scoops/scoop-context.ts
@@ -746,7 +746,12 @@ export class ScoopContext {
       }
     }
 
-    // Create default CLAUDE.md if missing
+    // Create default CLAUDE.md if missing. Best-effort: scoops with
+    // pure-replace `writablePaths: []` (or no overlap with the memory
+    // path) have no writable location for this file, and that's a
+    // legitimate configuration — a read-only / audit-style scoop
+    // simply runs without a persisted memory file. Swallowing the
+    // EACCES keeps init on the happy path for zero-write sandboxes.
     const memoryPath = this.scoop.isCone
       ? '/workspace/CLAUDE.md'
       : `/scoops/${this.scoop.folder}/CLAUDE.md`;
@@ -765,7 +770,19 @@ Created: ${new Date().toISOString()}
 ## Context
 (Add important context here)
 `;
-      await this.fs.writeFile(memoryPath, defaultMemory);
+      try {
+        await this.fs.writeFile(memoryPath, defaultMemory);
+      } catch (err) {
+        const code = (err as { code?: string })?.code;
+        if (code === 'EACCES') {
+          log.debug('Skipping default memory write (sandbox is read-only)', {
+            folder: this.scoop.folder,
+            path: memoryPath,
+          });
+        } else {
+          throw err;
+        }
+      }
     }
   }
 
--- a/packages/webapp/src/scoops/scoop-context.ts
+++ b/packages/webapp/src/scoops/scoop-context.ts
@@ -161,6 +161,7 @@ export class ScoopContext {
         env: Object.keys(secretEnv).length > 0 ? secretEnv : undefined,
         browserAPI: browser,
         jshDiscoveryFs: this.skillsFs ? effectiveSkillsFs : undefined,
+        allowedCommands: this.scoop.config?.allowedCommands,
       });
       log.info('WasmShell initialized', { folder: this.scoop.folder });
       const skills = await loadSkills(effectiveSkillsFs, this.skillsDir);
--- a/packages/webapp/src/scoops/scoop-management-tools.ts
+++ b/packages/webapp/src/scoops/scoop-management-tools.ts
@@ -187,7 +187,7 @@ export function createScoopManagementTools(config: ScoopManagementToolsConfig): 
               type: 'array',
               items: { type: 'string' },
               description:
-                'VFS paths the scoop can READ (not write). Pure replace — what you set is what you get. Omit to use the default ["/workspace/"] which exposes the shared skills tree. Pass [] for a scoop that reads nothing. Trailing slash recommended (e.g. "/shared/data/").',
+                'VFS paths the scoop can READ (not write). Pure replace — what you set is what you get. Omit to use the default ["/workspace/"] which exposes the shared skills tree. Pass [] for no extra read-only paths. Note: the scoop\'s writablePaths are always readable too, so a true read-nothing sandbox also requires writablePaths: []. Mounts remain readable regardless. Trailing slash recommended (e.g. "/shared/data/").',
             },
             writablePaths: {
               type: 'array',
--- a/packages/webapp/src/scoops/scoop-management-tools.ts
+++ b/packages/webapp/src/scoops/scoop-management-tools.ts
@@ -165,7 +165,7 @@ export function createScoopManagementTools(config: ScoopManagementToolsConfig): 
       tools.push({
         name: 'scoop_scoop',
         description:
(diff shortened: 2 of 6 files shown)


trieloff:
Addressed both Copilot doc nits in [ef3d449](https://github.com/ai-ecoverse/slicc/commit/ef3d449):

1. \`visiblePaths\` tool schema — reworded away from "reads nothing" claim. Now documents that \`writablePaths\` entries (and mounts) are also readable, and that a true read-nothing sandbox requires \`writablePaths: []\` as well.
2. \`ScoopConfig.writablePaths\` docstring — corrected to state that read access is the union of \`writablePaths\` and \`visiblePaths\`, with write access limited to \`writablePaths\` only. Previously implied only \`visiblePaths\` governed reads.

No behavior change. Live tests (visiblePaths, allowedCommands, empty-writablePaths) all pass; the read-only-sandbox init fix (\`scoop-context.ts\`) held up under scenario 3.

Retargeting base from \`feat/scoop-command-allowlist\` to \`main\` since #433 is in the queue; this PR will land right after.

PR #453, opened 2026-04-20
fix(swift-server): fail fast on port conflicts to stop silent port-walking

### LINKED ISSUE #451: Auto-restarted slicc-server climbs port numbers and spawns a duplicate Chrome
## Summary

When `slicc-server` is restarted (manually, by crash, or by the new Sliccstart auto-restart supervisor in `#TBD`), the fresh server often binds `5711`/`5712`/... instead of the original `5710`, and — if it launches its own Chrome — it launches a **second** Chrome on CDP `9223` next to the original on `9222`. The user-visible symptom is a blank Chrome tab (still pointed at the now-dead `localhost:5710`) plus a separate orphan Chrome window.

## Repro

1. Start Sliccstart, click Start on Chrome. Confirm `slicc-server` is on `5710` and Chrome is on CDP `9222` pointed at `http://localhost:5710`.
2. `kill -9 $(pgrep -f 'Resources/slicc-server')` (or let the process crash naturally).
3. Observe the replacement.

## Observed state (captured live)

```
Google      93992  ...  TCP 127.0.0.1:9222 (LISTEN)                                  # old Chrome, points at dead :5710
slicc-ser   94526  ...  TCP 127.0.0.1:5711 (LISTEN)                                  # new slicc-server, bumped port
slicc-ser   94526  ...  TCP 127.0.0.1:61151->127.0.0.1:9223 (ESTABLISHED)            # talks to its own new Chrome
Google      94533  ...  TCP 127.0.0.1:9223 (LISTEN)                                  # second Chrome, a fresh window
```

## Root cause (best current understanding)

Two independent issues, both on the `slicc-server` side, not the launcher:

1. **`findAvailablePort` walks upward on `EADDRINUSE`.** `packages/swift-server/Sources/Utilities/PortResolver.swift` increments `port += 1` on `EADDRINUSE`. On fast restart the still-connected Chrome keepalive sockets and/or TIME_WAIT can cause the bind to fail for a few seconds, after which the server happily settles on `5711` instead of waiting/retrying `5710`.
2. **`ChromeLauncher` always launches a new Chrome.** When a slicc-server starts with `--cdp-port=9222` but `9222` is already a LISTEN socket held by a prior Chrome, `ChromeLauncher` falls back to the next CDP port and spawns a second Chrome, instead of adopting the existing Chrome on the original port (or refusing to start and asking the supervisor to clean up first).

## Suggested directions

- **Short-term:** on startup, if `PORT` (e.g. 5710) is in use for a short grace window, wait-and-retry with a small timeout before bumping. Better yet: refuse to bump at all when started from Sliccstart and return a clear error — the launcher can then kill the orphan Chrome and retry the whole spawn.
- **Medium-term:** `ChromeLauncher` should detect a live `slicc-owned` Chrome on the requested CDP port (via `/json/version` user-agent or a probe) and reuse it instead of launching another.
- **Launcher responsibility:** `Sliccstart` should track the Chrome it asked slicc-server to manage so it can tear down the old Chrome before auto-restarting the server. Today Sliccstart only tracks the `slicc-server` Process, so orphan Chromes outlive the restart.

## Context

Uncovered while verifying PR #450 (`fix(swift-server): stop ChromeOutputMonitor busy-loop on pipe EOF`) and implementing a follow-on Sliccstart auto-restart supervisor. The supervisor works — it detected the SIGKILL as a crash and respawned slicc-server in ~0.8 s — but the port-walk + duplicate-Chrome behavior above makes the restart user-invisible only if the original port is still available, which isn't always the case.

Log excerpt from the live repro:

```
SliccProcess: process exited: Google Chrome status=9 reason=signal
SliccProcess: auto-restart: Google Chrome in 1.000000s (attempt 1/5)
SliccProcess: spawn: pid=94526 for Google Chrome
slicc.chrome-launcher: [slicc_server] Launched Chrome at /Applications/Google Chrome.app/...
slicc.chrome-launcher: [slicc_server] chrome stderr: DevTools listening on ws://127.0.0.1:9223/...
```

## Labels

Suggest: `bug`, `area:swift-server`, `area:swift-launcher`.

trieloff:
Fixes #451.

## Problem

`slicc-server` hid two separate conflicts behind silently-wrong state instead of surfacing them, producing the blank-Chrome + duplicate-Chrome mess tracked in #451.

1. **`PortResolver` walked the port space on `EADDRINUSE`.** If a launcher explicitly set `PORT=5710` and something briefly held `5710` (stale keepalive from a just-killed predecessor, TIME_WAIT, etc.), the server would happily settle on `5711` — leaving the already-spawned Chrome pointed at the now-dead `http://localhost:5710` and the user staring at a blank tab.
2. **`ChromeLauncher` always called `process.run()` even when a Chrome was already on the requested CDP port.** Because slicc-server uses the default user-data-dir, Chrome's single-instance IPC makes the freshly-spawned process forward its URL to the existing Chrome and exit 0. The only error surfaced was the misleading `Chrome exited with code 0 before reporting its CDP port.`

### `PortResolver` (Sources/Utilities/PortResolver.swift)

- `findAvailablePort` gains a `strict: Bool` parameter. When `true`, `EADDRINUSE` on the preferred port throws `PortResolverError.preferredPortUnavailable(port:)` instead of incrementing. Default stays `false` so unrelated callers keep their current behaviour.

### `ServerCommand` (Sources/CLI/ServerCommand.swift)

- `resolveServePort` now passes `strict: true` when `PORT` was explicitly set via the environment. If `PORT` is absent, permissive walking is preserved so casual dev runs still pick a free neighbour.

### `ChromeLauncher` (Sources/Browser/ChromeLauncher.swift)

- `launch` pre-flights `http://127.0.0.1:<cdpPort>/json/version` before `process.run()`. A live CDP endpoint produces the new `ChromeLauncherError.chromeAlreadyRunning(port:browser:)` with a message that names the offending browser build and tells the user to quit it first — no more duplicate Chrome windows, no more misleading cross-process error.
- The probe is strict: both a `Browser` identifier and a `webSocketDebuggerUrl` must be present, so a random HTTP dev server on the same port is not mistaken for Chrome.

```swift
if let existing = await probeExistingChrome(cdpPort: config.cdpPort) {
    logger.warning("Chrome already on CDP port \\(config.cdpPort): \\(existing)")
    throw ChromeLauncherError.chromeAlreadyRunning(port: config.cdpPort, browser: existing)
}
```

Error message users will now see:

> A Chrome instance is already running on CDP port 9222 (Chrome/147.0.7727.101). Quit it before starting slicc-server again.

### Tests

- `PortResolverTests`: `strict` mode throws `preferredPortUnavailable` on an occupied preferred port and returns the preferred port when free.
- `ServerCommandTests`: `resolveServePort` asks for `strict` mode when `PORT` is explicitly set and keeps permissive mode otherwise.
- `ChromeLauncherTests`: `probeExistingChrome` recognises a real CDP response, ignores connect errors, and rejects non-CDP HTTP. `launch` short-circuits with `chromeAlreadyRunning` and never invokes the `processFactory` when the port is occupied.

All swift-server tests pass: **122 / 122** (was 116).

## What this does NOT ship

Per #451 discussion, these follow-ons stay out of this PR:

- **Adopt-existing-Chrome mode** (medium-term fix in #451). Has real lifecycle questions (who terminates Chrome on server exit?) and deserves its own change.
- **Sliccstart-side Chrome teardown before auto-restart.** Belongs in the launcher, not the server. Preserved locally on a separate branch.
- **Retry-with-grace-window on the serve port.** Strict-fail + supervisor-driven cleanup is more deterministic than hoping `TIME_WAIT` clears in 2 s.

## Behaviour matrix

| Situation | Before | After |
| --- | --- | --- |
| `PORT=5710`, 5710 is free | bound 5710 | bound 5710 |
| `PORT=5710`, 5710 is in use | **silently** bound 5711 | `preferredPortUnavailable(5710)` |
| `PORT` unset, 5710 is in use | walked to 5711 | walked to 5711 (unchanged) |
| Chrome already on CDP 9222 | spawned dup Chrome, got misleading exit-0 error | `chromeAlreadyRunning(port: 9222, browser: \"Chrome/…\")` |
| CDP port 9222 empty | launched Chrome (unchanged) | launched Chrome (unchanged) |
| Some non-Chrome HTTP server on 9222 | spawned dup Chrome → same misleading error | launched Chrome (probe rejects non-CDP JSON) |

### DIFF
--- a/packages/swift-server/Sources/Browser/ChromeLauncher.swift
+++ b/packages/swift-server/Sources/Browser/ChromeLauncher.swift
@@ -104,8 +104,17 @@ struct ChromeLauncher: Sendable {
             FileManager.default.homeDirectoryForCurrentUser.path
         },
         processFactory: @escaping @Sendable () -> Process = { Process() },
-        fetchData: @escaping @Sendable (URL) async throws -> (Data, URLResponse) = {
-            try await URLSession.shared.data(from: $0)
+        fetchData: @escaping @Sendable (URL) async throws -> (Data, URLResponse) = { url in
+            // Bound every internal HTTP probe (CDP pre-flight, waitForCDP)
+            // with an explicit 2 s request timeout. Without this the default
+            // URLSession.shared request timeout of ~60 s can stall Chrome
+            // launch for a full minute when something is listening on the
+            // CDP port but hung at the HTTP layer — well past our own
+            // launchTimeout budget.
+            var request = URLRequest(url: url)
+            request.timeoutInterval = 2
+            request.cachePolicy = .reloadIgnoringLocalCacheData
+            return try await URLSession.shared.data(for: request)
         }
     ) {
         self.logger = logger
--- a/packages/swift-server/Sources/Browser/ChromeLauncher.swift
+++ b/packages/swift-server/Sources/Browser/ChromeLauncher.swift
@@ -54,6 +54,14 @@ enum ChromeLauncherError: LocalizedError, Sendable {
     case chromeExitedBeforeReportingPort(Int32)
     case timedOutWaitingForPort(TimeInterval)
     case cdpUnavailable(Int)
+    /// A Chrome instance was already listening on the requested CDP port
+    /// when we tried to launch our own. Spawning a second Chrome with the
+    /// same user-data-dir would have made the new process hand its URL
+    /// off to the existing one and exit 0 immediately, producing the
+    /// misleading `chromeExitedBeforeReportingPort` error. Fail fast
+    /// instead so the caller (launcher / supervisor / user) can clean
+    /// the orphan up.
+    case chromeAlreadyRunning(port: Int, browser: String?)
 
     var errorDescription: String? {
         switch self {
@@ -67,6 +75,9 @@ enum ChromeLauncherError: LocalizedError, Sendable {
             return "Timed out waiting for Chrome CDP port (\(Int(timeout * 1000))ms)."
         case .cdpUnavailable(let port):
             return "Chrome CDP endpoint did not become ready on port \(port)."
+        case .chromeAlreadyRunning(let port, let browser):
+            let tail = browser.map { " (\($0))" } ?? ""
+            return "A Chrome instance is already running on CDP port \(port)\(tail). Quit it before starting slicc-server again."
         }
     }
 }
@@ -162,6 +173,17 @@ struct ChromeLauncher: Sendable {
         }
         guard fileExists(executable) else {
             throw ChromeLauncherError.invalidChromeExecutable(executable)
+        }
+
+        // Fail fast if another Chrome is already listening on the requested
+        // CDP port. Spawning our own Chrome with the same user-data-dir in
+        // that situation is a trap: the new process hands its URL off to
+        // the existing instance via Chrome's single-instance IPC, then
+        // exits 0 — and the caller sees the misleading "Chrome exited
+        // before reporting its CDP port" error.
+        if let existing = await probeExistingChrome(cdpPort: config.cdpPort) {
+            logger.warning("Chrome already on CDP port \(config.cdpPort): \(existing)")
+            throw ChromeLauncherError.chromeAlreadyRunning(port: config.cdpPort, browser: existing)
         }
 
         let process = processFactory()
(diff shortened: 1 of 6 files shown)


trieloff:
Addressed all three Copilot review points in 77424ff:

1. **Bounded probe timeout.** The default `fetchData` now builds a `URLRequest` with `timeoutInterval: 2` so a hung HTTP service on the CDP port cannot stall startup for the 60 s URLSession default. Tightens `waitForCDP` too, as a bonus.
2. **`LocalizedError` conformance.** `PortResolverError` now has `errorDescription` for every case, so CLI users see "Port 5710 is already in use. Quit the process holding it…" instead of the generic "operation could not be completed" message.
3. **IPv6 false-positive in strict mode.** Strict now probes IPv4 only (since Hummingbird binds `127.0.0.1`). An IPv6-only occupier on `::1:<preferred>` no longer hard-fails the launch. New regression test `testStrictModeIgnoresIPv6OnlyOccupierBecauseServerBindsIPv4` pins this behavior.

Tests: 124 / 124 green (was 122).

PR #462, opened 2026-04-21
perf(scoops): use RestrictedFS for scoop skill discovery

trieloff:
Scoops now use their RestrictedFS for skill discovery instead of the unrestricted sharedFs. This bounds the compatibility skill BFS scan by the scoop's ACL - it can only see directories it has access to.

**Problem**: When creating a new scoop, the compatibility skill discovery (`discoverCompatibilitySkillCandidates`) does a BFS from `/` looking for `.agents/skills/` and `.claude/skills/` directories. If you have mounted directories with many folders, this is very slow.

**Solution**: Pass the scoop's `RestrictedFS` instead of `sharedFs` for skill discovery. The scan naturally stops at ACL boundaries. The cone still gets full unrestricted access.

This is a minimal, targeted fix. A follow-up PR will add a MountIndex for faster mounted directory walks.

### DIFF
--- a/packages/webapp/src/scoops/orchestrator.ts
+++ b/packages/webapp/src/scoops/orchestrator.ts
@@ -825,12 +825,15 @@ export class Orchestrator {
       getBrowserAPI: () => this.callbacks.getBrowserAPI(),
     };
 
+    // Cone gets unrestricted skillsFs for full compatibility skill discovery.
+    // Scoops use their RestrictedFS (passed as `fs`) so skill discovery is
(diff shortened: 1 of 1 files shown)


trieloff:
Closing based on review feedback. The fix doesn't actually solve the mount traversal problem because RestrictedFS includes mounts in getAllPrefixes(), and it breaks cache sharing across scoops. PR #463 (MountIndex) provides the proper solution for fast mount walks.

PR #472, opened 2026-04-23
feat(ui): inline 'Log in again' link on AuthError bubbles

trieloff:


trieloff:
Addressed all six review comments.

**Copilot:**
- `chat-panel.ts`: `data-action` now reads `meta.actionHint` rather than being hard-coded to `'reauth'`, so future action hints flow through without further edits.
- `chat-panel.ts`: 'Log in again' is now a real `<button type="button">` (was `<a role="button">`) so Space/Enter work for keyboard users with no extra key handling; `aria-disabled` tracks the busy flag; the status element gets `role="status"` + `aria-live="polite"` so screen readers announce login progress.
- `chat.css`: reset native button chrome so the element still reads as a link + focus-visible ring for keyboard nav.
- tests: `panel.dispose()` in `afterEach` (and on the first panel before constructing `reloaded`) — unregisters the document keydown listener + agent subscription so listeners don't leak across tests.
- tests: replaced `setTimeout(..., 10)` with `vi.waitFor` — persistence polling no longer depends on a fixed delay and won't flake on slower runners.

**Codex (P2):**
- `main.ts`: CLI `onError` now pushes a persisted auth-error `ChatMessage` into the per-scoop buffer *before* emitting the live event, so the inline 'Log in again' action survives `handleScoopSelect → loadMessages(buffer)` scoop switches.
- `offscreen-bridge.ts`: extension `onError` pushes the same message into the offscreen `messageBuffers` and calls `persistScoop`, so the bubble survives panel reconnects and later buffer-driven session writes. `BufferedChatMessage` extended with the optional `authAction` field.

Rebased onto `main` (now includes the merged #474 CDP fix) and force-pushed as `859aa223`. `npm run typecheck`, `npm test` (2885 pass), `npm run build -w @slicc/webapp`, `npm run build -w @slicc/chrome-extension` all green.

trieloff:
Popup-blocker fix pushed as `ee42bd29`.

**Problem:** Adobe's `onOAuthLogin` does `await fetchProxyConfig(...)` before calling the launcher, so by the time `launchOAuthCli` ran `window.open(authorizeUrl, ...)` the user activation from the "Log in again" click had already expired and Chrome silently blocked the popup. The bubble would flip to "Opening login window…" and just sit there forever.

**Fix:**
- `createOAuthLauncher(prewarmedPopup?)` now accepts an already-open window handle. `launchOAuthCli` navigates that window via `popup.location.href = authorizeUrl` instead of calling `window.open` a second time.
- `ChatPanel`'s click handler synchronously calls `window.open('about:blank', ...)` BEFORE any `await`, then hands the handle to `createOAuthLauncher`. The popup stays inside the user-gesture grace window and Chrome lets it through.
- Extension mode skips prewarming entirely (uses `chrome.identity.launchWebAuthFlow`, no popup).
- If the browser explicitly denies the popup, the chat bubble now surfaces "Your browser blocked the login popup…" instead of hanging silently.
- Tests cover both the prewarmed path (no `window.open` call, navigates existing handle) and the closed-handle fallback.

Verified locally with dev server on :5720. Note: a companion fix for a CDP feedback-loop crash that was making the dev server unusable long enough to actually trigger this path is in #477 — reviewing #472 in isolation is fine, but testing end-to-end needs #477 merged or cherry-picked first.

PR #474, opened 2026-04-23
fix(cdp-proxy): disable ws maxPayload cap to survive CDP feedback loop

trieloff:


trieloff:
Addressed both review notes (Codex P2 + Copilot):

- Reverted `maxPayload: 0` on the `WebSocketServer` — the feedback loop only grows on the Chrome→proxy direction, so raising the cap on client→proxy traffic just widens the DoS surface for anything on localhost that can reach `ws://127.0.0.1:PORT/cdp`.
- Kept `maxPayload: 0` on `new WebSocket(url, ...)` (the actual fix — that's where the oversized payloads arrive).
- Rewrote the comment so it no longer references a non-existent `CDP_WS_MAX_PAYLOAD` constant; the reasoning is now stated inline and cross-links between the two sockets.

Force-pushed as `e79741b5`. Verified locally: dev server on :5720 still brings up Chrome and returns `Target.getTargets` cleanly, no `RangeError: Max payload size exceeded`.

PR #477, opened 2026-04-23
fix(cdp-proxy): drop oversized Chrome→Client frames before stringifying

trieloff:


trieloff:
Addressed both Copilot comments and also switched to a more targeted approach.

**Comment 1 ("maxPayload: 0 referenced but not set"):**
Rebased onto `main` — the branch was originally cut before #474 landed, so `new WebSocket(url)` had no options. After the rebase the `maxPayload: 0` option is in place again (it already existed on main via #474) and the comment matches reality.

**Comment 2 ("String(data) doesn't handle ArrayBuffer/Buffer[] correctly"):**
Added a `toBuffer(data)` helper that normalises the ws library's polymorphic payload into a single `Buffer` before any peek/forward. Previously `String(data)` would coerce `ArrayBuffer` to `"[object ArrayBuffer]"` and `Buffer[]` to comma-joined stringified fragments, corrupting CDP frames whenever the socket happened to emit in those modes. Now everything flows through `buf.toString()` from the normalised Buffer.

**Bonus:** while debugging a real browser session with this change the blanket 16 MiB size drop was starving Chrome's internal debugger UI and re-triggering the 'Debugger paused in another window' overlay (it drops legitimately-large events like screenshots). Switched to a **method-based filter** that only drops the two self-referential events actually in the feedback loop — `Network.webSocketFrameReceived` and `Network.webSocketFrameSent`. Peek the first 256 KiB of each frame, check the JSON head, drop only those two methods. A 64 MiB hard cap remains as a safety net.

Force-pushed as `577d437d`.

PR #483, opened 2026-04-24
fix(scoops): add retry logic and fatal error handling for LLM requests

trieloff:
## Summary

Fixes scoops getting stuck when configured with non-existent models (e.g., `opus-4.5` instead of `opus-4.6`).

### Retry Logic with Exponential Backoff
- Added 3-attempt retry with exponential backoff (1s, 2s, 4s) for transient errors
- Retryable errors: rate limits (429), server errors (5xx), network issues

### Non-Retryable Error Detection
- Immediate failure for auth errors (401, 403), invalid models, billing issues
- New `isNonRetryableError()` and `isRetryableError()` helper functions

### Fatal Error Notification
- Added `onFatalError` callback that bypasses `scoop_mute` preferences
- Force-unmutes scoops on fatal error so cone is always notified
- Resolves any pending `scoop_wait` promises with null
- Sends `scoop-error` channel message to cone

### Skill Documentation
- Updated scoop-management skill with model verification guidance
- Added reminder to run `models` before specifying a model for a scoop
- Documented the new error handling behavior

## Testing
- Updated existing test for prompt failure to use non-retryable error
- All 56 scoop-context tests pass
- 4 pre-existing failures in scoop-context.tools.test.ts (unrelated to this change)

### DIFF
--- a/packages/vfs-root/workspace/skills/scoop-management/SKILL.md
+++ b/packages/vfs-root/workspace/skills/scoop-management/SKILL.md
@@ -161,6 +161,8 @@ scoop_unmute({ scoop_names: ["scraper"] })
 
 ## Model Selection for Scoops
 
+**IMPORTANT: Always run `models` to verify available models before specifying a model for a scoop.** Model availability depends on the configured provider and API key. Specifying a non-existent model will cause the scoop to fail immediately with an unrecoverable error.
+
 Use `models --json` to discover available models before creating scoops. The `scoop_scoop` tool accepts a `model` parameter.
 
 Intelligence, speed, and cost are independent dimensions. Use `models --json` to compare them and pick the best tradeoff for each task:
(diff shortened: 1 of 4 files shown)


trieloff:
Review comments addressed:

1. **Capture agent + try/finally cleanup + disposal re-checks** — already landed in `1763fe93`: `const agent = this.agent` captured once at the top of `prompt()`, the retry loop is wrapped in `try { ... } finally { this.isProcessing = false; ... }`, and `if (this.disposed) return;` runs before each attempt and after each `await`.

2. **Abortable backoff sleeps / stop() cancellation** — addressed in new commit `db44081a`. Introduced an `AbortController` per `prompt()` invocation and an `abortableSleep(ms, signal)` helper that resolves early on abort. `stop()` and `dispose()` now call `promptAbortController.abort()` before touching the agent, so pending backoff delays cancel immediately instead of waiting for `setTimeout` to fire. Abort state is checked at every retry boundary and before final error reporting, so `onFatalError` is not fired when the user explicitly stops the scoop.

3. **Reset `didStreamDeltas` per attempt** — already landed in `1763fe93`: the flag is reset at the top of each retry attempt inside the loop.

4. **Unit tests for `isNonRetryableError` / `isRetryableError`** — already landed in `1763fe93`: see the `describe('isNonRetryableError', ...)` and `describe('isRetryableError', ...)` blocks in `scoop-context.test.ts`. The new commit adds additional tests for `abortableSleep` and `stop()`/`dispose()`-during-backoff.

5. **Clear `scoopResponseBuffer` on fatal error** — already landed in `1763fe93`: `this.scoopResponseBuffer.delete(jid)` is called alongside `this.pendingCompletions.delete(jid)` in the orchestrator's `onFatalError` handler.

Verification: `npm run typecheck`, `npx vitest run packages/webapp/tests/scoops/scoop-context.test.ts` (77/77 passing), `npm run build -w @slicc/webapp` all green.

PR #484, opened 2026-04-24
fix(ui): make a href links clickable in inline sprinkles

trieloff:
Inline sprinkle iframes use `sandbox="allow-scripts allow-same-origin"` (CLI) or `sandbox="allow-scripts"` (extension nested iframe), which blocks top-level navigation. As a result, plain `<a href>` clicks inside agent-authored ```shtml blocks did nothing.

## Fix

- **`packages/webapp/src/ui/inline-sprinkle.ts`**: The in-iframe bridge click handler now also walks up to the nearest `<a href>`, `preventDefault()`s the click, resolves the URL against the iframe's base, and posts an `inline-sprinkle-open-link` message to the parent. Internal `#` anchors keep default (in-iframe) behavior; `javascript:` URLs are dropped. The CLI and extension parent handlers open the URL via `window.open(url, '_blank', 'noopener,noreferrer')`, gated to `http(s):`/`mailto:` schemes for safety.
- **`packages/chrome-extension/sprinkle-sandbox.html`**: Added `inline-sprinkle-open-link` to the child→parent relay list so extension-mode clicks reach the side panel.

## Verification

- `npx prettier --write` on touched files
- `npm run typecheck` — green
- Targeted sprinkle tests (17/17 passed): `packages/webapp/tests/ui/inline-sprinkle.test.ts`, `packages/chrome-extension/tests/sprinkle-sandbox.test.ts`
- `npm run build` and `npm run build -w @slicc/chrome-extension` — both green

### DIFF
--- a/packages/chrome-extension/sprinkle-sandbox.html
+++ b/packages/chrome-extension/sprinkle-sandbox.html
@@ -523,7 +523,7 @@ addEventListener('message', function(e) {
       'sprinkle-stat', 'sprinkle-mkdir', 'sprinkle-rm',
       'sprinkle-storage-set', 'sprinkle-storage-remove', 'sprinkle-storage-clear',
       'sprinkle-fetch-script',
-      'inline-sprinkle-lick', 'inline-sprinkle-height'
+      'inline-sprinkle-lick', 'inline-sprinkle-height', 'inline-sprinkle-open-link'
     ];
     if (bridgeTypes.indexOf(msg.type) !== -1) {
       parent.postMessage(msg, '*');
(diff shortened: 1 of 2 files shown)


trieloff:
Thanks for the review. The extra scoop retry/fatal-error changes were unintentionally pulled in because this branch was cut from `fix/scoop-error-handling-retry` (tracked in PR #483) instead of `main`. I've rebased `fix/inline-sprinkle-clickable-links` directly onto `origin/main` and force-pushed, so this PR now contains only the single inline-sprinkle link-click commit. The three scoop-related review comments (onFatalError doc, redundant null-check, SKILL.md wording) apply to PR #483 and should be addressed there.

PR #488, opened 2026-04-24
fix(server): load Keychain secrets before launching the browser

trieloff:
## Problem

When `sliccstart` launches `slicc-server`, the macOS Keychain access prompts for the `ai.sliccy.slicc` entries appear **after** Chrome / Electron has taken window focus. They end up visually stacked behind whichever browser tab the user has on top. Multiple secrets (e.g. four `GWS_*` entries from the Google Workspace skill) produce four sequential dialogs the user has to alt-tab through.

## Fix

Hoist the `SecretInjector(sessionId:envFileSecrets:)` construction before the Chrome / Electron launch block. The keychain reads now happen while the user's terminal / Finder is still in front, so the dialogs are visible immediately and can be dismissed before the browser opens.

## Why not coalesce N prompts into 1?

Investigated; the answer is no — captured in the in-source comment so future readers don't repeat the experiment:

- `kSecReturnAttributes` + `kSecReturnData` + `kSecMatchLimitAll` together returns `errSecParam` (-50). _Tested empirically._
- `kSecMatchItemList` requires `kSecMatchLimitOne` per Apple's docs, so feeding back persistent refs from a first attributes-only query also can't be batched.

The number of prompts is driven by each item's ACL, not the call shape:

- Items added with `security add-generic-password -A` (or by `slicc-server` itself, which trusts its own binary) show **no prompt**.
- Items added with the default ACL prompt **once per item** until the user clicks _Always Allow_.

A follow-up commit on the skills repo will recommend `-A` in the GWS install recipe so the four-prompts-on-first-launch problem disappears outright for new installs.

## Tests

`swift test` — all 124 swift-server tests pass.

🤖 Generated with [Factory](https://factory.ai)

### DIFF
--- a/packages/swift-server/Sources/CLI/ServerCommand.swift
+++ b/packages/swift-server/Sources/CLI/ServerCommand.swift
@@ -90,44 +90,24 @@ struct ServerCommand: AsyncParsableCommand {
         var browserLabel = config.electron ? "Electron" : "Chrome"
         var overlayInjector: ElectronOverlayInjector?
 
-        // Load secrets from the Keychain BEFORE the browser launches.
-        //
-        // Without this hoist, `SecretInjector.loadSecrets()` ran several
-        // hundred milliseconds after Chrome / Electron took window focus,
-        // and every macOS Keychain access prompt for an `ai.sliccy.slicc`
-        // entry appeared *behind* the browser window — one dialog per
-        // secret, all stacked under whichever Chrome tab the user happened
-        // to have on top. Pulling the load forward keeps the prompts in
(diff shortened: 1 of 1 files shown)


trieloff:
Closing — superseded by #493 (single-blob Keychain layout). The N-prompt scaling problem this PR addressed no longer exists; users now get exactly one ACL prompt per signed-binary lifetime regardless of secret count.

The single residual prompt still pops behind the browser on first launch, so I'm opening a slimmed-down replacement that just does the minimal `SecretInjector` hoist (no `BrowserLaunchPlan` enum, no obsolete N-prompt commentary).

PR #493, opened 2026-04-26
feat(secrets): collapse Keychain storage into a single .env blob

trieloff:
## Summary

- Replaces the per-secret Keychain layout with one item under service `ai.sliccy.slicc`, account `__envfile__`, whose value is an `.env`-formatted blob (`KEY=value` paired with `KEY_DOMAINS=...`). Same format as node-server's `~/.slicc/secrets.env`. One item = one macOS access prompt instead of N.
- Adds a Sliccstart **Settings** scene (Sliccstart > Settings…, `⌘,`) with a table-based secrets manager — add, edit, delete, hostname-pattern validation, error surface for Keychain failures.
- Centralises env-file parsing in a shared `EnvFileFormat` so the `--env-file` CLI override and the Keychain blob can't drift.

## Why

Each secret was its own Keychain item, and macOS prompted for the user's password once per item access at `SecretStore.list() + get(name:)` time. Consolidating to a single item means one prompt per signed-binary lifetime, regardless of how many secrets the user stores.

## Layout

```
KEYCHAIN
└── service: ai.sliccy.slicc
    └── account: __envfile__
        value: |
            GITHUB_TOKEN=ghp_...
            GITHUB_TOKEN_DOMAINS=api.github.com,*.github.com
            OPENAI_KEY=sk-...
            OPENAI_KEY_DOMAINS=api.openai.com
```

Both Sliccstart (settings UI) and swift-server (read at startup) hit this same item. `SecretStore`'s public API (`get`/`set`/`delete`/`list`) is unchanged, so `SecretInjector`, `APIRoutes`, and `ServerCommand` work without further touch-ups.

## Files

**swift-server**
- `Sources/Keychain/EnvFileFormat.swift` (new) — parser/serializer + blob↔Secret helpers, mirrors `node-server/src/secrets/env-file.ts`.
- `Sources/Keychain/SecretStore.swift` (rewritten) — single-item R-M-W with `NSLock`.
- `Sources/CLI/ServerCommand.swift` — `parseEnvFileSecrets` collapsed to a 4-line wrapper around the shared parser.
- `Tests/EnvFileFormatTests.swift` (new) — 10 tests.

**Sliccstart**
- `Sliccstart/Models/Secrets.swift` (new) — `Secret` type, Keychain accessors, parallel `EnvFileFormat`.
- `Sliccstart/Views/SettingsView.swift` (new) — `Table` + add/edit/delete sheets, validates name characters, blocks duplicates, requires ≥1 hostname pattern.
- `Sliccstart/SliccstartApp.swift` — registers `Settings { SettingsView() }`.
- `SliccstartTests/EnvFileFormatTests.swift` (new) — 6 tests pinning Sliccstart's parser to the same behaviour.

## Notes

- **No migration**: per the request, legacy per-name Keychain entries are not folded in — the feature isn't widely used yet.
- **Tradeoff**: all secrets share one ACL. Anyone allowed to read the blob reads all of them. This was implicitly true before too (all entries shared the same service identifier), but worth naming.
- **Format compatibility**: the blob format matches node-server's `~/.slicc/secrets.env` exactly, so `swift-server --env-file ~/.slicc/secrets.env` works for cross-runtime sharing.

## Test plan

- [x] `swift build` clean in both packages
- [x] `swift test` — 136 swift-server tests pass (10 new), 18 Sliccstart tests pass (6 new)
- [ ] Manually open Sliccstart > Settings…, add a secret, verify it shows up in `slicc-server`'s `GET /api/secrets`
- [ ] Verify only one Keychain prompt fires when slicc-server starts with multiple secrets stored
- [ ] Sanity-check that `--env-file` still overrides Keychain entries with the same name

### DIFF
--- a/packages/swift-launcher/Sliccstart/Models/Secrets.swift
+++ b/packages/swift-launcher/Sliccstart/Models/Secrets.swift
@@ -1,0 +1,173 @@
+import Foundation
+import Security
+
+/// A SLICC secret managed by the Settings UI. `name` doubles as the stable
+/// identity for SwiftUI lists and selection.
+struct Secret: Equatable, Identifiable, Hashable {
+    var name: String
+    var value: String
+    var domains: [String]
+
+    var id: String { name }
+}
+
+enum SecretsError: LocalizedError {
+    case keychainError(status: Int32)
+    case emptyDomains
+    case emptyName
+    case duplicateName(String)
+
+    var errorDescription: String? {
+        switch self {
+        case .keychainError(let status):
+            return "Keychain error (\(status))"
+        case .emptyDomains:
+            return "At least one hostname pattern is required."
+        case .emptyName:
+            return "Name must not be empty."
+        case .duplicateName(let name):
+            return "A secret named \"\(name)\" already exists."
+        }
+    }
+}
+
+/// Read/write the single Keychain item that holds every SLICC secret.
+///
+/// Service: `ai.sliccy.slicc`
+/// Account: `__envfile__`
+/// Value: `.env`-formatted blob, identical layout to swift-server's
+/// `SecretStore` and node-server's `~/.slicc/secrets.env`.
+enum SecretsKeychain {
+    static let service = "ai.sliccy.slicc"
+    static let account = "__envfile__"
+
+    static func readBlob() -> String {
+        let query: [String: Any] = [
+            kSecClass as String: kSecClassGenericPassword,
+            kSecAttrService as String: service,
+            kSecAttrAccount as String: account,
+            kSecReturnData as String: true,
+            kSecMatchLimit as String: kSecMatchLimitOne,
+        ]
+        var result: AnyObject?
+        let status = SecItemCopyMatching(query as CFDictionary, &result)
+        guard status == errSecSuccess,
+              let data = result as? Data,
+              let text = String(data: data, encoding: .utf8) else {
+            return ""
+        }
+        return text
+    }
+
+    static func writeBlob(_ content: String) throws {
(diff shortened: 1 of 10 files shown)


trieloff:
Addressed all six review comments in 289088dd.

**Issue A — data loss on Keychain read failure** (Codex P1×2, Copilot×2)
- `readBlob()` in both `SecretStore` (swift-server) and `SecretsKeychain` (Sliccstart) now treats `errSecItemNotFound` as the only legitimate `""`; every other status throws.
- `SecretStore.mutate` propagates the read failure, so `set`/`delete` refuse to overwrite the blob after a transient error.
- `SettingsView.unlock()` catches and surfaces the error via the existing alert; `unlocked` stays `false` so the overlay remains. The `+` button bails before opening the editor, and `save()` refuses to write from an unloaded snapshot.

**Issue B — N+1 Keychain reads in `SecretInjector`** (Copilot)
- New `SecretStore.all()` returns every secret in one read+parse. `SecretInjector.loadSecrets` switches to it; the unused `loadedNames` set went with it.

**Issue C — hostname pattern validation** (Copilot)
- `EnvFileFormat.isValidHostnamePattern(_:)` accepts `*`, `*.host.tld`, and exact `host.tld` shapes (per-label `[A-Za-z0-9_-]`, no leading/trailing dash, no empty labels). `SecretEditorSheet` runs every non-empty entry through it and surfaces the first invalid pattern in the validation banner; Save stays disabled.

**Tests**: swift-server 138 (+2), Sliccstart 20 (+2). All passing.

PR #497, opened 2026-04-27
feat(local-models): models tab + SwiftLM auto-download + Local provider

trieloff:
## Summary

End-to-end local-LLM support: a Models tab in Sliccstart that pulls MLX-quantized models from HuggingFace, runs them through SwiftLM, and exposes them as a no-config "Local (SwiftLM)" provider in the chat panel.

- **Sliccstart Models tab** (between Startup and Secrets). Lists models already in the HF cache, plus a curated "Suggested" catalog. Each row supports Run/Stop/Delete or Install/Cancel with progress.
- **SwiftLM auto-download** at `~/.slicc/SwiftLM/<version>/SwiftLM`. Pinned tag in `SwiftLMVersion.swift` with a Renovate `customManagers` marker, so SwiftLM `b###` bumps land as PRs. Quarantine attribute is stripped post-extract so the Hardened Runtime parent can spawn the ad-hoc-signed child without Gatekeeper detours.
- **`SwiftLMProcess`** sets sane CLI defaults (`--ctx-size` from the model's declared `max_position_embeddings`, `--max-tokens 32768`, `--cors *`, always-on `--thinking` and `--turbo-kv`) and **auto-passes `--vision`** when the chosen model is multimodal. Detection mirrors SwiftLM's own `ModelArchitectureProbe`.
- **`packages/webapp/src/providers/built-in/swiftlm.ts`** — `requiresApiKey: false`, `requiresBaseUrl: false`, `getModelIds` reads a localStorage cache that `register()` refreshes from `/v1/models` + `/health` at boot. Streaming tunnels through `/api/fetch-proxy` (same pattern as bedrock-camp) so the OpenAI SDK's `credentials: 'include'` doesn't fight SwiftLM's wildcard CORS. When SwiftLM stops, the cache clears and the provider silently disappears from the dropdown.
- **Client-side `ThinkSplitter`** mirrors SwiftLM's `<think>` / `<|channel>thought<channel|>` tag tracking and forces a default thinking phase for Qwen 3.x models (whose tokenizer emits the rationale before the tool call without wrapping it). Strips the orchestrator's `[time] sender: ` prefix from outgoing user messages so the model doesn't read it as a transcript.

## Suggested catalog

| Model | Size | Notes |
|---|---|---|
| `mlx-community/Qwen3.6-35B-A3B-4bit` | 18 GB | **Recommended.** MoE (3 B active per token); fast on Apple Silicon; reliable tool calls + thinking. |
| `mlx-community/Qwen3.6-27B-4bit` | 14 GB | Dense fallback if 35B-A3B doesn't fit. Generates much slower (every token activates all 27 B params). |

## Why a custom installer (not SPM)

SwiftLM cannot be consumed as a Swift Package dependency: its `Package.swift` declares `mlx-swift` and `mlx-swift-lm` as **local-path** packages backed by git submodules of forked Apple repos, and SPM does not init submodules of dependencies. The pre-built artifact is `.tar.gz` (raw Mach-O + `.metallib`), not the `.xcframework.zip` that `.binaryTarget` requires. So we download the release tarball at runtime. Renovate still bumps the pinned version via the `customManagers` regex on `SwiftLMVersion.swift`.

## Capability matrix (verified against SwiftLM b602)

| Capability | Status | Notes |
|---|---|---|
| Tool calls | ✓ | OpenAI `tools` ↔ `tool_calls` round-trip verified end-to-end through the chat panel against Qwen3.6-35B-A3B-4bit and Qwen3.6-27B-4bit. |
| Thinking traces | ✓ | `delta.reasoning_content` plus client-side `<think>` / `<\|channel>` splitter; default-on for Qwen 3.x. |
| Multimodal (VLM) | ✓ (code path) | `--vision` switches SwiftLM to `VLMModelFactory`; `/health.vision` exposed to the webapp. No VLM in the suggested catalog yet. |

## Models we tried and dropped

- **`mlx-community/gemma-4-26b-a4b-it-4bit`** — narrates `Then I'll call the bash tool…` instead of emitting a `tool_calls` fragment when given the SLICC system prompt. Confirmed ecosystem-wide bug (SwiftLM #69, mlx-lm #1096/#1125, oMLX #617, ollama #15368).
- **`mlx-community/gemma-4-31b-mxfp4`** — ships without `chat_template` in `tokenizer_config.json`; swift-transformers rejects every chat completion with `missingChatTemplate`. The `-it-4bit` sibling has the template but degenerates with newline loops on the SLICC prompt.
- **`mlx-community/DeepSeek-V4-Flash-*`** — bottlenecked upstream, not by SwiftLM. There is no `DeepseekV4.swift` in `mlx-swift-lm` (only `DeepseekV3.swift`); SwiftLM's own `Sources/SwiftLM/DeepseekV3DFlash.swift` covers V3 + DFlash but not V4. The actual blocker is [`ml-explore/mlx-lm` issue #1233](https://github.com/ml-explore/mlx-lm/issues/1233) ("Add model support for DeepSeek-V4"), which is **still open with no PR** as of 2026-05-07 — the `model_type: deepseek_v4` value isn't in the registry, so any V4 quant fails at architecture lookup before any weight-key issue surfaces. The earlier note in this PR about `attn_hc` / `ffn_hc` naming was inaccurate. Re-add a V4 quant once mlx-lm issue #1233 lands, then mlx-swift-lm picks it up, then SwiftLM cuts a release with the bumped submodule.

## Test plan

- [x] `swift build` clean in swift-launcher
- [x] `npm run typecheck` clean
- [x] `npm run test` clean
- [x] `npm run build -w @slicc/webapp` clean
- [x] Click "Run" in the Models tab, SwiftLM auto-downloads to `~/.slicc/SwiftLM/b602/`, port 5413 binds, "Local (SwiftLM)" shows up in the provider list
- [x] End-to-end tool-call round-trip through the chat panel (`bash` tool against Qwen3.6-35B-A3B-4bit + Qwen3.6-27B-4bit)
- [ ] Confirm one Renovate PR cycle bumps `SwiftLMVersion.pinned` cleanly when SwiftLM cuts a new release


trieloff:
Closing this. Not ready for prime time, and the risk of footguns for users who turn it on outweighs the value at the current model state.

## Why we're parking

Built an end-to-end eval (`packages/local-models-eval/`, separate commit on this branch) that drives a SLICC-shaped agent loop through `pi-agent-core` + `pi-ai` against a running SwiftLM, with realistic context padding sourced from the actual `packages/vfs-root/**/*.md` corpus (~25 K tokens, basically production scale). The eval results across all the variations we tried:

| Padding | parallel_math | file_exploration | edit_file_round_trip | write_then_run |
|---|---|---|---|---|
| 0       | PASS | PASS | PASS | XFAIL |
| 8 K     | PASS | PASS | **FAIL** | XFAIL |
| 25 K    | 4-bit PASS · bf16 **FAIL** | PASS | **FAIL** | XFAIL |

`edit_file_round_trip` (read → edit → verify, the most common dev workflow) **reliably fails at production context size** on Qwen 3.6 35B-A3B-4bit — the model loses the multi-round thread under realistic skill+memory load. `write_then_run` is a Qwen 3.6 family bug we couldn't tune away after trying:

- `--repeat-penalty 1.1` (vendor recommendation): 0/4 reliable
- Qwen-recommended sampling (`temp=0.6, top_p=0.95, top_k=20`): 0/5
- `--thinking off`: 1/5
- SLICC-aligned tool result wording: ~3/5 (best result; flaky)
- Dense Qwen 3.6 27B-4bit: same fail mode + ~6× slower
- bf16 (mlx-community/Qwen3.6-35B-A3B-bf16, 65 GB): equal-or-worse reliability + 2.3× slower clean, ~equal at scale

bf16 actually regressed `parallel_math` from PASS → FAIL at 25 K padding — the higher-precision model takes more reasoning steps per round and runs out of round budget. The blog-post recommendation for bf16 was for one-shot quality, not agent-loop discipline.

## What works in this branch (kept on the branch for the next iteration)

The work is solid and worth picking back up when the next model lands:

- **Sliccstart Models tab** + SwiftLM auto-download with version pinning + Renovate `customManagers` rule for auto-bumps
- **Adjustable ctx-size** capped at 75 % of physical RAM (`ContextWindowPolicy.resolve`) — fixes the 120 GB peak we hit on first try with Qwen 3.6 35B at the model's full 256 K declared window
- **App Management permission** moved from 2 s polling timer to `NSApplication.didBecomeActive` observer (already merged separately as #585)
- **Orphan SwiftLM port reclaim** with byte-for-byte binary path verification, so a force-killed parent doesn't strand its child holding 5413
- **Auto-run-on-launch** model setting (Models tab picker) gated by the same 64 GB RAM floor as the tab itself
- **Eval suite** at `packages/local-models-eval/` — TS on `pi-agent-core` so the call shape matches production, four scenarios with pytest-style xfail discipline, `--pad-to N` for production-scale context. **Use this as a tripwire on the next SwiftLM bump or model swap** — when the XFAILs flip to XPASS, the model has caught up and this work can be re-opened.

## When to re-open

Watch for:

- **Qwen 3.7 / Qwen 4** — Qwen iterates fast and historically fixes specific tool-call regressions per release
- **DeepSeek V4** — bottlenecked on [`mlx-explore/mlx-lm` issue #1233](https://github.com/ml-explore/mlx-lm/issues/1233) (still open, no PR); when that lands, `mlx-swift-lm` and SwiftLM will pick it up downstream
- **Gemma 5 / Llama 4** with first-class MLX quants
- A small (~7 B) tool-call-tuned MLX model — none exist on mlx-community today but would likely beat a 35 B generalist on agent loops

Run the eval at `--pad-to 25000` against the new model. If `edit_file_round_trip` and `write_then_run` come back clean, this PR is ready to revive.

PR #515, opened 2026-04-28
feat(onboarding): deterministic post-welcome flow with dedicated connect dip

trieloff:
Replaces the LLM-driven onboarding handler with a deterministic state machine owned by the webapp. The cone agent only gets pulled in **after** an LLM provider is configured.

## Flow

1. **Welcome wizard** (unchanged) collects the profile.
2. **OnboardingOrchestrator** intercepts the wizard's onboarding-complete BEFORE it reaches the cone:
   - persists profile + /shared/.welcomed marker
   - kicks off upskill recommendations --install silently in the background
   - posts three deterministic sliccy lines via addSystemMessage (named/anonymous greeting, role-flavoured pitch, AI-confession)
3. Orchestrator follows up with **connect-llm.shtml** — a two-step provider picker + credentials form populated from the live pi-ai catalogue.
4. Dip emits connect-attempt. Orchestrator validates the key against the provider's models endpoint (best-effort for ~9 providers, skip-and-save fallback for the rest), saves credentials, sets the selected model, then fires the new onboarding-complete-with-provider lick to the cone.
5. **SKILL.md is shrunk** to a single trigger — the cone greets + comments on the chosen provider/model in 1-5 sentences.

## New modules

- scoops/onboarding-messages.ts — randomized intro stubs
- scoops/api-key-validator.ts — best-effort live probes for OpenAI / Anthropic / Google / Groq / Cerebras / Mistral / x-ai / DeepSeek / OpenRouter
- scoops/onboarding-orchestrator.ts — state machine
- shared/sprinkles/welcome/connect-llm.shtml — provider picker + credentials form
- ui/dip.ts: adds broadcastToDips() and an in-iframe slicc-message CustomEvent relay so the parent can talk back to dips

## Tests

39 new unit tests cover the messages, validator, and orchestrator. Full webapp suite (3205 tests) + typecheck + webapp build all green.

## Known limitations / follow-ups

- [ ] Chrome-extension lick handler (mainExtension path in ui/main.ts ~L627) still uses the legacy LLM-driven flow — needs a parallel orchestrator wiring to match. **Working on this next.**
- [ ] Mid-flow reload isn't persisted: if the user closes the tab between wizard finish and connect, the orchestrator state is lost.
- [ ] The provider list inside the dip is the FULL pi-ai catalogue — should be reviewed for overwhelm.


trieloff:
Superseded by #519, which contains all of this PR's commits plus the extension split-layout, deterministic onboarding, and Haiku picker filter.

PR #521, opened 2026-04-29
feat: jumphost relay enables GitHub OAuth in extension

catalan-adobe:
## Summary

- **Worker `/auth/callback` relay** now branches on `state.source`. `local` (default, back-compat) keeps the existing localhost behavior; `extension` redirects to `<extension-id>.chromiumapp.org<path>` with strict `^[a-p]{32}$` ID validation to prevent open-redirect via subdomain injection.
- **GitHub provider** sends `redirect_uri=<worker>/auth/callback` in both modes and encodes `{source: 'extension', extensionId, path: '/github', nonce}` for extension. The worker bounces it to chromiumapp.org, where `chrome.identity.launchWebAuthFlow` intercepts.
- **Net effect**: GitHub OAuth Apps (single-callback-URL) work for the extension without a second OAuth App or migration to a GitHub App. The single registered URL stays `https://www.sliccy.ai/auth/callback`.
- **Also fixes** `resolveClientId()` in extension mode: previously the relative `fetch('/api/runtime-config')` 404'd under `chrome-extension://`, leaving `clientId` empty and the OAuth flow throwing. Now it goes straight to the worker.

## Why not simpler approaches

- *Pin the chromiumapp URL on the OAuth App* — GitHub OAuth Apps allow exactly one callback URL with strict host matching (per their docs). The chromiumapp.org and sliccy.ai hosts can't share one app. This is the constraint Adobe sidesteps by having Adobe IMS register the chromiumapp URL directly on the IMS app.
- *Two OAuth Apps* — workable but means a second client ID/secret and a worker-side branch on which app to use per request.
- *Convert to a GitHub App* — supports up to 10 callback URLs but changes the auth model (installations, fine-grained perms) and the user-token shape; would risk breaking GitHub Models API access via pi-ai.

The jumphost is the smallest change that keeps the existing OAuth App and adds extension support behind it.

## Test plan

- [x] Worker `auth-relay.test.ts`: 15 cases covering both branches, unknown `source`, malformed extensionId, subdomain injection attempts, port/path validation. Uses a script-execution test seam (`runRelay()`) so we exercise the relay's actual JS, not just string-match the HTML.
- [x] Full repo test suite: 3145 passing
- [x] Typecheck clean (`npm run typecheck`)
- [x] Extension build clean (`npm run build -w @slicc/chrome-extension`)
- [ ] **Staging deploy via this PR** (`worker-staging.yml` triggers on PR) — once the staging URL bot-comment lands, manually verify the GitHub login flow in the extension by setting `localStorage['slicc.trayWorkerBaseUrl']` to the staging worker URL.

## Files

| File | Change |
|---|---|
| `packages/cloudflare-worker/src/index.ts` | `OAUTH_RELAY_HTML`: source-aware branching with strict extensionId validation |
| `packages/cloudflare-worker/tests/auth-relay.test.ts` | Replaced "localhost-only invariant" test; added script-execution seam + 9 new behavior tests |
| `packages/webapp/providers/github.ts` | Extension `resolveClientId` branch; redirect URI through worker for both modes; state encodes source/extensionId/path; nonce verification now in both modes |
| `packages/webapp/src/providers/oauth-code-exchange.ts` | Export `getWorkerBaseUrl` (was private) |

Adobe is intentionally untouched — its IMS already accepts the chromiumapp URL directly.

### DIFF
--- a/packages/cloudflare-worker/src/index.ts
+++ b/packages/cloudflare-worker/src/index.ts
@@ -36,16 +36,29 @@ try {
   var raw = params.get('state') || hashParams.get('state');
   if (!raw) throw new Error('Missing state parameter');
   var state = JSON.parse(atob(raw));
-  var port = Number(state.port);
+  var source = state.source || 'local';
   var path = state.path || '/auth/callback';
   var nonce = state.nonce || '';
-  if (!port || port < 1024 || port > 65535) throw new Error('Invalid port: ' + port);
   if (!path.startsWith('/')) throw new Error('Invalid path');
-  // Forward all original query params (except state, which we consumed) to
-  // localhost so authorization codes (?code=xxx) survive the relay.
+  // Forward all original query params (except state, which we consumed) so
+  // authorization codes (?code=xxx) survive the relay.
   params.delete('state');
   params.set('nonce', nonce);
-  var target = 'http://localhost:' + port + path + '?' + params.toString();
+  var query = '?' + params.toString();
+  var target;
+  if (source === 'local') {
+    var port = Number(state.port);
+    if (!port || port < 1024 || port > 65535) throw new Error('Invalid port: ' + port);
+    target = 'http://localhost:' + port + path + query;
+  } else if (source === 'extension') {
+    // Chrome extension IDs are 32 chars in [a-p]. Strict format check prevents
+    // open-redirect via subdomain injection (e.g. "evil.com.").
+    var extensionId = state.extensionId || '';
+    if (!/^[a-p]{32}$/.test(extensionId)) throw new Error('Invalid extensionId');
+    target = 'https://' + extensionId + '.chromiumapp.org' + path + query;
+  } else {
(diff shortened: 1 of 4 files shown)


catalan-adobe:
End-to-end staging test passed.

Tested the GitHub OAuth flow in the Chrome extension against the staging worker:

- Loaded the unpacked extension built from this branch
- Set `localStorage['slicc.trayWorkerBaseUrl']` to the staging URL and reloaded the extension to clear the cached client ID
- Triggered GitHub login

The full chain worked: extension → GitHub authorize (staging client ID + staging redirect_uri, accepted by the staging OAuth App) → staging worker `/auth/callback` → new relay decoded `state.source = 'extension'` → redirected to `<extension-id>.chromiumapp.org/github` → `chrome.identity.launchWebAuthFlow` captured the redirect → token exchanged → user logged in.

Ready for review.

PR #525, opened 2026-04-29
test(worker): smoke-test OAuth broker credentials on deploy

catalan-adobe:
## Summary
- Adds a smoke test in `packages/cloudflare-worker/tests/deployed.test.ts` that POSTs a fake authorization code to `/oauth/token` and asserts GitHub returns `bad_verification_code`.
- GitHub validates `(client_id, client_secret)` before the code, so this distinguishes a missing/stale `GITHUB_CLIENT_SECRET` from a valid configuration.
- Runs against both prod and staging via the existing `describeIfConfigured` gate in `worker.yml` and `worker-staging.yml` — neither workflow needs changes.

## Why
A misconfigured `GITHUB_CLIENT_SECRET` deployed silently with a green checkmark earlier this cycle because the existing deployed-test only exercises tray flow, not the OAuth broker. This closes that gap.

## Test plan
- [x] `npx prettier --write packages/cloudflare-worker/tests/deployed.test.ts` (no changes — already formatted)
- [x] `npm run typecheck` passes
- [x] `WORKER_BASE_URL=https://www.sliccy.ai npx vitest run --project cloudflare-worker packages/cloudflare-worker/tests/deployed.test.ts` — 3/3 pass against prod
- [ ] Staging smoke test passes when this PR triggers `worker-staging.yml`

### DIFF
--- a/packages/cloudflare-worker/tests/deployed.test.ts
+++ b/packages/cloudflare-worker/tests/deployed.test.ts
@@ -232,6 +232,22 @@ describeIfConfigured('deployed tray worker', () => {
     const body = await response.text();
     expect(body).toContain('<!DOCTYPE html>');
   }, 15_000);
+
+  // GitHub validates (client_id, client_secret) before the code, so a fake-code
+  // probe distinguishes "credentials wrong" from "credentials fine, code fake".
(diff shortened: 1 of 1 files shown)


trieloff:
good idea.

PR #529, opened 2026-04-30
fix(release): centralize version on package.json

catalan-adobe:
## Summary

- The released artifacts (`sliccy@2.20.0`, `v2.20.0`) had drifted from three different stale placeholders carried in git: `package.json` `0.1.0`, `manifest.json` `0.2.0`, `vfs-root/shared/version.json` `0.0.0-dev`. `@semantic-release/npm`'s in-process bump was never written back, so the published webapp bundle (built before semantic-release ran) embedded the dev placeholder permanently — the upgrade-detection lick was effectively dead in the published CLI tarball.
- Make root `package.json` the single source of truth. Inject `__SLICC_VERSION__` and `__SLICC_RELEASED_AT__` via Vite `define` in the webapp and chrome-extension builds; `upgrade-detection.ts` reads the build-time constant instead of `/shared/version.json` from the VFS, dropping the `fs` argument and the now-vestigial `0.0.0-dev` sentinel branch. `dist/extension/manifest.json`'s version is always overwritten from `package.json` in the `copy-extension-assets` plugin (committed source carries `0.0.0` as a clear sentinel).
- Add `@semantic-release/git` so `package.json` and `package-lock.json` bumps land on `main`. The `.releaserc.json` `prepareCmd` now runs a full `npm run build` under `SLICC_RELEASED_AT` so the webapp bundle is rebuilt with the next version (previously only the chrome-extension was). `sign-and-package.sh` reads the version from `package.json` directly. Deleted: `sync-release-version.ts`, `vfs-root/shared/version.json`, `assertMatchingVersions`, and their tests. Bumped root `package.json` to `2.20.0` to match the current release.
- Drive-by: removed six pre-existing `(event as any)` casts in `offscreen.ts` that the linter flagged when this change touched the file — `LickEvent` already declares the accessed fields.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test` (3260 passed)
- [x] `dist/extension/manifest.json` shows `"version": "2.20.0"` after build
- [x] Webapp bundle (`dist/ui/assets/index-*.js`) inlines the literal `2.20.0`; no `__SLICC_VERSION__` literals leak
- [x] `npx prettier --check` on tracked files


catalan-adobe:
@trieloff, this is a proposal to set Slicc's artifacts version in one place (root package.json) and have packages use it.
Also have the release flow bump the version back into package.json

trieloff:
Thank you. Merging right away.

trieloff:
@catalan-adobe the addition of the git plugin prevents semantic release to go through because of the branch protections. I'm trying to relax them.

PR #568, opened 2026-05-05
fix(chat): preserve tool call collapse state and fix cluster gap

vdua:
## Summary

- **Preserve open state across rebuilds**: `updateMessageEl` and `reflowToolClusters` replace/rebuild DOM on every tool call, collapsing any `<details>` the user had expanded. Two new Sets on `ChatPanel` (`userOpenedToolCalls`, `userOpenedClusters`) record user intent via `toggle` listeners; element creation restores open state from those Sets.
- **Stable cluster key**: uses the first tool call's ID only (not all IDs joined) so the key stays the same as the cluster grows.
- **Fix visual gap after clustering**: when `reflowToolClusters` moves tool calls from continuation `msg-group` elements into a cross-message cluster, those groups become empty shells that create a visible gap. They are now marked `msg-group--emptied` (`display: none`) to close the gap, while staying in the DOM so `unwrapToolClusters` can restore calls on the next reflow.

## Test plan

- [ ] Run several bash commands in sequence — a Working cluster should appear with no gap between it and subsequent output
- [ ] Expand the Working cluster mid-run — it stays open when the next tool call fires
- [ ] Expand an individual tool call mid-run — it stays open on subsequent calls
- [ ] Collapse an expanded item — it stays collapsed
- [ ] Refresh the page — open state is not required to survive (by design)

### DIFF
--- a/docs/superpowers/specs/2026-05-05-tool-call-collapse-fix-design.md
+++ b/docs/superpowers/specs/2026-05-05-tool-call-collapse-fix-design.md
@@ -1,0 +1,60 @@
+# Tool Call Collapse State Preservation
+
+**Date:** 2026-05-05
+**Status:** Approved
+
+## Problem
+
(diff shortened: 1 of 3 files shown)


vdua:
closing the PR since it is a duplicate of #566 and #567

PR #581, opened 2026-05-06
fix(extension): make tray host actually start

trieloff:
## Summary

Two bugs in the Chrome extension prevented `host` from ever surfacing a join URL on a clean install — the agent reported "the host is inactive and no join URL is available" instead of bootstrapping a tray. This is what Eric hit on 2.29.1.

- **Offscreen never resolved a worker URL.** `packages/chrome-extension/src/offscreen.ts` called `resolveTrayRuntimeConfig` without `defaultWorkerBaseUrl`, so a fresh extension with no `VITE_WORKER_BASE_URL`, no stored config, and no URL override returned `null` and `LeaderTrayManager` was never constructed. Standalone passes a default; the extension silently did not. Now uses the production worker in production builds, staging in dev builds — mirroring standalone.
- **`createTrayFetch` lost the Window binding.** In the extension branch it returned the bare `fetch` reference. `LeaderTrayManager` invokes it as `this.fetchImpl(...)`, which rebinds `this` to the manager instance and trips the browser's "Illegal invocation" check on the global fetch. Wrapped in an arrow so the call site can't rebind `this`. Captured live: `Leader tray join failed: Failed to execute 'fetch' on 'Window': Illegal invocation`.

The "WebRTC flakiness / missing auto-reconnect" theory in the original thread was wrong — `LeaderTrayManager` already has a full reconnect loop. The leader simply never started, so there was nothing to reconnect.

## Test plan

- [x] `npx vitest run --project webapp packages/webapp/tests/scoops/tray-leader.test.ts` — 13/13 pass (includes 2 new `createTrayFetch` tests covering the method-call invocation invariant)
- [x] `npm run typecheck`
- [x] `npm run test` — 3541 passed / 3 skipped
- [x] `npm run build`
- [x] End-to-end smoke: fresh extension profile in Chrome for Testing now auto-attaches to the production tray worker and IndexedDB shows a real `leader-tray-session` with a `trayId` on first run.

## Merge note

Use a merge commit, not squash — per the repo policy avoiding the squash-merge-queue regression bug.

### DIFF
--- a/packages/chrome-extension/src/offscreen.ts
+++ b/packages/chrome-extension/src/offscreen.ts
@@ -17,6 +17,8 @@ import {
 } from '../../../packages/webapp/src/scoops/agent-bridge.js';
 import { LeaderTrayManager } from '../../../packages/webapp/src/scoops/tray-leader.js';
 import {
+  DEFAULT_PRODUCTION_TRAY_WORKER_BASE_URL,
+  DEFAULT_STAGING_TRAY_WORKER_BASE_URL,
   hasStoredTrayJoinUrl,
   resolveTrayRuntimeConfig,
   TRAY_JOIN_STORAGE_KEY,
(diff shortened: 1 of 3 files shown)


trieloff:
Superseded by #582, which now targets main and includes both the bug fixes (createTrayFetch binding, missing defaultWorkerBaseUrl) and the menu wiring.

PR #588, opened 2026-05-06
fix(webapp): show provider name in model dropdown when multi-provider

catalan-adobe:
## Summary

- The chat model dropdown rendered only `model.name`, so users with two or more provider accounts (e.g. Anthropic + Bedrock both serving Claude Sonnet) could not tell which provider — and which cost center — a selection was billing to.
- `getAllAvailableModels()` already exposes `providerName`; the flattening in `renderModelSelector` (`packages/webapp/src/ui/chat-panel.ts`) was discarding it. Now it carries through, and when more than one provider is configured the provider name is rendered as a muted secondary label on both the trigger pill and each menu item.
- Single-provider users see no visual change — the labels only appear when `new Set(providerIds).size > 1`.

Screenshot 2026-05-06 at 15 46 57

## Changes

- `packages/webapp/src/ui/chat-panel.ts` — extend the flattened `allModels` list with `providerName`, compute `showProviderLabel`, and append the muted label to the trigger and each menu item when multi-provider.
- `packages/webapp/src/ui/styles/chat.css` — three new selectors (`.chat__model-btn-provider`, `.chat__model-menu-left`, `.chat__model-menu-provider`) reusing existing S2 tokens; trigger gets a thin divider, menu sub-label is smaller and muted.
- `packages/webapp/tests/ui/chat-panel-model-selector.test.ts` — new regression test covering the single-provider (no labels) and multi-provider (labels on trigger + every menu item) cases.

## Test plan

- [x] `npm run typecheck`
- [x] `npx vitest run packages/webapp/tests/ui/` — 567/567 pass (incl. the 2 new cases)
- [x] `npx prettier --check` on touched files — clean
- [x] `npm run build` — succeeds (pre-existing warnings unrelated)
- [x] Manual UI verification with two provider accounts configured — provider names render on the pill and menu items; single-provider mode unchanged

### DIFF
--- a/packages/webapp/src/ui/chat-panel.ts
+++ b/packages/webapp/src/ui/chat-panel.ts
@@ -1274,12 +1274,18 @@ export class ChatPanel {
     const currentProvider = getSelectedProvider();
 
     // Flatten all models with their provider info
-    const allModels: Array<{ providerId: string; id: string; name: string; reasoning?: boolean }> =
-      [];
+    const allModels: Array<{
+      providerId: string;
+      providerName: string;
+      id: string;
+      name: string;
+      reasoning?: boolean;
+    }> = [];
     for (const group of groups) {
       for (const model of group.models) {
         allModels.push({
           providerId: group.providerId,
+          providerName: group.providerName,
           id: model.id,
           name: model.name,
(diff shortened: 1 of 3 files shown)


trieloff:
looks great.

PR #596, opened 2026-05-07
fix(ios): add NSCameraUsageDescription to resolve App Store rejection

trieloff:
## Summary

- Adds `NSCameraUsageDescription` to `project.yml` (XcodeGen source) and directly to both Debug and Release build configurations in `project.pbxproj`
- Resolves App Store rejection **ITMS-90683**: the WebRTC dependency references camera APIs, which requires a purpose string in Info.plist even though the app does not actively use the camera

## Test plan

- [ ] Build and archive `SliccFollower` — confirm `NSCameraUsageDescription` appears in the generated `Info.plist` inside the `.ipa`
- [ ] Upload to App Store Connect and verify no ITMS-90683 warning


trieloff (commented):
Addressed all three review comments in commit 82c8c77: updated `NSCameraUsageDescription` to explicitly state the app does not access the camera and that the key is only present due to the bundled WebRTC framework referencing camera APIs.

PR #604, opened 2026-05-08
feat(chat): stream dip preview into iframe as agent writes shtml

catalan-adobe:
## Summary

https://github.com/user-attachments/assets/4b82132b-7cea-4aca-9bba-20caf7d3eca4

- Streams fenced ```shtml blocks into a sandboxed iframe as the agent writes them, replacing the previous "raw markup typing out then sudden swap" experience with a live preview that takes shape in place.
- **The streaming preview is non-interactive** until the agent finishes the block (`pointer-events: none` on the iframe + `cursor: progress` on the segment container). Half-typed `slicc.lick()` calls and unfinished forms can be looked at but not clicked. Full interactivity returns when stream-end hydration mounts the final trusted dip.
- Adds a segment-based renderer in `chat-panel` that keeps each shtml block in a stable container the iframe never has to leave — re-parenting an iframe destroys its `contentWindow` in WHATWG-compliant browsers, so the previous per-frame `innerHTML` rebuild was reloading the iframe every animation frame and never finishing.

## How it works

**Streaming path** (`renderStreamingSegmented` in `chat-panel.ts`):
- `splitContentSegments(content)` slices `msg.content` at every ```shtml fence into ordered prose / shtml segments.
- Prose segments update via `innerHTML` only when the text changes.
- Shtml segments hold a draft iframe mounted exactly once. Updates flow via `postMessage` (`dip-draft-update`) — the iframe's body is replaced in place, lucide icons re-rendered, custom elements upgraded.

**Final hydration is unchanged**: stream-end disposes drafts and the existing `hydrateDips()` mounts the trusted final iframe with the full bridge.

**Security**: drafts are never registered in `trustedDipWindows` and the bridge ignores `dip-readfile`/`dip-exists`/`dip-stat` for them. Lick + auto-height + link-open work the same as final dips so partial UI is still wireable, within the same security envelope as inline shtml.

**Extension mode**: `mountDraftDip` returns `null` because `sprinkle-sandbox.html` doesn't yet have a draft-update path. The placeholder stays visible during streaming and final hydration runs as before — clean follow-up.

## Files

- `packages/webapp/src/ui/dip.ts` — adds `mountDraftDip`, `extractShtmlBlocks`, `splitContentSegments`, `DraftDipInstance`, `ContentSegment`. Refactors `mountDip` to share scaffolding via a `buildDipSrcdoc` helper.
- `packages/webapp/src/ui/chat-panel.ts` — replaces `updateStreamingContent` with the segment renderer. Adds a `drafts` map alongside `dips`, fanned into `disposeDipsForMessage` and `disposeAllDips`.
- `packages/webapp/src/ui/message-renderer.ts` — `renderAssistantMessageContent` gains an `isStreaming` flag that swaps closed shtml code blocks for a "Pouring a dip…" placeholder. Used as the first-frame affordance before content arrives.
- `packages/webapp/src/ui/styles/chat.css` — `.msg__dip-pending` placeholder (mirrors `.tool-call--running` row), `.msg__seg--shtml` container with the non-interactive cue.
- `packages/webapp/tests/ui/dip.test.ts` — 18 new tests (`mountDraftDip` lifecycle, `extractShtmlBlocks`, `splitContentSegments`).
- `packages/webapp/tests/ui/message-renderer.test.ts` — 6 new tests for the streaming placeholder swap.

## Test plan

- [ ] Hard reload, ask the agent to emit a sizable shtml block. Verify the iframe content streams in progressively without flicker.
- [ ] Try clicking buttons inside the dip while it's still streaming — clicks should not register, cursor should show `progress`.
- [ ] Wait for the stream to finish — buttons become interactive, lick events fire correctly.
- [ ] Ask the agent for a message with multiple shtml blocks interleaved with prose — each block should stream independently into its own iframe; prose between them should appear normally.
- [ ] `npm run typecheck` — clean
- [ ] `npx vitest run tests/ui/` — 615 tests pass

### DIFF
--- a/packages/chrome-extension/sprinkle-sandbox.html
+++ b/packages/chrome-extension/sprinkle-sandbox.html
@@ -375,6 +375,17 @@ addEventListener('message', function(e) {
     return;
   }
 
+  // Streaming-draft body updates: the parent posts incremental shtml
+  // content while the agent is still writing the block. The nested
+  // iframe's DRAFT_BRIDGE_EXTENSION listener swaps document.body.innerHTML
+  // on each message. We just relay — the child does the actual work.
+  if (e.source === parent && msg.type === 'dip-draft-update') {
+    if (window.__slicc_childIframe && window.__slicc_childIframe.contentWindow) {
+      window.__slicc_childIframe.contentWindow.postMessage(msg, '*');
+    }
+    return;
+  }
+
   if (msg.type === 'sprinkle-render') {
     // Apply initial theme class so CSS variables resolve correctly on first paint.
     if (typeof msg.isLight === 'boolean') {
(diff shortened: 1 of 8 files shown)


trieloff:
That looks nice.

PR #610, opened 2026-05-10
feat(webapp/ui): LLM-suggested labels for tool clusters and prompt placeholder

trieloff:
## Summary

- Adds `quickLabel()`, a one-shot wrapper around pi-ai's `completeSimple` that reuses the active provider's credentials and prefers a cheaper sibling in the same model family (Claude → haiku, GPT/o-series → mini/nano, Gemini → flash, Grok → fast/mini), falling back to the active model if nothing qualifies. Returns `null` on any failure so call sites can quietly keep their static fallback.
- Wires it into the chat input placeholder: after each agent turn ends, the static `"What shall we build?"` is replaced with a follow-up suggested from the last 3 user turns + last finalized assistant response. Aborted on the next turn; reverts to the default if the user has started typing.
- Wires it into the tool-call cluster preview: replaces `bash, bash, bash` with a short imperative phrase describing what the calls accomplish. Uses tool **inputs only** (never return values) so the label is stable while the run is still in progress. Cached by sorted tool-call ids so reflow re-renders reuse the prior result.

## Hardening (cluster prompt)

When bash inputs contain expressions like `console.log(1+2)` or `print(1+1)`, the LLM was treating them as code to evaluate and replying with the *result* (`"3"`, `"2"`). Two layers now prevent that:

1. The system prompt explicitly forbids executing/computing/answering inputs and includes worked examples covering an arithmetic-in-bash case.
2. An `isUsefulClusterLabel` filter rejects degenerate replies (< 6 chars, no letters, single word) so they fall back to the static comma-joined preview instead of being displayed and cached.

## Adobe-provider compatibility

Ad-hoc UI calls don't have a scoop session, so they get their own daily-rotated `X-Session-Id` anchored to a fixed `ui-quick-llm` localStorage key. This preserves the proxy's session-grouping invariant without leaking scoop folder names.

## Test plan

- [x] `npx vitest run packages/webapp/tests/ui/quick-llm.test.ts` — 13 tests covering family-based model picking (Claude/GPT/Gemini), active-model fallback, `modelId` override, Adobe header gating, prompt/system forwarding, and the three null-return paths.
- [x] `npx vitest run packages/webapp/tests/ui/` — all 604 UI tests pass.
- [x] `npm run typecheck` clean.
- [x] `npx prettier --check` clean.
- [ ] Manual: send a few prompts in `PORT=5720 npm run dev`, confirm the textarea placeholder rotates after each agent turn and that 3+-tool-call clusters get a meaningful imperative label (with `bash, bash, bash` fallback when the LLM call fails or returns a degenerate result).

### DIFF
--- a/packages/webapp/src/scoops/llm-session-id.ts
+++ b/packages/webapp/src/scoops/llm-session-id.ts
@@ -43,10 +43,28 @@ function safeLocalStorage(): Storage | null {
   }
 }
 
-function getDailyUuid(coneJid: string): string {
+/**
+ * Daily-rotated UUID anchored to a free-form key. Shared across every
+ * Adobe `X-Session-Id` producer in the codebase so rotation cadence
+ * and fallback semantics stay identical:
+ *
+ *  - Keyed on `DAILY_UUID_KEY_PREFIX + anchor` in `localStorage` when
+ *    available.
+ *  - Falls through to an in-memory `Map` when storage is locked or
+ *    unavailable. The in-memory cache rotates on the next call past
+ *    UTC midnight, exactly like the persisted path.
+ *  - Returns a fresh `crypto.randomUUID()` if both reads/writes fail,
+ *    so the caller still gets a usable id (worst case: the id rotates
+ *    every call).
+ *
+ * Anchors are arbitrary identifying strings — a scoop's cone JID for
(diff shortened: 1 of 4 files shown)


trieloff:
We should persist the labels for completed clusters in IndexedDB, so that we don' fall back to the default for them.

PR #613, opened 2026-05-10
fix(shell/sprinkle): proxy SprinkleManager from kernel-worker to page over BroadcastChannel

trieloff:
## Summary

After PR #607 the standalone shell runs in a \`DedicatedWorker\` that has no \`window\`. Every \`sprinkle\`, \`open <foo>.shtml\`, and post-install \`upskill\` call returned \`"sprinkle: sprinkle manager not initialized"\`. This adds a same-origin BroadcastChannel bridge so the worker-side shell can reach the real page-side \`SprinkleManager\`.

The wire shape mirrors the extension's existing chrome.runtime proxy — same op vocabulary, same caching semantics on the proxy side — so debugging across floats reads the same.

## What changes

**New module** \`packages/webapp/src/scoops/sprinkle-bridge-channel.ts\`:
- \`createSprinkleManagerProxyOverChannel()\` — worker-side proxy. Same surface the shell consumes (\`refresh\`, \`available\`, \`opened\`, \`open\`, \`close\`, \`sendToSprinkle\`, \`openNewAutoOpenSprinkles\`).
- \`installSprinkleManagerHandlerOverChannel(manager)\` — page-side dispatcher. Forwards manager errors as \`error\` strings.

**Wiring**:
- \`kernel-worker.ts\` publishes the proxy on \`globalThis.__slicc_sprinkleManager\`.
- \`mainStandaloneWorker\` constructs the real \`SprinkleManager\` (mirroring the extension panel's wiring), installs the channel handler, and disposes it on \`beforeunload\`.
- \`sprinkle-command.ts\`, \`open-command.ts\`, \`upskill-command.ts\`: read from \`globalThis.__slicc_sprinkleManager\` instead of \`window.__slicc_sprinkleManager\` so the same lookup works in both realms.

**Untouched**: extension offscreen still uses its existing chrome.runtime proxy (\`packages/chrome-extension/src/sprinkle-proxy.ts\`).

## Test plan

- [x] 9 new bridge tests cover: round-trip dispatch, error forwarding, fire-and-forget sends, timeout-on-no-handler, response/request isolation (handler ignores its own responses), no-op fallback when \`BroadcastChannel\` is unavailable.
- [x] Existing sprinkle / open / upskill command tests updated to publish on \`globalThis\`. 75/75 affected tests pass.
- [x] \`npm run typecheck\` clean (4 tsconfigs).
- [ ] Manual standalone (\`PORT=5720 npm run dev\`): \`sprinkle list\` shows available sprinkles; \`sprinkle open welcome\` opens the panel; \`open /workspace/skills/foo/foo.shtml\` works.

### DIFF
--- a/packages/webapp/src/kernel/kernel-worker.ts
+++ b/packages/webapp/src/kernel/kernel-worker.ts
@@ -220,6 +220,17 @@ async function boot(init: KernelWorkerInitMsg): Promise<void> {
     logger: console,
   });
 
+  // Publish a sprinkle-manager proxy on the worker's globalThis so the
+  // `sprinkle` / `open` / `upskill` shell commands can reach the real
+  // page-side manager. The bridge uses a same-origin BroadcastChannel;
+  // the page bootstrap (`mainStandaloneWorker`) installs the matching
+  // handler. Extension offscreen has its own chrome.runtime-based
+  // proxy in `offscreen.ts` and never goes through this path.
+  const { createSprinkleManagerProxyOverChannel } =
+    await import('../scoops/sprinkle-bridge-channel.js');
+  (globalThis as Record<string, unknown>).__slicc_sprinkleManager =
+    createSprinkleManagerProxyOverChannel();
+
   // Take the process manager from the kernel host so scoop-turns
   // (registered by `ScoopContext`) and shell execs (registered by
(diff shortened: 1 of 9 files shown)


trieloff:
I can see this working.

PR #644, opened 2026-05-13
fix(context-compaction): strip orphaned toolResult blocks after compaction cut

QuentinVecchio:
## Summary

Corrupt persisted sessions — where the first stored message is a `toolResult` with no preceding assistant turn — caused a permanent Bedrock 400 on the user's next prompt. This exports `stripOrphanedToolResults` and calls it at session restore so broken sessions self-heal silently on the next load.

## Why

Bedrock (and Claude in strict mode) rejects a conversation that opens with a `toolResult`:

  > messages.0.content.0: unexpected tool_use_id found in tool_result blocks

The orphan arises from a browser crash mid-save, or from sessions written before the walk-back guard was introduced. There was no recovery path; the session was permanently broken for the affected user.

`stripOrphanedToolResults` unconditionally drops any leading `toolResult` messages — no `toolCallId` matching needed, because a `toolResult` at the head is orphaned by definition (there is no preceding assistant turn).

It is called in two places:
  - **`scoop-context.ts` (load-bearing):** applied to `saved.messages` before the `Agent` is constructed. This is the fix that unblocks users. `droppedOrphans` is logged for field telemetry.
  - **`createCompactContext` / `compactContext` (defense-in-depth):** applied after the compaction cut. The walk-back guard already prevents a leading `toolResult` here by construction, so the strip is normally a no-op. It is kept as a safety net against future changes to the cut algebra and is documented as such in the docstring.

The integration test in `context-compaction.test.ts` was renamed from "strips orphaned toolResults left at the head after compaction cut" to "walk-back guard keeps assistant+toolResult pair together across the cut" — because the strip is a no-op on that fixture; it's the walk-back guard that makes the test pass.

## Cross-cutting impact

No behavior change outside the session-restore path. The compaction strips are no-ops under normal operation. No API surface changes. The new `stripOrphanedToolResults` export is the only public-facing addition.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test` — `<N>` passing, no new failures
- [x] `npm run build`
- [x] **New / changed behavior is covered by a unit test** in
      `packages/*/tests/` mirroring the changed `src/` path
      (e.g. `npx vitest run packages/webapp/tests/<area>/<file>.test.ts`)
- [ ] Manual smoke (CLI): `npm run dev` + …
- [x] Manual smoke (extension): load unpacked `dist/extension/` + …
- [ ] Manual smoke (Electron / Sliccstart / worker) — if relevant
- [ ] N/A — explain below

## Documentation

- [x] No documentation changes needed

## Checklist

- [x] No secrets, credentials, OAuth client secrets, or tokens in the diff (incl. logs, fixtures, `.env*`, snapshots)

### DIFF
--- a/packages/webapp/src/core/context-compaction.ts
+++ b/packages/webapp/src/core/context-compaction.ts
@@ -39,23 +39,25 @@ function hasRole(message: AgentMessage, role: string): boolean {
 }
 
 /**
- * Drop any `toolResult` messages at the HEAD of `messages` whose `toolCallId`
- * has no matching `toolCall` block in the immediately preceding assistant message.
+ * Drop any `toolResult` messages at the HEAD of `messages`.
  *
- * This happens after compaction slices away the assistant turn that contained
- * the matching `tool_use` blocks. Bedrock (and Claude in strict mode) rejects
- * such orphaned `tool_result` blocks with a 400 error.
+ * A leading `toolResult` is orphaned by definition: there is no preceding
+ * assistant message that contains its `toolCallId`. This arises in two
+ * call sites:
  *
- * The function is intentionally conservative: it only removes leading orphans
- * so the invariant "every toolResult follows an assistant message that contains
- * its toolCallId" holds for the messages the LLM will see after the summary
- * message is prepended.
+ *  1. **Session restore** (`scoop-context.ts`): IndexedDB can persist a
+ *     corrupt session whose first message is a `toolResult` (e.g. a browser
+ *     crash mid-save, or sessions written before the walk-back guard was
+ *     introduced). Without stripping it, Bedrock rejects the next prompt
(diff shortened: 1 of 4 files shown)


karlpauls (approved):
The PR description's retry-classification claim:
  
  > Strip orphaned toolResult blocks that appear at the head of the kept-messages slice after context compaction, **and classify structural tool history errors as non-retryable**.
  > ...
  > This was also matching `isRetryableError`, causing 3 pointless retries with the same broken history.
  
  There's no `isRetryableError` change in the diff (`git diff origin/main` shows only `context-compaction.ts`, `scoop-context.ts`, and the two test files — no retry classifier). If that piece is landing separately, please either link the follow-up PR or amend this PR's description so reviewers don't expect it here.
  
  Otherwise LGTM 🚢

PR #649, opened 2026-05-13
feat(compaction): extract memory on compact + freeze cone session

trieloff:
## Summary

- **Compaction → memory extraction.** When the agent compacts older messages, run a second LLM call sharing the same conversation prefix (embedded in the system prompt so Anthropic's prompt cache hits). Bullets are appended to `/shared/CLAUDE.md` under a dated heading. Cone-only — scoops keep their own user-curated memory. Best-effort: memory failure never blocks compaction.
- **"Clear chat" → "New session".** Before clearing, freeze the cone's chat to `/sessions/<timestamp>-<slug>.json` with an LLM-generated title and prepend an entry to `/sessions/index.json`. Scoops intentionally survive a "New session" so the fresh cone inherits the existing roster. `clear-chat` message now carries `target: 'cone' | 'all'`; the UI button uses `'cone'`.
- **Frozen sessions sidebar (standalone).** New section in the scoops rail rendering the index newest-first with snowflake glyphs. Click → Files tab + `FileBrowserPanel.revealPath` expands ancestors and selects the archive JSON. Extension UX deferred (rail is hidden there).

Inspired by @kpauls' compaction UI brainstorming.

## Test plan

- [ ] `npm run typecheck` ✅
- [ ] `npm run test` ✅ (4249 passing — including new `session-freezer.test.ts` and rewritten `context-compaction.test.ts`)
- [ ] `npm run build` ✅
- [ ] `npm run build -w @slicc/chrome-extension` ✅
- [ ] Manual: stuff context past 200K threshold; confirm `/shared/CLAUDE.md` gains a dated `## Auto-extracted (…, compaction)` block and conversation is summarized.
- [ ] Manual: click "New session" on a non-trivial cone chat; confirm `/sessions/<ts>-<slug>.json` and `/sessions/index.json` appear after reload, scoops are still present and runnable, and sidebar shows the new entry.
- [ ] Manual on Adobe path: both LLM calls carry `X-Session-Id` (check compaction logs / network).
- [ ] Manual on a non-Anthropic provider: both calls complete (no cache hit, but correct).
- [ ] Memory call mocked to throw → compaction still applies summary.

## Out of scope

- Resuming / thawing frozen sessions (the freezer is write-only from the UI).
- Extension-mode sidebar surface.
- Per-scoop auto-memory writes.
- Freezing scoop sessions on "New session" (scoops survive by design).
- Pruning / GC of `/sessions/`.

### DIFF
--- a/packages/chrome-extension/src/messages.ts
+++ b/packages/chrome-extension/src/messages.ts
@@ -245,6 +245,18 @@ export interface ScoopStatusMsg {
 }
 
 /**
+ * Fired by the offscreen agent's compaction transformer as it enters
+ * and leaves the summarize / memory-extract LLM phases. The panel
+ * renders a ghost-bubble affordance while the state is non-idle so the
+ * user knows why the agent is silent. `'idle'` clears the affordance.
+ */
+export interface CompactionStateMsg {
+  type: 'compaction-state';
+  scoopJid: string;
+  state: 'summarizing' | 'extracting-memory' | 'idle';
+}
+
+/**
  * Subset of `ScoopConfig` (see `packages/webapp/src/scoops/types.ts`)
  * carried across the offscreen → panel boundary. The panel only needs
  * the persisted-per-scoop bits that drive the UI affordances (model
@@ -440,6 +452,7 @@ export type OffscreenToPanelMessage =
   | OffscreenReadyMsg
   | AgentEventMsg
   | ScoopStatusMsg
+  | CompactionStateMsg
   | ScoopListMsg
   | StateSnapshotMsg
   | ErrorMsg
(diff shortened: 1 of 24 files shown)


packages/chrome-extension/src/messages.ts
```diff
@@ -78,6 +78,12 @@ export interface RequestScoopMessagesMsg {
 
 export interface ClearChatMsg {
   type: 'clear-chat';
+  /**
+   * Which sessions to clear. `'cone'` (default) wipes only the cone — used
+   * by the "New session" button so scoops survive. `'all'` wipes every
+   * scoop's session as well; reserved for future "reset everything" flows.
+   */
+  target?: 'cone' | 'all';
```
trieloff: Don't add reserved functionality for future scope.

PR #670, opened 2026-05-17
fix(workflow): enqueue cleanup PR into the merge queue

trieloff:
## Why

The `Planning Artifact Cleanup` workflow opens a PR (most recently #652) to remove `docs/superpowers/**` planning artifacts from `main`, then runs `gh pr merge --auto` on it. With a required merge queue on `main`, that command **does not** add the PR to the queue — it only flips the GitHub "Auto-merge" toggle, which then attempts a direct merge and is silently blocked by the `merge_queue` ruleset.

Result: PR #652 has been sitting `MERGEABLE` / `CLEAN`, all checks green, auto-merge on, but never enters the queue. Every subsequent push to `main` opens a new identical cleanup branch that never lands either.

## What

Replace the `gh pr merge --auto` step with one that:

1. Reads the PR's GraphQL node id.
2. Polls `mergeable` + `mergeStateStatus` + `isInMergeQueue` (up to 30 minutes) so we wait for CI to finish on the cleanup branch.
3. Calls the `enqueuePullRequest` GraphQL mutation explicitly — the same operation the "Merge when ready" UI button triggers.

Handles the obvious edge cases:

- already in the queue from a prior run → exit 0
- `DIRTY` / `BLOCKED` merge state → exit 0 (the cleanup PR has a real problem; surface it via the unmerged PR rather than spinning)
- polling timeout → exit 1 so the workflow surfaces the failure

## Verification

- `npx prettier --write` ran clean on the file.
- The change is workflow-only; no app code paths touched.
- After this merges, the next push to `main` that produces planning-artifact changes will exercise the new step end-to-end. PR #652 itself can be unstuck manually once by clicking "Merge when ready" (or by triggering the workflow via `workflow_dispatch` once this is on `main`).

### DIFF
--- a/.github/workflows/planning-artifact-cleanup.yml
+++ b/.github/workflows/planning-artifact-cleanup.yml
@@ -105,15 +105,27 @@ jobs:
               exit 0
             fi
 
+            # mergeStateStatus reference:
+            #   CLEAN / HAS_HOOKS  -> green, ready to enqueue
+            #   UNSTABLE           -> required checks pending or failing; keep waiting
+            #   BLOCKED            -> usually transient while CI is still running on a
+            #                         fresh push (queue + required checks not yet green);
+            #                         keep waiting until the polling timeout
+            #   UNKNOWN            -> GitHub still computing; keep waiting
+            #   DIRTY / BEHIND     -> needs human attention (conflict or rebase)
+            #   DRAFT              -> shouldn't happen for this workflow, but treat as terminal
             case "$merge_state" in
-              CLEAN|UNSTABLE|HAS_HOOKS)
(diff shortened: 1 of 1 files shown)


trieloff:
Addressed in b48618ea:

- **Codex (P1 — UNSTABLE)**: removed UNSTABLE from the break-out case; we now only enqueue on CLEAN or HAS_HOOKS. UNSTABLE keeps polling.
- **Copilot (BLOCKED)**: moved BLOCKED into the retryable bucket. It's almost always transient on a fresh-push cleanup branch (queue + required checks not yet green). DIRTY / BEHIND / DRAFT remain terminal because those need human attention.

PR #671, opened 2026-05-17
fix(swift-server): launch Chrome via LaunchServices so macOS TCC grants camera/mic access

trieloff:
## Why

Camera and microphone access were silently broken in Slicc-launched Chrome — for example, joining a Google Meet would render the camera permission UI, but the actual `navigator.mediaDevices.getUserMedia()` call hung forever.

The root cause was macOS TCC. `packages/swift-server/Sources/Browser/ChromeLauncher.swift` spawned Chrome with Foundation `Process` (raw `posix_spawn`), which left `slicc-server` — and therefore its GUI parent `com.slicc.sliccstart` — in Chrome's TCC responsibility chain. `log show --predicate 'subsystem == "com.apple.TCC"'` against a running Slicc made the smoking gun obvious:

```
tccd: Handling access request to kTCCServiceCamera,
  from Sub:{com.slicc.sliccstart}
  Resp:{TCCDProcess: identifier=com.slicc.sliccstart, pid=27009,
    binary_path=/private/var/folders/.../X/com.google.Chrome.code_sign_clone/.../Google Chrome}
  ReqResult(Auth Right: Unknown (None), promptType: 1, DB Action: None, …)
```

The launcher bundle has no `NS{Camera,Microphone}UsageDescription`, so TCC could neither prompt nor persist a decision, and the request returned `Unknown (None)` with no DB action. Direct spawn also caused macOS to use a `code_sign_clone` of Chrome — distinct from the canonical `/Applications/Google Chrome.app` — so any prior TCC grant the user had for Google Chrome did not apply.

## What

1. **Route Chrome through LaunchServices.** When the resolved executable lives inside a `.app` bundle (the canonical Chrome and Chrome-for-Testing paths both do), launch via:
   ```
   /usr/bin/open -n -a <Google Chrome.app> -W --args <existing Chromium flags>
   ```
   This is the same pattern `ElectronLauncher` already uses. LaunchServices owns the new process, Chrome becomes its own TCC responsible process, and the user's `/Applications/Google Chrome.app` privacy grant applies normally.
2. **Keep a direct-exec fallback** for bare-binary paths so existing tests still cover the path.
3. **Replace stderr CDP-port scraping with HTTP polling** on the LaunchServices path (stderr now belongs to `open`, not Chrome). `waitForCDPReady()` watches the requested port and the `open` exit code so LaunchServices failures fail fast.
4. **Add `NS{Camera,Microphone}UsageDescription` to Sliccstart's Info.plist** so the bundle can legally prompt if/when TCC still attributes work to the launcher (e.g., direct invocations or future code paths).
5. **Extend the eslint Node-globals scope** to `packages/swift-launcher/**` and `packages/swift-server/**` helper scripts so the existing `assemble-app.mjs` no longer fails lint when touched.

## Verification

- `swift build -c debug` on `packages/swift-server` succeeds.
- New unit tests cover `resolveAppBundle(forExecutable:)` and `buildOpenLaunchArgs()`, pinning the `-n -a <bundle> -W --args …` prefix so future refactors cannot accidentally drop `--args` (which would silently swallow Chrome's flags into `open`'s own option parser).
- Existing `ChromeLauncherTests` still pass; `testLaunchFailsFastWhenCdpPortIsAlreadyServingChrome` still exercises the pre-flight probe and never spawns Chrome.
- `npx eslint packages/swift-launcher/assemble-app.mjs` is clean after the config scope extension.
- Local manual repro: with this branch, joining Google Meet in a Sliccstart-launched Chrome no longer hangs in `getUserMedia`; first-run users see the standard macOS Camera/Microphone prompt attributed to Google Chrome.

## Risks / Limitations

- Slicc's existing graceful-shutdown path already kills the `open` parent rather than Chrome itself when CDP `Browser.close` fails (this matches the long-standing Electron path), so killing the launcher does not force-kill Chrome. Out of scope for this PR; the primary CDP-driven shutdown path is unchanged.
- We rely on a fixed CDP port handed off by the existing port-resolution code, which already calls `findAvailablePort()` before launch; we no longer recover from Chrome picking a different port via stderr (it should not, because the port is free). If that ever regresses, `waitForCDPReady` will time out on the wrong port rather than silently succeed.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -94,6 +94,8 @@ export default tseslint.config(
       'packages/node-server/scripts/**/*.{mjs,js,ts}',
       'packages/cloudflare-worker/src/**/*.ts',
       'packages/dev-tools/tools/**/*.{mjs,js,ts}',
+      'packages/swift-launcher/**/*.{mjs,js,ts}',
+      'packages/swift-server/**/*.{mjs,js,ts}',
       'vite.config*.ts',
     ],
     languageOptions: {
--- a/packages/swift-launcher/assemble-app.mjs
+++ b/packages/swift-launcher/assemble-app.mjs
@@ -128,6 +128,10 @@ const infoPlist = `<?xml version="1.0" encoding="UTF-8"?>
     <false/>
     <key>NSSupportsSuddenTermination</key>
     <false/>
+    <key>NSCameraUsageDescription</key>
+    <string>Slicc launches Google Chrome to host the assistant UI. Chrome — not Slicc — uses the camera for sites you visit (Google Meet, Zoom, etc.). Grant access if you want camera-enabled sites to work inside Slicc.</string>
+    <key>NSMicrophoneUsageDescription</key>
+    <string>Slicc launches Google Chrome to host the assistant UI. Chrome — not Slicc — uses the microphone for sites you visit (Google Meet, Zoom, etc.). Grant access if you want microphone-enabled sites to work inside Slicc.</string>
 </dict>
 </plist>
 `;
--- a/packages/swift-server/Sources/Browser/ChromeLauncher.swift
+++ b/packages/swift-server/Sources/Browser/ChromeLauncher.swift
@@ -54,6 +54,14 @@ enum ChromeLauncherError: LocalizedError, Sendable {
     case chromeExitedBeforeReportingPort(Int32)
     case timedOutWaitingForPort(TimeInterval)
     case cdpUnavailable(Int)
+    /// `/usr/bin/open` returned a non-zero exit code while trying to start
+    /// Chrome via LaunchServices. We use `open` so that LaunchServices —
+    /// not slicc-server / Sliccstart — becomes Chrome's parent and TCC
+    /// responsible process; this matters for camera/microphone access on
+    /// macOS, where TCC otherwise blames the closest GUI ancestor for the
+    /// hardware request and silently drops it if the ancestor has no
+    /// `NS{Camera,Microphone}UsageDescription`.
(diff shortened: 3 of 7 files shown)


trieloff:
Addressed the Codex P2 in 623969f.

The LaunchServices spawn path now discovers Chrome's real PID via NSWorkspace after launch and threads it through to `GracefulShutdownHandler.closeBrowser` and the atexit last-resort registry as a separate `browserKillPid`. The SIGKILL fallback prefers `browserKillPid` when present and falls back to `process.processIdentifier` for the direct-exec path (bare-binary CHROME_PATH on test runs / non-darwin), so the kill targets Chrome instead of `open` whenever `Browser.close` over CDP fails.

PID discovery: snapshot `NSWorkspace.shared.runningApplications` filtered by the resolved Chrome `.app` bundle URL before `process.run()`, then poll for a new PID after `waitForCDPReady` succeeds. `-n` in `buildOpenLaunchArgs` always forces a fresh instance so the set-difference is unambiguous. The `runningPidsForBundle` callback is injected so tests can simulate the new-PID arrival without spawning Chrome.

Verification:
- `swift test` (Xcode 6.3.2): 215/215 pass (+5 new — immediate / polled / timed-out PID discovery, and `browserKillPid` preferred vs falls back).
- `swift-coverage-check`: lines 52.37% / functions 51.93% / regions 46.99% (floors 40 / 40 / 35).
- `swift build -c release` clean; rebuilt Sliccstart.app contains `buildOpenLaunchArgs`, `browserKillPid`, `openLaunchFailed`, `/usr/bin/open` symbols, plus the camera/mic NSUsageDescription strings.

PR #672, opened 2026-05-17
fix(node-server): launch Chrome via LaunchServices on macOS to unbreak camera/mic

trieloff:
## Why

Follow-up to #671. PR #671 fixed the Sliccstart-launched Chrome path; this PR applies the same fix to the Node CLI runtime (`npm run dev`, the packaged Electron float, anywhere `packages/node-server` launches Chrome).

`packages/node-server/src/index.ts` spawned Chrome with `child_process.spawn` (a thin `posix_spawn` wrapper). That left whichever terminal launched `node` — `Terminal.app`, `iTerm2.app`, `Code.app`, `Warp.app`, etc. — sitting in Chrome's macOS TCC responsibility chain. On machines where the terminal app had no prior camera/microphone grant (or, for some terminals, no `NS{Camera,Microphone}UsageDescription` at all), `navigator.mediaDevices.getUserMedia()` in Google Meet / Zoom / etc. silently hangs.

Most developers don't hit this because their terminal already has the grants from earlier dev work, but fresh dev machines and CI smoke runs are exposed, and the underlying architecture issue is the same one #671 fixed for Sliccstart. Symptom and macOS `tccd` log shape are identical: `promptType: 1, DB Action: None`, plus a `code_sign_clone` of Chrome under `/var/folders/…/com.google.Chrome.code_sign_clone/…` that doesn't match the user's TCC entry for canonical `/Applications/Google Chrome.app`.

## What

1. **Add `planChromeSpawn()` and `resolveChromeAppBundle()`** in `packages/node-server/src/chrome-launch.ts`. On darwin, when the Chrome executable resolves to a `.app` bundle (both canonical Chrome and Chrome-for-Testing do), return a plan that runs:
   ```
   /usr/bin/open -n -a <Chrome.app> -W --args <existing Chromium flags>
   ```
   LaunchServices owns the new process, Chrome becomes its own TCC responsible process, and the canonical Chrome privacy grant applies.
2. **Keep a direct-exec fallback** on Linux and Windows (no LaunchServices equivalent, no TCC inheritance problem) and for bare-binary paths on darwin.
3. **Switch `index.ts` from `waitForCdpPortFromStderr` to `waitForCdpPort`** so CDP port discovery works in both modes. `waitForCdpPort` already races stderr scraping and `DevToolsActivePort` polling — direct-exec wins via stderr, LaunchServices wins via the active-port file (because `open`, not Chrome, owns our stderr pipe).
4. **Unit tests** pin the `-n -a <bundle> -W --args …` prefix shape so a refactor that drops `--args` (which would swallow Chrome's flags into `open`'s own option parser) fails the build, and cover the Linux / Windows / bare-binary fallback paths.

## Verification

- `npm run typecheck` ✅
- `npx vitest run packages/node-server/tests/chrome-launch.test.ts` ✅ (40/40 pass, +8 new)
- `npm run build -w @slicc/node-server` ✅
- `npx prettier --write` on touched files: clean
- `npx eslint` on touched files: 0 errors (12 pre-existing warnings in the test file, unrelated)
- Manual repro on darwin: `npm run dev` against a Sliccstart-style fresh-terminal-without-camera-grant scenario now shows the standard macOS Camera/Microphone prompt attributed to Google Chrome instead of hanging in `getUserMedia`.

## Risks / Limitations

- Same `open -W` caveat #671 inherited from the existing `ElectronLauncher` flow: SIGTERM/SIGKILL of the `launchedBrowserProcess` targets `open`, not Chrome. The primary shutdown path is CDP `Browser.close` (unchanged), and `open -W` does keep `launchedBrowserProcess.isAlive`/`on('exit')` tracking accurate because it stays alive as long as Chrome stays alive.
- The stderr-vs-active-port-file race is unchanged on Linux/Windows; we just stopped restricting it to stderr-only on macOS.

### DIFF
--- a/packages/node-server/src/chrome-launch.ts
+++ b/packages/node-server/src/chrome-launch.ts
@@ -174,6 +174,76 @@ export function buildChromeLaunchArgs(options: {
   return args;
 }
 
+/**
+ * Walk up from a Chrome executable path
+ * (`…/Foo.app/Contents/MacOS/Foo`) to its enclosing `.app` bundle so we
+ * can hand it to `/usr/bin/open -a`. Returns `null` on non-darwin
+ * platforms or bare-binary paths so the caller falls back to a direct
+ * exec (Linux/Windows have no LaunchServices equivalent).
+ */
+export function resolveChromeAppBundle(
+  executablePath: string,
+  platform: NodeJS.Platform = process.platform
+): string | null {
+  if (platform !== 'darwin') return null;
+  const parts = executablePath.split('/');
+  for (let i = parts.length - 1; i >= 0; i--) {
+    if (parts[i]?.toLowerCase().endsWith('.app')) {
+      return parts.slice(0, i + 1).join('/');
+    }
+  }
+  return null;
+}
+
+export interface ChromeSpawnPlan {
+  command: string;
+  args: string[];
+  /**
+   * `true` when the spawn is routed through `/usr/bin/open` so
+   * LaunchServices owns the new Chrome process. The caller should rely
+   * on `DevToolsActivePort` for CDP port discovery in this mode because
+   * `open`'s stderr never carries Chrome's `DevTools listening on …`
+   * banner.
+   */
+  usesLaunchServices: boolean;
+}
+
+/**
+ * Decide how to spawn Chrome so macOS TCC attributes camera/microphone
+ * requests to Chrome itself rather than to whatever terminal launched
+ * `node`. On darwin, when we can resolve an enclosing `.app` bundle for
+ * the Chrome executable, route the spawn through
+ * `/usr/bin/open -n -a <bundle> -W --args …` so LaunchServices becomes
+ * Chrome's parent and TCC responsible process. Without this hop,
+ * `getUserMedia()` calls in Google Meet, Zoom, etc. hang forever on
+ * machines where the terminal app has never been granted camera/mic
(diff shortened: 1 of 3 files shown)


trieloff:
Addressed the Copilot and Codex P2 feedback in 929dee4.

Added `clearStaleDevToolsActivePort(userDataDir)` and call it from `index.ts` right before spawn. Profile dirs are reused across runs (both the dev `/tmp/browser-coding-agent-chrome` profile and the persistent `.qa/chrome/<profile>` profiles) and Chrome never proactively clears the file on shutdown, so a previous crash / SIGKILL'd run would let the active-port-file poller resolve to a stale port the instant we spawn the next Chrome. ENOENT is the success case; other unlink errors are swallowed so a failure to clear can't block the launch.

This was the same race I hit in the earlier smoke test on this PR (`CDP did not become available on port 50169` when Chrome was actually on 59622). Re-ran the smoke test after the fix:

- Chrome launched via `/usr/bin/open -n -a … -W --args …`
- node-server logged `Chrome CDP listening on port 61310`
- `DevToolsActivePort` file contents matched (61310)
- `curl /json/version` on 61310 returned Chrome/148.0.7778.167
- Zero `CDP did not become available` errors in the log

`vitest` 43/43 pass (+3 new for the helper — deletes existing file, no-op on ENOENT, no-throw on missing directory). `npm run typecheck` clean.

PR #673, opened 2026-05-17
fix(node-server): validate DevToolsActivePort port via /json/version before trusting it

trieloff:
Follow-up to #672. Hardens `waitForCdpPortFromActivePortFile` with a live `/json/version` probe before accepting any port the file claims is Chrome's.

## Why

#672 deletes a stale `DevToolsActivePort` before spawn, which fixes the common case (previous run crashed, profile dir reused, file still says port 50169 even though that's a dead TCP port). But two narrower races slip past:

1. **Unlink loses to Chrome's early write.** `clearStaleDevToolsActivePort` is best-effort; on a fast cold start Chrome can land its own `DevToolsActivePort` before our unlink syscall completes. If Chrome's first published port is itself transient (renderer-process rebind, watchdog restart, whatever you saw in the smoke test where the file flipped 50169 → 59622), we still capture the transient one.
2. **Defense in depth for reused profiles.** Even with the pre-spawn unlink, any future code path that hands `waitForCdpPortFromActivePortFile` a profile dir Chrome hasn't owned yet has the same trust-the-file problem.

Both reviewers on #672 (Copilot and Codex P2) specifically suggested *either* pre-deletion *or* freshness validation. #672 did the deletion; this PR adds the validation.

## What

1. **`probeCdpAlive(port)`** — stdlib-only (`node:http`) single-shot probe of `http://127.0.0.1:<port>/json/version`. Returns `true` only on 2xx + valid JSON + non-empty `webSocketDebuggerUrl` (the CDP fingerprint; won't false-positive on a dev server / health endpoint / IDS scanner that happens to be on the port). Every failure mode (refused, timeout, non-JSON, oversized body) collapses to `false`.
2. **`waitForCdpPortFromActivePortFile`** now calls `probeCdpAlive` on the port it parsed from the file. On `false`, it doesn't resolve — falls back to the next poll tick and re-reads the file, so the freshly-spawned Chrome's overwrite eventually wins. Probe is injectable via `{ verifyPort }` for unit tests.
3. **`waitForCdpPort`** forwards `verifyPort` to the file leg. End-to-end semantics: if the file leg keeps probing a stale port that never goes live, the stderr leg can still win the race (covered by a new test).

## Why a probe instead of file-mtime / write-order tracking

- File-mtime is unreliable on macOS APFS at sub-second resolution and outright wrong inside Docker volumes.
- `fs.watch` works fine but doesn't help with the transient-bind case where Chrome's *own* second write is the right one. "Wait for the next change" sometimes resolves us to the same dead port.
- A live HTTP probe is the only check that actually answers the question we care about: "is there a CDP server accepting connections on this port *right now*?"

## Verification

- `npm run typecheck` clean.
- `npx vitest run packages/node-server/tests/chrome-launch.test.ts`: 50/50 pass (+7 new):
  - probe accepts valid CDP payload
  - probe rejects empty port
  - probe rejects non-CDP HTTP (the false-positive guard)
  - probe rejects non-2xx
  - file leg keeps polling when `verifyPort` returns false, eventually picks up the live overwrite
  - file leg times out when `verifyPort` never returns true
  - end-to-end race forwards `verifyPort` through `waitForCdpPort`
- `npm run build -w @slicc/node-server` clean.
- Manual smoke test: seeded `/tmp/browser-coding-agent-chrome/DevToolsActivePort` with `1\n/devtools/browser/totally-bogus-stale-id` then ran `npm run dev`. Chrome launched via `/usr/bin/open` as expected, node captured the real port (61592), `/json/version` answered, zero `CDP did not become available` entries in the log.

## Risks

- Adds a per-poll HTTP roundtrip (capped at 500 ms timeout, typically <5 ms on localhost). Negligible against the existing 15 s launch budget.
- The probe is HTTP-only, no WebSocket handshake. `/json/version` is the canonical Chromium readiness check and is what `puppeteer` / `playwright` use too.

## Base

This PR is stacked on `fix/macos-tcc-chrome-camera-node-server` (#672) because it depends on `clearStaleDevToolsActivePort`. Once #672 merges to main this PR's base will retarget to main automatically (or I'll rebase if needed).

### DIFF
--- a/packages/node-server/src/chrome-launch.ts
+++ b/packages/node-server/src/chrome-launch.ts
@@ -1,5 +1,6 @@
 import { existsSync, readdirSync } from 'fs';
 import { mkdir, readFile, unlink, writeFile } from 'fs/promises';
+import { request as httpRequest } from 'http';
 import { homedir } from 'os';
 import { dirname, join } from 'path';
 import type { ChildProcess } from 'child_process';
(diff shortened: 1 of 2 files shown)


trieloff:
Pushed `eabbe538` addressing all four review points:

| Reviewer | Severity | Comment | Fix |
| --- | --- | --- | --- |
| Codex | P1 | `waitForCdpPortFromActivePortFile` awaits `verifyPort` without try/catch; an out-of-range port (e.g. `70000`) in DevToolsActivePort makes `probeCdpAlive` reject with `ERR_SOCKET_BAD_PORT` and hangs the file leg. | (a) `probeCdpAlive` now range-checks port `1..65535` before `http.request`, (b) wraps `httpRequest(...)` in try/catch for any other sync throw, (c) `waitForCdpPortFromActivePortFile` wraps `await verifyPort` in try/catch as defense in depth. |
| Copilot | — | `httpRequest` can throw synchronously for invalid options; "all errors collapse to false" contract was violated. | Same fix as above; JSDoc updated to make the failure-collapse contract explicit and enumerate the failure modes. |
| Copilot | — | Port reuse — a stale file pointing at a port now serving an unrelated Chrome/CDP would still pass; compare against the file's line 2 WS path. | New `ProbeCdpAliveOptions.expectedWebSocketPath`. `waitForCdpPortFromActivePortFile` reads line 2 (`/devtools/browser/<uuid>`) and forwards it; the probe parses `webSocketDebuggerUrl` and rejects on path mismatch. |
| Copilot | — | Timeout message doesn't distinguish "file never appeared" from "file appeared but port never answered CDP". | Track `sawCandidate`/`lastCandidatePort`; on timeout emit either `Timed out waiting for DevToolsActivePort at <path>` or `Port N from DevToolsActivePort ... never answered CDP`. |
| Copilot | — | Probe tests don't cover timeout/oversized-body/invalid-port failure paths. | Added vitest cases for: out-of-range port (70000, -1, 0, NaN, 3.14), slow/hanging server, oversized response body (>16 KiB), WS-path match, WS-path mismatch (port-reuse case), poller forwarding the WS path, distinct timeout messages, and synchronously-throwing verifier. 57/57 tests pass (+7 new). |

Verification:
- `npm run typecheck` clean.
- `npx vitest run packages/node-server/tests/chrome-launch.test.ts`: 57/57 pass.
- `npm run test:coverage:node-server`: 68.66/60.19/72.17/70.49 vs 65/55/65/65 thresholds.

Internal signature changes (no public-API impact):
- `probeCdpAlive(port, options?)` instead of `probeCdpAlive(port, timeoutMs?)`.
- `verifyPort` seam now `(port, expectedWebSocketPath: string | null) => Promise<boolean>`.

Issue #682, opened 2026-05-19
Extension-leader follower sync: offscreen.ts `workerBaseUrl` branch never constructs LeaderSyncManager


karlpauls:
## Summary

When a Chrome extension runs as a **tray leader** (the user has set a worker base URL but not a join URL), followers can connect via WebRTC but the leader broadcasts nothing — no chat snapshots, no agent events, no scoops list, no sprinkles, no federated CDP. The data channel opens, then sits silent.

Surfaced as a follow-up to the `feat/browser-follower-sprinkle-sync` work (PR for that branch), which fixed the standalone-leader sync gaps but explicitly left this extension-leader path for a separate PR because the scope is much larger (~150-200 LoC of wiring vs ~10 LoC for the standalone fix).

## Where

`packages/chrome-extension/src/offscreen.ts` — the `if (trayRuntimeConfig?.workerBaseUrl)` branch in `syncTrayRuntime` (currently ~lines 416-457).

## What's missing

Compare with the standalone-leader implementation at `packages/webapp/src/ui/page-leader-tray.ts:startPageLeaderTray`. The extension-leader path is missing every one of these:

1. **No `LeaderSyncManager` constructed.** Standalone does `sync = new LeaderSyncManager(syncOptions)` at `page-leader-tray.ts:160`; the extension branch never builds one.
2. **`onPeerConnected` just logs.** Standalone calls `sync.addFollower(peer.bootstrapId, channel, { runtime, connectedAt })`; the extension branch drops the channel.
3. **No agent-event subscription.** Standalone wires `agentHandle.onEvent → sync.broadcastEvent` (`page-leader-tray.ts:219`); the extension branch never subscribes.
4. **No periodic broadcast interval.** Standalone runs `sync.broadcastScoopsList()` + `sync.broadcastSprinklesList()` every ~5 s; the extension branch has no equivalent.
5. **No `LeaderSyncManagerOptions` wiring at all.** Standalone passes `getMessages` / `getMessagesForScoop` / `getScoopJid` / `getScoops` / `getSprinkles` / `readSprinkleContent` / `onSprinkleLick` / `onFollowerMessage` / `onFollowerAbort` / `browserAPI` / `browserTransport` / `vfs`; none of these are provided in extension mode.
6. **No `sendToSprinkle → broadcastSprinkleUpdate` hook** (the equivalent of the fix from this PR's `7bc7430c`), so the agent's `sprinkle.update` push also wouldn't cross the wire.
[...]

PR #683, opened 2026-05-19
feat: browser-follower sprinkle sync + standalone-leader broadcast fixes

karlpauls:
## Why

The standalone webapp (sliccy.ai) and the Chrome extension can both join a leader as **followers**, but follower-side sync was incomplete: sprinkles never rendered, leader chat input never propagated, agent-pushed sprinkle updates never reached followers. iOS was the only follower with full sprinkle parity, and its `LeaderToFollowerMessage` decoder + view layer were quietly papering over leader-side wiring gaps that affected every follower equally.

This PR closes the standalone-leader half of that story end-to-end and gives the in-browser follower (standalone + extension) the same UX as iOS.

### 1. Follower-side sprinkle sync (the headline feature)

- **`FollowerSyncManager`** gains receive-side handlers for `sprinkles.list` / `sprinkle.content` / `sprinkle.update` and outbound methods `refreshSprinkles` / `fetchSprinkleContent` / `sendSprinkleLick` / `clearSprinkleCache` / `cancelSprinkleFetch`. Chunked-content reassembly + concurrent-fetch dedupe via a waiter list, modeled on iOS `AppState.swift`.
- **`SprinkleFollowerController`** (`packages/webapp/src/ui/sprinkle-follower-controller.ts`) is the page-side counterpart: reconciles the leader's open-state into the local layout, fetches `.shtml` content on demand, renders through the shared `SprinkleRenderer`, bridges `lick` / `stopCone` / `close` back to the leader.
- **`follower-sprinkle-bridge.ts`** in the Chrome extension connects the offscreen-hosted `FollowerSyncManager` to the panel-hosted controller over `chrome.runtime` messages. `PanelFollowerSprinkleProxy` implements the controller's `SprinkleFollowerSync` interface; `connectOffscreenFollowerSprinkleBridge` is the symmetric offscreen-side adapter.

### 2. Standalone-leader broadcast gaps that the work surfaced

All pre-existing `LeaderSyncManager` paths that were tested but never called:

- **`SprinkleManager.sendToSprinkle` broadcast** — agent's live sprinkle pushes only updated the leader's local renderer. New `setSendToSprinkleHook()` setter, wired in `main.ts:mainStandaloneWorker` to `pageLeaderTray.sync.broadcastSprinkleUpdate`.
- **Leader's own chat input** — agent responses streamed live but the prompt that triggered them only reached followers on snapshot refresh. New `ChatPanel.setOnLocalUserMessage()` callback, wired to `broadcastUserMessage`.
- **Follower-A → Follower-B re-broadcast** — multi-follower setups were one-way. `onFollowerMessage` re-broadcasts; sender dedupes its own echo via `sentMessageIds`.
- **Path-scrubbing** — `broadcastUserMessage` now runs leader-local `/tmp/attachment-*` paths through `stripLocalPathsForRemote` before going on the wire.

### 3. Hot-join via settings dialog (CR-9)

The merge from `main` brought a runtime `slicc:tray-join` listener that re-wires the follower without a page reload. As shipped, it omitted the `addSprinkle`/`removeSprinkle` callbacks the boot path passes, so a hot-join silently dropped follower sprinkle sync until reload. Fixed alongside half-state error-UX feedback via a new `slicc:tray-join-failed` CustomEvent and dialog listener.

### 4. Shared error-handling infrastructure

Three pieces of new infrastructure that consolidate patterns drift had introduced across leader/follower/extension paths:

- **`ThrottledErrorTracker`** (`packages/webapp/src/scoops/throttled-error-tracker.ts`) — shared throttle for the periodic CDP target-refresh paths in `page-leader-tray`, `page-follower-tray`, and `chrome-extension/src/offscreen.ts`. Configurable throttle window + recovery debounce; monotonic clock (`performance.now`); `finally`-block state mutations so double-throw failures don't lock the tracker.
- **`LeaderSyncManager.broadcastToAllFollowers`** — consolidates seven previously-unguarded broadcast loops. Each per-follower send is wrapped so a single dead channel doesn't abort the iteration. Per-bootstrapId throttle prevents log flooding when a follower channel is stuck.
- **`dispatchTrayJoinWithFailureFeedback`** (`packages/webapp/src/ui/provider-settings.ts`) — wires a one-shot listener for `slicc:tray-join-failed` with `requestId` filtering (so double-Connect attempts don't cross-bleed errors), detached-element guards, and proactive cancel.

### 5. Generalized disconnect cleanup + safety hooks

- `FollowerSyncManager.close()` is now idempotent and gated by `disconnected`, preventing the channel-close cascade from re-running teardown and spuriously firing reconnect logic after a clean shutdown.
- `rejectPendingRequests` extends the original sprinkle-only cleanup to `tabOpenResolvers`, `fsResolvers`, `cdpChunkBuffers`, and `remoteTransports` — so any caller awaiting one of those past a disconnect rejects cleanly instead of hanging forever.
- `fetchSprinkleContent` gets a per-waiter timeout (default 15s, configurable via `sprinkleFetchTimeoutMs`) so the standalone follower can't deadlock on a stuck leader. Constructor validates the option (`RangeError` on NaN / Infinity / negative).
- `narrowMsg<T>` renamed to `discriminateMsg<T>` so the name matches the contract (discriminator-only check, not full type guard).

### 6. Documentation

- Root `CLAUDE.md` module map now lists `packages/ios-app/`
- New `packages/ios-app/CLAUDE.md` with a **Protocol Mirror Invariant** 5-step checklist for the TS↔Swift wire
- New "Multi-Browser Sync (Tray) Architecture" section in `docs/architecture.md` with a per-message matrix showing which follower implements what
- Tray-sync write-up in `packages/webapp/CLAUDE.md`

## Test coverage

**4506 tests pass** + 3 skipped. Typecheck + builds + prettier all clean.

New / expanded test files:
- `tests/scoops/tray-follower-sync.test.ts` — sprinkle protocol full round-trip, chunking, dedup, cache invalidation, disconnect cleanup symmetry (open-tab + fs request + CDP), fetch timeout, sibling waiter survival
- `tests/scoops/tray-leader-sync.test.ts` — 12 sprinkle tests + path-scrub + multi-follower re-broadcast
- `tests/ui/sprinkle-follower-controller.test.ts` — open-state reconciliation, render-race ordering, dispose
- `tests/ui/sprinkle-manager.test.ts` — `setSendToSprinkleHook` (constructor + setter + detach)
- `tests/ui/chat-panel-local-message.test.ts` — leader chat echo hook contract
- `tests/scoops/throttled-error-tracker.test.ts` — 10 contract tests with clock injection
- `tests/ui/provider-settings-tray-join.test.ts` — dispatcher helper (requestId filter, dismiss-timer cancel, detached element, auto-cleanup)
- `tests/ui/page-leader-tray.test.ts` — throttle integration test pinning the warn→error escalation
- `tests/scoops/tray-webrtc.test.ts` — connect-failure log level + cancel guard
- `chrome-extension/tests/follower-sprinkle-bridge.test.ts` — round-trip fetch + lick, malformed envelopes, timeout, dispose

## Review process

The PR went through 13 rounds of multi-agent review (`/pr-review-toolkit:review-pr` running code-reviewer / pr-test-analyzer / silent-failure-hunter / type-design-analyzer / comment-analyzer in parallel). Each round triggered a fix-sweep commit. Notable findings across rounds:

- **R1–R5** (pre-merge): chunked-content cache poisoning, render-race order inversion, type-design corrections (`cancelSprinkleFetch` made required), `cacheEpoch` mechanism for the cache-write-races-list race, exception isolation in hook callers, path scrubbing on user attachments
- **R6–R8**: hook-wiring attribution corrections, fetch-timeout safety, disconnect-cleanup symmetry (R7 generalized the R4 sprinkle-only cleanup to all four pending-request maps), prod log-level escalations (`warn` → `error` where prod gate would otherwise suppress real failures), `close()` idempotency gate
- **R9** (post-merge): hot-join sprinkle wiring (CR-9), `slicc:tray-join` handler hardening, observability log-level fixes
- **R10–R12**: teardown ordering, half-state UI feedback via CustomEvent, recovery-log MTTR signal, `ThrottledErrorTracker` extraction (with `-Infinity` initial-state fix, recovery debounce against flapping), broadcast-loop guard, per-follower error throttle
- **R13** (final): `finally`-block state resets in `ThrottledErrorTracker` for double-throw resilience, transition-audit log levels, `clearTimeout` in dispatcher `cancel()`

## Known follow-ups (out of scope)

- **Extension-leader missing `LeaderSyncManager`** — filed as #682. The extension's `offscreen.ts:syncTrayRuntime` `workerBaseUrl` branch creates `LeaderTrayManager` + `LeaderTrayPeerManager` but never a `LeaderSyncManager`, so when the extension acts as a *leader*, followers can connect but the leader broadcasts nothing. Pre-existing, needs panel↔offscreen RPC layer extension similar to the existing `sprinkle-proxy.ts`. Both follow-up sprinkle/chat broadcast fixes in this PR will trivially port to extension-leader once that path is built out.
- **`sprinkle.cancel` wire envelope** — when a follower cancels an in-flight fetch (timeout or controller dispose), the leader keeps chunking until done; the follower silently drops the late chunks. Not a correctness bug, just wasted bandwidth on cancellation. Would need a Swift mirror update on the iOS side.

## Test plan

- [ ] Standalone leader ←→ standalone follower: chat syncs bidirectionally, sprinkles render on follower, agent-pushed sprinkle updates propagate live, leader's own chat shows on follower without refresh, attachments don't leak `/tmp/attachment-*` paths
- [ ] Hot-join via settings dialog ("Connect" button): chat AND sprinkles sync immediately without reload; an invalid join URL surfaces an error message in the dialog rather than silently dismissing
- [ ] Multi-follower (3+ browsers): follower A's chat reaches follower B and C; sender doesn't double-display
- [ ] Standalone leader ←→ Chrome extension follower (extension paneled, paste join URL via account dialog): same checks as above through the panel↔offscreen bridge
- [ ] iOS follower against standalone leader: unchanged (sanity check — iOS code path was the reference)
- [ ] Disconnect mid-fetch: a `fetchSprinkleContent` call rejects cleanly within 15s if the leader stops responding, and the controller's `opening` lock releases


trieloff:
Great PR

PR #729, opened 2026-05-21
fix: wire imgcat media preview in kernel-worker terminal mode

### LINKED ISSUE #731: imgcat is dead in the panel terminal (both floats) — build the kernel-transport renderer or deprecate
## Background

After the inline-orchestrator removal (kernel-worker is now the default for standalone too), every panel-terminal command runs inside `WasmShellHeadless`. `imgcat` lives in `packages/webapp/src/shell/supplemental-commands/imgcat-command.ts` and fails immediately because:

1. `imgcat-command.ts:34` guards on `typeof window === 'undefined'` and bails with `imgcat: browser APIs are unavailable in this environment`.
2. Even if you removed that guard, `WasmShellHeadless.renderMediaPreview` (`wasm-shell-headless.ts:405`) throws `terminal preview is unavailable in headless mode`. The page-side `WasmShell` subclass overrides it with a real DOM renderer, but the panel terminal never uses that subclass anymore.

Repro (both standalone and extension):

```
/ $ imgcat /tmp/test.png
imgcat: browser APIs are unavailable in this environment
```

PR #729 sidesteps this by switching the file-browser's media button from `imgcat` to `open`, which routes through `panelRpc.call('window-open', ...)` and opens a preview tab. That fixes the user affordance but leaves `imgcat` itself broken everywhere a user would type it.

## What the existing JSDoc already promises

`wasm-shell-headless.ts:401-403`:

> *The terminal RPC will add a third implementation that emits a `terminal-media-preview` envelope over the kernel transport.*

That third implementation was never built. The headless shell should emit a `terminal-media-preview` envelope (mirroring how `terminal-session-host` already streams output back to the panel) so the page-side terminal renderer can hydrate the bytes inline.

## Options

1. **Build the planned envelope.** Add a `terminal-media-preview` message type to the kernel transport, wire `WasmShellHeadless.renderMediaPreview` to emit it, and let the page-side `RemoteTerminalView` invoke the same DOM renderer `WasmShell.renderMediaPreview` already has. Drop the `typeof window` guard in `imgcat-command.ts` once `renderMediaPreview` actually works headless.
2. **Deprecate `imgcat` for the panel surface** and document `open <path>` as the supported affordance. Keep `imgcat` only for the agent's headless shell where it can still render bytes inline into the agent's view (if that path even still works — worth checking).

Option 1 preserves inline rendering, which is genuinely better UX than popping a new tab for every preview. Option 2 is the cheap path.

## Also clean up while we're in there

- `terminal-panel.ts:204` still says `Run imgcat to preview media here.` — misleading after #729 lands (and stale today in any case).
- `shell/terminal-protocol.ts:150` references `imgcat` as a worker-side command that produces inline media; the comment will need updating to reflect whichever option above is chosen.
- `help-command.ts:65` lists `imgcat` under `Browser & UI` — keep or drop alongside the fix.

## Test coverage

There is no test pinning the headless behavior of `imgcat` today (`tests/shell/supplemental-commands/imgcat-command.test.ts` doesn't exist). Whichever option is chosen, add a test that locks in the behavior so this doesn't regress silently again.

## Related

- PR #729 — file-browser switches to `open` (workaround; doesn't fix `imgcat` itself)
- `packages/webapp/CLAUDE.md` kernel-host section — describes the kernel-worker-as-default switch that turned `imgcat` into dead code in the panel terminal

froesef:
## Summary
- `imgcat` in the panel terminal (kernel-worker mode) threw "terminal preview is unavailable in headless mode" because the worker-side `WasmShellHeadless` had no `renderMediaPreview` override
- Added `PanelTerminalShell` subclass that overrides `renderMediaPreview` to emit `terminal-media-preview` messages over the kernel transport
- `RemoteTerminalView` now handles `terminal-media-preview` events and renders image/video previews in the panel
- Wired the preview host into `mountRemoteShell` so the preview tab works in kernel-worker mode

Fixes: #731 

## Test plan
- [ ] Run `npm run dev` (standalone mode)
- [ ] Open the Files panel, find an image file
- [ ] Click the CAT button on the image — should open in the preview tab without errors
- [ ] Run `imgcat /path/to/image.png` in the terminal — same result
- [ ] Verify video files also preview correctly


karlpauls:
Thanks, I think that is a possible way forward, however, the real issue is that imgcat is not working. @trieloff, wdyt - go with this PR or fix #731 or both?

trieloff:
I'd say fix the imgcat implementation.

froesef:
I fixed imgcat. It's displaying the image now in the preview tab:
image

karlpauls:
Cool, can we add some tests maybe?

PR #742, opened 2026-05-22
fix(playwright-cli): bridge follower tabs and teleport to worker via panel-RPC

vdua:
## Summary

`playwright-cli tab-list` silently returned only local CDP tabs when a tray was connected — follower tabs from other browsers were invisible. `teleport --list` and auto-teleport always errored "not connected to a tray" even with followers online. Both broke because the worker-side `BrowserAPI` has no `trayTargetProvider` set (page-side only), and the module-level setter hooks (`setPlaywrightTeleportBestFollower` / `setPlaywrightTeleportConnectedFollowers`) were imported in `main.ts` but never called.

## Why

The worker/page split in standalone mode means the agent's shell commands run in a `DedicatedWorker` with a separate `BrowserAPI` instance from the page. `trayTargetProvider` and the live `LeaderSyncManager` follower state both live page-side (forced there by `RTCDataChannel` non-transferability). The existing panel-RPC bridge (`BroadcastChannel`-based) already handles similar worker→page calls (`tray-reset`, `tray-leave`, `oauth-extras-set`) — this PR follows the same pattern.

**Repro:**
1. `npm run dev`, start a tray session (`host`), connect a second browser as follower
2. Run `playwright-cli tab-list` — only local tabs shown (follower tabs missing)
3. Run `playwright-cli teleport --list` — "not connected to a tray"

Two new panel-RPC ops:

- **`list-remote-targets`**: page handler calls `browser.listAllTargets()` on the page-side `BrowserAPI` (fully wired with `trayTargetProvider`), filters to composite targetIds (`runtimeId:localTargetId`), returns them. `getActionablePages()` always calls this and deduplicates by `targetId`. This is intentionally unconditional: today the worker-side `BrowserAPI` in `kernel/host.ts` never has `setTrayTargetProvider` called on it, so without the RPC supplement `listAllTargets()` silently returns local-only. Deduplication handles the hypothetical future case where the worker instance also gets wired — the RPC result would overlap with local results, and the dedupe discards the overlap rather than double-listing follower tabs.

- **`list-tray-followers`**: page handler reads `pageLeaderTray?.sync.getConnectedFollowers()` + `getBestFollowerForTeleport()`. Used by `teleport --list` and auto-teleport runtime selection.

The dead setter/getter mechanism (`setPlaywrightTeleportBestFollower` / `setPlaywrightTeleportConnectedFollowers`) is removed — it was never called in standalone, so the RPC path is now the sole path. The original `if (getter) { use getter } else { use RPC }` dual-path is collapsed to RPC-only.

The previous version had a heuristic in `getActionablePages()`: skip the RPC supplement if any composite (`runtimeId:localTargetId`) ID was already present in the local `listAllTargets()` result, on the assumption that the worker was already returning remote targets. This heuristic is removed for two reasons. First, stale composite IDs cannot appear in the page-side result: `LeaderSyncManager.removeFollower()` calls `registry.removeRuntime(runtimeId)` on disconnect, and `getConnectedEntries()` further filters to bootstrapIds still in `this.followers`, so the page-side `list-remote-targets` result is always current. Second, the heuristic was checking the worker's local result (not the page result), which can contain a composite ID from a previous remote-tab attach even after the follower disconnects — so the heuristic could skip the RPC supplement and silently drop all current follower tabs (re-introducing the original bug). Deduplication by `targetId` is kept as a correctness net for the future-wired-worker case only.

## Cross-cutting impact

- **Floats exercised**: Standalone CLI only. Extension float's offscreen document has a DOM and the same `BrowserAPI` instance the tray wires — `getPanelRpcClient()` returns null there, all new code paths are skipped. No extension regression.
- **Shell contexts**: Only the worker shell (`playwright-cli`). Panel terminal and side-panel shell unaffected.
- **Other callers of `getActionablePages`**: internal to `playwright-command.ts`. The new RPC call adds ≤3ms on the hot path when no follower is connected (page handler returns `[]` immediately); 3s timeout only fires if the page handler hangs.
- **Persisted state**: none — both ops are pure reads of in-memory state.
- `FloatType` import in `panel-rpc.ts` noted with a comment flagging potential future cleanup into a `wire-types.ts` if the domain-type list grows.

## Test plan

- [x] `npx prettier --write <changed-files>` — passed
- [x] `npm run test` — 198 passing in affected suites, no new failures
- [x] **New tests**: `panel-rpc-handlers.test.ts` covers `list-tray-followers` (no callback → empty defaults, with callback → forwarded result) and `list-remote-targets` (no callback → empty, composite-ID filter). `playwright-command.test.ts` adds merge+dedupe tests for `getActionablePages` and updates the teleport suite to mock `getPanelRpcClient` (replacing the deleted setter pattern).
- [x] Manual smoke (CLI): requires a two-browser tray setup — repro steps above
- [ ] N/A — extension, Electron, Sliccstart paths unaffected

**Pre-existing failures**: `lick-ws-bridge.test.ts` and `websocat-command.test.ts` fail with `CloseEvent is not defined` on `main` at `777ffa94` — unrelated to this PR.

## Risk & rollback

- Blast radius: standalone mode only, `playwright-cli` shell commands only.
- Rollback: revert this PR cleanly — no persisted state or schema changes.

## Checklist


karlpauls:
I like the idea - why not for the extension?

trieloff:
@vdua are you still working on this one?

vdua:
@trieloff Unfortunately couldn't work on this last week. I have a little bandwidth this week. Will pick it up tomorrow as it is blocking some of my work for which I use slicc.

trieloff:
No problem. Just merge when ready.

vdua:
closing this PR for #831 as the conflicts resolution increased the scope of the PR (teleport, first broke it then fixed it and then broke it 3 merge commits). It was better to start off with a fresh PR

PR #761, opened 2026-05-26
docs(context-compaction): document estimateTokens accounting + regression tests

trieloff:
## Summary

Receipt-matcher diagnosis of the scoop-compaction loop (**~35 MB / ~8.85M estimated tokens from 10 inlined JPEGs in a single turn**) initially raised the suspicion that `estimateTokens` in `@earendil-works/pi-coding-agent` was failing to count tool-result text payloads — which would have explained why compaction never triggered until well after the context was lost.

Verified directly against the un-mocked estimator in `node_modules/@earendil-works/pi-coding-agent/dist/core/compaction/compaction.js`: tool-result text payloads **are** counted (including the multi-megabyte blobs `open --view` could emit before the sibling fix). So no wrapper is needed — but the assumption is load-bearing and should not drift silently.

## Change

- Adds a ~14-line comment at the import site of `context-compaction.ts` documenting exactly what `estimateTokens` accounts for, per message role, and where the source of truth lives.
- Adds `packages/webapp/tests/core/context-compaction-real-estimator.test.ts` — a regression guard that exercises the un-mocked estimator end-to-end so any upstream behavior change in pi-coding-agent will surface as a failing test.

No production-code behavior changes.

## Verification

```
npm run typecheck
npx vitest run packages/webapp/tests/core/context-compaction-real-estimator.test.ts
# → 4 passed
```

## Related

- `fix/open-view-size` — fixes the root cause that produced the 8.85M-token payload in the first place (independent).
- `fix/compaction-hopeless-branch` — adds a short-circuit for the resulting compaction loop. Both this PR and that one touch the import-site comment with identical text; merge order does not matter, but landing this one first keeps the hopeless-branch PR's diff focused on the behavior change.


trieloff:

PR #762, opened 2026-05-26
fix(compaction): short-circuit hopeless contexts with elision

trieloff:
## Summary

Receipt-matcher diagnosis of the scoop-compaction loop measured **~35 MB / ~8.85M estimated tokens from 10 inlined JPEGs in a single scoop turn**. Two failure modes followed:

1. The token total was already so far over the context window that even running the LLM summarization call was hopeless — serializing the conversation into the summary prompt itself blew context.
2. The naive-drop fallback repeated forever, because the same oversized tool result kept landing inside the keep-recent window on every loop.

## Change

Adds a "hopeless" branch ahead of the existing compaction path in `createCompactContext`:

- New config field `hopelessMultiplier` (default **4**). When `totalTokens > contextWindow * hopelessMultiplier`, take the hopeless branch.
- `elideHopelessMessages` rewrites any single `toolResult` or assistant message above `(contextWindow - reserveTokens) / 2` to a short stub.
  - Preserves role, `toolCallId`, `toolName`.
  - For assistant messages, keeps `toolCall` content blocks intact so the agent loop's tool-call/tool-result pairing invariant still holds.
- If the elided message list is now below the soft compaction threshold, return immediately — no LLM call, no naive drop.
- Otherwise fall through to the existing naive-drop path with the elided messages, with the LLM block skipped (we already know it's hopeless).

41 tests, covering: threshold boundary, elision of toolResult, elision of assistant + toolCall pair, role/id preservation, byte accounting, and the early-return short-circuit.

## Verification

```
npm run typecheck
npx vitest run packages/webapp/tests/core/context-compaction.test.ts
# → 41 passed
```

## Related / merge order

- `fix/open-view-size` — the root cause that produced the 8.85M-token payload. Independent.
- `chore/context-compaction-estimator-docs` — documents the import-site assumption that tool-result text is counted by `estimateTokens`. This PR also includes that comment hunk (identical bytes), so the two PRs can land in either order with no conflict; the docs PR is the cleaner standalone read and ideally lands first.

This PR's base is `main` so it can be merged independently of the docs PR.

### DIFF
--- a/packages/webapp/src/core/context-compaction.ts
+++ b/packages/webapp/src/core/context-compaction.ts
@@ -96,6 +96,14 @@ export interface CompactionConfig {
   reserveTokens?: number;
   keepRecentTokens?: number;
   /**
+   * Multiplier of `contextWindow` above which compaction is treated as
+   * hopeless: the LLM summary call is skipped entirely because passing
+   * the conversation as a prompt would itself blow context. Oversized
+   * tool results / assistant tool-call payloads are elided in-place
+   * before any naive drop. Defaults to 4.
+   */
+  hopelessMultiplier?: number;
+  /**
    * HTTP headers forwarded to the LLM provider for the summarization and
    * memory-extraction requests. Used by the Adobe LLM proxy path to attach
    * `X-Session-Id` so compaction calls land in the same session as the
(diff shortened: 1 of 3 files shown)


trieloff:

PR #769, opened 2026-05-27
chore: remove planning artifacts from main

trieloff:
Automated follow-up after a merge-queue landing to remove planning artifacts that should stay visible in the original PR but not remain on `main`.


trieloff:

PR #770, opened 2026-05-27
Bound auto-extracted memory and redirect to /workspace/CLAUDE.md

trieloff:
Fixes two bugs in the auto-memory system:

## Bug 1: Wrong file

Auto-extracted memory from compaction and the New-session freezer was writing to `/shared/CLAUDE.md` — the agent-facing runtime instructions bundled from `packages/vfs-root/shared/CLAUDE.md`. It should write to `/workspace/CLAUDE.md`, the cone's private memory file.

Fix: new `Orchestrator.appendConeMemory` sink targeting `/workspace/CLAUDE.md`, used by both the compaction `onMemoryUpdates` wiring and `session-freezer.ts` (both its synchronous path and the boot-time pending-enrichment scan). The `update_global_memory` tool still targets `/shared/CLAUDE.md` — that's the explicit-edit surface and stays unchanged.

One-time **surgical** migration drops only `## Auto-extracted` regions out of `/shared/CLAUDE.md` and preserves the rest of the file verbatim (user-authored headers / footers / hand edits). No bundled-default reseed. Sentinel-guarded via `/workspace/.cone-memory-migrated`.

## Bug 2: Unbounded growth

Every freeze + every compaction appended a fresh `## Auto-extracted` block forever, and the whole file entered every cone turn's system prompt.

Fix: new `packages/webapp/src/scoops/cone-memory-budget.ts` module with:

- `computeBudget(N) = 4000 + 2000 * log2(N + 2)` chars, where `N` is the count of frozen sessions read from `/sessions/index.json`.
- `restructureConeMemory(...)` — when an append pushes the file past `budget * 1.25`, an LLM call consolidates the `## Auto-extracted` tail only. The user-authored header above the first `## Auto-extracted` heading is preserved verbatim.
- Best-effort: restructure failure logs and leaves the appended file in place for the next append to retry.

All writes to `/workspace/CLAUDE.md` are serialized via `memoryWriteChain` on the orchestrator so concurrent freezes (e.g. compaction + New-session race) can't lose blocks.

## Tests & gates

- 23 new test cases (`packages/webapp/tests/scoops/cone-memory-budget.test.ts` + migration & concurrency tests in `orchestrator.test.ts`).
- `npm run typecheck` ✓
- `npm run test -w @slicc/webapp` ✓ 4505/4505
- `prettier --check` ✓ on all touched files

## Non-blocking follow-ups

1. `cone-memory-budget.ts` re-implements the `completeSimple` call inline instead of importing `runOneOffCompactionCall` from `context-compaction.ts`. Same guarantees today (X-Session-Id, provider plumbing, headers all forwarded), but the two call sites can drift — small refactor candidate.
2. No post-call size enforcement on the restructure output — relies on the next append to re-trigger if the LLM exceeded budget. Matches the "best-effort" spec language.

---


trieloff:

PR #783, opened 2026-05-28
feat(shell): add biome linter/formatter as fourth WASM build tool

trieloff:
Follow-up to #781. Adds `biome` as the fourth member of the `Build tools` category, alongside `tsc`, `esbuild`, and `test`.

## Architecture

Mirrors the `esbuild` shape:

- **`biome-runtime.ts`** — lazy Biome loader, memoized per realm.
  - **Browser** (standalone + extension): fetches `biome_wasm_bg.wasm` (~6 MB) from a versioned `unpkg.com/@biomejs/wasm-web@<version>` URL, caches via the Cache Storage API, compiles to a `WebAssembly.Module`, hands it to the wasm-bindgen `__wbg_init`, then constructs `@biomejs/js-api/web`. Passing a compiled `WebAssembly.Module` (vs. a fetch URL) sidesteps the blob-URL / extension-origin CSP differences the same way `esbuild-wasm.ts` does.
  - **Node / vitest**: `@biomejs/wasm-nodejs` loads the wasm synchronously via `fs.readFileSync`, so `new Biome()` works the moment the dynamic import resolves.
  - Failed init clears the cached promise so a retry re-attempts the load instead of poisoning the session.
  - CDN URL is pinned to the installed `@biomejs/wasm-web` `package.json` `version` field, so Renovate bumps roll the wasm asset URL in lockstep.

- **`biome-command.ts`** — subcommands matching the upstream Biome CLI as closely as the Workspace API allows:
  - `biome lint   [paths...]` — diagnostics only
  - `biome format [paths...]` — print formatted output (`--write` to persist)
  - `biome check  [paths...]` — lint + format together
  - `biome ci     [paths...]` — strict mode, exits non-zero on any warning too
  - Mutating flags: `--write`, `--apply` (→ `safeFixes`), `--apply-unsafe` (→ `safeAndUnsafeFixes`)
  - Stdin: `--stdin-file-path <path>` so the workspace picks the right parser
  - Path arguments may be files or directories; directories are walked recursively and filtered to the known JS/TS/JSON/CSS/GraphQL/HTML/Svelte/Vue/Astro extensions (`node_modules`, `.git*` skipped).

## Tests

- 13 pure-logic tests always run: argv parsing, lintable-file detection, path expansion, dispatch (help text), unknown-option and flag-shaped-next-token rejection.
- 3 live WASM smoke tests gated behind `SLICC_TEST_HEAVY_WASM=1`: `format --write` rewrites a TS file, `lint` of clean code exits 0, `--version` returns a semver.

```
SLICC_TEST_HEAVY_WASM=1 npx vitest run packages/webapp/tests/shell/supplemental-commands/biome-command.test.ts
# Tests  16 passed (16)
```

## Verification

- ✅ Prettier, typecheck, `vitest run --project webapp` (4678 tests pass), `build -w @slicc/webapp`, `build -w @slicc/chrome-extension`.
- ✅ `commands` lists `biome` under `Build tools` alongside `tsc`, `esbuild`, `test`.

## Notes

Based on `js-bundler-wasm-tools` (the previous PR). Merge that one first; this one will rebase cleanly onto `main` once #781 lands.

---


trieloff:
@codex review

Follow-up commit `031ede57` addresses both prior P2 findings:

- `check` / `ci` without `--write` now emits a `<path>: file is not formatted` diagnostic and increments the error (or warning, for `ci`) count when `formatContent` would change a file.
- `renderDiagnostics` now counts both `'warn'` and `'warning'` severity strings; Biome's WASM build emits `warn`, so warning-level diagnostics no longer silently pass `ci`.

Added two heavy-WASM integration tests verifying both subcommands exit 1 on unformatted input without `--write` and do not mutate the source.

karlpauls:
Heads up — this PR is the source of the current `release` failure (`wrangler deploy --dry-run` → `Asset too large`).

wasm-bindgen's `new URL('biome_wasm_bg.wasm', import.meta.url)` fallback in `@biomejs/wasm-web` makes Vite statically emit the **33 MB** `biome_wasm_bg.wasm` into `dist/ui/`, which is over Cloudflare's 25 MiB per-asset cap:

```
✘ [ERROR] Asset too large.
  …biome_wasm_bg-DQn8Ios_.wasm with a size of 32.8 MiB.
```

`biome-runtime.ts` always fetches that wasm from the CDN and passes a precompiled module, so the bundled copy is dead weight (it also silently added 33 MB to the extension build).

Fix in flight: #787 — strips the dead asset in both the webapp and extension builds and repoints the reference at the CDN, plus adds a `wrangler deploy --dry-run` gate to the `cloudflare-worker` job so this fails the PR next time instead of only the release.

PR #791, opened 2026-05-29
feat(webapp): add OpenAI Codex (ChatGPT subscription) provider

trieloff:
## Summary

Adds an external OAuth provider that streams **GPT-5 Codex** models through a user's **ChatGPT Plus/Pro/Business/Team** subscription — no API key required.

- **Auth**: reuses the public Codex-CLI OAuth client via slicc's intercepted-redirect flow (CDP `Fetch.requestPaused`), the browser equivalent of pi-ai's `:1455` loopback server. PKCE via Web Crypto, token exchange/refresh against `auth.openai.com`, silent renew, and 60s pre-expiry refresh.
- **Streaming**: delegates to pi-ai's `openai-codex-responses` API, forced to `sse` (browser `WebSocket` can't set the required auth headers).
- **Models**: the 6-model Codex catalog (`gpt-5.2` … `gpt-5.5`), default `gpt-5.3-codex`.
- **Account identity**: label reads `email (Plan)` (e.g. `lars@example.com (Team)`) instead of the raw account UUID.
- **Avatar**: the header avatar now shows the user's **Gravatar** (keyed on the SHA-256 of the verified email, `d=404`), gracefully falling back to email-aware initials. OpenAI's own endpoints don't expose a usable picture (`userinfo` returns only `sub`; `chatgpt.com/backend-api/me` is Cloudflare-gated), so Gravatar is the portable, email-keyed source.

## Avatar initials fix

`initialsFromLabel()` (new, `packages/webapp/src/ui/avatar-initials.ts`) strips a trailing `(Plan)` suffix and derives initials from emails (local part + next segment/domain), so `lars@trieloff.net (Team)` becomes **LT** instead of the previous broken `L(`. `Layout.getInitials` now delegates to it.

## Tests

- `tests/ui/avatar-initials.test.ts` — initials derivation (email, dotted local part, plain names, single token, blanks).
- `tests/providers/openai-codex.test.ts` — config + 6-model catalog, PKCE authorize URL + code exchange + account persistence, Gravatar `userAvatar`, and OAuth state-mismatch rejection.

## Verification

- `npm run typecheck` — pass
- provider + ui suites — **211 tests pass**
- `npm run build -w @slicc/webapp` — pass
- prettier/eslint — clean (pre-commit hooks ran)

Verified end-to-end over CDP: live ChatGPT Team login captured, a real Codex stream returned output, and the header avatar renders the user's Gravatar (falls back to `LT` initials when no Gravatar exists).

### DIFF
--- a/packages/webapp/providers/openai-codex.ts
+++ b/packages/webapp/providers/openai-codex.ts
@@ -73,16 +73,15 @@ const CODEX_API: Api = `${PROVIDER_ID}-openai` as Api;
 // ── Models ─────────────────────────────────────────────────────────
 //
 // Mirrors the `openai-codex` catalog shipped in pi-ai's
-// `models.generated`. `thinkingLevelMap` is forwarded as an extra key —
-// the metadata layer in `provider-settings.ts` spreads unknown keys onto
-// the model record, and the codex stream reads `model.thinkingLevelMap`
-// to map reasoning effort.
+// `models.generated`. `thinkingLevelMap` is a `ModelMetadata` field that
+// `applyModelMetadata` (in `provider-settings.ts`) merges onto the model
+// record — the codex stream reads `model.thinkingLevelMap` to translate
+// the reasoning effort, so the `minimal` → `low` remap below must survive
+// the metadata merge (the chatgpt Codex backend rejects `minimal`).
 
-type CodexModelDef = { id: string; name: string } & ModelMetadata & {
-    thinkingLevelMap?: Record<string, string>;
-  };
+type CodexModelDef = { id: string; name: string } & ModelMetadata;
 
-const CODEX_THINKING_LEVEL_MAP: Record<string, string> = { xhigh: 'xhigh', minimal: 'low' };
+const CODEX_THINKING_LEVEL_MAP: Record<string, string | null> = { xhigh: 'xhigh', minimal: 'low' };
 
 const CODEX_MODELS: CodexModelDef[] = [
   { id: 'gpt-5.5', name: 'GPT-5.5', input: ['text', 'image'] },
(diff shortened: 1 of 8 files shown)


trieloff:
Addressed the review feedback in `07ee31a8`:

**1. Preserve Codex reasoning-level mapping (Codex P2 / Copilot)** — Root cause confirmed: `applyModelMetadata()` copied only a fixed field set and dropped `thinkingLevelMap`, so the resolved `gpt-5.3-codex` inherited pi-ai's base map `{off:null, xhigh:'xhigh'}` and our `minimal → low` remap was lost (verified live). Fix: added `thinkingLevelMap` to `ModelMetadata` and `applyModelMetadata` now merges it per-level over the base map (same pattern as `compat`). Live check now returns `{off:null, xhigh:'xhigh', minimal:'low'}`. This also fixes the same latent drop for the `xai-grok` provider, which relies on the field too.

**2. Require OAuth state on the callback (Codex P2 / Copilot)** — Changed `if (returnedState && returnedState !== state)` to strict `if (returnedState !== state)`, so a callback that omits `state` is rejected exactly like a mismatch, matching the Codex CLI's strict-equality CSRF binding. Added a test for the missing-`state` case.

**3. JWT base64url decoding (Copilot, both occurrences)** — Replaced the two `atob(parts[1])` calls with a single `decodeJwtPayload()` helper that restores the base64url alphabet (`-`/`_` → `+`/`/`) and pads before `atob`. The login test now encodes the fixture token as real base64url (with a marker that forces `+`/`/` → `-`/`_`) to exercise the path.

Also: default model is now **GPT-5.5** (per follow-up request).

Verification: `npm run typecheck`, webapp build, and the provider + ui suites (**1094 tests**) all pass; new/updated tests cover each fix.

PR #805, opened 2026-05-29
feat(OAuth): add OAuth logout UI with account switching support

bstopp:
## Summary

Adds a **Logout** button to OAuth account rows in the provider settings panel (between Edit and Delete). Clicking Logout revokes the token at the IdP, attempts to clear the IdP browser session, and leaves the row in a `loggedOut` state with a **Login** button that re-opens a fresh OAuth flow. Delete and the avatar "Sign out" action both flow through the same logout sequence. This closes the account-switching gap where deleting a GitHub/Adobe account and reconnecting would silently re-authorize the same user.

## Why

Users with multiple accounts for a single OAuth provider (e.g. two GitHub accounts) had no way to switch: the only option was to delete the local account entry, but reconnecting would land back on the same account because the IdP browser session was never cleared. There was also no logout-without-delete action anywhere in the UI.

**Layering:**
- `providers/types.ts` — `Account.loggedOut?: boolean`; `ProviderConfig.getOAuthLogoutUrl?`; `OAuthLoginOptions.forceReauth?`
- `providers/oauth-service.ts` — `openIdpLogoutUrl()` (opens IdP logout URL in popup, waits for user to close it); `launchOAuthCli` gains `popup.closed` polling so cancelling the OAuth popup resolves immediately instead of waiting 2 minutes
- `provider-settings.ts` — `logoutOAuthAccount()` orchestrates revocation → IdP session clear → local token wipe → `loggedOut: true`; `removeAccount` calls it first for OAuth providers; `renderAccountsList` renders the toggle button
- `layout.ts` — avatar "Sign out" now calls `logoutOAuthAccount` for OAuth accounts (keeps the row) vs `removeAccount` for API-key accounts (removes the row)
- `providers/github.ts` — `getOAuthLogoutUrl` returns `https://github.com/logout`
- `providers/adobe.ts` — no `getOAuthLogoutUrl` (IMS logout requires POST body, can't be driven via popup URL); uses `prompt=login` on the authorize URL when `forceReauth: true` instead

**Non-obvious design choices:**

- **IdP logout at logout time, not login time.** The `getOAuthLogoutUrl` popup opens when the user clicks Logout, not when they click Login. This means by the time they hit Login the session is already gone. The alternative (open at login time) was rejected because it makes the Login action feel heavier.

- **Wait for user to close the IdP logout popup, don't auto-close.** GitHub's `/logout` is a human-facing page that requires clicking "Sign out" — auto-closing after N seconds would leave the session intact. The popup waits for `popup.closed` with a 3-minute hard timeout.

- **Adobe uses `prompt=login` instead of a popup.** `POST /ims/revoke` (existing `onOAuthLogout`) handles token revocation. IMS doesn't accept `access_token` as a query parameter on its logout endpoint, so a GET-navigable popup URL isn't viable. `prompt=login` on the authorize URL achieves the same result: IMS forces credential entry regardless of the active SSO session.

- **`loggedOut` flag on `Account`.** The row is retained in localStorage with tokens stripped and `loggedOut: true`. This lets the UI distinguish "never configured" from "configured but signed out" without a second storage key.

- **GitHub `/logout` is best-effort.** GitHub's real logout is a CSRF-protected POST. Opening `https://github.com/logout` in the popup is a GET navigation that presents the logout page — the user needs to click through — so it works in practice but isn't a guaranteed programmatic sign-out.

- **`popup.closed` polling in `launchOAuthCli`.** Pre-existing issue surfaced by this feature: closing the OAuth popup mid-flow left the result poller running for the full 2-minute timeout. Fixed alongside this work.

**Data shape change:** `Account` in `slicc_accounts` (localStorage) gains an optional `loggedOut: boolean` field. No migration needed — old entries without the field behave as `loggedOut: false` (logged in).

## Cross-cutting impact

- **Floats exercised:** CLI (node-server) is the primary target. Extension runtime: `openIdpLogoutUrl` is a no-op (`isExtension` guard), `launchOAuthCli` is not used by the extension. Worker runtime: `logoutOAuthAccount` and `openIdpLogoutUrl` both no-op when `typeof window === 'undefined'`. Electron/Sliccstart follow the CLI path.

- **Other callers of changed functions:**
  - `removeAccount` — now calls `logoutOAuthAccount` first for OAuth providers. Non-OAuth callers (`addAccount` replace flow, `clearAllSettings`, MCP delete) are unaffected — guarded by `configToRemove?.isOAuth`. The second replica DELETE that `removeAccount` issues after `logoutOAuthAccount` already cleared it hits the existing `status !== 404` guard and is silent.
  - `launchOAuthCli` — the new `closedTimer` interval fires every 500ms. Existing tests use fake timers; `mockPopup.closed` is `undefined` (falsy) in all pre-existing mocks so the interval never fires early.
  - `onOAuthLogin` on Adobe — signature gains an optional third `options` param. All existing callers that omit it are unaffected.

- **Persisted state:** `slicc_accounts` entries for OAuth providers may now carry `loggedOut: true` with token fields absent. `selected-model` is cleared by `removeAccount` (unchanged). No new localStorage keys.

- **Async lifecycle:** `openIdpLogoutUrl` clears both `setInterval` and `setTimeout` handles on resolution. `launchOAuthCli`'s new `closedTimer` is cleared in the shared `cleanup()` function that all resolution paths call.

- **Security:** `access_token` is no longer passed as a URL query parameter to Adobe IMS (the initial `getOAuthLogoutUrl` approach would have exposed tokens in browser history/server logs — caught and removed before shipping).

## Test plan

- [ ] `npx prettier --write <changed-files>` — CI runs `prettier --check .` as a lint gate
- [ ] `npm run typecheck`
- [x] `npm run test` — **4482 passing**, no new failures
- [x] `npm run build`
- [x] **New behavior covered by unit tests:**
  - `packages/webapp/tests/providers/oauth-service.test.ts` — `openIdpLogoutUrl` (4 tests: popup open, timeout force-close, user-close resolves early, worker no-op); `launchOAuthCli` popup-closed cancellation (1 test)
  - `packages/webapp/tests/ui/provider-settings.test.ts` — `logoutOAuthAccount` (6 tests); `removeAccount` OAuth wiring (2 tests)
- [x] Manual smoke (CLI): `npm run dev`
  - GitHub row shows Logout button between Edit and Delete
  - Logout → `/logout` popup opens → sign out → close popup → row shows "Logged out — was @username", Login button appears
  - Login → OAuth popup opens fresh, no pre-selected account → reconnect as different account
  - Delete on a logged-out OAuth account → row removed
  - Avatar → Sign out on an OAuth account → row stays in logged-out state
  - Adobe: Logout (no popup, token revoked) → Login → IMS prompts for credentials (`prompt=login`)
- [x] Manual smoke (extension): `openIdpLogoutUrl` is a no-op; logout/login button still appears; `onOAuthLogout` token revocation still runs

**Pre-existing failures:** none known.

## Documentation

## Risk & rollback

- **Blast radius:** OAuth login/logout flows on all floats. API-key providers are untouched.
- **Rollback:** revert cleanly. Any `loggedOut: true` entries left in localStorage degrade gracefully — they display as accounts with no token, same as a newly added account that hasn't logged in yet.
- **CI cannot catch:** actual GitHub `/logout` GET behaviour (may not clear session in all browser profiles); Adobe IMS `prompt=login` round-trip; `popup.closed` detection in real Electron/Chrome contexts.

## Checklist

- [x] `OAuthLoginOptions.forceReauth` and `ProviderConfig.getOAuthLogoutUrl` are additive optional fields; no existing callers broken
- [ ] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs


trieloff (commented):
Address the copilot review comments, then go for it.

PR #861, opened 2026-06-04
Pyodide root-cause perf: warm worker pool + lazy FS

trieloff:
## Summary

Fixes the Pyodide performance problem at the root by eliminating the eager "copy the whole cwd + /tmp tree into MEMFS on every invocation" model. Tracked in the workspace spec.

This is a draft — multi-wave work in progress (opened overnight per request).

### Wave 1 — Warm, reusable Pyodide worker pool (committed)
Replaces the one-shot "spawn a fresh worker per python invocation, then terminate" model with a pool of warm, reusable Pyodide realms.
- New py-realm-pool.ts: 1 warm idle, up to 2 parallel, FIFO queue when saturated, SIGKILL eviction + lazy replacement, idle-TTL drop.
- Reusable-realm protocol: a worker accepts a new realm-init after realm-done instead of being terminated; per-run interpreter state reset (sys.modules baseline + clean __main__ + MEMFS scratch).
- Per-checkout host re-attach to the current CommandContext.
- Keeps the existing eager VFS-to-Pyodide sync (runs against the reused worker).

Gates: typecheck, lint, and the webapp realm test suite (229 tests) all green.

### Wave 2 — JSPI-backed lazy filesystem for Python (in progress)
Spike-first: confirm Pyodide 0.29.4 + Emscripten exposes a JSPI-suspendable custom-FS hook. If viable, mount the whole VFS tree lazily over the realm RPC channel (no cwd//tmp subset restriction — reads are on-demand). Capability-gated with an eager-copy fallback (scoped to cwd//tmp) for non-JSPI runtimes. If the spike fails, fall back to incremental delta-sync on the warm pool. No SharedArrayBuffer / cross-origin isolation.

### Wave 3 — Harden the eager-copy fallback path
Surface real FsError causes instead of the generic "unreadable from VFS" warning.

## Test plan
- npm run lint
- npm run typecheck
- npm run test -w @slicc/webapp -- realm
- npm run test

### DIFF
--- a/packages/webapp/src/kernel/realm/py-realm-pool.ts
+++ b/packages/webapp/src/kernel/realm/py-realm-pool.ts
@@ -1,0 +1,205 @@
+/**
+ * `py-realm-pool.ts` — a pool of warm, reusable Pyodide realms.
+ *
+ * Replaces the one-shot "spawn a DedicatedWorker per `python`
+ * invocation, then `worker.terminate()`" model: the interpreter and
+ * its FS survive between runs, so only the first call pays the
+ * ~1-2 s `loadPyodide` cold boot. Defaults: keep 1 warm idle worker,
+ * scale to 2 in parallel, queue (FIFO) beyond that.
+ *
+ * The pool is signal-agnostic — the runner (`runInPooledRealm`)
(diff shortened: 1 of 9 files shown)


trieloff:
Closing. We'll have a more comprehensive fix for this.

PR #864, opened 2026-06-04
fix(chat): fix markdown element contrast in user message bubbles

### LINKED ISSUE #862: fix(chat): inline code and URL links unreadable in user message bubbles (contrast <1.5:1)
## Summary

Two CSS cascade bugs make inline `<code>` tokens and URL links **unreadable inside user message bubbles** in the chat panel. Both affect the core interaction surface for a developer tool.

---

### What happens

`<code>` rendered inside a user bubble (`.msg--user`) shows **white text on a light-gray background**. The contrast ratio is ~1.3:1 — WCAG AA requires 4.5:1 minimum.

**Computed styles on the `code` element inside `.msg--user`:**
| Property | Value |
|---|---|
| `color` | `rgb(255, 255, 255)` — white |
| `background-color` | `rgb(225, 225, 225)` — light gray |
| Contrast ratio | **1.3:1** (catastrophic failure) |

### Root cause

Two rules with **equal specificity** (`0,1,2`) set the `background` property:

**`packages/webapp/src/ui/styles/markdown.css` line 84–88:**
```css
.msg :not(pre) > code {
  background: var(--s2-gray-200);  /* rgb(225,225,225) */
  padding: 2px 6px;
  border-radius: var(--s2-radius-s);
}
```

**`packages/webapp/src/ui/styles/chat.css` line 87–90:**
```css
.msg--user :not(pre) > code {
  background: rgba(255, 255, 255, 0.18);  /* intended: translucent white on purple */
  color: #fff;
}
```

`markdown.css` loads after `chat.css`, so its gray background wins on equal specificity. `color: #fff` from `chat.css` still applies (no color rule in `markdown.css` for this selector), producing the accidental white-on-gray combination.

### Fix

Increase the specificity of the chat.css rule by one class:

```css
/* chat.css line 87 — before */
.msg--user :not(pre) > code {

/* chat.css line 87 — after */
.msg.msg--user :not(pre) > code {
```

This raises specificity to `0,2,2`, unconditionally winning over `markdown.css`'s `0,1,2` rule.

---

### What happens

Links (`<a>`) inside user bubbles render as **browser-default blue** (`rgb(30, 102, 245)`) against the purple bubble background (`rgb(113, 85, 250)`). Both colors are dark cool-hue with near-identical luminance.

**Contrast ratio: 1.04:1** — essentially invisible to anyone with blue-purple color deficiency. The underline is the only differentiator.

### Root cause

**`packages/webapp/src/ui/styles/chat.css` lines 79–83** correctly defines:
```css
.msg--user .msg__content a,
.msg--user .msg__content a:visited {
  color: #c8bbff;  /* light lavender — correct intent */
  text-decoration: underline;
}
```

But the selector requires a `.msg__content` wrapper in the DOM. The actual `<a>` elements sit inside `<p>` tags **directly under `.msg--user`**, with no `.msg__content` intermediate — the rule never matches.

### Fix (two options)

**Option A** — Fix the selector to match the actual DOM structure:
```css
/* chat.css lines 79–80 — before */
.msg--user .msg__content a,
.msg--user .msg__content a:visited {

/* after */
.msg--user a,
.msg--user a:visited {
```

**Option B** — Use white (`#fff`) for maximum contrast (4.77:1 against the purple bubble, already passes WCAG AA):
```css
.msg--user a,
.msg--user a:visited {
  color: #fff;
  text-decoration: underline;
}
```

---

## Steps to Reproduce

1. Open any SLICC session in the standalone chat
2. Type a message containing a backtick-wrapped word, e.g.: `how is this \`code\` ?`
3. Type a message containing a URL, e.g.: `check out https://example.com`
4. Observe: the inline code token is a near-invisible gray pill; the URL link is indistinguishable from the surrounding bubble text by color alone

## Affected files

- `packages/webapp/src/ui/styles/chat.css` — lines 79–90
- `packages/webapp/src/ui/styles/markdown.css` — lines 84–88

## Labels

`bug` `accessibility` `chat-ui`

Image

vdua:
## Summary

Fixes #862. Multiple markdown elements were unreadable inside user message bubbles (`.msg--user`) due to CSS cascade conflicts: `markdown.css` is imported after `chat.css`, so at equal specificity its generic `.msg` / `.msg__content` rules won over the user-bubble overrides. (One offender was intra-file: `.theme-light .msg__content a` sits later in `chat.css` and beat the user-bubble link rule.)

The user bubble is purple (`#7155fa`) in **both** themes, but the losing cascade left it with light-theme backgrounds/text, producing catastrophic contrast.

## Elements fixed

| Element | Before | After |
|---|---|---|
| Inline `code` | white on light-gray (~1.3:1) | white on translucent-white overlay |
| Fenced `pre` block | lavender text on light-gray bg | lavender on dark overlay |
| Links | browser-default blue on purple (~1.04:1) | white + styled underline (4.79:1, WCAG AA) |
| Table `th` bg | white on light-gray (~1.1:1) | white on dark overlay |
| Table `th`/`td` borders | invisible light-gray | `rgba(255,255,255,0.25)` |
| Blockquote text | dark-gray on purple (~1.3:1) | `rgba(255,255,255,0.8)` |

## Approach

Raised the user-bubble overrides to `.msg.msg--user` (class-doubling specificity bump) so they win regardless of stylesheet load order, and added rules for table/blockquote which previously had none. All overrides are theme-independent, matching the always-purple bubble.

Note: chose white over the issue's suggested lavender `#c8bbff` for links — lavender only reaches 2.74:1 against the purple bubble (fails AA); white reaches 4.79:1.

## Test plan

- [x] `npm run build -w @slicc/webapp` — clean
- [x] Verified each element via `?ui-fixture=1` + injected probes (computed styles + screenshots)
- [ ] Light and dark theme spot-check

### DIFF
--- a/packages/webapp/src/ui/styles/chat.css
+++ b/packages/webapp/src/ui/styles/chat.css
@@ -99,6 +99,16 @@
 .msg.msg--user pre code {
   color: #e8e0ff;
 }
+.msg.msg--user .msg__content th {
+  background: rgba(0, 0, 0, 0.2);
+}
+.msg.msg--user .msg__content th,
+.msg.msg--user .msg__content td {
+  border-color: rgba(255, 255, 255, 0.25);
+}
+.msg.msg--user .msg__content blockquote {
+  color: rgba(255, 255, 255, 0.8);
+}
 .msg--user .msg__attachments {
   margin-bottom: 8px;
 }
--- a/packages/webapp/src/ui/styles/chat.css
+++ b/packages/webapp/src/ui/styles/chat.css
@@ -92,11 +92,11 @@
   background: rgba(255, 255, 255, 0.18);
   color: #fff;
 }
-.msg--user pre {
+.msg.msg--user pre {
   background: rgba(0, 0, 0, 0.25);
   border-color: rgba(255, 255, 255, 0.1);
 }
-.msg--user pre code {
+.msg.msg--user pre code {
   color: #e8e0ff;
 }
 .msg--user .msg__attachments {
--- a/packages/webapp/src/ui/styles/chat.css
+++ b/packages/webapp/src/ui/styles/chat.css
@@ -76,15 +76,15 @@
   line-height: 18px;
   letter-spacing: 0;
 }
-.msg--user .msg__content a,
-.msg--user .msg__content a:visited {
-  color: #c8bbff;
+.msg.msg--user .msg__content a,
+.msg.msg--user .msg__content a:visited {
+  color: #fff;
   text-decoration: underline;
 }
-.msg--user .msg__content a:hover {
+.msg.msg--user .msg__content a:hover {
   color: #fff;
 }
-.msg--user :not(pre) > code {
+.msg.msg--user :not(pre) > code {
   background: rgba(255, 255, 255, 0.18);
   color: #fff;
 }
--- a/packages/webapp/src/ui/styles/chat.css
+++ b/packages/webapp/src/ui/styles/chat.css
@@ -80,9 +80,13 @@
 .msg.msg--user .msg__content a:visited {
   color: #fff;
   text-decoration: underline;
+  text-decoration-color: rgba(255, 255, 255, 0.6);
+  text-decoration-thickness: 1px;
+  text-underline-offset: 2px;
 }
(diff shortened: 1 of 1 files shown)


vdua:
image

image

PR #872, opened 2026-06-04
fix: persist Chrome profiles to ~/.slicc/chrome-profiles

bstopp:
## Summary

Chrome profiles were written to `$TMPDIR` (falling back to `/tmp`), which macOS clears on reboot, silently wiping all session history. Both `node-server` and `swift-server` now write profiles to `~/.slicc/chrome-profiles/`, a stable user-owned directory that survives reboots. A one-time migration copies any existing legacy profile to the new location on first launch after upgrade.

## Why

After running two SLICC instances simultaneously (e.g. sliccstart + `npx sliccy`), the default cone would lose all session history after a reboot. Root cause: sliccstart's swift-server launches as a GUI app and may not inherit the shell's `$TMPDIR`, causing it to fall back to `/tmp` — which macOS purges on restart.

**Repro:**
1. Open sliccstart, do some work (sessions created)
2. Run `npx sliccy` for a second cone
3. Reboot
4. Reopen sliccstart

Expected: previous sessions are present
Actual: session list is empty

- `resolveDefaultChromeUserDataDir()` (`chrome-launch.ts`) and `resolveUserDataDir()` (`ChromeLauncher.swift`) now default to `~/.slicc/chrome-profiles/` via `homedir()` / `homeDirectoryProvider()` instead of `$TMPDIR ?? /tmp`. The per-port suffix logic (`-<N>` for non-5710 ports) is unchanged.
- `index.ts` no longer passes `tmpDir: process.env['TMPDIR']` to `resolveChromeLaunchProfile` — that env-var pass-through was the production path that kept profiles in temp.
- The `tmpDir` override parameter is kept on both functions as a test escape hatch (existing tests pass explicit paths; none touch the real home directory).
- `migrateLegacyDefaultChromeProfile(newDir, candidates)` copies the first found legacy candidate to `newDir` if `newDir` doesn't exist yet. Non-destructive — old profile is left in place. Handles all port variants because the caller derives `candidates` from `basename(userDataDir)`, which already includes any port suffix.
- `legacyChromeCandidates(profileDirName)` builds the ordered candidate list: `$TMPDIR` first, then `/tmp`, deduplicating when they're the same path.
- Migration is skipped for hosted mode (`--hosted`) and QA profiles (`leader`/`follower`/`extension`) — hosted has its own `CHROME_USER_DATA_DIR`, and QA profiles never lived in tmp.

## Cross-cutting impact

- **Floats exercised**: CLI (`node-server`) and Sliccstart (`swift-server`). Chrome extension float is unaffected (no Chrome profile management). Electron float is unaffected (uses separate CDP attach path).
- **Other callers**: `resolveChromeLaunchProfile` is called in one place in `index.ts`; `ChromeLaunchConfig.userDataDir` is set at the call site in `ServerCommand.swift` (swift-server CLI entry) — that call site already uses `resolveUserDataDir()` so picks up the new default automatically.
- **Persisted state side-effects**: IndexedDB (`browser-coding-agent`, `slicc-fs`, etc.) lives inside the Chrome profile directory and moves with it. No schema changes; the data is copied verbatim.
- **Migration rollback**: if a user downgrades, the old code will create a fresh profile in `$TMPDIR` — sessions accumulated in `~/.slicc/chrome-profiles/` won't be visible. No automated undo.

## Test plan

- [ ] `npm run typecheck` — passes (pre-existing `just-bash ByteString` error on `main`, unrelated)
- [ ] `npm run test` — 6285 passing, 3 pre-existing failures in `packages/cherry/` (reproduce on `main`)
- [ ] `npm run lint` — clean
- [ ] **New behavior covered by unit tests** in `packages/node-server/tests/chrome-launch.test.ts`:
  - `defaults to ~/.slicc/chrome-profiles when no tmpDir override is given`
  - `uses an explicit tmpDir override when provided`
  - `migrateLegacyDefaultChromeProfile > copies a legacy profile to the new stable location`
  - `migrateLegacyDefaultChromeProfile > skips migration when the new profile already exists`
  - `migrateLegacyDefaultChromeProfile > is a no-op when no legacy profile exists`
  - `migrateLegacyDefaultChromeProfile > tries candidates in order and stops at the first match`
  - Swift equivalents in `ChromeLauncherTests.swift` (build blocked by pre-existing `swift-nio-ssl` C++ header issue on macOS 15, unrelated to this PR)
- [ ] Manual smoke (CLI): `npm run dev` — confirmed Chrome launches, profile appears at `~/.slicc/chrome-profiles/browser-coding-agent-chrome`
- [ ] Manual smoke (Sliccstart): relevant — swift-server change means sliccstart profiles now survive reboot

**Pre-existing failures:**
- 3 failures in `packages/cherry/tests/` reproduce on `main` (Vite transform error, unrelated)
- Swift test suite fails to compile on macOS 15 due to `swift-nio-ssl` missing `<memory>` C++ header — pre-existing, reproduces on `main`

## Documentation

- [x] No documentation changes needed — profile location is an implementation detail not referenced in user-facing docs or agent skills

## Risk & rollback

- **Blast radius**: all floats that use a default Chrome profile (CLI + Sliccstart). QA profiles, hosted mode, and extension float are unaffected.
- **Rollback**: revert this PR. Users who already launched after upgrade will have a profile at `~/.slicc/chrome-profiles/` that a reverted build won't find — they'd get a fresh profile in `$TMPDIR`. The `~/.slicc/chrome-profiles/` directory would need manual cleanup to fully undo.
- **CI cannot catch**: actual Chrome profile persistence across reboot; migration behavior with a real legacy profile in `/tmp`.

## Checklist

- [x] No public API / tool / shell-command changes
- [x] Checked CLI and Sliccstart paths; extension and Electron floats unaffected

### DIFF
--- a/packages/node-server/src/chrome-launch.ts
+++ b/packages/node-server/src/chrome-launch.ts
@@ -1,6 +1,6 @@
 import type { ChildProcess } from 'child_process';
 import { existsSync, readdirSync } from 'fs';
-import { cp, mkdir, readFile, unlink, writeFile } from 'fs/promises';
+import { cp, mkdir, readFile, rm, unlink, writeFile } from 'fs/promises';
 import { request as httpRequest } from 'http';
 import { homedir } from 'os';
 import { dirname, join } from 'path';
@@ -146,8 +146,16 @@ export async function migrateLegacyDefaultChromeProfile(
   if (existsSync(newDir)) return;
   for (const candidate of candidates) {
     if (existsSync(candidate)) {
-      await cp(candidate, newDir, { recursive: true });
-      console.log(`Migrated Chrome profile: ${candidate} → ${newDir}`);
+      try {
+        await cp(candidate, newDir, { recursive: true });
+        console.log(`Migrated Chrome profile: ${candidate} → ${newDir}`);
+      } catch (err) {
+        await rm(newDir, { recursive: true, force: true }).catch(() => {});
+        console.warn(
+          `Chrome profile migration failed (${candidate} → ${newDir}); continuing with a fresh profile.`,
+          err
+        );
+      }
       return;
     }
   }
--- a/packages/node-server/src/chrome-launch.ts
+++ b/packages/node-server/src/chrome-launch.ts
@@ -112,7 +112,7 @@ export function resolveQaProfilesRoot(projectRoot: string): string {
 }
 
 export function resolveDefaultChromeUserDataDir(
-  profilesDir = join(homedir(), '.slicc', 'chrome-profiles'),
+  profilesDir = join(homedir(), '.slicc', 'profiles'),
   servePort?: number
 ): string {
   const suffix = servePort && servePort !== 5710 ? `-${servePort}` : '';
--- a/packages/node-server/src/chrome-launch.ts
+++ b/packages/node-server/src/chrome-launch.ts
@@ -1,6 +1,6 @@
 import type { ChildProcess } from 'child_process';
 import { existsSync, readdirSync } from 'fs';
-import { mkdir, readFile, unlink, writeFile } from 'fs/promises';
+import { cp, mkdir, readFile, unlink, writeFile } from 'fs/promises';
 import { request as httpRequest } from 'http';
 import { homedir } from 'os';
 import { dirname, join } from 'path';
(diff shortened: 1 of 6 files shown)


karlpauls:
can we make it just `profiles` instead of `chrome-profiles`?

trieloff:
@codex what are idiomatic locations for these profiles under macOS, Linux and Windows?

bstopp:
Do you want me to update the PR to put the profile in the recommended paths based on the platform?

trieloff:
Yes, please.

PR #897, opened 2026-06-06
fix(kernel): seed chat buffers from restored agent history on boot

trieloff:
## Summary

On reload, the kernel host's chat buffer (`OffscreenBridge.messageBuffers`) started empty on every boot. The first post-reboot `persistScoop` call overwrote the canonical UI session store with a truncated copy, causing the chat panel to show only the messages recorded since the last boot.

This PR seeds each registered scoop's buffer from its restored canonical `AgentMessage[]` inside `createKernelHost`, right after `orchestrator.init()` and before `kernel-worker-ready` is signaled, so the UI store is fully repaired before the panel can mount and read it. Additionally, `persistScoopAwait()` (an awaitable variant of `persistScoop`) is used in the seed loop to guarantee the IDB write lands before the worker signals ready.

## Why

`browser-coding-agent / session-cone` was being silently truncated on every reload — the canonical `agent-sessions` store was intact (669 messages), but the chat panel rendered from the truncated UI store (6 messages). The race between `handleRequestScoopMessages` hydration and the first `persistScoop` call was consistently lost because the hydration path is fire-and-forget with no retry.

- **`KernelFacade.seedBuffersFromAgentState()`** — new method on the host facade, implemented in `OffscreenBridge`. Non-destructive: only fills buffers that are still empty.
- **`persistScoopAwait()`** — awaitable variant of `persistScoop`; the fire-and-forget `persistScoop()` now delegates to it via `void`. The seed loop awaits it so the UI store is durably repaired before `kernel-worker-ready`.
- **`buildBufferFromAgentMessages()`** — extracted shared helper (agent→chat translation), used by both `seedBuffersFromAgentState` and `handleRequestScoopMessages` to avoid duplication.
- Applies to **both** floats — `kernel-worker.ts` (standalone) and `offscreen.ts` (extension) both boot through `createKernelHost`.

## Cross-cutting impact

- **Floats exercised**: standalone kernel worker + Chrome extension offscreen.
- **Persisted state side-effects**: `browser-coding-agent / session-cone` is repaired (written once at boot if empty) rather than truncated.
- **Async lifecycle**: `persistScoop` callers outside the seed path are unchanged (still fire-and-forget); only the boot-time seed loop awaits `persistScoopAwait()`.
- **CI flake fixed**: `tests/ui/file-browser-panel.test.ts > renders via RemoteVfsClient + VfsRpcHost` intermittently failed under CI load due to fixed `await tick(20)` waits for a real-MessageChannel RPC round-trip. Replaced with a polling `waitFor()` helper so assertions are deterministic.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test` — `vitest run offscreen-bridge.test.ts` ✅ (122 passed, 4 new); `vitest run tests/kernel` + translator ✅ (675 passed); `file-browser-panel.test.ts` ✅ (4 passed × 5 runs)
- [x] `npm run build -w @slicc/webapp` and `-w @slicc/chrome-extension` ✅
- [x] `npm run lint` ✅
- [x] **New / changed behavior is covered by a unit test** — `seedBuffersFromAgentState` suite in `offscreen-bridge.test.ts`: seeds empty buffer and persists it; does not overwrite a buffer with live messages; skips scoops with no restored messages; no-op when orchestrator is unbound.

**Pre-existing failures**: none — `file-browser-panel.test.ts` flake (from #896) is resolved by the `waitFor()` polling fix in this PR.

## Documentation

- [x] No documentation changes needed

## Risk & rollback

- Blast radius: chat panel history display on reload across both floats.
- No feature flags. Revert this PR cleanly reverts the fix; the truncation behaviour returns but no data migration is needed (canonical `agent-sessions` is untouched by this PR).
- The seeding is non-destructive: buffers that already have live messages are never overwritten.

## Checklist

- [x] No secrets, credentials, OAuth client secrets, or tokens in the diff (incl. logs, fixtures, `.env*`, snapshots)

### DIFF
--- a/packages/chrome-extension/src/offscreen-bridge.ts
+++ b/packages/chrome-extension/src/offscreen-bridge.ts
@@ -863,6 +863,70 @@ export class OffscreenBridge implements KernelFacade {
   }
 
   /**
+   * Translate a scoop's restored canonical `AgentMessage[]` into the
+   * buffered chat shape. Returns `null` when there is no context or no
+   * agent messages yet. Lazy-imports the translator so it doesn't pull
+   * pi-ai types into the bridge's hot path until needed. Shared by
+   * {@link handleRequestScoopMessages} and {@link seedBuffersFromAgentState}.
+   */
+  private async buildBufferFromAgentMessages(
+    scoop: RegisteredScoop
+  ): Promise<BufferedChatMessage[] | null> {
+    const context = this.orchestrator?.getScoopContext(scoop.jid);
+    if (!context) return null;
+    const agentMessages = context.getAgentMessages();
+    if (agentMessages.length === 0) return null;
+    const { agentMessagesToChatMessages } = await import(
+      '../../../packages/webapp/src/scoops/agent-message-to-chat.js'
+    );
+    const chatMessages = agentMessagesToChatMessages(agentMessages, {
+      source: scoop.isCone ? 'cone' : (scoop.name ?? scoop.folder),
+    });
+    return chatMessages.map((m) => ({
+      id: m.id,
+      role: m.role,
+      content: m.content,
+      attachments: m.attachments,
+      timestamp: m.timestamp,
+      source: m.source,
+      channel: m.channel,
+      toolCalls: m.toolCalls?.map((tc) => ({
+        id: tc.id,
+        name: tc.name,
+        input: tc.input,
+        result: tc.result,
+        isError: tc.isError,
+      })),
+      isStreaming: false,
+    }));
+  }
+
+  /**
+   * Seed each registered scoop's chat buffer from its agent's restored
+   * canonical history at boot, BEFORE any post-boot turn can run
+   * `persistScoop`. The bridge's `messageBuffers` otherwise start empty
+   * on a fresh boot, so the first agent turn after a reload would
+   * persist only the new messages and overwrite the full conversation
+   * in the `browser-coding-agent` UI store — the "only the last few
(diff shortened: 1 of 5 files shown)


trieloff:
Addressed both the review and the CI failure in 5d3264c9:

**@Copilot (seeded persist was fire-and-forget):** Correct — `seedBuffersFromAgentState()` is awaited inside `createKernelHost` precisely so `browser-coding-agent` is repaired before `kernel-worker-ready`, but `persistScoop()` doesn't await `SessionStore.saveMessages()`, so the await didn't actually guarantee the write landed. Added an awaitable `persistScoopAwait()` (the fire-and-forget `persistScoop()` now just delegates to it via `void`), and the seed loop `await`s it. The UI store is now durably repaired before the worker signals ready / the panel mounts.

**CI failure (`webapp` job):** The failing test was `tests/ui/file-browser-panel.test.ts > renders via RemoteVfsClient + VfsRpcHost`, a pre-existing flake (from #896, now on `main`) — it waited a fixed `await tick(20)` for a real-MessageChannel RPC round-trip (`readDir → stat → render`), which intermittently lost under CI load (it passes locally). Replaced the fixed waits with a polling `waitFor()` helper so the assertions are deterministic. Verified green across 5 consecutive local runs.

Verification: `npm run lint`, `npm run typecheck`, `offscreen-bridge.test.ts` (122 passed), `file-browser-panel.test.ts` (4 passed × 5 runs) all green.

PR #905, opened 2026-06-07
docs: document VFS mount empty directory visibility pitfall under Pyodide

ddttom:
Documents the behavior where mounted VFS paths appear empty inside sandboxed Pyodide/Python environments.


trieloff:
Hi Tom, that PR looks like it has a bunch of unrelated changes. I will try and see if we can fix the root issue rather than just document it.

trieloff:
Filed #908 to track the root cause (mounted VFS directories appearing empty inside the Pyodide `python3` realm) so it can be fixed rather than documented. Thanks for surfacing the behavior.

ddttom:
Ta, Slicc did the PR. I just directed it

On Mon Jun 8, 2026, 07:20 AM GMT, Lars Trieloff ***@***.***> wrote:
[image]trieloff left a comment (ai-ecoverse/slicc#905) <https://github.com/ai-ecoverse/slicc/pull/905#issuecomment-4646318493>

trieloff:
Par for the course. But unless we are ready to run SLICC inside SLICC, it would have been hard enough to reproduce and too hard to fix.

PR #919, opened 2026-06-08
Fix Pyodide mounted-directory access (read + write) — closes #908

### LINKED ISSUE #908: bug: mounted VFS directories appear empty inside Pyodide/Python (`python3`)
## Summary

Mounted VFS directories (local FS Access, S3, and DA mounts created via the `mount` command) are **invisible/empty** when read from inside the Pyodide `python3` / `python` realm. `bash` tools (`read_file`, `ls`, `cat`) see mounted content correctly because they route through `VirtualFS`, but Python sees the mount point as an empty directory. Writes from Python into a mounted path are also silently lost.

Filed to track the underlying root cause after #905 (which only proposed documenting the behavior and carried unrelated changes) was closed with the intent to fix the root issue rather than document it.

Related/closed PR: #905

## Repro

1. `mount` a local directory, S3 bucket, or da.live source at some VFS path (e.g. `mount add s3 …` → mounts under `/mnt/<name>` or similar).
2. Confirm `bash` sees content: `ls /mnt/<name>` and `read_file /mnt/<name>/somefile` work.
3. Run Python over the same path:
   ```
   python3 -c "import os; print(os.listdir('/mnt/<name>'))"
   ```
   Observed: `[]` (empty). Expected: the mounted entries.
4. `python3 -c "open('/mnt/<name>/x.txt','w').write('hi')"` appears to succeed but the write never reaches the mount backend and is not visible to `read_file`.

## Root cause

The Pyodide realm does **not** sync the VFS through `VirtualFS`. For performance it reaches **directly into the OPFS subtree** the kernel worker owns (`OPFS-root/slicc-fs/<path>`) via the in-tree `OPFS_SYNC_FS` Emscripten plugin:

- `packages/webapp/src/kernel/realm/py-realm-shared.ts` — `mountOpfsDirsAndSyncIn()` resolves each sync dir to an OPFS `FileSystemDirectoryHandle` and `prewalkOpfsTree()`s it, then `pyodide.FS.mount(OPFS_SYNC_FS, …)`.
- `packages/webapp/src/kernel/realm/opfs-sync-fs.ts` — builds the Pyodide-FS tree straight from the OPFS prewalk snapshot.

But **mount points are not stored in OPFS**. `VirtualFS.mount()` / `mountInternal()` only create an **empty placeholder directory** in the backing store so that ancestor path resolution works (`packages/webapp/src/fs/virtual-fs.ts:863-898`, `:1058-1065`):

> "All reads and writes under that path are transparently bridged to the real directory handle — no copying occurs. A placeholder directory is created in LightningFS so that ancestor paths resolve correctly."

The actual mounted content is served lazily by the `MountBackend` implementations (`backend-local.ts` FS Access handle, `backend-s3.ts` HTTP+SigV4, `backend-da.ts` HTTP+IMS) and is **only reachable when a read routes through `VirtualFS.findMount()`** (`readDir` / `readFile` / `stat` in `virtual-fs.ts`).

Since the OPFS prewalk bypasses `VirtualFS` entirely, it never calls `findMount()` — it just sees the empty placeholder directory in OPFS. Hence Python sees an empty mount. Symmetrically, Python writes flush back to the OPFS placeholder (`flushOpfsRealmMounts`), never to the backend.

Note: the legacy `vfs.walkTree` RPC sync path *does* go through `ctx.fs` (the VFS adapter) and would route through mount backends — but that path is only used in Node test environments without `navigator.storage.getDirectory`. **Production always takes the OPFS fast path**, so production Python always sees empty mounts.

## Affected paths

- `packages/webapp/src/kernel/realm/py-realm-shared.ts` (`mountOpfsDirsAndSyncIn`, `mountOpfsChild`, `flushOpfsRealmMounts`)
- `packages/webapp/src/kernel/realm/opfs-sync-fs.ts` (OPFS prewalk plugin)
- `packages/webapp/src/shell/supplemental-commands/python-command.ts` (`computePyodideMountDirs`, OPFS dbName handoff)
- `packages/webapp/src/fs/virtual-fs.ts` (mount registry / `findMount` / placeholder dirs)
- `packages/webapp/src/fs/mount/backend*.ts` (the content actually lives here)

## Acceptance criteria

- `python3 -c "import os; print(os.listdir('<mountpoint>'))"` returns the same entries that `bash` `ls` shows for local, S3, and DA mounts.
- Reading a file under a mount from Python returns its real content.
- Writing a file under a mount from Python (where the backend supports writes) persists through the mount backend and is visible to `read_file` afterwards (or, if out of scope, fails loudly rather than silently discarding).
- Behavior is correct in both standalone (CLI) and extension floats.
- No measurable regression to `python3` cold-start / sync time for the common case of non-mounted dirs.

## Notes / possible approaches (to be designed)

- Detect mount points overlapping the Pyodide sync dirs and sync their content through `VirtualFS` (the `MountBackend`) rather than the OPFS prewalk — e.g. fall back to the `walkTree`/RPC path for those subtrees, or materialize the backend listing into the Pyodide FS.
- Alternatively, route Python file ops for mounted subtrees back over the VFS RPC channel on demand.
- Whichever approach: keep the OPFS fast path for the (common) non-mounted majority of the tree to preserve current performance.

trieloff:
## Summary

Mounted VFS directories (local FS Access, S3, DA) were invisible from the Pyodide `python3`/`python` realm, and writes silently flushed to an empty OPFS placeholder. The realm synced the VFS by reaching directly into the OPFS subtree via the `OPFS_SYNC_FS` prewalk, bypassing `VirtualFS.findMount()` — mounts exist in OPFS only as empty placeholder directories, so Python saw nothing.

This PR makes mounts **visible and writable** from Python, using the same snapshot-and-sync pattern the OPFS plugin already relies on to bridge the async→sync Emscripten boundary.

Closes #908. (Supersedes the closed, docs-only PR #905.)

## How it works

The OPFS path is not synchronous all the way to disk — it preloads file bytes async **before** Python runs, serves sync from memory **during** the turn, and flushes async **after**. We apply the same time-shifting to async mount backends:

- **Read sync (eager materialization):** For mount points overlapping the Pyodide sync dirs, the realm materializes the subtree into a MEMFS overlay over the OPFS placeholder via the `vfs` RPC (`readDir`/`stat`/`readFile` → `VirtualFS.findMount → MountBackend`). Listing is summed **before** any body fetch.
- **Write-back:** After the turn, `flushRealmMountWriteBack` diffs MEMFS against the materialize snapshot and routes creates/modifies/removes through the `vfs` RPC back to the backend; a backend rejection surfaces a clear stderr warning (never silent).
- **`--remote-mount-cap=<size>` flag** (default `5m`) bounds bytes eagerly materialized from **remote** (S3/DA) mounts. Over-cap fails fast with an actionable error; `--remote-mount-cap=0` disables remote mounts; **local FS Access mounts are exempt**.

True lazy sync (block the worker per read) would require `SharedArrayBuffer` + `Atomics.wait` and cross-origin isolation, which this app does not have — so eager materialization is the correct approach.

## Changes

- `shell/supplemental-commands/python-command.ts` — `--remote-mount-cap` flag, human-size parsing, overlapping-mount + kind detection passed via `RealmInitMsg`.
- `kernel/realm/realm-types.ts` — mount-points (path + kind) + `remoteMountCapBytes` on `RealmInitMsg`.
- `kernel/realm/py-realm-shared.ts` — eager materialization (list→cap→fetch) and diff-aware write-back flush.
- `fs/virtual-fs.ts` — `listMountPoints` helper.
- `docs/pitfalls.md`, `docs/shell-reference.md` — documented hybrid mount visibility + the cap flag; removed the old "empty-mount" pitfall framing.

## Tests

- `tests/kernel/realm/py-realm-mount-materialize.test.ts` — local/s3/da read+listdir parity, cap enforcement, `=0` skip, local exemption, mixed, per-mount listing failure.
- `tests/kernel/realm/py-realm-mount-write-back.test.ts` — multi-backend write/create/remove/rename persistence + backend-rejection warnings.
- `python-command.test.ts` — size parser + overlap detection.
- **#908 tests: 50/50 passing.**

## Verification

- `npm run lint` ✅
- `npm run typecheck` ✅ (all 6 tsconfigs)
- `npm run build` ✅ + `npm run build -w @slicc/chrome-extension` ✅
- `npm run test` (full suite): a pre-existing repo-wide `@zenfs/core` ESM module-frozen failure exists on `origin/main` too (reproduced on a clean worktree) — **not introduced by this branch**. Recommend a separate issue to track it.

## Follow-ups (non-blocking)

- Separate bug for the pre-existing `@zenfs/core` test-environment regression on main.
- Optional live-Pyodide end-to-end smoke test (current tests use a fake `pyodide.FS` shim).

---


trieloff:
Superseded by #922, which fixes #908 with **Model A (async-only `slicc.fs`)** instead of eager materialization. The eager approach here regressed badly on large mounts — thousands of sequential `vfs` RPCs during realm boot hung the realm — and there's no synchronous escape hatch for async backends without `SharedArrayBuffer` + cross-origin isolation. #922 makes the contract explicit (sync access bombs with friendly guidance; use `await slicc.fs.*`). Closing in favor of #922.

PR #934, opened 2026-06-09
fix(git-command): Add support for starting point in branch creation

QuentinVecchio:
## Summary

git checkout -b <branch> <start-point> now correctly creates the branch at the specified commit (previously it ignored the start point).

As part of the changes, also refactored the git-commands to pass linter tests

## Why

The branch-setup.sh script in of1-demo-skills-v3 was failing with "Invalid checksum in GitIndex buffer" after git checkout -b <branch> origin/main followed by git add -A. Root cause: isomorphic-git's git.branch({ checkout: true }) only updates the HEAD symref — it does NOT update the working tree or index. A subsequent git add -A would read the stale index and crash.

Repro:
1. Clone a repo in SLICC's VFS
2. git checkout -b feature origin/main
3. git add -A

Expected: clean add
Actual: Invalid checksum in GitIndex buffer

- Pass the startpoint (commit or branch) as object in git.branch call

## Cross-cutting impact

- Floats exercised: All (same GitCommands class used everywhere).
- Other callers: checkout is only called from the git supplemental command dispatcher — no other internal callers.
- Persisted state: No new state; only the git index (.git/index) is affected, which is the fix.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run test` — `<N>` passing, no new failures
- [x] `npm run build`
- [x] **New / changed behavior is covered by a unit test** in
      `packages/*/tests/` mirroring the changed `src/` path
      (e.g. `npx vitest run packages/webapp/tests/<area>/<file>.test.ts`)
- [] Manual smoke (CLI): `npm run dev` + …
- [x] Manual smoke (extension): load unpacked `dist/extension/` + …
- [] Manual smoke (Electron / Sliccstart / worker) — if relevant
- [ ] N/A — explain below

## Documentation

- [ ] `README.md` (user-facing behavior)
- [ ] Relevant `CLAUDE.md` (developer / package architecture)
- [ ] `docs/` (agent reference: tools, commands, skills, patterns)
- [ ] `packages/vfs-root/shared/CLAUDE.md` or `packages/vfs-root/workspace/skills/<skill>/SKILL.md` (agent-facing)

## Checklist

- [ ] Commits are focused and package-local where possible
- [ ] No hand-edited files under `dist/`
- [ ] No secrets, credentials, OAuth client secrets, or tokens in the diff (incl. logs, fixtures, `.env*`, snapshots)
- [ ] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs
- [ ] If this changes a contract used by other floats (CLI ↔ extension ↔ tray), I checked the followers/leader/offscreen paths

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -356,7 +356,6 @@
         "packages/webapp/src/fs/mount/backend-local.ts",
         "packages/webapp/src/fs/virtual-fs.ts",
         "packages/webapp/src/git/diff.ts",
-        "packages/webapp/src/git/git-commands.ts",
         "packages/webapp/src/git/git-config.ts",
         "packages/webapp/src/kernel/host.ts",
         "packages/webapp/src/kernel/panel-rpc.ts",
--- a/packages/webapp/src/git/git-commands.ts
+++ b/packages/webapp/src/git/git-commands.ts
@@ -933,7 +933,10 @@ Available commands:
     }
 
     if (createBranch) {
-      await git.branch({ fs: this.lfs, dir: cwd, ref, checkout: true });
+      const bIdx = args.indexOf('-b');
+      const afterB = args.slice(bIdx + 1).filter((a) => !a.startsWith('-'));
+      const startPoint = afterB.length > 1 ? afterB[1] : undefined;
+      await git.branch({ fs: this.lfs, dir: cwd, ref, object: startPoint, checkout: true });
       return {
         stdout: `Switched to a new branch '${ref}'\n`,
         stderr: '',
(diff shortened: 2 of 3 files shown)


trieloff:
@QuentinVecchio is there a chance that we turn this into an upstream PR and a patch file for us? Look at what we are doing with zerofs

QuentinVecchio:
> @QuentinVecchio is there a chance that we turn this into an upstream PR and a patch file for us? Look at what we are doing with zerofs

@trieloff the fix is actually in our git wrapper not in the isomorphic-git (their implementation already exposed the `starting point` as `object`)

PR #958, opened 2026-06-10
feat(ui): add-menu — replace attachment button with searchable context picker

usman-khalid:
## Summary

Replaces the attachment-only Paperclip button on the chat composer with a `+` "Add context" button that opens a searchable menu. Users can attach files (upload or drag-and-drop), take a screenshot, and add **references** to VFS files/folders, skills, frozen sessions, and scoops. References render as dismissible chips in the composer and compile into a hidden `[context]` preamble prepended to the agent message on send. The preamble is stripped from the reconstructed transcript display so it never leaks into chat bubbles.

## Why

The attachment button was an all-or-nothing file picker. Agents can already resolve VFS paths, skill names, session files, and scoop JIDs with their own tools — this surfaces those as first-class, typeahead-searchable inputs from the composer, giving users a faster way to point the agent at context without pasting paths by hand.

**Layering:**

- `packages/webapp/src/ui/add-menu/` — five new modules: `add-item.ts` (type definitions), `add-menu.ts` (DOM component), `capture.ts` (screen capture via `getDisplayMedia`), `preamble.ts` (compile + strip `[context]` block), `search-providers.ts` (VFS walk, skill/session/scoop providers, aggregator).
- `packages/webapp/src/ui/create-lucide-icon.ts` — extracted shared SVG builder that was duplicated in `chat-panel.ts`, `lick-view.ts`, and `tool-call-view.ts`.
- `packages/webapp/src/ui/chat-panel.ts` — `setAddMenu()` wires the menu into the composer; `addReference()` manages pending refs; `sendMessage()` compiles the preamble before handing text to the agent.
- `packages/webapp/src/ui/main.ts` — `wireAddMenu()` constructs the aggregator for both extension and standalone-worker boot paths.

**Non-obvious choices:**

- *References ride as a hidden text preamble, not new tool parameters.* The agent resolves each locator with its own tools (`read_file`, `bash cat`, etc.) — no new tool surface needed.
- *VFS walk is cached for 3 s with in-flight dedup.* Over the OPFS RPC path a fresh walk per keystroke would be dozens of round-trips; the TTL keeps typeahead responsive while still picking up filesystem changes.
- *`[context]` block stripped in `agentMessagesToChatMessages`* so the hidden preamble never appears in the rebuilt transcript when messages are loaded from session storage.
- *`WALK_LIMIT = 500`* bounds the breadth-first walk on large VFS trees.

**Design choices rejected:**

- Embedding file content inline in the message was considered but abandoned — it would bloat the context window unpredictably. A path reference lets the agent fetch exactly what it needs.
- A dedicated `add_reference` tool was considered but rejected in favour of the preamble text convention, which works without any model-side changes.

**New data shape:**

- `ChatMessage.references?: AddItem[]` — persisted by the session freezer so chips re-render on reload without the raw preamble being present.

## Cross-cutting impact

**Floats exercised:**
- CLI (standalone-worker): `wireAddMenu` called from `mainStandaloneWorker`, screenshot via `captureScreenshot()` (direct `getDisplayMedia`).
- Chrome extension: `wireAddMenu` called from `mainExtension`, screenshot routed through `captureViaPopup` because `getDisplayMedia` cannot prompt from the side panel.
- Electron: reuses standalone-worker boot path, no separate change needed.
- Sliccstart / worker: N/A — UI feature only.

**Shell contexts (extension):** side-panel `WasmShell` and offscreen `WasmShell` are untouched. The add-menu is side-panel UI only.

**Other callers of changed code:**
- `agentMessagesToChatMessages` — the `stripContextPreamble` call is guarded by `!lickChannel`, so lick messages are unaffected.
- `renderPendingAttachments` — now also renders reference chips; attachment chip rendering is unchanged.
- `lick-view.ts` / `tool-call-view.ts` — `iconNodeToSvg` replaced with the shared `createLucideIcon`; pixel output is identical (same attributes, same size).

**Persisted state side-effects:**
- `ChatMessage.references` is a new optional field written by the session freezer. Old sessions without it load fine (field is absent → no chips rendered). No migration needed.

**Async lifecycle:**
- `AddMenu.dispose()` cancels the debounce timer and removes all listeners.
- `ChatPanel.destroy()` now calls `addMenu?.dispose()`.
- In-flight `aggregator.search()` after dispose modifies a detached DOM subtree — harmless.

**Security:** all label/locator text is set via `textContent` (not `innerHTML`). Newlines in locators/labels are collapsed in `collapseNewlines()` to prevent `[context]` block corruption.

**Bundle size:** `lucide` icons were already imported in `chat-panel.ts`. This PR adds `Plus`, `Monitor`, and `Upload` — negligible delta.

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run test` — all passing, no new failures
- [x] `npm run build`
- [x] New modules covered by unit tests mirroring `src/` layout: `add-menu.test.ts`, `capture.test.ts`, `preamble.test.ts`, `search-providers.test.ts`, `create-lucide-icon.test.ts`, `chat-panel-add-integration.test.ts`
- [x] Manual smoke (CLI): `npm run dev` — `+` button opens menu, search filters VFS files/skills/sessions, chips appear and dismiss, Enter sends with preamble, transcript hides preamble on reload.
- [x] Manual smoke (extension): load unpacked `dist/extension/` — screenshot routes through popup, all other paths as above.
- [ ] Manual smoke (Electron): not verified; reuses standalone-worker path.

**Pre-existing failures:** none observed on `main`.

## Documentation

- [x] No documentation changes needed — this is a UI-only feature with no new shell commands, tools, or agent-facing contracts.

## Risk & rollback

- **Blast radius:** webapp UI only — CLI, extension, and Electron chat composer. No VFS, agent core, or protocol changes.
- **No feature flags.** `setAddMenu` is called unconditionally in both boot paths; rolling back means reverting this PR.
- **Rollback is clean** — revert restores the Paperclip button. The `references` field in frozen sessions is ignored by old code (optional field, no migration). `ChatMessage` shape is backwards-compatible.
- **CI cannot catch:** real `getDisplayMedia` picker UX and the extension popup capture path require manual verification with a loaded extension.

## Checklist

### DIFF
--- a/coverage-thresholds.json
+++ b/coverage-thresholds.json
@@ -33,6 +33,7 @@
         "packages/webapp/src/shell/**",
         "packages/webapp/src/fs/**",
         "packages/webapp/src/core/proxy-error.ts",
+        "packages/webapp/src/ui/add-menu/**",
         "packages/node-server/src/**"
       ]
     },
--- a/packages/webapp/src/scoops/agent-message-to-chat.ts
+++ b/packages/webapp/src/scoops/agent-message-to-chat.ts
@@ -36,6 +36,7 @@ import type {
   ToolResultMessage,
   UserMessage,
 } from '@earendil-works/pi-ai';
+import { stripContextPreamble } from '../ui/add-menu/preamble.js';
 import { isLickChannel, LICK_CHANNELS, type LickChannel } from '../ui/lick-channels.js';
 import type { ChatMessage, ToolCall as UiToolCall } from '../ui/types.js';
 import { HIDDEN_TOOL_NAMES } from './hidden-tools.js';
@@ -99,7 +100,7 @@ export function agentMessagesToChatMessages(
         const msg: ChatMessage = {
           id: idSeed(),
           role: 'user',
-          content: env.body,
+          content: lickChannel ? env.body : stripContextPreamble(env.body),
           timestamp: m.timestamp,
         };
         if (lickChannel) {
--- a/packages/webapp/src/ui/add-menu/add-item.ts
+++ b/packages/webapp/src/ui/add-menu/add-item.ts
@@ -12,3 +12,18 @@ export interface AddItem {
    *  file,folder → VFS path; skill → name; session → /sessions/<file>; scoop → jid. */
   locator: string;
 }
+
+/** User-facing category label for a reference kind. "session" reads as
+ *  "conversation" to match the add-menu search placeholder; everything else
+ *  uses its own name. */
+const REFERENCE_KIND_LABELS: Record<AddItemKind, string> = {
+  file: 'file',
+  folder: 'folder',
+  skill: 'skill',
+  session: 'conversation',
+  scoop: 'scoop',
+};
+
+export function referenceKindLabel(kind: AddItemKind): string {
+  return REFERENCE_KIND_LABELS[kind];
+}
--- a/packages/webapp/src/ui/add-menu/add-item.ts
+++ b/packages/webapp/src/ui/add-menu/add-item.ts
@@ -1,0 +1,14 @@
+/** Kinds of thing the add-menu can reference. All are reachable via the
+ *  agent's tools, so they ride along as references (not inline content). */
+export type AddItemKind = 'file' | 'folder' | 'skill' | 'session' | 'scoop';
+
+export interface AddItem {
+  kind: AddItemKind;
+  /** Display name: filename / skill name / session title / scoop name. */
+  label: string;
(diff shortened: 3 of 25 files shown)


usman-khalid:
Let me know if I should be addressing the `cognitive-complexity` fixes in:

 - packages/webapp/src/ui/chat-fixture.ts
 - packages/webapp/src/ui/chat-panel.ts

trieloff:
This will likely conflict with #961

trieloff:
Addressed in #961

PR #960, opened 2026-06-10
bug(welcome-prompt): Fix dupe Welcome message

### LINKED ISSUE #601: bug: Welcome wizard is duplicated on first run of `npx sliccy`
### Summary

"Welcome to SLICC — let's get you set up." and the wizard are duplicated when `npx sliccy` is run:

Image

### Float (runtime)

Standalone CLI / `npm run dev`

### Surface (best guess is fine)

Sprinkles / dips (`.shtml` rendering)

### Reproduction

```markdown
On macOS, run `$ npx sliccy`

Version in use on my machine: `sliccy@2.35.0`
```

### Suspected root cause / location (optional)

```markdown

```

### Logs / error output (optional)

```shell
This may be related:


404 GET /shared/sprinkles/welcome/welcome.shtml 2ms
NotFoundError: Not Found
    at createHttpError (/Users/blefebvr/.npm/_npx/96f86eb5ac0d0e25/node_modules/send/index.js:861:12)
    at SendStream.error (/Users/blefebvr/.npm/_npx/96f86eb5ac0d0e25/node_modules/send/index.js:168:31)
    at SendStream.pipe (/Users/blefebvr/.npm/_npx/96f86eb5ac0d0e25/node_modules/send/index.js:468:14)
    ...
```

### SLICC version / commit (optional)

2.35.0

### OS / browser / Node (optional)

macOS 14.4.1

### Pre-flight

bstopp:
## Summary

Closes: #601 

  Fixes two related welcome-flow bugs: (1) the "let's get you set up" prompt appeared twice on a fresh install, and (2) the prompt re-appeared on every restart when the user had not yet completed the profile form. After this PR, the welcome shows exactly once and a restart does not re-trigger it.

  ## Why

  **Repro (Bug 1 — double fire):**
  1. Run `npx sliccy`, start your default profile,
  2. Run `npx sliccy`, to start a new profile
  3. Expected: welcome dip appears once on the new profile
     Actual: welcome dip appears twice (stacked), one from the initial boot and one from the SW-registration page reload.

  **Repro (Bug 2 — restart re-welcome):**
  1. Run `npx sliccy` on a fresh profile → welcome appears, do **not** submit the profile form
  2. Quit and restart `npx sliccy`
  3. Expected: previous session restored, no second welcome.
     Actual: welcome re-appears on every restart.

  Two independent defects, both in the onboarding boot path:
  
  **Defect 1 — `handleFirstRun()` stage not advanced**
  `OnboardingOrchestrator.handleFirstRun()` had a guard `if (this.stage !== 'idle') return` but never set `this.stage` after posting, so any second call found stage still `'idle'` and re-posted the welcome. Fix: set `this.stage = 'collect-profile'` before the first `postDipReference` call.

  Side-effect fix: `handleOnboardingComplete()`'s entry guard also checked `!== 'idle'` and would have silently no-oped when stage was `'collect-profile'`. Widened to `!== 'idle' && !== 'collect-profile'` so the normal flow `handleFirstRun() → handleOnboardingComplete()` still works. The `'idle'` entry path is retained for the reload case where the dedup ledger suppressed `handleFirstRun()` but the user reaches `handleOnboardingComplete()` via the persisted connect-llm dip.

  **Defect 2 — stale-clearing block in `runFirstRunDetection`**
  The block intended to handle "user wiped VFS but not localStorage" would fire whenever `detectWelcomeFirstRun` returned `isFirstRun: true` AND `'first-run'` was already in the localStorage ledger — which is exactly the state after any page-reload or process restart mid-onboarding. The block cleared the dedup entry and called `handleFirstRun()` again.

  Fix: replace clear+refire with an early `return`. If `'first-run'` is already in the ledger, the welcome was shown in a prior session — do nothing.

  Known limitation: a user who manually clears only IndexedDB (not localStorage) will not see the welcome again without also clearing `localStorage['slicc:welcome-flow-fired']`. This is a very narrow edge case and preferable to the common-case regression.

  ## Cross-cutting impact

  - **Floats exercised**: CLI (`npx sliccy` / `npm run dev`) and Chrome extension (`mainExtension`). Both boot paths call `runFirstRunDetection` via
    `setup-standalone-trailers.ts` and `main.ts:181` respectively — both are fixed.
    Electron / Sliccstart / worker floats all reach `runFirstRunDetection` through
    the same boot chain; no float-specific code changed.
  - **Shell contexts**: no shell contexts touched.
  - **Other callers of changed functions**: `handleOnboardingComplete` is only called from `setup-welcome-flow.ts:dispatchWelcomeBranch`. The widened guard  (`'idle' || 'collect-profile'`) has no effect on that caller's behavior because both stages are now valid entries and the caller never passes stage as an argument.
    `handleConnectReady` and `handleConnectAttempt` accept `stage !== 'complete'` —  unchanged.
  - **Persisted state side-effects**: no changes to IndexedDB stores. The `slicc:welcome-flow-fired` localStorage key is now never cleared mid-session by the boot path; it is only written (added) when the welcome fires for the first time.
  - **`/shared/.welcomed` marker**: unaffected. Written by `handleOnboardingComplete` as before. The marker-based fast-path in `detectWelcomeFirstRun` is unchanged.
  - **`hasWelcomeLickInHistory` history check**: unaffected — still present as defence-in-depth for returning users whose marker was written.

  ## Test plan

    - [x] `npm run lint` — clean
  - [x] `npm run test` — 6333 passing, 10 pre-existing skips, no new failures
  - [x] `npm run build` — clean (Swift deprecation warnings are pre-existing on `main`)
  - [x] **New unit tests** in `packages/webapp/tests/scoops/onboarding-orchestrator.test.ts`
        (`describe('handleFirstRun')`) and
        `packages/webapp/tests/ui/boot/setup-onboarding.test.ts`
        (`'suppresses re-fire when first-run is already in the dedup ledger'`)
  - [x] Manual smoke (CLI):
    1. `rm -rf "$HOME/Library/Application Support/Slicc/profiles/default"`
    2. `npx sliccy` — welcome appears **once**
    3. Close without submitting profile form; restart — welcome does **not** reappear
    4. Complete full onboarding; restart — opens to existing session, no welcome
  - [ ] Manual smoke (extension): load unpacked `dist/extension/`, open side panel on

  **Pre-existing failures**: none observed on `main`.

  ## Documentation

  ## Risk & rollback

  - **Blast radius**: welcome flow only. No LLM calls, no account storage, no CDP
    involved. Regression would mean new users don't see the welcome on first install
    (detectable immediately).
  - **Rollback**: clean revert with no persisted-state migration needed. The
    `slicc:welcome-flow-fired` localStorage key exists in both old and new code;
    reverting leaves the key in place but the old code would clear and refire on it,
    which is the pre-fix behavior.
  - **CI cannot catch**: the double-fire requires a page reload between the first
    `handleFirstRun()` call and the dedup check — covered by the manual smoke steps
    above.

  ## Checklist
  
  - [x] No public API / tool / shell-command / lick-channel changes
  - [x] Contract change (`handleOnboardingComplete` guard widened) is internal to
        `OnboardingOrchestrator`; checked all callers (one: `dispatchWelcomeBranch`)

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -359,7 +359,6 @@
         "packages/webapp/src/net/link-header.ts",
         "packages/webapp/src/providers/built-in/azure-openai.ts",
         "packages/webapp/src/scoops/db.ts",
-        "packages/webapp/src/scoops/onboarding-orchestrator.ts",
         "packages/webapp/src/scoops/scoop-management-tools.ts",
         "packages/webapp/src/scoops/tray-leader-sync.ts",
         "packages/webapp/src/scoops/tray-webrtc.ts",
--- a/packages/webapp/src/scoops/onboarding-orchestrator.ts
+++ b/packages/webapp/src/scoops/onboarding-orchestrator.ts
@@ -270,29 +270,13 @@ export class OnboardingOrchestrator {
     // accounts. Without this, picking azure-openai from the welcome
     // dip used to write an account missing deployment + api-version
     // and break at the first chat request.
-    const providerEntry = (() => {
-      try {
-        return this.deps.getProviderCatalogue().providers.find((p) => p.id === provider);
-      } catch {
-        return undefined;
-      }
-    })();
-    if (providerEntry?.requiresDeployment && !deployment?.trim()) {
+    const requiredFieldError = this.validateProviderRequiredFields(provider, baseUrl, deployment);
+    if (requiredFieldError) {
       this.deps.broadcastToDip({
         type: 'slicc-connect-result',
         ok: false,
         kind: 'failed',
-        message: `${providerEntry.name} requires a deployment name.`,
-      });
-      this.stage = 'awaiting-connect';
-      return;
-    }
-    if (providerEntry?.requiresBaseUrl && !baseUrl?.trim()) {
-      this.deps.broadcastToDip({
-        type: 'slicc-connect-result',
-        ok: false,
-        kind: 'failed',
-        message: `${providerEntry.name} requires a base URL.`,
+        message: requiredFieldError,
       });
       this.stage = 'awaiting-connect';
       return;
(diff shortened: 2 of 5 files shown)


trieloff:
> Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

I'll give this one some extra scrutiny.

PR #983, opened 2026-06-12
fix(skills): stop kernel-worker OOM from monkeypatching the sudo-fs Proxy

karlpauls:
## Problem

A fresh report (commit `e60b2620`): the prompt *"using stardust create a more contemporary version of the page in the current tab"* drives the kernel worker to ~4GB and crashes it with `Ineffective mark-compacts near heap limit` — within a minute. Reproduces in **both** standalone and the extension.

## Root cause

The skill-discovery compatibility cache (`skills/catalog.ts`) invalidates itself by **monkeypatching fs write methods in place** — `mutableFs.writeFile = async (...) => { await original.apply(fs, args); cache.delete(key); }` for `writeFile`/`mkdir`/`rm`/`rename`/`mount`/`unmount`.

The agent shell hands discovery the **sudo-fs `Proxy`** (`fs/sudo-fs.ts`, the gated handle backing both file tools and the shell). That Proxy is **get/set-asymmetric**:

- `get('writeFile')` → a *gating override* that delegates dynamically to `target.writeFile`.
- `set('writeFile', …)` → writes straight **through to the wrapped target** (no `set` trap).

So when discovery runs over the gated handle, reassigning `gatedFs.writeFile`:

1. captures `original = gatedFs.writeFile` (via `get`) = the **override**, then
2. writes the new wrapper onto the **target** (`rawFs.writeFile = wrapper`).

Now the override (which re-reads `target.writeFile` each call) calls the wrapper, whose captured `original` is the override → **unbounded `override ↔ wrapper` async recursion**. The next gated write — stardust `upskill` → `playwright-cli` writing `/.playwright/session.md` — triggers it and the worker balloons until V8 OOMs.

This is intermittent by nature: boot-time `loadSkills` uses the *unwrapped* fs, so it only fires on a **first-time, in-shell skill install** (when the agent runs `upskill`/`skill list` over the gated handle). Once the skill is installed, the path isn't re-run — which is exactly the reported "sometimes" behavior.

### Forensics

CDP heap sampling during the balloon showed the sudo write-override invoked **20,000×** while `VirtualFS.writeFile` was **never reached** — the hallmark of a wrapper cycle that never bottoms out. Stack capture confirmed the frames alternate `catalog-hook ↔ sudo-override`.

### Where it was introduced

This is a *collision* of two changes, latent until both were present:

- The **in-place fs-method monkeypatch** in `catalog.ts` (`installCompatibilityCacheInvalidationHooks`) is old and was harmless on its own — it predates the package split and traces back through #187 (the monorepo restructure).
- The cycle only became reachable when **#829 (`feat(sudo): explicit native approval for sensitive agent actions`, merged 2026-06-06)** introduced `createSudoFs` and wired the **get/set-asymmetric sudo-fs `Proxy`** in as the agent shell's fs handle (`scoops/scoop-context.ts` → `WasmShell`). From that point, an in-shell `upskill`/`skill list` handed skill discovery a Proxy whose methods cannot be safely reassigned, and the latent monkeypatch turned into the override↔wrapper recursion.

So this is **not a regression introduced just before the report** — it has been live since the sudo feature merged (~2026-06-06), which matches the "it could have been there before" observation. #829 is the change to anchor the fix against; the monkeypatch is the pre-existing fragility it collided with.

## Fix

- `fs/sudo-fs.ts`: the Proxy advertises a `MONKEYPATCH_UNSAFE_FS` marker (a `Symbol.for` registry symbol) through its `get` trap.
- `skills/catalog.ts`: `getCompatibilitySkillCandidates` detects the marker and **skips both the in-place hooks and the cache** for such a handle, always re-discovering instead. The agent's skill view stays correct after an in-shell install, with no cycle. `installCompatibilityCacheInvalidationHooks` keeps a defense-in-depth guard so any future caller is safe too.

### Bonus hardening (second OOM vector)

- `shell/vfs-adapter.ts`: `readdir` now reports symlinks per Node `Dirent`/`lstat` semantics (`isFile`/`isDirectory` both `false`, `isSymbolicLink: true`) instead of resolving the target. Previously a symlink-to-directory looked like a plain directory, so `find`/`grep -r`/`ls -R` followed it — and a symlink **cycle** recursed forever, allocating path objects until OOM. POSIX tools don't follow symlinks by default; this restores that and is cheaper (no per-entry stat). It also makes the discovery BFS correctly skip symlinks.

## Verification

Reset-flail reproduction on the exact reported prompt (skills removed + cone IDB cleared so the agent installs stardust fresh):

- **Before:** heap balloons 18MB → 1.6GB+ and the worker dies within ~20s.
- **After:** the agent installs the full stardust bundle (`extract`/`direct`/`prototype`) and runs **33 `playwright-cli` `session.md` writes** — the precise writes that crashed before — with a **35.6MB peak heap**; worker survives indefinitely. Zero `RangeError`/OOM in logs.

## Gates

- ✅ `lint` ✅ `typecheck` ✅ `test` (7593 passed) ✅ `test:coverage:webapp` ✅ `build` (webapp) ✅ `build` (chrome-extension)
- New regression tests in `tests/skills/catalog.test.ts`: discovery over a real `createSudoFs` Proxy must not clobber the wrapped target's methods, a gated write must reach the real fs exactly once, and results match the unwrapped path. Symlink-Dirent tests in `tests/shell/vfs-adapter.test.ts`.

## Cross-runtime parity

The crash and fix are in the shared `webapp` core (kernel worker), so both the standalone (DedicatedWorker) and extension (offscreen) floats are fixed by the same change. No float-specific code touched.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -395,7 +395,6 @@
         "packages/webapp/src/shell/supplemental-commands/webhook-command.ts",
         "packages/webapp/src/shell/supplemental-commands/websocat-command.ts",
         "packages/webapp/src/shell/wasm-shell.ts",
-        "packages/webapp/src/skills/catalog.ts",
         "packages/webapp/src/skills/install-from-drop.ts",
         "packages/webapp/src/tools/search-tools.ts",
         "packages/webapp/src/ui/ansi.ts",
--- a/packages/webapp/CLAUDE.md
+++ b/packages/webapp/CLAUDE.md
@@ -161,6 +161,7 @@ Deep reference: `docs/kernel/process-model.md`.
 - Path: `packages/webapp/src/skills/`
 - Discovers install-managed native skills from `/workspace/skills/`.
 - Also discovers compatible read-only skill roots under `.agents/skills/*/SKILL.md` and `.claude/skills/*/SKILL.md`.
+- **Never monkeypatch an fs method in place on a get/set-asymmetric Proxy.** `catalog.ts`'s compatibility-skill cache is invalidated by wrapping `writeFile`/`mkdir`/`rm`/… in place. The agent shell hands discovery the **sudo-fs `Proxy`** (`fs/sudo-fs.ts`), whose `get` returns a gating override while `set` writes through to the wrapped target — so reassigning a gated method clobbers the target's real method and leaves the override delegating to the new wrapper, whose captured `original` is that same override: an unbounded `override↔wrapper` async recursion that OOMs the kernel worker on the next gated write (stardust `upskill` → `playwright-cli` writing `/.playwright/session.md`). The sudo Proxy therefore advertises `MONKEYPATCH_UNSAFE_FS` (a `Symbol.for` registry marker) and `getCompatibilitySkillCandidates` skips both the hooks **and** the cache for it (always re-discovers). Boot-time `loadSkills` uses the unwrapped fs, so it is unaffected — which is why the crash only reproduced on first-time, in-shell skill installs.
 
 ### Sprinkle Rendering
 
--- a/packages/webapp/src/fs/sudo-fs.ts
+++ b/packages/webapp/src/fs/sudo-fs.ts
@@ -31,6 +31,22 @@ import { FsError } from './types.js';
 
 const log = createLogger('sudo:fs');
 
+/**
+ * Marker the sudo-fs `Proxy` advertises through its `get` trap so callers that
+ * monkeypatch fs methods *in place* (e.g. the skill-discovery cache-invalidation
+ * hooks in `skills/catalog.ts`) can detect this handle and refuse to touch it.
+ *
+ * The Proxy is get/set-asymmetric: `get` returns a gating override for a gated
+ * method, while `set` writes straight through to the wrapped target. Reassigning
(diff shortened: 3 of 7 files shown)


karlpauls:
Thanks for the review. Confirming the two observations — both were deliberate:

1. **Re-walk on every gated `discoverSkillCandidates`.** Intentional. The gated (sudo-fs Proxy) path can't be hooked for cache invalidation without the recursion, so the only correct alternatives are (a) cache without invalidation — which would show a stale skill list right after an in-shell `upskill`, the worst time — or (b) always re-walk. I chose correctness. It's bounded (the BFS skips symlinks now that `entryToDirent` reports them as links, and prunes `.git`/`.slicc`) and only runs on the cone-init / `reloadSkills` / `upskill` cadence. If a hotter caller ever appears, the right move is a generation-counter cache keyed off the VFS/`FsWatcher` rather than re-introducing the in-place monkeypatch on the Proxy.

2. **`Symbol.for` cross-realm visibility.** Intended — that's exactly why it's a registry symbol rather than a module-private one: the catalog can detect the marker with no `skills → fs/sudo-fs` import, and it carries no authority (it only tells a patcher "don't reassign my methods"), so cross-realm observability is harmless.

No code changes from the review.

PR #1027, opened 2026-06-16
fix(webapp): restore leader-followers shim so standalone host shows live followers

benpeter:
## Problem

Standalone `host` runs in the kernel worker, where the live `connectedFollowersGetter` is never set (it is page/offscreen-only). So it reads the connected-follower list solely from the `slicc.leaderTrayFollowers` localStorage shim — but **nothing wrote that key**. The page-side writer was lost in the migration to the wc-shell UI, so:

- a standalone leader's `host` never reflected live followers, and
- it kept showing stale "follower" fossils from older builds that survived `nuke` / `host reset` (nothing refreshed the key — the frozen `connected Nh ago` timestamps were the tell).

The sibling `slicc.leaderTrayStatus` mirror works correctly because the page writes it on boot and on every status change (`wc-tray.ts`); the followers mirror simply lost its writer.

## Fix

Restore the page-side writer, symmetric with the status mirror:

- `writeConnectedFollowersToShim()` next to the reader in `host-command.ts`.
- `wc-tray.ts` calls it on every follower-count change, when the leader stops (→ `[]`), and on boot (seeds/clears any stale value).
- `installPageStorageSync` forwards the page write into the worker's Map-backed shim, so worker-side `host` reads the live list.

No change to the read path, and extension mode is unaffected (the offscreen leader sets the live getter and never hits the shim fallback).

## Verification

- TDD: 2 new tests in `host-command.test.ts` (worker-side read reflects a mirrored list; an empty list clears a stale value).
- `npm run lint:ci`, full `npm run typecheck` (incl. the no-DOM worker guard), `npm run test`, the webapp coverage gate, and `build -w @slicc/webapp` + `build -w @slicc/chrome-extension` — all green.

### DIFF
--- a/packages/webapp/src/shell/supplemental-commands/host-command.ts
+++ b/packages/webapp/src/shell/supplemental-commands/host-command.ts
@@ -64,7 +64,7 @@ export function getConnectedFollowersWithFallback(): ConnectedFollowerInfo[] {
 }
 
 // Same reason: the module-level getter is only set on the page thread.
-// Fall back to the localStorage shim value written by onFollowerCountChanged.
+// Fall back to the localStorage shim value written by writeConnectedFollowersToShim.
 function getFollowersWithFallback(): ConnectedFollowerInfo[] {
   if (connectedFollowersGetter) return connectedFollowersGetter();
   try {
@@ -76,6 +76,29 @@ function getFollowersWithFallback(): ConnectedFollowerInfo[] {
     // ignore parse errors
   }
   return [];
+}
+
+/**
+ * Mirror the leader's connected-follower list into the
+ * `slicc.leaderTrayFollowers` shim so the standalone kernel-worker `host`
+ * command — which has no live `connectedFollowersGetter` — sees it via
+ * `getFollowersWithFallback`. The page calls this on every follower-count
+ * change, on boot (to clear a stale value), and when the leader stops;
+ * `installPageStorageSync` forwards the write into the worker's Map-backed
+ * shim. Symmetric with the `slicc.leaderTrayStatus` mirror written in
+ * `wc-tray.ts`. Best-effort: a missing `localStorage` (Node/tests without an
+ * override) or a quota/serialization error is swallowed.
+ */
+export function writeConnectedFollowersToShim(
+  followers: ConnectedFollowerInfo[],
+  storage: Pick<Storage, 'setItem'> | undefined = (globalThis as { localStorage?: Storage })
+    .localStorage
+): void {
+  try {
+    storage?.setItem(LEADER_FOLLOWERS_STORAGE_KEY, JSON.stringify(followers));
+  } catch {
+    // best-effort mirror — ignore quota / serialization failures
+  }
 }
 
 // When `host reset` runs from the kernel-worker's terminal, the worker's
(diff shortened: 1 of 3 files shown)


trieloff:
Thanks Ben, I've added you to the repo, so next PR does not need a fork.

trieloff:
Claude, when you review this, consider if this highlights also a lack of wiring in the UI, we should show accurate follower state there, and this bug makes me suspicious about that.

trieloff (commented):
Correct fix, clean approach — the symmetric write/read pattern mirrors what slicc.leaderTrayStatus already does.

packages/webapp/src/shell/supplemental-commands/host-command.ts
```diff
@@ -78,6 +78,29 @@ function getFollowersWithFallback(): ConnectedFollowerInfo[] {
   return [];
 }
 
+/**
```
trieloff, line 81: Style nit: CLAUDE.md says "never write multi-paragraph docstrings or multi-line comment blocks — one short line max." This block is 10 lines. The WHY is genuinely non-obvious and worth keeping, but it should collapse to something like:

```ts
// Page-side mirror so the kernel-worker `host` (no live getter) reads the live list via getFollowersWithFallback.
```

packages/webapp/tests/shell/supplemental-commands/host-command.test.ts
```diff
@@ -95,6 +98,70 @@ describe('host command', () => {
...
+      expect(getConnectedFollowersWithFallback()).toEqual(followers);
+    } finally {
+      if (original === undefined) delete (globalThis as { localStorage?: Storage }).localStorage;
+      else (globalThis as { localStorage?: Storage }).localStorage = original;
+    }
+  });
+
+  it('writing an empty follower list clears a stale shim value', () => {
+    setConnectedFollowersGetter(null);
+    const store = new Map<string, string>([
```
trieloff, line 139: The fake-localStorage boilerplate (`store`, `fakeLs`, save/restore of `globalThis.localStorage`) is identical in both new tests. A small helper at describe-scope would remove ~20 lines of duplication:

```ts
function withFakeLocalStorage(fn: (ls: Storage) => void) {
  const store = new Map<string, string>();
  const fakeLs = {
    getItem: (k: string) => store.get(k) ?? null,
    setItem: (k: string, v: string) => { store.set(k, v); },
    removeItem: (k: string) => { store.delete(k); },
  } as unknown as Storage;
  const original = (globalThis as { localStorage?: Storage }).localStorage;
  (globalThis as { localStorage?: Storage }).localStorage = fakeLs;
  try { fn(fakeLs); }
  finally {
    if (original === undefined) delete (globalThis as { localStorage?: Storage }).localStorage;
    else (globalThis as { localStorage?: Storage }).localStorage = original;
  }
}
```
Non-blocking — tests are correct as-is.

trieloff:
Checked the UI wiring re @trieloff's question: the UI was always correct. The floatbar label ("tray · N followers") is updated directly by the `onFollowerCountChanged` live callback (no shim involved), and the avatar menu reads from an in-memory `leaderTrayRuntimeStatus` module variable that's updated in real-time on the page thread. Neither path touches `slicc.leaderTrayFollowers`.

The detailed follower list (runtimeIds, connected-at timestamps) is only rendered by the `host` shell command, which runs in the kernel worker where the live getter is never set — that's the exact path this PR fixes.

So no UI gap to plug here. The bug was isolated to the worker thread.

PR #1082, opened 2026-06-19
chore(docs): Add Biome Gates to Verificaitons

bstopp:
This would split out the verification docs to their own file (keeping `CLAUDE.md` under 30k) but also inform Claude that it needs to perform the biome checks to prevent those from causing errors when PRs are opened.


bstopp:
@trieloff  & @karlpauls - def would like your input on this, if it doesn't align with your doc goals.

trieloff (approved):

PR #1084, opened 2026-06-19
Thin-extension CDP-bridge architecture (breaking release)

trieloff:
## Summary

Breaking release that converts the Chrome extension from a self-contained fat app into a **thin CDP-bridge**. The UI is now served from the hosted webapp (`www.sliccy.ai`, or `localhost:5710` in dev) and connects to a local bridge; the extension ships only a service-worker relay + content-script bootstrapper. Eliminates all same-host API assumptions and unifies browser-permission prompts behind a single cross-realm Permission Surface.

## Why

The fat extension forced same-host API assumptions, duplicated CDP plumbing across panel/offscreen, and bundled an offscreen agent engine that complicated MV3 lifecycle handling. Moving the UI to a hosted origin and shrinking the extension to a CDP bridge removes that duplication, lets the agent run in a single place, and unifies permission prompts behind one user-gesture surface.

## Approach / Implementation notes

- **Thin bridge / leader-follower** — pinned "home" tab is the tray leader (runs the agent); injected iframes auto-follow. `chrome.debugger` is proxied from the leader through the SW.
- **Origin split** — hosted UI (`:8787` wrangler harness / prod `sliccy.ai`) vs. local bridge/UI (`:5710`). Agent-initiated HTTP routes through `createProxiedFetch()`; cross-origin `/api` is bridge-token gated.
- **Permission Surface** — `<slicc-permissions>` web component handles user-gesture-gated USB / HID / Serial / FileSystem pickers in the leader tab.
- **Dead-code cleanup (Wave 9)** — removed the fat-extension UI (`index.html`/`offscreen.html`), offscreen agent engine, `electron-overlay`, duplicate CDP proxies (`DebuggerClient`/`OffscreenCdpProxy`/`PanelCdpProxy`), `bridge-transport`, and dropped the `sidePanel` + `offscreen` MV3 permissions.

## Cross-cutting impact

- **Floats exercised**: Chrome extension (thin-bridge), CLI / webapp (hosted-leader origin), worker (`sliccy.ai` UI host). Electron and Sliccstart not directly modified in this branch.
- **Shell contexts** (extension): offscreen `AlmostBashShell` removed entirely with the offscreen agent engine; agent shell now runs in the hosted UI realm. Side-panel `AlmostBashShell` removed with the side-panel UI; sandbox iframe path unchanged.
- **MV3 permissions reduced**: kept `debugger` + `<all_urls>` + `externally_connectable: www.sliccy.ai`; dropped `sidePanel` and `offscreen`.
[...]

PR #1085, opened 2026-06-19
slicc-5: thin-extension CDP bridge + ipk/ipx package manager (breaking release)

trieloff:
## Summary

The **slicc-5** breaking release. It folds three bodies of work into one linear branch:

1. **Thin-extension CDP bridge (#1084):** the Chrome extension becomes a thin CDP-emulator/bridge; the full UI + agent engine is served from the hosted webapp (`www.sliccy.ai`, or `localhost:5710` in dev), driven by the extension as a tray **leader** in a pinned tab. Removes the fat-extension UI, offscreen agent engine, electron overlay, and duplicate CDP plumbing.
2. **`ipk`/`ipx` package manager (#1066):** an npm/npx-style package manager for the `jsh` runtime. Packages install from the public npm registry into a real `node_modules`; the realm's module surface moves to Node-standard semantics (`require('sliccy:…')` for SLICC capability bridges, real CJS/ESM `require`/`import`).
3. **Total runtime-CDN elimination + a unified permission surface:** every shell command that previously pulled its wasm/JS from a CDN at runtime now resolves its backing assets from `ipk`-installed `node_modules` via the preview Service Worker (standalone) or the bundled `sandbox.html` (extension). All gesture-gated browser permissions route through a single in-tab `<slicc-permissions>` surface with parity across standalone, extension, and electron.

## Why combined

All three are breaking changes landing in the same release window. The thin-extension cleanup removes the offscreen agent shell; `ipk`'s realm refactor and the CDN-elimination both target the surviving floats (kernel-worker + `sandbox.html` iframe). Shipping them together avoids three consecutive breaking releases and redundant migration windows.

## Breaking changes

- **Extension:** no longer bundles the webapp UI or an offscreen agent engine; loads the hosted UI. MV3 permissions reduced — kept `debugger` + `<all_urls>` + `externally_connectable: www.sliccy.ai`; dropped `sidePanel` + `offscreen`.
- **jsh realm:** SLICC capability bridges are no longer bare globals — `.jsh` skills referencing bare `exec`/`http`/`browser`/`fs`/etc. must move to `require('sliccy:…')` / `require('fs')`. Out-of-tree skills relying on the old bare globals break until updated.
- **No runtime CDN for tooling:** the heavy commands below are **inert until their backing package is installed** with `ipk add`. Running one without its package exits non-zero with a guidance error pointing at the exact `ipk add` line — zero network traffic on the not-installed path. Model weights for `hear`/`say` must be fetched with `hf download` before use; there is no silent CDN auto-fetch (`allowRemoteModels=false`).

## Commands touched (runtime-CDN elimination)

| Command | Before | After |
| --- | --- | --- |
| `esbuild` | esm.sh / CDN wasm | `ipk add esbuild-wasm` → wasm + JS API from VFS `node_modules`; inert + guidance error otherwise |
| `biome` | CDN wasm | `ipk add @biomejs/wasm-web @biomejs/js-api` → from `node_modules`; inert otherwise |
| `tsc` / `test` | CDN | `ipk add typescript` → shared `getTypeScript()` loader from `node_modules`; inert otherwise |
| `convert` / `magick` | CDN wasm | ImageMagick wasm from `ipk`-installed `node_modules`, no CDN fallback |
| `ffmpeg` | CDN core | `ipk add @ffmpeg/core` required; no CDN; camera/video capture routed through `<slicc-permissions>` |
| `node -e` / `.jsh` / `.bsh` | esm.sh `?bundle` + jsdelivr Function-constructor | esm.sh ripped; flow is now `ipk install` → `esbuild` bundle → run |
| `hear` (Whisper STT) | transformers.js CDN weights | library via `ipk`; weights via `hf download onnx-community/whisper-tiny` into `/workspace/models`; `allowRemoteModels=false` |
| `say` (Kokoro TTS) | transformers.js CDN weights | library via `ipk`; weights via `hf download onnx-community/Kokoro-82M-v1.0-ONNX`; `allowRemoteModels=false` |
| `python3` / `python` | Pyodide CDN | JS loader resolves from `ipk`-installed `node_modules/pyodide`; `PYODIDE_RUNTIME_CDN` retained **only** for Pyodide's wheel ecosystem (sanctioned exception) |
| `hf` (new) | — | downloads model repos from the Hugging Face Hub into the VFS via the captured `SecureFetch` proxy seam (same as `ipk`/`upskill`); default `--to /workspace/models/<repo>/` |
| `ipk` / `ipx` (#1066) | — | npm/npx-style installer + runner from the public registry into a real `node_modules` |

**Sanctioned runtime fetches (the only ones that remain):** model weights via the explicit `hf download` command, and Pyodide's wasm/stdlib/wheels via `PYODIDE_RUNTIME_CDN`. Everything else is local-VFS or fails closed with guidance.

## Unified permission surface (`<slicc-permissions>`)

All gesture-gated browser permissions route through one cross-realm surface (`getLeaderPermissionsSurface()` → `<slicc-permissions>`), with the gesture originating from a real keystroke/click in the leader tab. Documented in `docs/approvals.md`.

| Permission | Triggers | Standalone | Extension | Electron |
| Microphone | composer push-to-talk, `hear` live capture | in-tab prompt | in-tab prompt | in-tab prompt |
| Camera / video | `ffmpeg -f avfoundation`, composer photo/video capture → attachment | in-tab prompt | in-tab prompt | in-tab prompt |
| Screen-share | `getDisplayMedia` paths | in-tab prompt | in-tab prompt | in-tab prompt |
| USB | `usb request`, `esptool` (no `--port`) | terminal keystroke → picker | picker-popup window | terminal keystroke → picker |
| HID | `hid request` | terminal keystroke → picker | picker-popup window | terminal keystroke → picker |
| Serial | `serial request`, `esptool` | terminal keystroke → picker | picker-popup window | terminal keystroke → picker |
| File System Access | `mount`, folder drag-drop | terminal keystroke / drop → picker | picker-popup window | terminal keystroke → picker |

Deny/cancel on any row emits `slicc-permission-deny` with `reason: 'cancelled'` and surfaces a clean error (no silent no-op). Chromium-only device pickers are unavailable in the cloud/hosted-leader float.

## Automated verification (CI gates green on the combined branch)

`lint`, `typecheck` (7 projects), `test` (8,995+ tests, 0 failures), `test:coverage` (all package floors held), `build`, `build -w @slicc/chrome-extension`. Linear history enforced (no merge commits). `dist`-grep confirms no new runtime-CDN literals introduced by the elimination waves; remaining literals are the two sanctioned exceptions above plus inert/doc-link strings.

## Manual end-to-end test plan (Wave 13 — `:8787` wrangler harness)

CI cannot drive `requestDevice` / `getUserMedia` dialogs headlessly. After a release build, walk these against the local wrangler UI (`packages/cloudflare-worker/`, port `8787`) so the popup paths exercise the deployed `picker-popup.html`. Run each on **all three surfaces** (standalone, extension, electron) unless noted. ✅ pass · ⚠️ noted asymmetry · ❌ regression (fix before merge).

### Network (no runtime CDN)
- [ ] `ipk add esbuild-wasm` then `esbuild --version` / `--bundle` — works from `node_modules`; **before** install, `esbuild` exits non-zero with the `ipk add` hint and **zero** network traffic.
- [ ] `ipk add @biomejs/wasm-web @biomejs/js-api` then `biome check` — same install-gated behavior.
- [ ] `ipk add typescript` then `tsc --noEmit` and `test` — same.
- [ ] `ipk add` an ImageMagick build then `convert -resize 800x600 in.jpg out.jpg` — no CDN.
- [ ] `node -e` / a `.jsh` using a bare npm specifier — resolves via `ipk install` → `esbuild` bundle → run; **no** esm.sh/jsdelivr request in the network panel.
- [ ] Confirm via DevTools Network that the only external fetches are `hf download` (when invoked) and Pyodide wheel/wasm URLs.

### Permissions (each surface)
- [ ] `mount /workspace/scratch` — directory picker opens → mount appears in file tree.
- [ ] Drag a folder onto the panel — mount appears (drop is the gesture; no popup).
- [ ] `usb request` (terminal) and agent-issued `usb request` (cone card) — chooser → `usb list` shows the handle; deny → clean error.
- [ ] `hid request --vid 0x594d` — filters applied; multi-interface device registers every matching interface.
- [ ] `serial request` + `esptool flash …` without `--port` — routes through the serial picker; once granted esptool drives the same handle.
- [ ] Composer mic (push-to-talk) — `getUserMedia` prompt in the hosted leader tab.
- [ ] `ffmpeg -f avfoundation -i 0 -frames:v 1 photo.jpg` (after `ipk add @ffmpeg/core`) — `<slicc-permissions>` camera pre-prompt → Allow → photo lands in VFS; **deny** → clean `camera permission denied`, no capture.
- [ ] Composer photo/video capture — captured media enters the composer as an attachment.

### jsh runtime (#1066)
- [ ] `require('sliccy:exec')` / `require('fs')` work; a `.jsh` using the old bare `exec`/`fs` globals fails with a clear migration error.
- [ ] `ipk add <pkg>` / `ipx <pkg>` install + run from the public registry into `node_modules`; parity in both the kernel-worker (standalone) and `sandbox.html` iframe (extension).

### wasm / weights
- [ ] `hf download onnx-community/whisper-tiny` then `hear` — transcribes; **without** weights, `hear` surfaces a clean `model not found` with the `hf download` hint (no CDN fetch).
- [ ] `hf download onnx-community/Kokoro-82M-v1.0-ONNX` then `say "hello"` — synthesizes; composer audio mode records → transcribes → plays back.
- [ ] `python3 -c "print(1+1)"` — Pyodide loads from `node_modules/pyodide`; wheel install pulls only from the sanctioned `PYODIDE_RUNTIME_CDN`.

## Merge-prep note

`node-server` coverage floors on this branch lag `main` due to a post-branch auto-ratchet; a pre-merge rebase onto `origin/main` reconciles them.

## Deferred (post-release, tracked)

- SW `panel`/`offscreen` message-branch removal (behavioral refactor, still test-covered).
- knip barrel re-export warnings (refactor, not deletion).
- CRXJS build-system migration (eval, gated on the thin model being stable).
- Kokoro voice `.bin` URL is a pre-existing literal that would route through the proxy at synthesis time — flagged for Wave 13 manual smoke.

## Stabilization landings (post-branch hardening)

Follow-up hardening that landed on the branch after the initial slicc-5 body was written. These close a thin-bridge `/api` security-parity gap between the Node and Swift servers, fix a cross-origin overlay regression, and add the matching regression test.

### Thin-bridge `/api` security parity (Swift ↔ Node)

The Node server already gated cross-origin `/api/*` behind a **two-factor** check: an Origin allowlist **and** a per-process bridge token (`X-Bridge-Token`). The Swift server only checked the Origin allowlist, so any script on an allowlisted *remote* origin could call the local `/api` with no token. The Swift side now mirrors Node byte-for-behavior:

- `BridgeSecurity.swift` — `bridgeTokenHeader`, `isLoopbackBridgeOrigin` (localhost / 127.0.0.1 / ::1), and a constant-time `validateBridgeToken` (length-mismatch → false).
- `ThinBridgeCorsMiddleware.swift` — `403 bridge-token-required` for non-loopback allowlisted origins without a valid token; loopback exempt (same-origin); `OPTIONS` preflight short-circuits **before** the gate (`204` + PNA).

### BUG-F4 — CORS mounted whenever a bridge token is present

The Electron overlay loads cross-origin from the hosted leader and fetches `/api/runtime-config` from the local server. The thin-bridge CORS/PNA middleware was only mounted in canonical thin-bridge mode, so the overlay (e.g. under `--electron`) got no CORS headers. Both servers now share a `shouldMountThinBridgeCors` predicate that mounts the middleware whenever a bridge token exists — not just when `thinBridgeMode` is true.

### FIX-A — overlay runtime-config routes through the bridge

`fetchRuntimeConfig` now resolves through `resolveApiUrl` + `apiHeaders()` so the local runtime-config call carries the bridge headers (empty record in the same-origin case). The matching unit test (`tray-runtime-config.test.ts`) was updated to the new fetch shape.

### Commit chain

| Commit | Scope |
| --- | --- |
| `70ae2157` | Swift `BridgeSecurity` primitives (token header, loopback detection, constant-time validate) |
| `58156ef8` | Swift `/api` token gate in `ThinBridgeCorsMiddleware` + unit tests |
| `b867b25e` | BUG-F4: shared `shouldMountThinBridgeCors` (Node + Swift `ServerCommand` token threading) |
| `b34518a2` | FIX-A regression test alignment (`tray-runtime-config.test.ts`) |

### Verification (HEAD `b34518a2`)

Authoritative matrix green: `lint:ci`, `typecheck`, `test:coverage` (556 files / **9360 tests**, 0 failures, all floors held), `build`, `build -w @slicc/chrome-extension`, `swift build`, `swift test` (**434 / 0**, run with a full-Xcode `DEVELOPER_DIR` — bare `swift test` under Command Line Tools lacks `XCTest`).

The Swift `/api` gate is additionally covered at three levels:

- **Unit** — 23 `BridgeSecurityTests` + 8 `ThinBridgeCorsMiddlewareTests`.
- **HTTP layer** — 31/31 through Hummingbird's real request→response chain.
- **Live process** — 7/7 on a real swift-server rig (no Origin → 200; loopback → 200; remote no/wrong token → 403; remote correct token → 200 + ACAO; `OPTIONS` → 204 + `Access-Control-Allow-Private-Network: true`; disallowed origin → 200 with no CORS leak).


packages/chrome-extension/dnr-frame-ancestors.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "id": 1,
+    "priority": 1,
+    "condition": {
+      "urlFilter": "||sliccy.ai",
```
trieloff: [P1] Scope the frame-ancestors DNR override

When an extension user visits any page that frames a non-cherry `sliccy.ai` URL, this rule replaces the worker's `frame-ancestors 'none'` CSP with `frame-ancestors *` for all `sliccy.ai` subframes, not just the intended embedded launcher/cherry surface. That makes the hosted SLICC UI frameable by arbitrary third-party pages and bypasses the server-side anti-clickjacking policy.

packages/webapp/src/shell/supplemental-commands/ffmpeg-command.ts
```diff
@@ -335,6 +400,125 @@ export function isAvfoundationCapture(parsed: ParsedFfmpegInvocation): boolean {
...
+  // Page realm: ask the in-tab surface directly when one is mounted.
+  if (typeof window !== 'undefined') {
+    try {
+      const { getLeaderPermissionsSurface } = await import(
+        '../../ui/wc/wc-permissions-registry.js'
+      );
+      const surface = getLeaderPermissionsSurface();
+      if (surface) {
+        const result = await surface.prompt({ kinds, description });
+        if (result.status === 'granted') return { ok: true };
```
trieloff: [P2] Stop streams returned by the ffmpeg permission probe

When `ffmpeg` runs in a DOM-backed shell and requests camera/mic permission through the unified surface, `prompt()` returns live `MediaStream` grants. This branch only checks `status` and then opens the real capture path, so the priming stream is never stopped and a duplicate camera/mic stream can remain active after the command.

packages/webapp/src/ui/panel-rpc-handlers.ts
```diff
@@ -748,6 +871,77 @@ function buildRemoteCdpHandlers(options: StandalonePanelRpcHandlerOptions) {
...
+            break;
+          }
+          case 'camera':
+          case 'microphone':
+          case 'screenshare': {
+            // The MediaStream can't cross the bridge; the page retains it
+            // (the surface emits a `slicc-permission-grant` event the
+            // host can subscribe to for stream handling) and the worker
+            // only learns the kind succeeded.
+            out.push({ kind: grant.kind, ok: true });
```
trieloff: [P2] Release media grants before returning worker permission ok

When a worker uses `permission-request` only to gate a later capture, `surface.prompt()` has already opened camera/mic/screen streams that cannot cross the bridge. This branch records only `{ ok: true }` and leaves the page-side stream alive, so the permission probe can keep hardware capture active while the worker opens a second stream for the actual capture.

trieloff (commented):
The patch introduces a broad CSP override that makes the hosted UI frameable by arbitrary pages, and both new media permission-probe paths leak live capture streams. These are blocking correctness/security issues that should be fixed before merging.

packages/chrome-extension/dnr-frame-ancestors.json
trieloff: Fixed in `88208b05`. Narrowed the DNR `urlFilter` from the blanket `||sliccy.ai` to `||sliccy.ai/?cherry=1`, so the `frame-ancestors *` override now applies **only** to the `?cherry=1` follower-embed `sub_frame` surface — the `?slicc=leader` hosted UI keeps its server-side `frame-ancestors 'none'`. Added a DNR rule-shape test (`packages/chrome-extension/tests/dnr-frame-ancestors.test.ts`) asserting `urlFilter === '||sliccy.ai/?cherry=1'`, `resourceTypes === ['sub_frame']`, operation `set`, plus a regression note guarding against re-broadening to all sliccy.ai subframes.

packages/chrome-extension/src/bridge-sw.ts
```diff
@@ -0,0 +1,503 @@
...
+): Promise<Record<string, unknown>> {
+  const { method, params, sessionId } = req;
+
+  // Per-port `Target.*` shims map to chrome.tabs operations, so the leader
+  // sees a real CDP-like target surface without us hard-coding tab ids.
+  if (method === 'Target.getTargets') return cdpGetTargets(state, deps);
+  if (method === 'Target.attachToTarget') return cdpAttachToTarget(params ?? {}, state, deps);
+  if (method === 'Target.detachFromTarget') return cdpDetachFromTarget(params ?? {}, state, deps);
+  if (method === 'Target.createTarget') return cdpCreateTarget(params ?? {}, deps);
+  if (method === 'Target.closeTarget') return cdpCloseTarget(params ?? {}, state, deps);
```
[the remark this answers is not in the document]
trieloff: Fixed in `88208b05` via **Option B** (skip the watcher for this transport). Added `readonly isExtensionBridge?` to `CDPTransport` (mirrors the existing `superseded?` precedent) and set it on `ExtensionBridgeTransport`; `startNavigationWatcherForHost` now early-returns for that transport, so it no longer sends the sessionless `Target.setDiscoverTargets` this bridge would reject (no behavior regression — the watcher tore down immediately on this transport anyway). Chose B over synthesizing target-discovery in the bridge to keep the change contained and avoid forcing `chrome.debugger.attach` + per-tab banners.

Note: the *separate*, pre-existing SW→leader `navigate-lick` **delivery** gap (the thin-ext leader tab has no `onMessage` receiver for the handoff `Link` header) is out of scope for this PR and now tracked as #1127.

packages/webapp/src/ui/panel-rpc-handlers.ts
trieloff: Fixed in `88208b05`. The `permission-request` handler now stops all tracks of the camera/microphone/screenshare probe `MediaStream` (`track.stop()`) before recording `{ ok: true }`, so the page-side probe no longer keeps hardware capture alive while the worker opens the real stream for the actual capture. Test in `packages/webapp/tests/ui/panel-rpc-handlers.test.ts` updated to assert the probe tracks are stopped.

packages/webapp/src/shell/supplemental-commands/ffmpeg-command.ts
trieloff: Fixed in `88208b05`. `requestCapturePermission` now stops the page-realm probe stream's tracks after the grant check and before the real capture path opens, so no duplicate cam/mic stream is left active after the command. Extracted `stopProbeStreamTracks()` + `tryPageRealmCapturePermission()` helpers (which also brought the function back under the cognitive-complexity cap). Regression test added in `packages/webapp/tests/shell/supplemental-commands/ffmpeg-command.test.ts` asserting the probe tracks are stopped.

packages/chrome-extension/vite.config.ts
```diff
@@ -175,13 +309,21 @@ function buildSliccDiffPlugin() {
...
 function writeExtensionManifest(): void {
   const manifest = JSON.parse(readFileSync(resolve(Dirname, 'manifest.json'), 'utf-8'));
   manifest.version = rootPkg.version;
   if (process.env['SLICC_EXT_DEV']) {
     delete manifest.key;
+    if (manifest.externally_connectable?.matches) {
+      manifest.externally_connectable.matches = [
+        ...manifest.externally_connectable.matches,
+        'http://localhost:*/*',
+        'http://127.0.0.1:*/*',
```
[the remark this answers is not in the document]
trieloff: Fixed in `88208b05`. Removed the invalid `:*` port wildcards from the `SLICC_EXT_DEV=1` `externally_connectable.matches` augmentation: `http://localhost:*/*` / `http://127.0.0.1:*/*` → `http://localhost/*` / `http://127.0.0.1/*`. Chrome match patterns reject `:*`, so the dev manifest is now valid — verified the built dev manifest contains zero `:*` entries.

PR #1093, opened 2026-06-22
fix(kernel): silence cross-sibling VFS "drop unmatched response" spam

benpeter:
## Problem

In the standalone (kernel-worker) float the page runs **two** VFS clients on a **single** `KernelTransport`: the read-only `RemoteVfsClient` and the read+write `RemoteWritableVfsClient`. Both match `vfs-read-*-result` envelopes, so every response is delivered to *both* `handleResult` methods — and the one that didn't issue the request has no pending entry for it.

The old code treated "no pending entry" as an anomaly and logged a `drop unmatched response` debug line for **every** sibling response. Those debug lines are relayed page→node-server, so under tray-leader load the log filled with:

```
[remote-writable-vfs-client] drop unmatched response { ... }
```

(One line per read the *other* client issued.)

## Root cause

The two clients shared an overlapping id space and neither could tell its own responses from a sibling's, so the "unmatched ⇒ bug" assumption was wrong on a shared wire with two demuxers.

## Fix

- Give the clients **disjoint** request-id namespaces: `vfs-r-` (reader) and `vfs-w-` (writer) — deliberately chosen so neither is a prefix of the other.
- In each `handleResult`, only log a drop when the unmatched id is in **that client's own** namespace (a genuine duplicate/late host reply). A sibling's response is now ignored silently.

No behavior change beyond logging: the owning client still resolves/rejects exactly as before.

## Tests

Adds a regression block (`writable-vfs-client.test.ts`) that wires a reader + writer onto one shared transport and asserts that a read issued by one client does **not** produce a drop log on the other — covering both directions.

## Verification

- `vitest run` on both kernel client suites — 35 pass
- `npm run typecheck` — all targets clean
- `biome` / `prettier` — clean

Found while debugging a tray-leader log flood; this is the standalone-only spam half. (The other half of that investigation — a duplicate-tab CDP eviction war from Chrome session-restore after an unclean `Ctrl-C` — is a separate branch/PR.)

### DIFF
--- a/packages/webapp/src/kernel/remote-vfs-client.ts
+++ b/packages/webapp/src/kernel/remote-vfs-client.ts
@@ -108,6 +108,14 @@ interface PendingRequest {
 /** Default per-request timeout (ms). */
 const DEFAULT_REQUEST_TIMEOUT_MS = 30_000;
 
+/**
+ * Request-id namespace for the read client. Disjoint from the writable
+ * client's `vfs-w-` prefix (note: NOT a prefix of it) so the two clients
+ * — which share one transport and both match `vfs-read-*-result` envelopes
+ * — can each tell its own responses from a sibling's. See `handleResult`.
+ */
+const READ_REQUEST_ID_PREFIX = 'vfs-r-';
+
 class RemoteVfsClient implements RemoteVfsClientHandle {
   private readonly transport: KernelTransport<ExtensionMessage, PanelToOffscreenMessage>;
   private readonly log: NonNullable<RemoteVfsClientOptions['logger']>;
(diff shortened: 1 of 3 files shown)


trieloff:
We'll pick this up in #1085 (the slicc-5 thin-extension + ipk/ipx breaking release).

Verified compatible and clean: #1085 doesn't touch the VFS client files, it merges with no conflicts, and the semantic target is intact — on #1085 both `handleResult` paths still log the unconditional `drop unmatched response`, and the reader still uses the generic `vfs-` id prefix (which is a prefix of the writer's `vfs-w-`). Your disjoint `vfs-r-`/`vfs-w-` namespacing + own-namespace drop-log gate applies as-is. Folding it into #1085.

trieloff:
The VFS `vfs-r-`/`vfs-w-` disjoint-namespace fix and the drop-log gating are now in `main` (rebased into the slicc-5 / #1085 line, new SHAs). Merging this branch into `main` is a no-op. Closing as already-landed.

PR #1099, opened 2026-06-22
fix(tray): recover a stale stored tray at boot; hide stale followers when inactive

benpeter:
> Draft — found while testing leader setup on the integration branch (alongside #1093 and #1096). These are the two papercuts behind "host won't become a leader anymore / shows a dead follower."

## Symptoms

1. After a tray expired, a fresh `npm run dev` came up **`inactive`** instead of leading — and `host leave --leader <worker>` appeared to *hang* (~60–90s) before recovering. Logs showed boot auto-lead hitting `POST …/controller/<stale-tray> ← 502 (fetch failed)` → `Leader tray start failed` → inactive; the fresh-tray recovery only fired later on the explicit leave→lead.
2. `host` listed a follower "connected 3h ago" under `status: inactive` — a stale entry from the persisted shim shown as if live.

## Root causes

1. **Stale-tray recovery was too narrow.** Reusing the stored tray session goes through the node-server fetch-proxy. When the tray is gone the worker is unreachable *at the proxy layer*, surfaced as a generic "Proxy fetch failed" `Error` — **not** a `LeaderTrayHttpError`. `shouldRecreateTray` only matched `LeaderTrayHttpError` 403/404/410, so this threw and the leader stayed inactive instead of re-minting.
2. **Followers shown without a session.** `formatLeaderOutput` listed followers whenever the shim was non-empty, ignoring `status.state`/`status.session`. The shim persists across sessions, so a leftover entry showed under an inactive tray.

## Fixes

1. Introduce a typed **`TrayProxyFetchError`** (thrown by `createTrayFetch` on a tagged proxy/transport failure) and **broaden `shouldRecreateTray`** to also recover on it **and** on 5xx. The stored session is just a cache; the recovery retry passes `session=null` (not recreate-eligible), so it can't loop. Now a boot whose stored tray expired re-mints a fresh tray automatically instead of going inactive.
2. Gate the follower list on an active **`status.session`** (consistent with the `join_url` line) so stale followers never show under `inactive`.

## Cross-runtime parity

Tray-leader + host command are shared webapp code; no float-specific branches touched. N/A for swift/ios.

## Tests

- `tray-leader`: recovers to a fresh tray when the stored controller attach fails with a proxy-transport error (the actual boot case) **and** when it returns a 5xx; existing 410 recovery still passes.
- `host-command`: does not list stale followers when there's no active session.

Local verification: `lint` (biome+prettier), `typecheck` (all 7 targets), affected suites (87 pass), `build -w @slicc/webapp`. CI runs the full gate.

### DIFF
--- a/packages/webapp/src/scoops/tray-leader.ts
+++ b/packages/webapp/src/scoops/tray-leader.ts
@@ -670,8 +670,36 @@ class LeaderTrayHttpError extends Error {
   }
 }
 
+/**
+ * Thrown by {@link createTrayFetch} when the node-server `/api/fetch-proxy`
+ * itself failed to reach the tray worker (a tagged proxy/transport error, not a
+ * real upstream HTTP status). Typed so `shouldRecreateTray` can recognise it
+ * and mint a fresh tray instead of leaving the leader inactive — a dead stored
+ * tray surfaces here as "Proxy fetch failed", not as a LeaderTrayHttpError.
+ */
+export class TrayProxyFetchError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'TrayProxyFetchError';
+  }
+}
+
 function shouldRecreateTray(error: unknown): boolean {
-  return error instanceof LeaderTrayHttpError && [403, 404, 410].includes(error.status);
+  // A stored tray session is just a cache. If reusing it fails because the tray
+  // is gone (403/404/410), the worker is failing (5xx), or the proxy transport
+  // itself failed (worker unreachable → node-server returns a tagged proxy
+  // error), discard it and mint a fresh tray rather than leaving the leader
+  // inactive. `attachWithRecovery` only reaches here with a stored session and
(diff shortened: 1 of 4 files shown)


trieloff:
We'll pick this up in #1085 (slicc-5 breaking release).

Compatible as-is: it overlaps `tray-leader.ts` with #1085 but merges with no conflicts, and the semantic targets are intact — on #1085 `shouldRecreateTray` is still the narrow `error instanceof LeaderTrayHttpError && [403,404,410]` check, and `formatLeaderOutput` still lists followers gated only on `followers.length > 0` (not on `status.session`). Your typed `TrayProxyFetchError` recovery + session-gated follower list apply cleanly. Folding it into #1085.

trieloff:
The stale-tray boot recovery and stale-follower hiding are now in `main` (rebased into the slicc-5 / #1085 line). Merging this branch adds nothing. Closing as already-landed.

PR #1100, opened 2026-06-22
fix(wc): start the tray leader on boot even if sprinkle discovery stalls

benpeter:
> Draft — found while testing leader setup on the integration branch (with #1093, #1096, #1099). This is the one that makes a fresh `npm run dev` *deterministically* lead.

## Symptom

A fresh standalone `npm run dev` comes up `status: inactive` and never starts a tray leader — even though `/api/runtime-config` returns a worker and `slicc.trayWorkerBaseUrl` is seeded in localStorage. It's non-deterministic: sometimes it leads, sometimes not.

## Root cause (confirmed with boot instrumentation)

`mountWcUiLive` → `attachWcClient` wires the tray inside a single chain:

```js
const sprinkles = await wireWcSprinkles({...});   // ← awaited
boot.onClientReady(() => sprinkles.resync());
if (standalone && instanceId) await wireWcTray({...});  // ← tray leader (startInitialRole)
```

and `wireWcSprinkles` ended with `await resync()`, where `resync()` awaits `manager.refresh()` → VFS sprinkle discovery (`for await (... of fs.walk(root))`). When that discovery **stalls** (a kernel-gated walk that doesn't resolve), `wireWcSprinkles` never returns, so `wireWcTray` → `startInitialRole` never runs and the leader never starts. Temp `[boot-diag]` logging showed the trace stop dead at `calling wireWcSprinkles` with no `wireWcSprinkles done`, across every reload.

Unlike the freezer rail and sprinkles (both re-run on `onClientReady`), the tray wiring had **no recovery** — so a single stalled discovery stranded leadership permanently.

## Fix

Fire the initial `resync()` in the **background** instead of awaiting it:

```js
void resync().catch((err) => log.warn('WC shell: initial sprinkle resync failed', err));
return { manager, zone, resync };
```

Discovery is best-effort, already re-run on kernel-ready (idempotent), and the instant dock rail is seeded *before* it — so nothing downstream needs it to have completed. `wireWcSprinkles` now returns promptly, `wireWcTray` runs, and `startInitialRole` leads deterministically from the seeded worker URL.

This is the minimal change; it matches the code's own "instant rail before the (VFS-backed, kernel-gated) discovery resolves" design note.

## Cross-runtime parity

WC live shell (standalone/extension popout) only; no float-specific branches touched. N/A for swift/ios.

## Tests

`wireWcSprinkles` now has its first unit test: with an injected `fs` whose `walk` hangs, `wireWcSprinkles` still resolves within the window (pre-fix it blocks). Drives the real `SprinkleManager` over the jsdom `makeRefs()` harness.

Local verification: `lint` (biome+prettier), `typecheck` (all 7 targets), `wc-sprinkles` suite (21 pass). CI runs the full gate.

## Follow-up (separate)

This makes the **leader** deterministic, but the underlying *why does discovery stall* (sprinkles not loading on some boots) is a distinct bug worth its own investigation — the decoupling means it can no longer take the leader down with it.

### DIFF
--- a/packages/webapp/src/ui/wc/wc-sprinkles.ts
+++ b/packages/webapp/src/ui/wc/wc-sprinkles.ts
@@ -407,7 +407,13 @@ export async function wireWcSprinkles(deps: WireWcSprinklesDeps): Promise<WcSpri
         });
     }
   };
-  await resync();
+  // Fire the initial discovery+restore in the BACKGROUND — never block the
+  // caller on it. It is VFS-backed and kernel-gated, so a slow or stalled walk
+  // must not strand the rest of boot: `attachWcClient` sequences the tray
+  // leader wiring AFTER this returns, and the awaited resync used to hang there
+  // forever when discovery stalled (the leader never started). Hosts re-run
+  // resync() on kernel-ready as the recovery, and resync() is idempotent.
+  void resync().catch((err) => log.warn('WC shell: initial sprinkle resync failed', err));
   return { manager, zone, resync };
 }
 
--- a/packages/webapp/tests/ui/wc/wc-sprinkles.test.ts
+++ b/packages/webapp/tests/ui/wc/wc-sprinkles.test.ts
@@ -12,6 +12,9 @@ installWcDomStubs();
 // Registers the library elements so makeRefs gets a REAL slicc-tab-bar —
 // the tab-close regression test drives its actual event contract.
 import '@slicc/webcomponents';
+import type { VirtualFS } from '../../../src/fs/virtual-fs.js';
+import type { BootStageLogger } from '../../../src/ui/boot/types.js';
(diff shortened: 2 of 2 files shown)


trieloff:
We'll pick this up in #1085 (slicc-5 breaking release).

Compatible as-is: #1085 doesn't touch `wc-sprinkles.ts`, it merges clean, and the semantic target is intact — on #1085 `wireWcSprinkles` still ends with the blocking `await resync()` (right before `return { manager, zone, resync }`) that your fix replaces with a backgrounded `void resync().catch(...)`. Folding it into #1085.

trieloff:
Starting the tray leader on boot even when sprinkle discovery stalls is now in `main` (rebased into the slicc-5 / #1085 line). Merging this branch adds nothing. Closing as already-landed.

PR #1124, opened 2026-06-24
fix(wc-settings): add deployment field and fix stale provider dropdown

hannessolo:
## Summary

- **Missing deployment field**: the new `wc-settings` dialog never rendered the model-ID (deployment) input for providers that require it (e.g. Local LLM). Added the field with provider-supplied placeholder text, pre-population from any existing saved value, and validation on Save. The value is now passed through to `addAccount` so model IDs are actually persisted.
- **Optional API key not respected**: `optionalApiKey: true` providers (like Local LLM) were blocked from saving without an API key. Save now only requires the key when `config.requiresApiKey` is true.
- **Stale provider dropdown after delete**: the add-account provider `<select>` was built once when the dialog opened, so deleting an account didn't make that provider re-selectable until the dialog was closed and reopened. Fixed by rebuilding the add-account section inside every `renderList` call via an `addSectionSlot` container.

Screenshot 2026-06-24 at 16 25 07

## Test plan

- [ ] Open Accounts dialog → select **Local LLM (OpenAI-compatible)** → confirm Base URL, Model IDs, and API key (optional) fields all appear
- [ ] Fill in Base URL + Model IDs, leave API key empty → Save succeeds
- [ ] Remove the account → the same provider immediately reappears in the dropdown without closing the dialog
- [ ] Add it again → confirm it saves correctly and appears in the account list

### DIFF
--- a/packages/webapp/src/ui/wc/wc-settings.ts
+++ b/packages/webapp/src/ui/wc/wc-settings.ts
@@ -217,7 +217,7 @@ function buildAddSection(deps: ViewDeps): HTMLElement {
 
     const baseUrlInput = document.createElement('input');
     baseUrlInput.className = 'wcset__input';
-    baseUrlInput.placeholder = 'Base URL';
+    baseUrlInput.placeholder = config.baseUrlPlaceholder ?? 'Base URL';
     baseUrlInput.value = ps.getBaseUrlForProvider(pid) ?? '';
     if (config.requiresBaseUrl) controls.append(baseUrlInput);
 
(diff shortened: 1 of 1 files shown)


trieloff (approved):
Reviewed the diff and found no blocking issues. Focused verification: `npm test -- --run packages/webapp/tests/ui/wc/wc-settings.test.ts` passes.

trieloff:
@hannessolo this is approved from my side, and the focused wc-settings test passes. You can merge it yourself when you are ready.

hannessolo:
@trieloff I don't have permissions to merge :)

karlpauls:
@hannessolo, you should have a pending invite :-)

PR #1131, opened 2026-06-24
feat(cup): external-brain steering bridge over loopback HTTP — experiment

benpeter:
> **Status: experiment, rebased onto main.** Proving out a stable external Claude
> brain driving SLICC. The core flows are live-proven (see _Verification_); the shape
> (API surface, auth model, session semantics) may still evolve — see _What's ahead_.

## Skip the cone, use a cup

A normal SLICC instance runs its own agent — the **cone** — inside the browser. **Cup
mode** (`npm run cup`) boots the exact same SLICC, but with the cone switched off: the
browser, shell, VFS, and provider-authenticated mounts are all there, just with no agent
driving them. It's an empty ice-cream cup waiting for a scoop.

What fills it is **a Claude Code session running on your laptop** — a stable, long-lived
"brain" outside the browser. Cup exposes SLICC over loopback HTTP, so that brain can:

- run shell commands and read/write the VFS,
- drive the real browser (`playwright` commands) — navigate, click, screenshot,
- reach data through SLICC's mounts (S3, `da://`) and OAuth, **without ever holding the
  credentials**,
- **lead a tray** so a follower (e.g. the extension in your everyday Chrome) lends its
  real-profile, logged-in browser to the brain as a fully drivable federated target,
- and **answer the human typing in SLICC's chat panel.**

**Why you'd want this.** You get SLICC's whole capability surface — a real browser with
your logins, a shell, mounted data — driven by a brain that doesn't crash, compact, or
lose the thread the way an in-page agent loop can. The cone is great; but when you want
Claude Code itself to _be_ the agent and treat SLICC as its hands, you pour it into a cup.

And the operator experience is **one sentence**: open `claude` in the repo and say
_"be the brain for my SLICC."_ It leads by default, hands you the join URL, and answers
the chat — no ports, no session UUIDs, no curl, no status codes. To the human it just
talks like sliccy.

## What's in it

- **Cup mode** — `--cup` → `?cup=1` → `skipConeBootstrap` (no internal cone, exactly one
  CDP authority). Loopback steering API: `POST /api/shell/exec` (+ `?stream`),
  `GET /api/shell/session/:id`, `GET/POST /api/vfs/*`, `GET /api/targets`,
  `POST /api/lick/emit`. Sessions keyed by a caller `X-Slicc-Session` UUID (own
  `cwd`/`env`/device handles, GC'd after 5 min idle).
- **Lick-back** — the symmetric outbound mirror: a cone-less webapp's chat messages and
  orphaned `upgrade`/`sprinkle` licks route to whichever brain has **claimed** the channel
  (`POST /api/lickback/{claim,heartbeat,reply}`, `GET /api/lickback` SSE drain; cup-owned
  atomic claim + ~45s lease). The claimed brain streams replies back as a normal assistant
  turn — composer spinner, stop button, and all.
- **One-sentence bring-up** — the `slicc-cup` + `slicc-lickback-handler` skills turn the
  whole flow into one sentence. Deterministic bundled `node` scripts do the plumbing and
  speak exit codes instead of HTTP:
  - `cup-up` brings up (or attaches to) a drivable cup in one call — auto-detects dev vs
    prod from the git branch and waits for the real bridge (`/api/targets`), not a
    premature `/api/status`;
  - `cup-lead` fires `host lead` and polls until the join URL is live, in one call;
  - `cup-bootstrap` fetches SLICC's brain docs in one call; `cup-stop` cleans up a cup it
    launched;
  - `lickback-{claim,drain,next,reply,heartbeat}` own session/claim/SSE-drain/lease/reply.
    The claim is **self-healing**: it reaps an orphaned drain a prior session left holding
    the channel (port-scoped) before claiming, so a fresh brain never dead-locks on a
    phantom "already claimed". Unit + integration tested against a fake cup.
- **Detect-and-attach** — `GET /api/status` reports `{ cup, servePort, pid }` and boot
  writes `~/.slicc/cup.json` (mode 0600), so a second brain attaches to the running cup
  instead of spawning a parallel one.
- Skills: `slicc-cup` (steering + tray reference) + `slicc-lickback-handler` (the chat role).

## Security (trusted-localhost, spec §9)

- The `/api` CORS gate is fail-closed in cup too: a per-process bridge token gates `/cdp`
  and the cross-origin `/api` CORS gate; cross-origin requests without the token get `403`,
  loopback callers pass ungated — the supported steering path.
- A single loopback **`Host`-header allowlist** (one shared factory) on the steering +
  lick-back routes defends against DNS-rebinding → RCE. The `127.0.0.1` bind plus the Host
  guard is the trust boundary. The `/licks-ws` upgrade is Origin-gated in thin-bridge mode.

## Verification

- typecheck (8 tsc targets), full test suite, `lint:ci`, and both builds green in CI.
- Lick-back wire covered by `lickback-api` / `lickback-registry` / `lickback-channel`
  suites; the handler scripts by `packages/dev-tools/lickback-scripts/` (a fake cup over
  node:http). Cup is standalone-only (spec §11), so there's no cross-runtime parity surface.
- **Live-proven end to end:** the one-sentence "be the brain and lead" flow (cup + leader
  + chat handler), cold start from a repo with no cup running, federated driving of a
  follower's real-profile browser, and the brain using an in-SLICC workspace skill from
  the outside.

## What's ahead

- Generic attach to other Electron apps (relaunch-with-debugging / attach-if-port-open).
- Optional UX hardening: an app-tab filter on `/api/targets`, a no-responder watchdog on
  the optimistic chat turn.
- Banked: a multi-channel front skill, attachments over the lick-back wire, promoting the
  handler to a defined agent type only if spawn reliability ever demands it.


trieloff:
Wait, what???

PR #1140, opened 2026-06-25
fix(webcomponents): cache shader programs per mode + immediate repaint to kill cone↔freezer blue flicker

trieloff:
## Problem

When the background `<slicc-shader>` switches between modes (cone / scoop / freezer), the **blue of the freezer state lingers on screen for a few frames** before the new mode paints.

## Root cause (runtime-measured via CDP)

Two compounding bugs in `packages/webcomponents/src/freezer/slicc-shader.ts`:

1. **No program cache** — `#linkMode()` recompiled + relinked the GLSL program on *every* mode change. The cold compile blocks the main thread:

   | switch | ms |
   |--------|----|
   | → freezer (cold) | **40.1** |
   | → scoop (cold) | **17.8** |
   | → cone / warm | ~2 |

   During the ~40ms stall `requestAnimationFrame` cannot run, so the last-painted frame freezes on screen.

2. **Deferred repaint** — on a mode change in animated mode, `attributeChangedCallback` relinked but only called `#renderIfStatic()` (a no-op when animated), so the canvas wasn't repainted with the new program until the next rAF tick. The old program's last frame (blue freezer) lingered an extra frame.

## Fix

- **Cache the compiled+linked `WebGLProgram` per `ShaderMode`** (built once, reused on switch). Revisiting a mode is now a cached `useProgram` swap with no recompile/relink.
- **Repaint synchronously** (`#renderFrame()`) after a successful relink on a mode change, so the new program paints in the same task instead of showing the old blue frame.
- **Teardown safety**: `#dispose()` frees *all* cached programs and their shaders (not just the active one); `webglcontextlost`/`webglcontextrestored` handlers clear and rebuild the cache; listeners are removed on dispose.
- No change to the observation set, animation loop, uniforms, scroll/tint behavior, or the `no-webgl` fallback.

## Tests

Added 2 regression tests in `packages/webcomponents/tests/freezer/slicc-shader.test.ts`:
- switching to a mode, away, then back does **not** recompile (compile/createProgram call counts stay flat on revisit)
- a mode change triggers an **immediate** render with the new program (not deferred to rAF)

## Verification

- `npm run lint` ✅
- `npm run typecheck` ✅
- `npm run test -w @slicc/webcomponents` ✅ (73 files / 1597 tests)

Scoped to 2 files. Separate from the switcher ResizeObserver fix (#1136).

---

### DIFF
--- a/packages/webcomponents/src/freezer/slicc-shader.ts
+++ b/packages/webcomponents/src/freezer/slicc-shader.ts
@@ -201,6 +201,15 @@ function clampNum(v: number, lo: number, hi: number, fb: number): number {
   return Number.isFinite(v) ? Math.max(lo, Math.min(hi, v)) : fb;
 }
 
+/** A compiled+linked program kept per mode so a switch never recompiles. */
+interface CachedProgram {
+  program: WebGLProgram;
+  vs: WebGLShader;
+  fs: WebGLShader;
+  aPos: number;
+  loc: Partial<Record<UniformName, WebGLUniformLocation | null>>;
+}
+
 /** Parse a CSS color via getComputedStyle into a 0..1 rgb triple. */
 function colorToVec3(css: string, fallback: [number, number, number]): [number, number, number] {
   if (!css || typeof getComputedStyle !== 'function') return fallback;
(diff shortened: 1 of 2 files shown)


trieloff:
@karlpauls maybe this is it?

PR #1157, opened 2026-06-25
docs(shell): record arg-parser migration decisions for agent/websocat (#1150)

### LINKED ISSUE #1150: Follow-up to #1119: evaluate shared arg-parser for agent-command.ts and websocat-command.ts
Follow-up to #1119 / #1141.

PR #1141 migrated `git-commands.ts` and the four device commands (`usb`/`hid`/`serial`/`esptool`) onto the new shared `packages/webapp/src/shell/arg-parser.ts` (`mri` wrapper). Two other commands with heavy hand-rolled parsing were left as out-of-scope follow-ups. This issue tracks them.

## 1. `agent-command.ts` — investigated, NOT a clean fit (recommend leaving as-is)

Four migration attempts during #1141 stalled, then a code+test review showed a real semantic mismatch with the shared `mri`-based parser:

1. **Value-shadowing is inverted.** The shared parser's core feature is that a value-flag swallows its next token even when flag-looking (`commit -m --help` → message is `--help`). `agent-command` requires the *opposite*: it must **reject** flag-looking values. `--model --help`, `--thinking --help`, `--read-only --help`, and `--schema-b64 --help` must all error. Locked by tests in `agent-command.test.ts` (around lines 530, 682, 981, 1054).
2. **Context-sensitive positionals.** The 3rd positional (the prompt) is taken **literally** even when it looks like a flag: `agent . "*" -h` → prompt is `-h`; `agent . "*" "--thinking high"` → literal prompt. `mri` has no equivalent mode. `stopEarly` stops at the *first* positional, but tests require flags to still parse in the middle/after positionals (`agent-command.test.ts` around 493, 503, 656). No single `mri` mode satisfies both.

Forcing the shared parser would require wrapping it in enough agent-command-specific post-validation that it **adds** a layer rather than removing one — the opposite of #1119's "reduce complicatification" goal — and would risk ~10 calibrated tests (including bridge-ordering microtask timing).

**Recommendation:** leave `agent-command.ts` on its purpose-built parser. Close this part as wontfix unless the shared parser gains an opt-in "strict value" + "literal Nth positional" mode.

## 2. `websocat-command.ts` — unevaluated candidate

Has hand-rolled flag parsing similar to the device commands but has not been fit-checked against the shared parser. Needs a quick evaluation:

- [ ] Confirm whether its flags map cleanly onto `parseFlagArgs()` (boolean/string/alias spec, no value-rejection or literal-positional requirements).
- [ ] If clean, migrate and add/extend tests; otherwise document the conflict like `agent-command` above.

## Acceptance
- [ ] Decision recorded for `agent-command.ts` (migrate or wontfix).
- [ ] `websocat-command.ts` evaluated; migrated if it's a clean fit, otherwise reasoned out.
- [ ] No observable command behavior changes; webapp coverage stays at/above floor.

trieloff:
Closes #1150 (follow-up to #1141).

## Solution

PR #1141 left two commands' hand-rolled parsers untouched as "evaluate later". This PR records the migrate-or-wontfix decision for both. Pure docs — no behavior changes, no test changes.

- **`agent-command.ts` — wontfix, documented as JSDoc on `parseArgs`.** Two semantic mismatches with the shared mri-based parser, both locked by existing tests:
  1. Inverted value-shadowing — `--model --help` etc. must REJECT; the shared parser's core feature is the opposite.
  2. Context-sensitive literal Nth positional — `agent . "*" -h` takes `-h` as the prompt literally. `stopEarly` stops at the FIRST positional but flags must still parse mid-stream; no single mri mode satisfies both.
- **`websocat-command.ts` — wontfix, recorded here only.** Evaluated against the shared parser; not a clean fit. The validation surface (unsupported-flag rejection table, `-v` count flag, mutually-exclusive `-t`/`-b`, `--jsonrpc-omit-jsonrpc` implication, per-flag numeric range and bespoke `websocat:` error strings, cross-flag `--close-reason` requires `--close-status-code`, extra-positional rejection) dominates the file. Migration would replace ~200 lines of cohesive parsing with a small mri spec plus a larger post-validation block — net negative, and the same trap #1141 identified for agent-command. The in-file JSDoc was reverted because `websocat-command.ts` is on the function-size debt list (`check-touched-exemptions` rejects any modification without a refactor); the rationale lives here and in #1150 instead. Revisit only if the shared parser grows declarative support for these constraints.

## Testing

None needed — JSDoc-only edit (agent-command.ts) inside an existing comment block. CI runs the canonical verification pass.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01KVZ6RHM93PSVGVCMRPYB97WF&panel=chat)

### DIFF
--- a/packages/webapp/src/shell/supplemental-commands/agent-command.ts
+++ b/packages/webapp/src/shell/supplemental-commands/agent-command.ts
@@ -285,6 +285,28 @@ function validatePositionals(
  *   - Any other `-...` / `--...` token is an unknown-flag error.
  *   - Exactly three positional arguments are required; more is a too-many
  *     error.
+ *
+ * Why this is NOT on the shared `parseArgs` / `parseFlagArgs` from
+ * `shell/arg-parser.ts` (decision tracked in #1150, follow-up to #1141): two
+ * semantic mismatches with the `mri`-based shared parser:
+ *
+ *   1. **Value-shadowing is inverted.** The shared parser's core feature is
+ *      that a value-taking flag swallows its next token even when flag-looking
+ *      (`commit -m --help` -> message is `--help`). `agent` requires the
+ *      opposite: `--model --help`, `--thinking --help`, `--read-only --help`,
+ *      and `--schema-b64 --help` MUST error. Locked by tests in
+ *      `agent-command.test.ts` around lines 530, 682, 981, 1054.
+ *   2. **Context-sensitive positionals.** The 3rd positional (the prompt) is
+ *      taken literally even when it looks like a flag (`agent . "*" -h` ->
+ *      prompt is `-h`). The shared parser's `stopEarly` stops at the FIRST
+ *      positional, but tests require flags to still parse in the middle/after
+ *      positionals — no single `mri` mode satisfies both.
+ *
+ * Forcing the shared parser would require wrapping it in enough
+ * `agent`-specific post-validation that it ADDS a layer rather than removes
+ * one — the opposite of #1119's "reduce complicatification" goal. Revisit
+ * only if the shared parser gains an opt-in "strict value" plus
+ * "literal Nth positional" mode.
  */
 function parseArgs(args: string[]): ParsedArgs {
   const state: ParseState = {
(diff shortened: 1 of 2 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KVZ6RHM93PSVGVCMRPYB97WF)⚡</sup>

👋 I've got this PR.

PR #1166, opened 2026-06-25
fix: terminal Ctrl-C and kill -9 terminate realm-backed jobs (#1116)

### LINKED ISSUE #1116: kill -9 and Ctrl-C no longer terminate processes in the terminal (all surfaces)
## Summary

`kill -9 <pid>` and `Ctrl-C` (SIGINT) no longer terminate running processes in the SLICC terminal. Reproduces across **all surfaces** (node-server, swift-server, chrome-extension, electron) — this is shared shell behavior.

## Expected

- `Ctrl-C` interrupts the foreground process and returns to the prompt.
- `kill -9 <pid>` force-terminates the target process.

## Actual

- Neither signal stops the running process; the foreground command keeps running and the prompt does not return.

## Surfaces

All four floats (CLI / swift / extension / electron). Likely a shared `AlmostBashShell` foreground-job signal-handling + `ps`/`kill` supplemental-command regression rather than a per-float issue.

## Context

Found during PR #1085 (slicc-5) manual verification. Feature area: shell process control / job control.

trieloff:
## Summary

Fixes #1116 — terminal `Ctrl-C` (SIGINT) and `kill -9 <pid>` no longer terminated running processes in the SLICC terminal across all four floats.

## Root cause

Signals reached only the **outer `kind:'shell'`** process, never the **inner realm child** (`node -e` / `.jsh` / `python`) doing the actual work — so the job ran forever and the prompt never returned. Three coupled gaps:

1. `process-manager.ts` `signal()` had **no ppid fan-out** to children.
2. Realm children were **mis-parented** — `node`/`python` commands passed `undefined` PM-config, so the realm `ppid` defaulted to `1` instead of the shell pid.
3. `realm-runner.ts` `onSignal` honored **only SIGKILL on its own pid** — SIGINT/SIGTERM never terminated the opaque realm.

Originating commits: `bacb8cd7` (Phase 3.2 `kind:'shell'` execs) + `09644a01` (realm runtime).

## The fix (shared kernel/shell path → covers all four floats)

1. **ppid fan-out** in `ProcessManager.signal()` — cycle-safe BFS that delivers the signal to all live descendants, only true descendants (no over-signaling of siblings, the cone, or scoops).
2. **Shell→realm parent linkage** — threads the active shell pid through `terminal-session-host` → `AlmostBashShellHeadless` → `node`/`python`/`.jsh` so realms spawn parented to the shell pid.
3. **SIGINT/SIGTERM escalation** — `realm-runner` now escalates SIGINT (exit 130) and SIGTERM (exit 143) to `realm.terminate()`, not just SIGKILL (exit 137).

## Tests

- New regression test `packages/webapp/tests/kernel/issue-1116-terminal-signals.test.ts` drives the real signal path (TerminalSessionClient → TerminalSessionHost → AlmostBashShellHeadless → runInRealm → ProcessManager). Covers SIGINT, `kill -9`, and a control case. Failed on `main`; now passes 3/3.
- Added/updated `process-manager`, `realm-runner`, and `terminal-session` unit tests.

## Verification

Independently verified, merge-ready (high confidence):

- Reproduction test passes 3/3 and was **not** weakened.
- `lint` 0, `typecheck` 0, webapp suite **7829 passed / 5 skipped / 0 failed**, coverage gate **PASS** (all metrics above floor).
- Fan-out audited cycle/ordering-safe; exit codes correct (130/143/137).
- Cross-runtime parity: N/A — swift/ios don't mirror this kernel/shell code.

---

### DIFF
--- a/packages/webapp/src/kernel/terminal-session-host.ts
+++ b/packages/webapp/src/kernel/terminal-session-host.ts
@@ -293,7 +293,10 @@ export class TerminalSessionHost {
       : null;
     session.currentProcess = proc;
     try {
-      const result = await session.shell.executeCommand(msg.command, abort.signal);
+      // Pass the shell proc pid so realm-backed commands (`node` / `.jsh` /
+      // `python`) parent their realm child to it — a terminal signal to this
+      // shell pid then fans out to the realm via `pm.signal` (#1116).
+      const result = await session.shell.executeCommand(msg.command, abort.signal, proc?.pid);
       await this.emitExecSuccess(msg, result, abort, proc);
     } catch (err) {
       this.emitExecError(msg, err, abort, proc);
--- a/packages/webapp/src/kernel/terminal-session-host.ts
+++ b/packages/webapp/src/kernel/terminal-session-host.ts
@@ -318,7 +318,13 @@ export class TerminalSessionHost {
     abort: AbortController,
     proc: Process | null
   ): Promise<void> {
-    const exitCode = abort.signal.aborted ? 130 : result.exitCode;
+    // Derive the aborted code from the recorded signal (mirroring
+    // `emitExecError`) so a terminal `kill -TERM/-9 <shellPid>` that still
+    // returns normally from the shell reports 143/137 instead of being
+    // flattened to 130. SIGINT still maps to 130.
+    const exitCode = abort.signal.aborted
+      ? signalExitCode(proc?.terminatedBy ?? 'SIGINT')
+      : result.exitCode;
     if (!abort.signal.aborted) {
       // Gate output + exit emission. If the user (or another shell) sent
       // SIGSTOP between command launch and now, hold the buffer here until
--- a/packages/webapp/src/shell/almost-bash-shell-headless.ts
+++ b/packages/webapp/src/shell/almost-bash-shell-headless.ts
@@ -168,7 +168,8 @@ export interface HeadlessShellLike {
   syncJshCommands(): Promise<void>;
   executeCommand(
     command: string,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    shellPid?: number
   ): Promise<{ stdout: string; stderr: string; exitCode: number }>;
   executeScriptFile(
     scriptPath: string,
@@ -258,6 +259,27 @@ export class AlmostBashShellHeadless implements HeadlessShellLike {
   private pendingEnvWrites = new Map<string, string>();
 
   /**
+   * The `kind:'shell'` pid of the in-flight `executeCommand` call, set by
+   * the caller (`TerminalSessionHost.handleExec` passes the spawned shell
+   * proc's pid). Realm-backed commands (`node` / `.jsh` / `python`) parent
+   * their realm child to this pid via {@link buildJshProcessConfig}, so a
+   * terminal signal to the shell pid fans out to the realm (#1116). Cleared
+   * after each exec. Falls back to `options.getCurrentShellPid` (the scoop
+   * turn pid) when the caller doesn't supply one.
+   */
+  private activeShellPid: number | undefined;
+
+  /**
+   * Stable callback handed to realm-backed commands (`node` / `python`)
+   * via `createSupplementalCommands`. Resolves the per-exec jsh process
+   * config (PM, owner, parent pid) lazily at command-execution time. A
+   * bound class field rather than an inline constructor arrow so the
+   * (already large) constructor stays under the cognitive-complexity cap.
+   */
+  private readonly resolveJshProcessConfig = (): JshProcessConfig | undefined =>
+    this.buildJshProcessConfig();
+
+  /**
    * When sudo is wired with `defaultDisposition: 'require-approval'` the
    * policy is the single command-enforcement surface (the per-scoop sudoers
    * already encodes `allowedCommands` as `NOPASSWD Cmnd` grants, and any
@@ -333,6 +355,7 @@ export class AlmostBashShellHeadless implements HeadlessShellLike {
       scriptCatalog: this.scriptCatalog,
       browserAPI: options.browserAPI,
       getParentJid: options.getParentJid,
+      buildProcessConfig: this.resolveJshProcessConfig,
       // Thread the manager into `ps` / `kill`. When the
       // shell is constructed without one (extension offscreen,
       // inline standalone), the commands fall back to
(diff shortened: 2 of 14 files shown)


packages/webapp/src/kernel/terminal-session-host.ts
```diff
@@ -293,7 +293,10 @@ export class TerminalSessionHost {
       : null;
     session.currentProcess = proc;
     try {
-      const result = await session.shell.executeCommand(msg.command, abort.signal);
+      // Pass the shell proc pid so realm-backed commands (`node` / `.jsh` /
+      // `python`) parent their realm child to it — a terminal signal to this
+      // shell pid then fans out to the realm via `pm.signal` (#1116).
+      const result = await session.shell.executeCommand(msg.command, abort.signal, proc?.pid);
```
trieloff, line 299: **P2 — hardcoded 130 still swallows the real signal after this fan-out fix lands** (see `emitExecSuccess` at line 321, just below this hunk — unchanged, so I can't comment inline on it).

With the new fan-out, `kill -TERM <shellPid>` and `kill -9 <shellPid>` both abort the shell's controller (via `deliverSignal`). When `bash.exec` returns normally with the realm child's exit code, line 321 remaps it to 130 instead of 143/137:

```ts
const exitCode = abort.signal.aborted ? 130 : result.exitCode;
```

`emitExecError` (line 352) already does the right thing — use the same here:

```ts
const exitCode = abort.signal.aborted
  ? signalExitCode(proc?.terminatedBy ?? 'SIGINT')
  : result.exitCode;
```

Same issue Codex flagged separately. The test gap on the `kill -9` case (inline comment there) currently masks it — add an `expect(outcome.exitCode).toBe(137)` and this will fail until line 321 is fixed.

packages/webapp/tests/kernel/issue-1116-terminal-signals.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    const shellPid = w.pm.list().find((p) => p.kind === 'shell' && p.status === 'running')?.pid;
+    await client2.exec(`kill -9 ${shellPid}`);
+    const outcome = await raceExec(execPromise);
+    client2.close();
+    w.teardown();
+    expect(
+      outcome.timedOut,
+      `#1116: kill -9 ${shellPid} did not terminate the foreground job — ` +
+        'the exec never returned to the prompt'
+    ).toBe(false);
```
trieloff, line 155: **Test gap — the `kill -9` case never asserts the exit code.**

The control test pins `exitCode === 137`, but this test only asserts `outcome.timedOut`. After the fan-out fix the shell's controller IS aborted by the fan-out, so `emitExecSuccess` hits its hardcoded `abort.signal.aborted ? 130 : ...` branch (see inline comment on `terminal-session-host.ts:299`) and emits 130 instead of 137 — and nothing here catches it.

Suggested:

```ts
).toBe(false);
expect(outcome.exitCode, '#1116: kill -9 should surface 137, not 130').toBe(137);
```

Ideally add a third SIGTERM case pinning 143 too, since the three exit codes are the contract this PR commits to.

trieloff (commented):
## Review — solid fix, two follow-ups

Diagnosis is correct, architecture is sound, and the regression coverage drives the real signal path end-to-end (with a meaningful control case that isolates the defect to terminal→realm routing). I'd land it after two small fixes; one out-of-scope follow-up worth filing separately.

### What I verified

- ✅ All three coupled gaps from the issue are closed: PM had no fan-out, realm was misparented (`ppid: 1`), realm-runner ignored SIGINT/SIGTERM.
- ✅ Fan-out is cycle-safe, self-loop-safe, prunes dead branches, and pre-snapshots the ppid index so delivery mutations don't invalidate the walk. `process-manager.test.ts` pins all four invariants.
- ✅ Per-exec `activeShellPid` save/restore in `executeCommand` uses `try/finally`, so nested/thrown execs restore correctly.
- ✅ `buildJshProcessConfig`'s arrow-form `getParentPid` is re-evaluated per realm spawn (not captured at construction), so the per-exec pid wins when set.
- ✅ The control test (SIGKILL targeted directly at the realm pid) is exactly the right isolation.

### Issues to address

**🟡 P2 — exit-code mapping still wrong on the fan-out success path.** Inline comment on the `executeCommand` call site, but the real fix is at `terminal-session-host.ts:321` (`emitExecSuccess`). Same bug Codex flagged separately, plus the **test gap that masks it** (inline on the `kill -9` test).

**🟢 Follow-up (out of scope for #1116) — scoop bash-tool path is unfixed.** Codex's P1 stands: `ScoopContext.initShellAndSkills` constructs `AlmostBashShell` without `processManager` / `processOwner` / `getCurrentShellPid`, so the new `() => this.activeShellPid ?? this.options.getCurrentShellPid?.()` fallback returns `undefined` in the agent bash-tool path; realm children parent to pid 1 and Stop/drop still doesn't reach them. Worth a separate issue ("Parent agent bash realms under the turn/tool") — this PR is terminal-scoped per the issue.

### Nits (non-blocking)

- `collectLiveDescendants` allocates a fresh `childrenByPpid` index per `signal()` call. Fine in practice (signal volume is tiny); flag if it ever gets hot.
- The cycle-safe test corrupts `ppid` via a cast — fine, but the property is real and the test is the only regression pin; a one-line code comment in `deliverSignal` / `collectLiveDescendants` pointing at the test would help.
- The three SIGINT/SIGTERM/SIGKILL branches in `realm-runner` could collapse into a local `signalExitCode(sig)`, but the explicit form is also readable — keep as-is.

packages/webapp/tests/kernel/issue-1116-terminal-signals.test.ts
trieloff: Fixed in 996fa086. The `kill -9` case now asserts `exitCode === 137`, the SIGINT case asserts `130`, and I added a third `kill -TERM` case asserting `143` — pinning the full 130/143/137 contract. These would fail if the hardcoded-130 mapping in `emitExecSuccess` regressed (that mapping itself was fixed in bdcd15db). Test passes 4/4.

packages/webapp/src/kernel/terminal-session-host.ts
trieloff: Fixed in bdcd15db. `emitExecSuccess` now derives the aborted exit code via `signalExitCode(proc?.terminatedBy ?? 'SIGINT')` (mirroring `emitExecError`) instead of the hardcoded `130`, so `kill -TERM/-9 <shellPid>` surface 143/137 and SIGINT stays 130. The masking test gap you flagged is closed in 996fa086 (now asserts 137/143/130).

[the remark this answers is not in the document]
trieloff: Fixed in bdcd15db. `emitExecSuccess` now derives the code from `proc.terminatedBy` via `signalExitCode()` (same approach as `emitExecError`) rather than flattening every abort to 130, so SIGTERM→143 and SIGKILL→137 are preserved on the fan-out success path; SIGINT still 130. Exit-code assertions added in 996fa086.

packages/webapp/src/shell/almost-bash-shell-headless.ts
```diff
@@ -964,7 +1000,10 @@ export class AlmostBashShellHeadless implements HeadlessShellLike {
     return {
       processManager: this.options.processManager,
       owner: this.options.processOwner,
-      getParentPid: this.options.getCurrentShellPid,
+      // Prefer the per-exec shell pid (panel terminal) and fall back to
+      // the static `getCurrentShellPid` (scoop turn pid) so both the
+      // panel and the agent's bash tool parent realm children correctly.
+      getParentPid: () => this.activeShellPid ?? this.options.getCurrentShellPid?.(),
```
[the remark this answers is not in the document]
trieloff: Fixed in bdcd15db (we addressed this in-PR even though it's arguably out-of-scope). `ScoopContext.initShellAndSkills` now constructs the agent bash-tool `AlmostBashShell` with `processManager`, `processOwner` ({ kind: cone|scoop, scoopJid }), and `getCurrentShellPid: () => currentTurnProcess?.pid`, so realm children parent to the scoop-turn pid instead of `ppid: 1` and Stop/drop fan-out reaches them. Verified non-tautological: reverting the hunk makes the new scoop-realm-parenting tests fail (orphans at ppid:1 survive); restoring passes 32/32.

PR #1180, opened 2026-06-25
fix(webapp): prevent SW from double-proxying /api/* calls in thin-bridge mode

bstopp:
## Summary

  In thin-bridge mode (`npx sliccy`), the LLM-proxy service worker intercepts cross-origin fetches from the kernel DedicatedWorker — including direct calls to node-server endpoints like `/api/da-sign-and-forward`. Re-routing these through `/api/fetch-proxy` strips `X-Bridge-Token` (it's in
  `FETCH_PROXY_SKIP_HEADERS`) and decodes `X-Proxy-Origin` back to `Origin` on the internal request, causing the node-server's CORS middleware to reject with `bridge-token-required`.

  This PR extends the SW's bridge-local-API pass-through — previously only `/api/fetch-proxy` — to cover any `/api/*` path at the bridge origin, so DA and S3 mount calls reach the node-server with their original headers intact.

  > **Note:** #1168 collapses all node-server modes into thin-bridge permanently. Once that lands, this fix applies to every float, not just `npx sliccy`.

  ## Why

  **Repro:**
  1. `npx sliccy` (thin-bridge mode — UI served from `https://www.sliccy.ai`)
  2. `mount --source da://org/site /mnt/site`

  Expected: mount succeeds
  Actual: `bridge-token-required` (403) — the SW double-proxied the `/api/da-sign-and-forward` call, stripping the bridge token in the process

  The bug affects any direct `/api/*` call the kernel worker makes in thin-bridge mode: `/api/da-sign-and-forward`, `/api/s3-sign-and-forward`, and any future node-server endpoint called directly from the worker.

  - **`isBridgeFetchProxyTarget` (`kernel-worker-fetch-bypass.ts`)**: changed from exact path match on `/api/fetch-proxy` to prefix match on `/api/` at the bridge origin. This stamps `x-bypass-llm-proxy: 1` on all direct `/api/*` calls from the kernel worker so the SW skips re-interception.
  - **`isBridgeLocalApiUrl` (`llm-proxy-sw-config.ts`)**: new exported function — same `/api/` prefix + bridge origin check, extracted so the SW and its tests can share the logic independently of the kernel-worker path.
  - **`llm-proxy-sw.ts`**: switched the bridge pass-through guard from `isBridgeFetchProxyUrl` (exact `/api/fetch-proxy` match) to `isBridgeLocalApiUrl` (any `/api/*`). The pass-through re-fetches with `x-bypass-llm-proxy: 1` so the SW does not re-intercept the outgoing call.

  ## Cross-cutting impact

  - **Floats exercised**: currently thin-bridge only (`npx sliccy`). Once `remove-localhost-webapp-serving` lands and thin-bridge becomes the only mode, this fix applies to all floats.
  - **Other callers of `isBridgeFetchProxyTarget`**: only `shouldStampBypass` in the same file. `isBridgeFetchProxyUrl` (the exact-match variant) is still exported and used by existing tests — it is unchanged.
  - **Security**: the pass-through re-fetches with `x-bypass-llm-proxy: 1` but preserves all caller headers byte-for-byte, including `X-Bridge-Token`. No headers are added or removed beyond the bypass marker. The node-server's bridge-token gate still validates every non-loopback request.
  - **No persisted state changes**, no new dependencies, no IndexedDB or localStorage changes.

  ## Test plan

  - [x] `npm run lint` — clean
  - [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — 0 of 5 changed files on any debt list
  - [x] `npm run test` — 9594 passing, 15 skipped, no new failures
  - [x] `npm run test:coverage` — above thresholds
  - [x] `npm run build` — clean
  - [x] **New behavior is covered by unit tests**:
    - `npx vitest run packages/webapp/tests/kernel/kernel-worker-fetch-bypass.test.ts` — updated `isBridgeFetchProxyTarget` describe block with 4 new cases for `/api/da-sign-and-forward`, `/api/s3-sign-and-forward`, any `/api/*`, and non-`/api/` rejection
    - `npx vitest run packages/webapp/tests/ui/llm-proxy-sw-config.test.ts` — new `isBridgeLocalApiUrl` describe block with 6 cases
  - [x] Manual smoke (thin-bridge): `npm start` + `mount --source da://org/site /mnt/site` — mount succeeded; retry logic kicked in automatically after fix was applied
  - [ ] Manual smoke (CLI dev mode / extension): N/A — SW pass-through only fires when `getBridgeProxyOrigin()` returns non-null, which only happens when a bridge config is present

  **Pre-existing failures**: none observed — 15 skipped tests reproduce on `main` (timeout/interactive shell tests).

  ## Documentation

  - [x] No documentation changes needed — the SW routing architecture is already documented; thin-bridge-specific framing will be superseded by #1168 

  ## Risk & rollback

  - **Blast radius**: currently thin-bridge mode only; expands to all floats once #1168 lands. Previously broken calls to `/api/da-sign-and-forward` and `/api/s3-sign-and-forward` now pass through correctly — no regression path for previously working code paths.
  - **No feature flags or kill switches** — the fix is unconditional within thin-bridge mode.
  - **Rollback**: revert this PR cleanly. No persisted state migration. 
  - **CI cannot catch**: the full thin-bridge routing path (page → SW → node-server) requires a real browser with a live SW; unit tests cover the routing predicates but not the full chain.

  ## Checklist

  - [x] If this changes a contract used by other floats (CLI ↔ extension ↔ tray), I checked the followers/leader/offscreen paths — SW pass-through change is bridge-config-gated; extension and non-bridge CLI are unaffected

### DIFF
--- a/packages/webapp/src/kernel/kernel-worker-fetch-bypass.ts
+++ b/packages/webapp/src/kernel/kernel-worker-fetch-bypass.ts
@@ -23,22 +23,22 @@
  * wasm/asset payloads but works uniformly across CDNs and matches
  * the path `proxiedFetch` uses for everything else.
  *
- * Thin-bridge exception: the local node-server's own `/api/fetch-proxy`
- * is OUR infra endpoint, not a third-party CDN, and is reachable
- * cross-origin from a hosted-leader / wrangler-served UI. When the
- * worker realm's `proxiedFetch` calls it directly we must stamp the
- * bypass header so the page-installed SW skips re-interception
- * (re-proxying would clobber the caller's `X-Target-URL`). The
- * bridge already handles the resulting CORS preflight via
- * `createThinBridgeCorsMiddleware`, so the CDN-preflight rationale
- * above doesn't apply. The caller wires `getBridgeProxyOrigin` to
- * `() => getLocalApiBaseUrl()`-derived origin so the check stays
+ * Thin-bridge exception: any `/api/*` endpoint on the local node-server
+ * is OUR infra, not a third-party CDN, and is reachable cross-origin
+ * from a hosted-leader / wrangler-served UI. When the worker realm calls
+ * these endpoints directly (e.g. `/api/fetch-proxy`, `/api/da-sign-and-forward`,
+ * `/api/s3-sign-and-forward`) we must stamp the bypass header so the
+ * page-installed SW skips re-interception — re-proxying strips
+ * `X-Bridge-Token` and causes the node-server's CORS gate to reject with
+ * `bridge-token-required`. The bridge already handles the resulting CORS
+ * preflight via `createThinBridgeCorsMiddleware`, so the CDN-preflight
+ * rationale above doesn't apply. The caller wires `getBridgeProxyOrigin`
+ * to `() => getLocalApiBaseUrl()`-derived origin so the check stays
  * dynamic (the bridge origin is set AFTER `installFetchBypass` runs).
  */
 
 const BYPASS_HEADER = 'x-bypass-llm-proxy';
 const BYPASS_VALUE = '1';
-const FETCH_PROXY_PATH = '/api/fetch-proxy';
 
 // Track the global `fetch` signature so the wrapper composes
 // transparently and stays in lockstep with lib.dom updates.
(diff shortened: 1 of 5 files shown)


bstopp:
@claude - can you double check this again

bstopp:
how did the matrix test fail on the merge queue, but not in the PR?

PR #1195, opened 2026-06-26
fix(webapp): bridge worker silent token renewal to the page (#1181)

### LINKED ISSUE #1181: Adobe session expired surfaces as unrecoverable cone error when token expires mid-session in worker realm
## What the error is

The cone surfaces `Adobe session expired — please log in again` as a fatal error mid-session. The same payload arrives in RUM twice for each occurrence — once as the raw stream error under `source: llm`, and once wrapped by the cone non-retryable handler under `source: scoop:cone` as `Scoop "Cone" failed with unrecoverable error: Adobe session expired — please log in again`. Both fingerprints below describe the same underlying failure.

## Where it surfaces

- The literal string is thrown by `getValidAccessToken` in `packages/webapp/providers/adobe.ts:497` when the local `tokenExpiresAt` window has elapsed AND `silentRenewBackoff.run(() => silentRenewToken())` returned `null`.
- `silentRenewToken` in `packages/webapp/providers/adobe.ts:516-522` short-circuits with `if (typeof window === 'undefined') return null;` — the kernel-worker / offscreen agent realm has no DOM, so it cannot drive the IMS popup/iframe flow.
- `isNonRetryableError` (`packages/webapp/src/scoops/scoop-context.ts:108-126`) matches the `session expired|log in again` pattern, so `handleNonRetryableError` (`scoop-context.ts:763`) immediately wraps it as the `scoop:cone` "unrecoverable error" emit. There is no retry, no live re-auth from the cone.

## Suspected root cause

Single-shot OAuth bootstrap. `packages/webapp/src/ui/oauth-bootstrap.ts` re-pushes / silent-renews tokens once, before the kernel worker boots, and the worker reads them via the localStorage shim. Per `packages/webapp/CLAUDE.md` lines 238–239 this is the intentional design: the page does the renewal, the worker passively reads the result.

Trouble is the worker has no live channel to ask for a renewal after boot. When a long-lived session crosses the `tokenExpiresAt` boundary (Adobe IMS tokens are ~24h), the next worker-initiated stream calls `getValidAccessToken`, finds the token expired, calls `silentRenewToken`, gets `null` (no `window`), and surfaces the literal "session expired" message — even though the user's underlying IMS session cookie is almost certainly still valid and a page-side renewal would have succeeded.

This is the same architectural gap previously closed out in #1070 (which targeted the on-401 `Invalid Compact JWS` variant). That issue proposed on-401 silent renewal; this fingerprint pair is the local-clock-expired side of the same seam.

## Float

extension

## Occurrence stats

- sampled: 2, estimated: 20 (each fingerprint; both emit for every occurrence)
- firstSeen: 2026-06-25T18:00:46Z
- lastSeen: 2026-06-26T01:00:04Z

## Proposed fix

Add a panel-RPC `silent-renew` op so the kernel worker can ask the page realm to drive `silentRenewToken()` on demand, gated by the existing `silentRenewBackoff` so a genuinely revoked session doesn't loop. In the extension/thin-bridge float this routes worker → offscreen-bridge → hosted-leader page (the only realm with a usable `window`). The worker-side `silentRenewToken` would call this RPC instead of bailing on `typeof window === 'undefined'`.

Less invasive alternative: schedule a page-side `setTimeout` keyed off `tokenExpiresAt - RENEW_BUFFER_MS` to proactively renew before the worker ever notices, persisting the new token via the existing `saveOAuthAccount` path so the worker's localStorage shim picks it up. Avoids the RPC seam entirely but only helps when the page tab stays alive (it is the hosted leader, so it usually does).

Leaving to a human — both options touch the OAuth renewal seam and want careful handling around the `silentRenewBackoff` cooldown so a revoked session doesn't hot-loop renewal requests.

trieloff:
## Problem

Fixes #1181. In the kernel-worker realm there is no DOM, so once the cached Adobe IMS token crosses `tokenExpiresAt`, the silent-renewal path in `packages/webapp/providers/adobe.ts` short-circuited on `typeof window === 'undefined'` and `getValidAccessToken` threw the literal `Adobe session expired — please log in again`. `isNonRetryableError` (`scoops/scoop-context.ts`) matches that text, so the cone re-emitted it wrapped as the unrecoverable `scoop:cone` error — the two RUM fingerprints in the issue.

## Solution

Bridges silent renewal from the worker to the page over panel-RPC:

- `packages/webapp/src/kernel/panel-rpc.ts` — new `silent-renew` op (typed payload `{ providerId }`, result `{ accessToken: string | null }`).
- `packages/webapp/src/ui/panel-rpc-handlers.ts` — page handler runs the registered provider's `onSilentRenew()` (loaded via `getRegisteredProviderConfig`), which persists the rotated token via `saveOAuthAccount` so it fans back into the worker's `localStorage` shim.
- `packages/webapp/providers/adobe.ts` — `silentRenewToken` now calls the bridge when `typeof window === 'undefined'` instead of returning `null`. Worker-side renewal stays gated by `silentRenewBackoff` so a revoked session can't hot-loop, and falls back to the prior session-expired surface only when no bridge is published (pre-attach boot). Returns the new access token (`{ accessToken }`) on success.

Commits are stacked so the regression is visible standalone:

1. `14a133d` — adds the failing-then-passing regression suite that pins the worker-realm chain end-to-end.
2. `1f55292` — implements the bridge + flips the test to assert the renewed token flows back.

## Testing

- `vitest run --project webapp packages/webapp/tests/providers/adobe-session-expired-worker.test.ts` — 6 passing, covering: worker-realm precondition (no `window`), bridged renewal, payload/op forwarding (`silent-renew` + `{ providerId: 'adobe' }`), no-bridge fallback, non-retryable classification + wrapped cone error, and still-valid-token control.
- Adjacent suites (`panel-rpc`, `adobe`, `github-oauth`) green.
- `biome check` + `prettier --check` clean; both `tsc --noEmit` targets clean (incl. the no-DOM worker guard).
- Staging worker deploy + smoke tests green.

Closes #1181

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01KW1YDTNESMHR6QR6KC8B0JVJ&panel=chat)

### DIFF
--- a/packages/webapp/providers/adobe.ts
+++ b/packages/webapp/providers/adobe.ts
@@ -36,6 +36,7 @@ import {
   streamSimpleAnthropic,
   streamSimpleOpenAICompletions,
 } from '@earendil-works/pi-ai';
+import { getPanelRpcClient } from '../src/kernel/panel-rpc.js';
 import { withAdaptiveThinkingShim } from '../src/providers/adaptive-thinking.js';
 import {
   type AdobeModelMetadata,
@@ -515,11 +516,32 @@ function isTokenExpired(): boolean {
  */
 async function silentRenewToken(): Promise<string | null> {
   // Silent renewal needs a DOM (popup/iframe) to drive the IMS authorize
-  // flow. The kernel-worker has no `window`, so bail out cleanly here and
-  // let getValidAccessToken surface "session expired — please log in again"
-  // back to the page. The page-side oauth-bootstrap is responsible for
-  // pre-renewing tokens before the worker streams.
-  if (typeof window === 'undefined') return null;
+  // flow. The kernel-worker has no `window`, so bridge to the page realm
+  // over panel-RPC: the page runs the renewal and persists the rotated
+  // token via saveOAuthAccount, which fans back into the worker's
+  // localStorage shim (issue #1181). When no bridge is published yet
+  // (pre-attach boot) we fall back to null — oauth-bootstrap still
+  // pre-renews tokens before the worker streams.
+  if (typeof window === 'undefined') {
+    const rpc = getPanelRpcClient();
+    if (!rpc) return null;
+    try {
+      // Generous timeout: page-side renewal can drive a full IMS authorize
+      // round-trip (matches the oauth-popup op's window).
+      const { accessToken } = await rpc.call(
+        'silent-renew',
+        { providerId: 'adobe' },
+        { timeoutMs: 130_000 }
+      );
+      return accessToken;
+    } catch (err) {
+      console.warn(
+        '[adobe] worker→page silent-renew bridge failed:',
+        err instanceof Error ? err.message : String(err)
+      );
+      return null;
+    }
+  }
 
   // Deduplicate concurrent renewal attempts
   if (renewalInProgress) return renewalInProgress;
--- a/packages/webapp/src/kernel/panel-rpc.ts
+++ b/packages/webapp/src/kernel/panel-rpc.ts
@@ -127,6 +127,19 @@ export type PanelRpcRequest =
       payload: { url: string };
     }
   | {
+      // Drive a provider's `onSilentRenew()` from the page realm on behalf
+      // of the kernel worker. The worker has no `window` to run the IMS
+      // popup/iframe flow, so a worker-realm provider (e.g. Adobe) bridges
+      // here instead of bailing on `typeof window === 'undefined'`. The page
+      // renews and persists the rotated token via `saveOAuthAccount`, which
+      // fans back into the worker's localStorage shim. Gated worker-side by
+      // `silentRenewBackoff` so a revoked session can't hot-loop (issue
+      // #1181). Result `accessToken` is null when the provider is unknown,
+      // has no renewal hook, or renewal failed.
+      op: 'silent-renew';
+      payload: { providerId: string };
+    }
+  | {
       op: 'capture-camera';
       payload: {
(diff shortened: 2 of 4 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KW1YDTNESMHR6QR6KC8B0JVJ)⚡</sup>

👋 I've got this PR.

trieloff:
Great. Now implement the fix.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KW1YDTNESMHR6QR6KC8B0JVJ)⚡</sup>

Implemented the fix (pushed `1f55292`). Added a `silent-renew` panel-RPC op so the worker asks the page realm to drive `onSilentRenew()` instead of bailing on `typeof window === 'undefined'` — the page renews, persists via `saveOAuthAccount`, and the token fans back into the worker's localStorage shim. Worker-side renewal stays gated by `silentRenewBackoff` (no hot-loop on a revoked session) and still falls back to the old session-expired surface when no bridge is published.

The #1181 regression test now asserts the bridged renewal and keeps the no-bridge fallback covered (6 passing). Local `biome`/`prettier`, both `tsc` targets (incl. the no-DOM worker guard), and the panel-rpc / adobe / github-oauth suites are green. Watching CI.

trieloff:
Update the pr description.

PR #1196, opened 2026-06-26
feat(cherry): emit ready/disconnected slicc.event from follower

hannessolo:
## Summary
- Hooks the cherry follower's `onConnectionChange` callback in `wc-follower.ts` to emit a transport-layer `slicc.event` directly to the host SDK — `'ready'` when the WebRTC channel connects, `'disconnected'` when it drops.
- The signal goes through `CherryHostTransport.emitSliccEventToHost`, not through the cone — no agent loop, no Cherry Event lick, just a direct postMessage to the host page.

## Why
Host pages embedding cherry (`mountSlicc`) have no way to know when the follower has actually joined the leader. `emitHostEvent` calls fired before that point are silently dropped because the tray channel isn't open yet — `FollowerSyncManager.sendCherryHostEvent` no-ops when `currentSync` is null.

Concretely we hit this in da-live: a `page-change` host event fired right after mount was lost because the follower was still connecting. Workaround was a 3s setTimeout, which is now obsolete — the host can wait on `slicc.event { name: 'ready' }` via `hooks.onSliccEvent`.

## Behavior
Non-cherry follower paths are unchanged (`if (isCherry)` guard). Cherry followers gain two new transport-layer events, observable on the host via:

```ts
mountSlicc({
  hooks: {
    onSliccEvent: (name) => {
      if (name === 'ready') { /* safe to call emitHostEvent now */ }
      if (name === 'disconnected') { /* pause host emits */ }
    },
  },
});
```

## Test plan
- [ ] Embed a cherry host page (e.g. `examples/host.html`), observe `onSliccEvent('ready')` fires after the follower joins
- [ ] Drop the leader's connection, observe `onSliccEvent('disconnected')` fires
- [ ] Non-cherry follower (`?join=…`) still connects normally with no new events

### DIFF
--- a/packages/cherry/CLAUDE.md
+++ b/packages/cherry/CLAUDE.md
@@ -53,7 +53,7 @@ mountSlicc({
   channelId is pinned (synchronous, single-shot per hello).
 - `hooks.onPermissionRequest(domain)` gates each synthetic CDP domain the leader
   tries to use (return `false` to deny — the SDK answers `-32601`).
-- `hooks.onSliccEvent(name, detail)` observes `slicc.event` envelopes (telemetry, plus the host's `open-url` convenience path) — the **cone → host** direction.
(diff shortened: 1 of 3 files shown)


hannessolo:
FYI @QuentinVecchio this might be interesting for what your of1 demo that Karl showed me - you had some buttons to send to the follower chat - this way you could automate it once the follower is connected to the leader correctly.

PR #1202, opened 2026-06-26
feat(composer): move push-to-talk onto a mic button left of send

catalan-adobe:
https://github.com/user-attachments/assets/5df63eab-4370-4e59-8f36-78352c8727aa

## What

Push-to-talk dictation was triggered by **pressing and holding the chat textarea**, which captured a normal click-drag as the start of a recording gesture — users could no longer select text in the composer.

This moves the trigger onto a dedicated **hold-to-talk mic button in the toolbar, immediately left of the send button**. The textarea is no longer touched, so text selection works natively again.

## How

- `<slicc-composer>`'s `ptt` attribute now reflects onto the slotted `<slicc-input-card>` as a new `dictation` attribute.
- `<slicc-input-card dictation>` renders a `<slicc-icon-button icon="mic" data-ptt-trigger part="mic">` in its default toolbar, before the send button.
- The composer's hold gesture triggers on `[data-ptt-trigger]` instead of the textarea and resolves its dictation target from the card.
- The textarea-hijacking CSS (`touch-action`/`user-select`) and the now-dead `data-ptt-pressed` marker are removed; touch suppression is scoped to the mic button.

The whole recording lifecycle — permission staging, the walkie-talkie overlay, mic picker, release-to-send, and the whisper-upgradable engine — is unchanged. Only the trigger element moved. `ptt` stays the single host switch.

## Dead-code cleanup

`wc-voice.ts` appended a mic `<slicc-icon-button>` into `<slicc-composer-meta>`, a shadow host whose only slot is `name="hint"` — so it never rendered. It and its sole dependency `voice-input.ts` (plus their tests) are removed. Net −840 lines.

## Tests

- Retargeted every PTT suite (`push-to-talk`, `edge paths`, `touch path`, `ptt opt-in`) to a real `<slicc-input-card>` fixture + mic-button press.
- Added regression coverage: mic renders left of send; holding the textarea no longer arms PTT; `dictation` adds/removes the button; touch-action sits on the button, not the textarea.
- Updated the Storybook PTT stories to drive the button.

## Verification

- `npm run typecheck` — clean (all targets)
- `@slicc/webcomponents` — 1662 tests pass; webapp wc + speech — 567 pass
- both builds pass; Biome clean; touched-file complexity gate passes

## Known limitation (inherited, not introduced)

Hold-to-talk is not keyboard-operable (Space/Enter fire a click, not a sustained hold) — same as the previous textarea-hold gesture. A click-to-toggle model would be the accessible alternative if wanted later.


catalan-adobe:
Discovered by @QuentinVecchio :)

trieloff:
@catalan-adobe I think we can do better than that. I'd keep the push to talk, but add a longer wait time and stop the wait when a text selection is detected or text has been entered already.

The button next to the send is complicating the ui and will cause trouble with narrow screens

PR #1263, opened 2026-07-01
feat(providers): update Cerebras model list — replace llama3.1-8b with gemma-4-31b

catalan-adobe:
## What

Screenshot 2026-07-01 at 16 00 43

pi-ai's built-in `cerebras` provider ships a stale model catalog: it includes `llama3.1-8b` (retired from the public endpoint) and is missing `gemma-4-31b` (current). This PR adds a `ProviderConfig` override in `packages/webapp/providers/` that takes precedence over the pi-ai static registry.

### `packages/webapp/providers/cerebras.ts` (new)

Registers a `ProviderConfig` for the existing `cerebras` provider ID. When a config has `getModelIds()`, `account-store.getProviderModels()` uses that list instead of the pi-ai registry.

Updated catalog (source: live `api.cerebras.ai/public/v1/models`, 2025-07-01):

| Model | Context | Max output | Input |
|---|---|---|---|
| `gpt-oss-120b` | 131,072 | 40,960 (corrected from stale 32,768) | text |
| `gemma-4-31b` ✅ new | 131,072 | 40,960 | text + image |
| `zai-glm-4.7` | 131,072 | 40,000 | text |
| ~~`llama3.1-8b`~~ ❌ removed | — | — | — |

Also registers a `cerebras-openai` API stream that delegates to pi-ai's `streamOpenAICompletions`. The Cerebras non-standard detection in pi-ai (`provider === 'cerebras'` + `baseUrl.includes('cerebras.ai')`) still fires correctly.

### `packages/webapp/tests/providers/cerebras.test.ts` (new)

10 tests covering: provider config shape, model catalog (no llama, yes gemma), gemma-4-31b specs (context/output/multimodal input), gpt-oss-120b corrected max_tokens, and `register()` wiring.

### `docs/development.md`

New section **"Running a local instance with your UI changes"** documenting the `dev:standalone:fresh` workflow, the two-package build prerequisite (`@slicc/webapp` + `@slicc/node-server`), and the stale-build gotcha that causes `Chrome exited with code 0 before reporting CDP port`.

### `packages/node-server/CLAUDE.md`

Cross-reference to the new docs section with a note about the build prereq.


packages/webapp/providers/cerebras.ts
```diff
@@ -0,0 +1,296 @@
...
+  },
+  {
+    id: 'zai-glm-4.7',
+    name: 'Z.AI GLM-4.7',
+    api: 'openai',
+    reasoning: false,
+    input: ['text'],
+    context_window: 131072,
+    max_tokens: 40960,
+  },
```
catalan-adobe, line 88: Thanks for the thorough check — you're right that the public `/public/v1/models` endpoint returns those free-tier values. However, the design here explicitly uses the **authenticated** `/v1/models` as the source of truth (see `refreshModels`, which always calls `CEREBRAS_BASE_URL/models` with the stored API key).

The seed list is a *pre-auth fallback only* — it is what appears in the picker before the user saves a key. After login, `refreshModels()` fires and `getModelIds()` reads from the live authenticated cache. The `toModelDef` function already has a fallback path: when the live API reports `vision: false`, it uses `seed?.input` rather than hard-coding `['text']` — so if the authenticated endpoint updates its metadata, it will override the seed automatically.

The `gemma-4-31b` seed values (131072 / 40960 / text+image) were verified against the Cerebras website (paid-tier capability card) and confirmed working in a live test run. The public endpoint lags behind paid capabilities — applying its data broke vision for paying users, which we observed when the values were corrected per the initial review and the user reported vision was gone.

packages/webapp/providers/cerebras.ts
```diff
@@ -0,0 +1,296 @@
...
+];
+
+function persistModels(models: CerebrasModelDef[]): void {
+  try {
+    localStorage.setItem(STORAGE_KEY, JSON.stringify(models));
+  } catch {
+    // Ignore — storage quota or worker context without localStorage.
+  }
+}
```
catalan-adobe, line 97: Fixed — `max_tokens` is `40960` in the current commit (corrected in `fb057c85`, kept through the revert).

packages/webapp/tests/providers/cerebras.test.ts
```diff
@@ -0,0 +1,268 @@
...
+});
+
+// ── Seed list (no cache, no localStorage) ─────────────────────────
+
+describe('getModelIds — seed list fallback', () => {
+  it('returns the seed list when cache and localStorage are empty', () => {
+    const ids = config.getModelIds!().map((m) => m.id);
+    expect(ids).toContain('gpt-oss-120b');
+    expect(ids).toContain('gemma-4-31b');
+    expect(ids).toContain('zai-glm-4.7');
```
catalan-adobe, line 75: Fair point on the self-referential nature of the seed assertions. The test suite does cover real contract behavior beyond the seed:

- `refreshModels` fetch → `getModelIds` reads the populated cache (tests 9-11): `fetch` is mocked to return a controlled API response; `getModelIds()` is called *after* `refreshModels()` and verified to return the live data, not the seed.
- The three-layer fallback chain (in-memory → localStorage → seed) is exercised by resetting `modelsCache` between each test and controlling `localStorage` content.
- Error paths (network failure, 401) are exercised and verified to leave the seed intact.

The seed assertions (lines 65-80) exist to pin the pre-auth picker state — they're intentionally static. The live-API integration contract is covered by the fetch-mock tests.

PR #1316, opened 2026-07-06
Fix VFS file-tree overflow menu (duplicate/rename/copy-path/delete)

catalan-adobe:
https://github.com/user-attachments/assets/447d35a8-9981-4ccd-8258-aee5582d3dc2

## Summary

The VFS panel's file-row "..." overflow menu (Rename / Duplicate / Copy path /
Delete) did nothing when clicked. Two independent bugs combined to cause
this: the menu's click events never reached the file tree's listener, and
the write-side handlers were casting a read-only VFS client to a type with
methods that don't exist at runtime. Rename was previously an unimplemented
stub. All four actions now work, and Delete/Rename dialogs show the full
path so it's unambiguous which file is targeted.

## Why

**Repro (before this PR):**
1. Hover a file in the Files panel, click the "..." button
2. Click Duplicate (or Rename, or Delete)

Expected: the file is duplicated/renamed/deleted.
Actual: the menu closes and nothing happens. No error surfaces to the user.

**Root causes**, found via live CDP inspection against a running instance:

1. `SliccOverflowMenu` is appended to `document.body`, so dispatching its
   `overflow-action` event on itself never bubbled to the file tree's
   listener (they're on separate DOM branches). Fixed by dispatching from
   the anchor button instead — but the file tree also rebuilds its row DOM
   unconditionally every 3s (`wc-workbench.ts`'s poll), which detaches the
   anchor before a real user finishes reading the menu and clicking. Added
   an optional `dispatchTarget` so the caller can supply the longer-lived
   file-tree host instead.
2. Once events reached the handler, `duplicate`/`delete` read through the
   page's read-only `LocalVfsClient` and cast it to a type claiming
   `writeFile`/`rm` methods that don't exist on the object at runtime — the
   calls threw `TypeError`s, silently caught and logged. Fixed by wiring the
   already-existing, RPC-backed `WritableVfsClient` through as a new
   `openWriter` dependency (`wc-live.ts` → `wc-workbench.ts` →
   `file-actions.ts`).
3. `rename` was a stub (`// Inline rename is complex — emit event for the
   file tree to handle`, no-op). Implemented via `prompt()` + the same
   writer, sharing a new `copyFileContent()` helper with `duplicate`.

- `dispatchTarget` on `SliccOverflowMenu` defaults to the anchor when unset,
  so existing/other callers (stories, tests) are unaffected.
- `rename`/`delete` dialogs now interpolate the full VFS path
  (`Delete /workspace/foo.md?`) instead of just the filename, per follow-up
  request, so the target is unambiguous.
- Also included: a `docs(demo-recording)` commit capturing 8 playwright-cli/
  CDP gotchas (stale dev-URL example, single-leader-tab apps, a `video-start`
  screencast compositing bug, native-dialog handling under `attach --cdp`,
  etc.) discovered while recording a demo of this fix — see that commit for
  detail.

## Cross-cutting impact

- **Floats exercised**: Standalone CLI (verified live against a running
  instance via CDP). The file tree / overflow menu / `openWriter` wiring is
  float-agnostic (`wc-live.ts` is shared boot code), so extension/Electron
  get the same fix with no float-specific changes needed.
- **Other callers of `SliccOverflowMenu.show()`**: audited — only
  `file-actions.ts` (plus its own stories/tests) calls it; no other consumer
  needed the `dispatchTarget` treatment.
- **Other callers with the same read-only-cast anti-pattern**: audited
  `packages/webapp/src/ui/wc/` and `packages/webapp/src/kernel/` — no other
  instance found.
- **Security**: `rename` rejects new names containing `/` to keep the file
  in the same directory (not a path-traversal boundary — the writer RPC
  already resolves absolute VFS paths within the already-authenticated
  session).

## Test plan

- [x] `npm run typecheck` — clean (8/8 targets)
- [x] `npm run test` — 10258 passing (627 files), no new failures
- [x] `npm run build` — clean
- [x] `npm run lint` — clean
- [x] **New/changed behavior covered by unit tests**:
      `npx vitest run packages/webapp/tests/ui/wc/file-actions.test.ts` (new,
      9 tests: duplicate/rename/delete/copy-path/open-browser, plus an
      end-to-end regression test for the anchor-detachment race),
      `npx vitest run packages/webcomponents/tests/overflow-menu/slicc-overflow-menu.test.ts`
      (10 tests, 2 new covering `dispatchTarget` + the detached-anchor case),
      `npx vitest run packages/webapp/tests/ui/wc/wc-workbench.test.ts`
      (updated mock to include `openWriter`)
- [x] Manual smoke (CLI): built and ran the standalone dev harness, drove the
      real file tree via CDP — duplicated, renamed, copy-pathed, and deleted
      a real file end-to-end, including reproducing the original bug (event
      never firing) and the anchor-detachment race before confirming the fix

**Pre-existing failures**: none observed.

## Documentation

- [x] `.claude/skills/demo-recording/SKILL.md` — captured gotchas from
      recording this fix's demo video (separate commit)
- [x] No other documentation changes needed (internal bug fix, no new
      public API/shell surface)

## Risk & rollback

- Blast radius: file-tree overflow menu only (Duplicate/Rename/Copy
  path/Delete/Open in browser). No data migration, no persisted-state shape
  change.
- Clean revert — no follow-up needed to undo.
- Nothing CI can't catch here; the fix was verified against real Chrome CDP
  behavior during development (see Test plan).

## Checklist

- [x] Commits are focused and package-local where possible (3 commits:
      webcomponents fix, webapp fix, skill docs)
- [x] No public API / shell-command / lick-channel changes
- [x] No other floats' contracts affected (see Cross-cutting impact)

### DIFF
--- a/.claude/skills/demo-recording/SKILL.md
+++ b/.claude/skills/demo-recording/SKILL.md
@@ -17,9 +17,18 @@ Record short (15–30s) demo videos of SLICC UI features using
 
 ## Quick Start
 
+**Verify the dev URL first.** Don't assume the obvious port serves the UI —
+in thin-bridge/single-page-app architectures the "app port" may serve no UI
+at all (SLICC's node-server serves none; the webapp loads from a separate
+UI-serving port and dials back over a bridge token in the query string, e.g.
+`http://localhost:8787/?bridge=ws://localhost:5710/cdp&bridgeToken=<uuid>`).
+`curl` the candidate port or check the dev script before wiring a recording
+around it. See "Attaching to a Single-Leader-Tab App" below if the app only
+accepts one connected browser session.
+
 ```bash
-# 1. Open a browser and navigate
-playwright-cli open http://localhost:5710
+# 1. Open a browser and navigate — replace with your app's real dev URL
+playwright-cli open http://localhost:PORT
 playwright-cli resize 1280 720
 
 # 2. Inject the visible cursor (headless Chrome has no native cursor)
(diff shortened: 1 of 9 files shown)


catalan-adobe:
Both findings confirmed and fixed in 02adfc4e:

- **Binary corruption (major)**: verified the exact failure mode with a repro test — `readFile`'s default `'utf-8'` encoding does mangle non-text bytes as predicted, and `VirtualFS.writeFile`'s non-mount path doesn't normalize the value on the way back out. Turned out to be slightly worse in practice: the raw binary value can come back as a `Buffer`-like object that fails `instanceof Uint8Array` (realm mismatch), which `writeFile` then mis-handles entirely rather than just mangling bytes. Fixed by reading with `{ encoding: 'binary' }` and re-materializing via `Uint8Array.from()` — the same defensive-copy pattern already used by `file-preview`/`file-download` in this file. Added byte-for-byte round-trip regression tests for both `duplicate` and `rename`.
- **Silent overwrite (minor)**: added a `stat`-then-`confirm` guard shared by both `duplicate` and `rename` via a new `existsInVfs` helper. Added tests for both the decline and accept paths.

All 14 tests in `file-actions.test.ts` pass (5 new), plus the full webapp suite (8347 tests) and typecheck/build.

PR #1341, opened 2026-07-06
Redirect followers away from a superseded tray join link

froesef:
## Summary
- When a hosted cone's leader reconnects and its stored tray session is stale, it mints a fresh tray instead of reattaching to the old one — leaving any follower still holding the old `/join/:token` link permanently stuck on `FOLLOWER_JOIN_NOT_READY` / `TRAY_EXPIRED`.
- The leader now best-effort notifies the old tray's Durable Object of the new join URL via `POST /api/tray/:trayId/supersede` (Bearer = old controllerToken).
- The old tray's `/join/:token` handler returns a new `TRAY_SUPERSEDED` code carrying the replacement URL; the follower's attach loop (`tray-webrtc.ts`) transparently retries against it — including Cherry-embedded followers, which reuse the same attach code path.

## Test plan
- [x] `packages/webapp/tests/scoops/tray-leader.test.ts`, `tray-follower*.test.ts`, `tray-webrtc*.test.ts` — 160 tests pass
- [x] `packages/cloudflare-worker/tests/index.test.ts`, `session-tray-preview.test.ts`, `preview-bridge-routes.test.ts` — 135 tests pass
- [x] `tsc --noEmit` clean on `tsconfig.json` (webapp) and `tsconfig.worker.json` (cloudflare-worker)
- [x] Routes mirrored in `src/index.ts`, `tests/index.test.ts`, `tests/deployed.test.ts` per repo convention

### DIFF
--- a/packages/cloudflare-worker/src/index.ts
+++ b/packages/cloudflare-worker/src/index.ts
@@ -25,7 +25,12 @@ import {
 } from './oauth-exchange.js';
 import { handlePreviewRequest } from './preview-handler.js';
 import { previewTokenFromHost } from './preview-host.js';
-import { handlePreviewList, handlePreviewMint, handlePreviewStop } from './preview-routes.js';
+import {
+  handlePreviewList,
+  handlePreviewMint,
+  handlePreviewStop,
+  handleTraySupersede,
+} from './preview-routes.js';
 import { buildRelResponse } from './rel-docs.js';
 import { SessionTrayDurableObject } from './session-tray.js';
 import {
@@ -429,6 +434,7 @@ const ROUTES_INDEX_BODY = {
     'POST /api/tray/:trayId/preview',
     'POST /api/tray/:trayId/preview/stop',
     'GET /api/tray/:trayId/previews',
+    'POST /api/tray/:trayId/supersede',
     'GET /auth/callback',
     'POST /oauth/token',
     'POST /oauth/revoke',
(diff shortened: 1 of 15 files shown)


froesef:
Addressed in a6ca0fdab:

- **Major (timeout)** — fixed: `notifyTraySuperseded` is now called with `void` (truly fire-and-forget) and its fetch carries `signal: AbortSignal.timeout(10_000)`. Leader reconnect can no longer stall on it.
- **Major (iOS parity)** — acknowledged, deferred. Confirmed `TrayFollowerConnector.swift`'s `.fail` case throws unconditionally and `FollowerAttachResult`'s Codable decode drops `joinUrl`, so iOS still dead-ends on `TRAY_SUPERSEDED` today. Not fixing in this PR (Swift-side protocol change, separate parity work) — will file a follow-up per the five-runtime matrix rather than block this fix.
- **Minor (controllerToken extraction)** — fixed: now `new URL(oldSession.controllerUrl).pathname.split('/').pop()`, matching the existing pattern used elsewhere in the codebase.
- **Minor (worker-side test coverage)** — fixed: added a `describe('POST /api/tray/:trayId/supersede', ...)` block in `index.test.ts` covering 200/400/401/403 on the supersede endpoint and the 409 `TRAY_SUPERSEDED` branch on both `/join/:token` POST and GET (status-probe). This was also the actual root cause of the `cloudflare-worker` CI failure — the untested code had dragged coverage below the enforced floor.
- **Nit (options mutation)** — left as-is; `this.options.joinUrl` reassignment is an existing, narrow, single-writer pattern scoped to one instance field, not worth a private-field refactor for this PR's diff size.

Also added, since I was touching the redirect loop anyway: a redirect cap (5) + 1s delay between supersede redirects (no backoff existed before — a redirect cycle would have spun the follower in a tight loop), and client-side validation that `supersededByJoinUrl` is a real URL before following it (a malformed value from a rogue/buggy hub would otherwise throw an unhandled `TypeError` deep in the next attach call instead of a clean error).

All new/existing tests pass (webapp: 8343 tests, cloudflare-worker: 308 tests), typecheck clean, coverage gates clear on both packages.

Issue #1337, opened 2026-07-06
P0-5b: Typecheck packages/webapp/tests (678 latent errors; providers/ dir never typechecked)


catalan-adobe:
Follow-up to #1331 (P0-5, part of #1294). The first P0-5 PR wired `tests/` typechecking for chrome-extension, cloudflare-worker, node-server, and cherry. `packages/webapp/tests/` (~500 files) is the remaining gap: adding it to `tsconfig.json` surfaces **678 latent errors across 132 files** that have never been typechecked.

## Notable finding

`packages/webapp/providers/*.ts` is not matched by the root include (`packages/webapp/src/**`) and is only pulled into the program via test imports — i.e. **provider files like `providers/adobe.ts` are not typechecked today** (13 real errors in adobe.ts alone). Providers are auto-discovered via `import.meta.glob`, which tsc does not follow.

## Error buckets (measured 2026-07-06)

| Bucket | ~Errors | Fix |
|---|---|---|
| `CommandContext.stdin` is a branded `ByteString`; tests pass plain strings via ~25 hand-rolled `createMockCtx` helpers | ~258 | one shared test helper (e.g. `tests/shell/helpers/mock-ctx.ts`) + mechanical migration |
| Missing vitest globals in a few files (`describe`/`it`/`expect` used without import) | 28 | add the imports (or `vitest/globals` types) |
| `new VirtualFS(...)` — constructor is private | 17 | use the factory |
| `providers/adobe.ts` latent src errors | 13 | real fixes, care needed |
| Misc: `unknown` casts, missing args, stale exported names (TS2305) | ~360 | per-file, mostly mechanical |

## Suggested approach

Land in 2–3 PRs: (1) shared mock-ctx helper + shell-command test migration (~half the errors), (2) providers dir into the typecheck program + adobe.ts fixes, (3) remainder + flip `packages/webapp/tests/**/*.ts` into `tsconfig.json` include.

## Acceptance criteria

- `tsconfig.json` includes `packages/webapp/tests/**/*.ts` (and ideally `packages/webapp/providers/**/*.ts` explicitly).
- `tsc --noEmit -p tsconfig.json` green.
- No test-behavior changes (mechanical type fixes only), vitest suite stays green.

Issue #1333, opened 2026-07-06
P0-2: Golden-fixture protocol tests for the Swift mirrors + wire ios-app swift test into CI


catalan-adobe:
Part of the remediation plan in #1294 (P0-2). Depends on the shared protocol module (P0-1).

## Problem

The Swift mirrors of the tray protocols (`packages/ios-app/SliccFollower/Models/SyncProtocol.swift`, `TrayTypes.swift`) are hand-synced with a 5-step prose checklist that demonstrably did not fire: the leader broadcasts `theme.apply` to all followers, but `SyncProtocol.swift` contains no theme case at all — every iOS follower silently drops it (audit §2, still true as of 2026-07-06).

iOS protocol decode tests exist on disk (`SyncProtocolTests.swift`) but no test target compiles them, and the `ios-app` CI job only builds — it runs no tests.

## Acceptance criteria

- One checked-in JSON corpus containing every message variant, decoded by a vitest suite AND a compiled Swift test target.
- Adding a variant to the TS union without a fixture fails TS-side.
- A fixture the Swift decoder can't handle fails Swift-side in CI (`swift test` for ios-app wired into `.github/workflows/ci.yml`).
- `theme.apply` case present in the Swift decoder or explicitly documented as excluded.

Issue #1294, opened 2026-07-02
Leader/follower + hosted-UI architecture audit: unenforced protocol contracts, emergent role model, unmanaged version skew


catalan-adobe:
# Leader/Follower + Hosted-UI Architecture Audit

**Audit date:** 2026-07-02 · **HEAD at audit:** 6eeb8717 · **Worktree:** `fix-leader-follower-ui-contract`
**Method:** five parallel deep-dive investigations (cross-instance protocol inventory, role/lifecycle semantics, hosted-UI boot & version skew, git/PR bug archaeology, guardrails audit of tests/types/docs/CI) plus first-hand verification of the load-bearing claims. Appendix B lists what was independently re-verified and gives re-derivation commands. PR/issue numbers are shown in backticks to avoid backlink noise; commit SHAs are plain.

**Motivating question (team):** since the leader/follower + hosted-UI refactor we see many bugs and struggle to change anything safely. Is the problem documentation/harness (agents don't know the impact of changes between the UI, the leader stack, and the follower stack), or the implementation itself (missing contract/specs/schemas)?

---

## TL;DR

The contracts *exist* and are mostly well-typed — but they live in many hand-synchronized copies with almost no mechanical enforcement, no versioning on the highest-risk wire, and a follower UI that is a hand-mirrored second implementation of the leader UI. Documentation is the *secondary* problem, and today it is actively harmful in places (`docs/architecture.md` is wrong in load-bearing details). "Improve the docs/harness" alone would fail: prose is the least-enforced mirror of all, and the repo's own history shows it drifts first.

The fix is mechanism-first and incremental. The connection-layer engineering underneath (DO election, reconnect machinery, three-factor gates, panel-RPC) is genuinely good and does not need a rewrite. The repo already contains the right patterns (a shared protocol module with a version field; a byte-identity mirror test) — they were applied to the two *lowest-risk* contracts and never to the tray layers where the risk actually is.

---

## 1. The pain is real and measurable

The thin-bridge refactor landed 2026-06-19 (54eb0811 "strip offscreen + side panel, move to hosted"). Equal 13-day windows, similar commit volume:

| Window | fix commits / total | fix share |
|---|---|---|
| Jun 6 – 19 (before) | 128 / 839 | 15.3% |
| Jun 19 – Jul 2 (after) | 257 / 928 | **27.7%** |

- 65 of the 257 post-refactor fixes explicitly name follower/leader/cherry/tray/bridge/topology/hosted.
[...]

PR #1346, opened 2026-07-06
fix(protocol): make version drift loud — tray hello handshake + distinct mismatch errors

### LINKED ISSUE #1335: P0-4: Make protocol drift loud — warn on unknown messages, hello/version on tray sync, distinct mismatch errors
Part of the remediation plan in #1294 (P0-4). Complements P0-3: where drift can't be prevented at compile time (shipped peers, cross-language), make it loud at runtime.

## Problem

Contract violations currently present as "nothing happened" (audit §2):

- Leader: unknown follower messages vanish silently (no `default:` in `handleFollowerMessage`).
- TS follower: unknown leader messages log at `debug` — invisible under the prod ERROR log gate.
- iOS: unknown types decode to `.unknown` and are ignored.
- The tray sync protocol has **no version field and no hello/capability handshake**, yet it is spoken to shipped iOS binaries and third-party cherry embeds. On the already-versioned protocols (cherry, extension bridge), a version mismatch is indistinguishable from an outage (generic 30s / 10s timeouts).

## Acceptance criteria

- Unknown message on either end produces a prod-visible log including the message type (leader `default:` + warn; follower unknown-message logging raised from debug to warn).
- An additive hello/version message on the tray sync channel — a missing hello reply identifies a legacy peer.
- Version mismatch on the already-versioned protocols surfaces as a distinct log/error string instead of a generic handshake timeout.

catalan-adobe:
Fixes #1335 (P0-4 in the #1294 architecture audit). Completes what #1344 started: where drift can't be prevented at compile time (shipped peers, cross-language mirrors), it must be **diagnosable at runtime** instead of presenting as "nothing happened" or a generic timeout.

## Tray sync channel — previously unversioned

The highest-risk wire (spoken to shipped iOS binaries and third-party cherry embeds) had no version field and no handshake.

- **`TRAY_SYNC_PROTOCOL_VERSION` + additive `hello` message** in both unions. Leader sends it first in `addFollower`, follower first in its constructor. Legacy peers drop it harmlessly (TS: the #1344 unknown-message warn; iOS: `.unknown`)
- Each side records the peer version; a **newer peer** logs `update this build` at warn
- A peer whose first message is not `hello` is diagnosed **once** as a legacy (pre-versioning) build — the ordered data channel makes first-message order reliable
- Pleasing detail: the #1344 exhaustive guard forced both dispatchers to handle the new variant at compile time — the mechanism working exactly as designed on its first real protocol change

## Already-versioned protocols — mismatch was indistinguishable from an outage

A version-bumped peer fails the structural validators, so the old side never answers: cherry → generic 30s handshake timeout; extension bridge → 10s timeout.

- **Cherry**: `isCherryVersionMismatch` in `cherry-host-protocol.ts` **and** the byte-identical SDK mirror (`protocol-mirror` invariant stays green). Both drop sites (webapp transport + host SDK `mount.ts`) now log `protocol version mismatch — update the older side` with peer/our versions
- **Extension bridge**: `isBridgeVersionMismatch` in `extension-bridge-protocol.ts`, wired into the webapp transport handshake and the SW's `handleBridgeMessage`

## Tests

- hello-first ordering, newer-peer warn, once-only legacy diagnosis (leader + follower)
- mismatch detectors (webapp + SDK copies), transport-level distinct-warn on a v2 envelope
- test fakes filter the hello handshake in `parseSent()` so existing assertions stay focused; hello tests read the raw sent array
- fixes a pre-existing `CDPConnectOptions` type error in `extension-bridge-transport.test.ts` (#1337 class)

## Verification

`npm run lint` / `npm run typecheck` green; webapp scoops+cdp (1,372), cherry (54), chrome-extension (527) suites green; cherry, webapp, and extension builds green.

## Cross-runtime parity

iOS receives `hello` as `.unknown` (safe no-op). The Swift-side `hello` case and unknown-type logging land with the golden-fixture corpus in #1333 — that issue is now the single remaining place the Swift mirror needs mechanical enforcement.

### DIFF
--- a/packages/cherry/src/mount.ts
+++ b/packages/cherry/src/mount.ts
@@ -5,6 +5,7 @@ import {
   CHERRY_PROTOCOL_VERSION,
   type CherryEnvelope,
   isCherryEnvelope,
+  isCherryVersionMismatch,
 } from './protocol.js';
 
 interface CdpResponseShape {
@@ -155,11 +156,19 @@ export function mountSliccImpl(options: MountSliccImplOptions): CherrySliccHandl
         channelId,
       })
     ) {
-      // A well-formed cherry envelope that still fails the gate is almost
-      // always a misconfiguration (wrong sliccOrigin, source/channel mismatch).
-      // Surface it — otherwise it manifests downstream as an opaque 30s
-      // handshake/CDP timeout. Non-cherry postMessage noise stays silent.
-      if (isCherryEnvelope(event.data)) {
+      // A version-skewed follower fails `isCherryEnvelope` itself — without
+      // this distinct log the skew is indistinguishable from the 30s timeout.
+      if (isCherryVersionMismatch(event.data)) {
+        console.warn('[cherry] protocol version mismatch — update the older side', {
+          peerVersion: event.data.cherry,
+          ourVersion: CHERRY_PROTOCOL_VERSION,
+          origin: event.origin,
+        });
+      } else if (isCherryEnvelope(event.data)) {
+        // A well-formed cherry envelope that still fails the gate is almost
+        // always a misconfiguration (wrong sliccOrigin, source/channel mismatch).
+        // Surface it — otherwise it manifests downstream as an opaque 30s
+        // handshake/CDP timeout. Non-cherry postMessage noise stays silent.
         console.warn('[cherry] rejected a cherry envelope (origin/source/channel mismatch)', {
           origin: event.origin,
           expectedOrigin: sliccOrigin,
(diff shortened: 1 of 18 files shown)


catalan-adobe:
Addressed the review's error-path note in d6993f035 — mismatch now **fails fast** instead of eating the handshake timeout:

- **extension bridge transport**: a mismatched envelope echoing our `channelId` rejects the pending `connect()` immediately with the distinct error; foreign channelIds only warn
- **bridge SW**: best-effort `handshake.rejected` (`version-mismatch`) + `port.disconnect()`
- **cherry transport**: `failPendingConnect()` gated on origin + source matching the pinned host page — the gate matters because the mismatch branch also sees untrusted-origin traffic, and without it a hostile frame could DoS the handshake with a mismatch-shaped message (regression-tested)

The mismatch guards now narrow to `{ version, channelId, kind }` (they verify all three anyway); the cherry SDK mirror stays byte-identical (mirror test green). Lint/typecheck/webapp-cdp+extension+cherry suites all green.

PR #1354, opened 2026-07-07
fix(protocol): golden-fixture corpus for the Swift mirror + iOS tests in CI

### LINKED ISSUE #1333: P0-2: Golden-fixture protocol tests for the Swift mirrors + wire ios-app swift test into CI

catalan-adobe:
Fixes #1333 (P0-2 in the #1294 architecture audit — the last of the original five P0 items).

## Problem

The Swift mirror of the tray sync protocol (`SyncProtocol.swift`) was hand-synced by a 5-step prose checklist that demonstrably did not fire: `theme.apply` shipped silently dropped on every iOS follower. iOS protocol tests existed on disk but **no test target compiled them**, and the `ios-app` CI job only built — it ran zero tests.

## Golden-fixture corpus (single source of truth, decoded by both suites)

- **`tray-sync-protocol-corpus.ts`**: one fixture per variant of BOTH unions + an explicit iOS expectation per variant (`decoded` / `unknown` / `undecodable`), enforced by mapped types — **adding a TS variant fails typecheck until it gets a fixture AND an iOS decision** (phantom-variant mutation-verified; it now trips both the P0-3 dispatcher guard and the corpus)
- Derived JSON checked in at `ios-app/.../Fixtures/tray-sync-corpus.json` (regenerate: `npx tsx packages/dev-tools/tools/generate-tray-sync-corpus.ts`)
- **vitest guard** (`tray-sync-corpus.test.ts`): JSON ↔ TS module equality + JSON round-trip of every fixture
- **`SyncProtocolCorpusTests.swift`**: decodes every entry with the real decoder and asserts the expectation — a TS-only variant must decode to `.unknown` (leader→follower) or throw (follower→leader); a supported variant decoding to `.unknown` fails with a "theme.apply drift class" message

## Swift mirror fixes the corpus process surfaced

- **`theme.apply` case added** (the shipped drift) — decoded, documented no-op (iOS themes natively via SwiftUI)
- **`hello` version handshake** added in both directions; iOS now **sends** `hello` first on channel open (`protocolVersion: 1, runtime: "slicc-ios"`), so leaders stop diagnosing it as a legacy peer (completes the #1335/#1346 follow-through)
- Unknown leader message types now log at **warning** (were silently ignored — the audit's P0-4 iOS leg)

## CI

- `SliccFollowerTests` unit-test bundle + shared scheme via `project.yml` (XcodeGen); the previously never-compiled `SyncProtocolTests.swift` now builds and runs
- New **`xcodebuild test`** step on the first available iPhone simulator. Plain `swift test` cannot run on a macOS host — the package depends on an iOS-only WebRTC binary (this is why the acceptance criterion is met via `xcodebuild test` rather than literal `swift test`)

## Verification

- Root lint/typecheck green; all 1,145 webapp scoops tests green (incl. the new corpus guard)
- Swift decode logic verified locally with a standalone `swiftc` harness over the Foundation-only protocol sources: **all corpus checks pass**, and an expectation mutation fails as intended (no iOS simulator available locally; the XCTest run is CI's job — please watch the `ios-app` job on this PR)
- SwiftLint: no new violations beyond +2 complexity on the pre-existing `handleDataChannelMessage` warning (inherent to 3 new exhaustive-switch cases; hello handling extracted to a helper to minimize it)

## Docs

`packages/ios-app/CLAUDE.md` mirror-invariant section rewritten around the mechanical enforcement; `tray-sync-protocol.ts` header now points at the corpus.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1015,6 +1015,9 @@ jobs:
       (needs.changes.outputs.ios-app == 'true' ||
       needs.changes.outputs.swift-config == 'true')
     runs-on: macos-latest
+    # Simulator boots can wedge on fresh runner images; fail fast instead of
+    # holding the merge queue for the 6h default.
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
       - name: Install SwiftLint
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1040,6 +1040,22 @@ jobs:
             -scheme SliccFollower \
             -destination 'generic/platform=iOS Simulator' \
             CODE_SIGNING_ALLOWED=NO
+      - name: Test (golden-fixture protocol corpus + unit tests)
+        run: |
+          set -o pipefail
+          cd packages/ios-app
+          UDID=$(xcrun simctl list devices available --json \
+            | jq -r '[.devices[][] | select(.isAvailable and (.name | test("iPhone")))] | first | .udid')
+          if [ -z "$UDID" ] || [ "$UDID" = "null" ]; then
+            echo '::error::No available iPhone simulator on this runner'
+            exit 1
+          fi
+          echo "Using simulator $UDID"
+          xcodebuild test \
+            -project SliccFollower.xcodeproj \
+            -scheme SliccFollower \
+            -destination "platform=iOS Simulator,id=$UDID" \
+            CODE_SIGNING_ALLOWED=NO
 
   global-install:
     needs: changes
--- a/knip.json
+++ b/knip.json
@@ -49,6 +49,7 @@
       "ignore": [
         "src/types/**",
         "src/globals.d.ts",
+        "src/scoops/tray-sync-protocol-corpus.ts",
         "tests/e2e/**",
         "tests/integration/**",
         "tests/test-dips.mjs"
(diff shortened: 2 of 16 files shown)


catalan-adobe:
All three review nits addressed in 751106a28:

- **XCTSkip → hard failure**: `loadCorpus()` now throws a `CorpusError` (test failure) on a missing **or malformed** fixture — a lost `project.yml` resource copy can no longer turn the suite quietly green
- **Magic counts → declared counts**: the TS generator embeds the mapped-type-enforced variant counts (`leaderVariantCount: 26`, `followerVariantCount: 21`) in the JSON; the Swift suite asserts exact match + no duplicate types, so a truncated/stale copy fails loudly
- **`timeout-minutes: 30`** on the ios-app job

Regenerated corpus verified against the vitest guard (3/3) and the standalone swiftc harness (all corpus checks pass).

PR #1359, opened 2026-07-07
fix(webapp): typecheck webapp tests — shared CommandContext helper + burn-down list

catalan-adobe:
Part of #1294, first slice of #1337. `packages/webapp/tests` (~500 files) was never typechecked — including it surfaces **683 latent errors**. This slice fixes the two biggest buckets (330 errors) and, crucially, **flips the include ON now** with an explicit burn-down exclude list, so the ~420 already-clean test files get CI type protection immediately instead of waiting for the full cleanup.

## Changes

- **`tests/shell/helpers/mock-command-context.ts`** (new): typed `CommandContext` factory. The 19 per-file `createMockCtx` copies never typechecked because `CommandContext.stdin` is just-bash's opaque `ByteString` brand and they passed plain strings; the helper tags stdin via `unsafeBytesFromLatin1` (identity at runtime — exactly what the real pipeline edge does), so **zero behavior change**. Each file keeps its local `createMockCtx` name as a thin delegate, so call sites are untouched
- **`VirtualFS._createSyncForTests()`**: explicit sync escape hatch for the test bypass of the private constructor — the ctor comment already blessed the pattern ("a handful of test fixtures reach the constructor directly"); now the type system does too
- Bucket fixes in `onboarding-orchestrator` / `proxy-error` / `scoop-context` tests (typed payload aliases, missing imports, `readFile` encoding option, `find()`-result optional chaining) and the shared helpers (`fsa-test-helpers` circular alias → interface, stdin brands, implicit anys)
- **`tsconfig.json`**: `packages/webapp/tests/**/*.ts` included; the 116 remaining files are listed in `exclude` as a **burn-down** with a banner comment (fix a file → delete its line; new tests must typecheck). Five of those are clean provider tests that transitively pull the still-unchecked `providers/adobe.ts` + `xai-grok.ts` — the providers dir is the next slice

## Verification

- `tsc --noEmit -p tsconfig.json`: 683 → **0** errors in the included set
- Full webapp vitest suite: **8,404 passed**, zero failures (test behavior unchanged — mechanical type fixes only)
- lint (incl. knip production scan) green

## Remaining (next slices of #1337)

1. `providers/` into the typecheck program + `adobe.ts` (13 errors) / `xai-grok.ts` fixes — production code, deserves its own review
2. The 111-file burn-down tail (~350 errors: realm-runner, orchestrator, ui/wc, sudo, ipk buckets)


catalan-adobe:
Nit addressed in 3e87297c2 — the three already-fixed helper files are off the burn-down list (they're typechecked transitively via included importers, as you mutation-verified). List is now strictly 'still-broken': 113 entries.

PR #1362, opened 2026-07-07
fix(webapp): typecheck the providers dir — adobe.ts + xai-grok.ts fixes

catalan-adobe:
Part of #1294, second slice of #1337. **Stacked on #1359** (base branch `fix/1337-webapp-tests-typecheck`) — will be rebased onto `main` once that merges.

## Problem

`packages/webapp/providers/` was never typechecked: it sits outside the `src/**` include and is only reachable via `import.meta.glob`, which tsc doesn't follow. `adobe.ts` — the provider carrying all Adobe LLM traffic — had 13 latent type errors.

## Changes

- **`tsconfig.json`**: `packages/webapp/providers/**/*.ts` is now a first-class include root
- **`adobe.ts`** (13 errors): the header helpers (`withSliccVersionHeader`, `ensureSessionIdHeader`) constrained `headers` to `Record<string, string>` but pi-ai's real `StreamOptions.headers` is `ProviderHeaders = Record<string, string | null>` — widened to match; the raw `/v1/models` payload is typed once at the parse edge (`RawProxyModel`) instead of blind `Record<string, unknown>` field reads (same trust level as before); the anthropic stream branch gets the same cross-vocabulary options cast its openai sibling already had
- **`xai-grok.ts`** (2 errors): `withGrokConvHeader` widened to `ProviderHeaders`
- **`adobe-provider.test.ts`** (14 errors): the inlined mirror helpers now use a concrete `TestStreamOptions` bag — the repo's TS version applies excess-property checks against generic constraints for fresh object literals, so the generic mirrors never typechecked
- **Burn-down**: 6 more entries deleted (5 clean provider tests that only transitively pulled the broken providers + `adobe-provider.test.ts`); **109 remain**

## Verification

- `tsc --noEmit -p tsconfig.json` green (providers now in the program)
- All 406 provider tests pass; webapp build green; lint (incl. knip) green
- Zero runtime change — type-level widenings and one cast that mirrors the existing sibling-branch pattern


catalan-adobe:
Superseded by #1367 — GitHub closed this PR when the stacked base branch (`fix/1337-webapp-tests-typecheck`, merged as #1359) was deleted before the retarget landed. Same branch, rebased onto main, identical diff.

PR #1398, opened 2026-07-08
Fix async leader rollback in performTrayLeave

### LINKED ISSUE #1387: P1-2: performTrayLeave: roll back persisted leader config when async leader.start() fails
Part of #1294 (P1 tier). See the audit §4 last bullet: "`performTrayLeave`'s documented storage rollback only catches synchronous throws."

## Problem

When the user switches trays (leave + restart leader on a new worker), `performTrayLeave` persists the new leader config **before the leader has actually connected**. The production starter kicks the connection off fire-and-forget, so an async failure (worker down, auth rejected, WebRTC init failure) leaves `slicc.trayWorkerBaseUrl` pointing at a dead worker — the next page reload tries to revive a leader that can never start. This is one instance of the repo's largest bug class #4, "stale persisted state hijacking boot" (29 commits).

## Preflight

```bash
rg -n "void leader" packages/webapp/src/ui/page-leader-tray.ts       # fire-and-forget start
rg -n "startLeader" packages/webapp/src/ui/tray-leave-runtime.ts     # sync-only rollback
rg -n "startLeader:" packages/webapp/src/ui/wc/wc-tray.ts            # the dep wiring (~line 417)
```

## Current state (verified)

- `performTrayLeave` (`packages/webapp/src/ui/tray-leave-runtime.ts:90`): the leader-restart branch calls `deps.startLeader(url)` **synchronously**, writes `TRAY_WORKER_STORAGE_KEY` after it returns, and rolls back to dormant only when `startLeader` **throws synchronously**. Its own doc comment (file header, lines 8-13) claims failed startup leaves the runtime fully dormant — currently only true for sync throws.
- Production starter `startPageLeaderTray` (`packages/webapp/src/ui/page-leader-tray.ts:434`) returns its handle synchronously, then runs `void leader.start().then(updateUrlBar).catch(log.error)` (~lines 464-478). Async failures never reach `performTrayLeave`.
- `PageLeaderTrayHandle` (`page-leader-tray.ts:137`) exposes `stop()`, `reset()`, and the managers — **no readiness signal**.
- Dep wiring: `wc-tray.ts:417` (`startLeader: (workerBaseUrl) => startPageLeaderTray(leaderOptions(workerBaseUrl))`). The direct boot paths at `wc-tray.ts:301,314` are boot-time (not leave/restart) and are out of scope.

## Task

1. Add `ready: Promise<LeaderTraySession>` to `PageLeaderTrayHandle`, fulfilled from the **existing** `leader.start()` chain in `startPageLeaderTray` (do NOT call `start()` twice). Keep the existing `.catch(log.error)` so boot-path consumers that ignore `ready` see unchanged behavior (no unhandled rejection: derive `ready` from the same promise but attach the log-catch on a separate branch).
2. In `performTrayLeave`'s leader-restart branch (`tray-leave-runtime.ts`):
   - Widen `TrayLeaveDeps.startLeader`'s handle constraint so the returned handle carries `ready` (extend `TrayLeaveStoppable` or add a second generic bound — keep it minimal).
   - `await handle.ready` (bounded: race it with a timeout — pick the same order of magnitude the leader's own connect timeout uses; investigate `LeaderTrayManager.start` in `packages/webapp/src/scoops/tray-leader.ts` and reuse its bound rather than inventing one; if it already rejects on timeout, no extra race is needed — verify and state which in the PR).
   - Only write `TRAY_WORKER_STORAGE_KEY` after `ready` resolves.
   - On rejection: `handle.stop()` (wrapped in try/catch like the existing stop calls), `deps.setLeader(null)` + `deps.clearLeaderHooks()`, remove both storage keys, log via the existing `log.error` pattern, rethrow (preserving the documented "stop succeeded, restart failed" contract).
3. Update the file-header comment in `tray-leave-runtime.ts` to describe the now-true invariant.
4. Check the other `performTrayLeave` call sites (`ui/boot/setup-standalone-panel-rpc.ts`, `shell/supplemental-commands/host-command.ts`, `ui/wc/wc-nav.ts` — grep) compile against the widened dep type.

## Tests (files exist)

- `packages/webapp/tests/ui/tray-leave-runtime.test.ts`: add — async `ready` rejection → both keys removed, handle stopped, hooks cleared, error rethrown; async success → key written only after resolution (assert ordering, e.g. storage spy vs a deferred promise); sync-throw behavior unchanged.
- `packages/webapp/tests/ui/page-leader-tray.test.ts`: `ready` resolves with the session on success and rejects on start failure, while the legacy log-on-error behavior still fires once.

## Acceptance criteria

- A leader restart against an unreachable worker leaves storage fully dormant (both keys absent) and surfaces an error to the caller — demonstrated by a test.
- No double `leader.start()` call (assert via spy).
- `npm run lint`, `npm run typecheck`, targeted vitest, `npm run test:coverage:webapp` all green.

## Out of scope

- The boot-time leader starts in `wc-tray.ts:301,314` (they have their own recovery semantics).
- Cross-tab election (P1 sibling issue).

## Repo guardrails (apply to every step)

- Branch off fresh `origin/main`; one PR for this issue; PR body `Fixes #<this issue>`. If you intentionally deliver only a slice, say "first slice of #<this>" and file a scoped follow-up issue instead of leaving silent remainder.
- Merge queue enforces **linear history** — `git rebase origin/main`, never merge main into the branch.
- Before pushing, run in order: `npm run lint` (the most common CI failure), `npm run typecheck` (nine tsc targets — **test files are typechecked too**), targeted `npx vitest run <paths>`, and the touched package's coverage gate `npm run test:coverage:<pkg>` (floors live in `coverage-thresholds.json`; never lower them — add tests instead).
- `knip --production` runs inside lint: a new `src/**` module reachable only from tests fails it; wire it for real or add a `knip.json` ignore entry with justification.
- Tests live in `packages/*/tests/` mirroring `src/`. Bug fixes ship a regression test.
- Root `CLAUDE.md` has a 30,000-char cap enforced by `lint:docs`.
- If you touch any tray sync wire message shape: regenerate the golden corpus (`npx tsx packages/dev-tools/tools/generate-tray-sync-corpus.ts`) and follow `packages/ios-app/CLAUDE.md` "Protocol Mirror Invariant". Not expected for this issue unless stated.
- File:line references below were verified at `main` = 0073ad0c1 (2026-07-08). Run the **Preflight** greps first; if a claim no longer holds, adapt to current reality and note the discrepancy in the PR body rather than forcing the stale instruction.

catalan-adobe:
Fixes #1387

## Problem

When `performTrayLeave` switches trays, it calls `startLeader(url)` which returns a handle synchronously while `leader.start()` runs fire-and-forget. If the async connection fails (worker down, auth rejected, WebRTC init failure), storage is already persisted pointing at the dead worker — the next page reload tries to revive a leader that can never start.

### 1. Add `ready` promise to `PageLeaderTrayHandle`

Exposed on the handle, derived from the same `leader.start()` call (no double invocation). A separate fire-and-forget branch keeps the existing `.catch(log.error)` so boot-path consumers that ignore `ready` see unchanged behavior with no unhandled rejections.

### 2. Await `ready` before persisting storage

`performTrayLeave`'s leader-restart branch now:
- Installs the handle immediately (`setLeader` + `wireLeaderHooks`) so the leader can function during connection
- Awaits `handle.ready` — the leader's own connect timeout (10s, `LEADER_TRAY_CONNECT_TIMEOUT_MS`) bounds it; no extra race needed
- Only writes `TRAY_WORKER_STORAGE_KEY` after `ready` resolves
- On rejection: stops the handle, nulls the leader, clears hooks, removes storage keys, logs, and rethrows

### 3. Type changes

Added `TrayLeaveReadyHandle extends TrayLeaveStoppable` with a `ready: Promise<unknown>` field. Updated `TrayLeaveDeps` generic constraint. All existing call sites (`wc-tray.ts`, `setup-standalone-panel-rpc.ts`, `host-command.ts`) compile against the widened type since `PageLeaderTrayHandle` now satisfies `TrayLeaveReadyHandle`.

## Tests

- **`tray-leave-runtime.test.ts`**: async `ready` rejection → both keys removed, handle stopped, hooks cleared, error rethrown; async success → storage written only after resolution (ordering via deferred promise); stop-throw during rollback logs but surfaces the original error
- **`page-leader-tray.test.ts`**: `ready` resolves with session on success; rejects on start failure while the log-on-error branch still fires

## Verification

- `npm run lint` ✅
- `npm run typecheck` ✅ (zero errors, all 9 tsc targets)
- Targeted vitest (23 tests passing) ✅
- `npm run test:coverage:webapp` gate ✅ (exit 0)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -789,6 +789,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -806,6 +809,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -823,6 +829,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -840,6 +849,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -1956,7 +1968,7 @@
         "typebox": "1.1.38"
       },
       "bin": {
-        "pi-ai": "dist/cli.js"
+        "pi-ai": "./dist/cli.js"
       },
       "engines": {
         "node": ">=22.19.0"
@@ -2072,6 +2084,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2087,6 +2102,9 @@
       "integrity": "sha512-AGuJdgKsmJdm4Pych7kv3sqe591ERRaAHW3xjLooiFzn8J+PxUyof++7YZrB5Y5tpnTO+K18Og3taj2NpluCRQ==",
       "cpu": [
         "arm64"
+      ],
+      "libc": [
+        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -2104,6 +2122,9 @@
       "cpu": [
         "riscv64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2120,6 +2141,9 @@
       "cpu": [
         "x64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -2135,6 +2159,9 @@
       "integrity": "sha512-/DHn+1DrfL6oRaPPWXaOKvonFFrni666fxd+zFqiQEfvBH0tsHVWjq9iqBk0oDp0qaPA72lIMy5BptxISBEhZQ==",
       "cpu": [
         "x64"
+      ],
+      "libc": [
+        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -6045,6 +6072,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6062,6 +6092,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6079,6 +6112,9 @@
         "ppc64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6096,6 +6132,9 @@
         "s390x"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6113,6 +6152,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6130,6 +6172,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/package-lock.json
+++ b/package-lock.json
@@ -1968,7 +1968,7 @@
         "typebox": "1.1.38"
       },
       "bin": {
-        "pi-ai": "./dist/cli.js"
+        "pi-ai": "dist/cli.js"
       },
       "engines": {
         "node": ">=22.19.0"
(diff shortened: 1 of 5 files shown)


catalan-adobe:
Review follow-ups pushed:

| Finding | Commit |
|---|---|
| Unrelated `package-lock.json` churn (libc-field removals + pi-ai bin normalization — local npm-version artifact; no deps change on this branch) | `b3face416` — restored byte-identical to `origin/main`. Note: this environment *regenerates* the churn on some npm operations, which is how it slipped in — keep an eye on it before future pushes from this worktree |
| Success-path reentrancy hole: the `stillOurs` guard protected only the rollback. 'Switch to A' → 'Stop' inside the connect window, with A's connect racing the stop, resurrected the stopped leader's storage on next reload | `bfb06c5fa` — post-await supersession guard skips the storage write + logs at ERROR; **mutation-verified** (disabling the guard fails the new resolve-after-takeover test) |

The new branch tipped `performTrayLeave` over biome's cognitive-complexity limit (26 > 25), so the async-failure rollback moved to a `rollbackAfterReadyFailure` helper (same commit, behavior unchanged — all 25 targeted tests pass).

Re-verified locally: `npm run lint` exit 0, typecheck clean, tray-leave-runtime + page-leader-tray suites green.

Issue #1391, opened 2026-07-08
P1-6: UI self-skew recovery: vite:preloadError reload + extension pinned-tab refresh


catalan-adobe:
Part of #1294 (P1 tier — version-skew management, audit §5). Directly addresses #1330 (cloud cone crashes on stale assets after deploy).

## Problem

The hosted UI deploys continuously; Cloudflare Static Assets then drop the previous build's hashed chunks. Every long-lived page — the extension's pinned leader tab, the cloud sandbox's Chromium, any tab left open across a deploy — 404s on its next lazy `import()`: `Failed to fetch dynamically imported module: https://www.sliccy.ai/assets/<chunk>-<oldhash>.js`. Today there is **no recovery**: no `vite:preloadError` listener, no build-id probe, no extension-update tab refresh (all verified absent: `rg -rn "preloadError|buildId|build-id" packages/webapp/src` → zero). In the cloud float this kills the cone mid-session (#1330).

## Preflight

```bash
rg -rn "preloadError" packages/webapp/src packages/chrome-extension/src   # expect zero
rg -n "onUpdateAvailable" packages/chrome-extension/src                    # expect zero
gh issue view 1330 --json state                                            # confirm still OPEN
```

### (a) Page-realm recovery: `vite:preloadError` → guarded one-shot reload

- In early page boot (`packages/webapp/src/ui/main.ts` — before any lazy import can fire; find the earliest common code path shared by all floats), register `window.addEventListener('vite:preloadError', handler)`.
- Handler: set a `sessionStorage` loop-guard (e.g. `slicc.preload-reload-at` with a timestamp; skip reload if the last one was <60s ago — a still-broken deploy must not reload-loop), call `event.preventDefault()` (suppresses Vite's unhandled rejection), log at ERROR with the failing URL, then `location.reload()`.
- Preserve the full current URL (reload does; do not redirect).
- Safety check before building: confirm a mid-session reload is recoverable — the leader restores conversations from IndexedDB (`SessionStore`) and tray role from storage on boot. Read `ui/main.ts` boot + session-restore paths and state your conclusion in the PR body. If a reload would lose in-flight agent state, gate the reload on agent-idle (investigate the orchestrator's busy signal) and document the tradeoff.

### (b) Kernel-worker / cloud-cone path (the actual #1330 crash)

[...]

Issue #1330, opened 2026-07-06
Cloud cone crashes on stale assets after deploy


QuentinVecchio:
## Problem

When a new build is deployed while a cloud cone session is running, the cone crashes with:

```
Something went wrong
Scoop "Cone" failed after 3 attempts: Failed to fetch dynamically imported module: https://www.sliccy.ai/assets/anthropic-BOEkIcb-.js
```

The Vite-built chunk hash changed on deploy, but the running cloud cone still holds the old module graph and tries to lazy-import a chunk that no longer exists on the CDN. The 3-retry logic fails and the session is unrecoverable — the only fix is to kill and restart the cone.

## Impact

- Cloud cone sessions are killed by any deploy
- No graceful recovery — user must manually restart
- Local/extension floats can hard-refresh; cloud cones cannot

## Possible solutions

1. **Pin asset versions per session** — cloud-core stores the asset manifest hash at boot and serves from a versioned path (or CDN with immutable caching)
2. **Detect stale imports and auto-restart** — catch the dynamic import failure, notify the user, and auto-restart the cone against fresh assets
3. **Keep old chunks alive for a grace period** — deploy retains previous chunk hashes for N hours so in-flight sessions don't break
4. **Preload all lazy chunks at boot** — eliminate lazy imports for provider modules in cloud mode so the full module graph is resolved before any deploy can invalidate it

## Repro

1. Start a cloud cone session
2. Deploy a new build (chunk hashes change)
3. Trigger the cone to lazy-import a provider (e.g., Anthropic)
4. Observe the failure

karlpauls:
Verified the root cause and found it's **broader than cloud cones** — the same stale-chunk failure hits **any long-lived sliccy.ai UI tab after a deploy**, including the **extension's pinned leader tab** and its side-panel follower, plus any open standalone tab. The "local/extension floats can hard-refresh" note is optimistic: the extension leader tab is long-lived (spans deploys), and a side-panel-only user has no obvious "reload" and won't know to hard-refresh.

**Verified detail that changes the error mode + the fix:** the worker's `assets` binding uses `not_found_handling: "single-page-application"`, so a request for an old (removed) hashed chunk does **not** 404 — it returns **`index.html` as `200 text/html`**. Confirmed live against the worker:

```
GET /assets/browser-DEADBEEF404.js   → HTTP 200, content-type: text/html   (index.html body)
[...]

PR #1406, opened 2026-07-08
feat(chrome-extension): refresh pinned leader tab on extension update (#1391c)

catalan-adobe:
Fixes the extension part of #1391 (part c). Parts (a) and (b) were merged via #1364.

## Summary

Adds `chrome.runtime.onUpdateAvailable` handler in the service worker to refresh the pinned leader tab when the extension updates, ensuring the leader picks up the new hosted UI.

### Behavior

- Handler: reloads the pinned leader tab (so it picks up new hosted UI), then reloads the SW (`chrome.runtime.reload()`)
- Session-storage guard (60s) prevents rapid re-reloads
- Guard is stamped **only after** the leader tab reload succeeds — if the tab vanished, the guard stays clear so Chrome's next `onUpdateAvailable` retry can attempt the reload
- Missing leader tabs are tolerated; SW reload always proceeds even if no leader tab found
- Added `onUpdateAvailable` and `reload()` to `chrome.d.ts` types

## Tests

- 8 tests in `service-worker-update.test.ts`
- All existing test files updated with new `chrome.runtime` mock properties (549 tests pass)

## Verification

- `npm run typecheck` (9 targets) ✅
- `npm run test:coverage:chrome-extension` ✅
- `npm run lint` ✅

### DIFF
--- a/packages/chrome-extension/src/chrome.d.ts
+++ b/packages/chrome-extension/src/chrome.d.ts
@@ -151,6 +151,15 @@ interface ChromeAPI {
     onConnectExternal: {
       addListener(callback: (port: ChromeRuntimePort) => void): void;
     };
+    /**
+     * Fired when an update is available. Calling `chrome.runtime.reload()`
+     * from the listener applies it immediately.
+     */
+    onUpdateAvailable: {
+      addListener(callback: (details: { version: string }) => void): void;
+    };
+    /** Reload the extension (applies a pending update). */
+    reload(): void;
   };
   notifications: {
     create(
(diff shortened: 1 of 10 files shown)


catalan-adobe:
## Overlap with #1364

Parts (a) and (b) of this PR overlap significantly with #1364, which addresses the same #1330 crash with a more thorough approach:

- **instanceId-scoped** BroadcastChannel (prevents cross-tab interference)
- Boot-time `try/catch` around `boot()` in `kernel-worker.ts`
- `Worker error` event handler in `spawn.ts`
- UI toast integration in `wc-chat-controller.ts`

**Directly conflicting files**: `scoop-context.ts`, `setup-preload-error-reload.ts`, `main.ts`, and their tests.

**Plan**: Once #1364 merges, I'll rebase this PR down to **part (c) only** — the `chrome.runtime.onUpdateAvailable` extension leader-tab refresh, which is unique to this PR and not covered by #1364. That part has no file overlap with #1364.

PR #1454, opened 2026-07-09
feat(worker): retain hashed assets in R2 so tabs survive deploys (#1330) [blocked on ops]

karlpauls:
Root-cause follow-up to #1330 / PR #1364. Instead of the reactive page **reload** (which can't save an in-flight cone cycle and risks mixed-release UI), keep every content-hashed `/assets/*` chunk in a per-env **R2 bucket** so a long-lived tab keeps fetching its own build's chunks after a deploy — no crash, no reload.

## ⚠️ Blocked on ops — CI is expected to be RED until the buckets exist

This PR is **safe to open and review while red**, and **cannot merge until ops is done** (which is exactly the gate we want):

- The only red is the hard-fail **"Archive assets to R2 (staging)"** step, which needs the buckets + token. It runs *before* the staging deploy, so on failure the job stops there — **no deploy runs, nothing in staging/prod changes.**
- The prod release only runs on merge; a red PR can't merge, so the `r2_buckets` binding can't reach the prod release deploy until the buckets exist.
- Once ops completes the steps below and CI re-runs: the upload succeeds → staging deploys → CI goes green → mergeable, and the eventual prod release deploys cleanly.

## Ops steps (require Cloudflare admin on the AEM Demo account)

1. Create both buckets: `wrangler r2 bucket create slicc-asset-archive` and `… -staging`.
2. Apply the 14-day lifecycle to each: `wrangler r2 bucket lifecycle add <bucket> retention-14d --expire-days 14`.
3. Grant the deploy `CLOUDFLARE_API_TOKEN` **R2 Object Read & Write** on **both** buckets (CI uses staging; the release uses prod).

Full runbook is in `packages/cloudflare-worker/CLAUDE.md` (Ops Runbook section).

## What it does

- **Serving** (`serveAssetWithArchiveFallback`, `src/index.ts`): ASSETS-first; on a hashed-`/assets/*` miss (ASSETS returns the SPA shell), serve the archived bytes from R2 with an immutable `Cache-Control`; on any R2 miss/error, fall back to the shell (the shipped #1364 reload) — **never 500s, safe even before the bucket exists**. Conditional (304/412) and Range (206) are intentionally not implemented (immutable content-hashed chunks); full `200`/bodyless `HEAD`.
- **Upload** (`scripts/upload-*.mjs`): re-puts the full current asset set to the bucket before every deploy (`--remote`, hash-invariant asserted, bounded concurrency + retries), wired into all four deploy paths (publish-worker.sh, worker.yml, ci.yml, worker-staging.yml).
- **GC**: 14-day age-based lifecycle (option A). Fully covers the observed vendor-chunk crash; build-unique lazy chunks past the window fall back to the reload. Option B (manifest-touch, full coverage) is documented as an additive future upgrade — no re-spec needed.
- Tests: worker unit (present/miss/HEAD/cache/never-500/gate) + a **staging-only** deployed archive smoke; **384 worker tests pass**, coverage ≥ floor.

Built spec → plan → codex/cursor review loops → subagent-driven implementation; final whole-branch review "Ready to merge: Yes". Rebased onto current `main`. (Spec/plan commits under `docs/superpowers/` are kept per convention — the release scrubs them.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,6 +542,19 @@ jobs:
           path: coverage/
           if-no-files-found: ignore
 
+      # #1330 retention: archive this build's content-hashed assets to the
+      # STAGING R2 bucket BEFORE any deploy attempt so long-lived tabs can
+      # recover gone lazy chunks. Hard-fail gate (NOT continue-on-error) — a
+      # build must never go live unarchived. Shares the deploy step's non-fork
+      # guard so it never runs without Cloudflare secrets on fork PRs.
+      - name: Archive assets to R2 (staging)
+        if: |
+          (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false) ||
+          github.event_name == 'push'
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ vars.CLOUDFLARE_ACCOUNT_ID }}
+        run: node packages/cloudflare-worker/scripts/upload-assets-to-r2.mjs slicc-asset-archive-staging --dir dist/ui/assets
       # Deploy to staging on PRs (not forks) and pushes to main.
       #
       # The staging Worker (slicc-tray-hub-staging) has Worker Versions enabled.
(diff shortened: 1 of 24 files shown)


catalan-adobe:
Ops prerequisites completed:
- ✅ `slicc-asset-archive` bucket created
- ✅ `slicc-asset-archive-staging` bucket created  
- ✅ 14-day lifecycle rule applied to both
- ✅ `CLOUDFLARE_API_TOKEN` updated with R2 permissions

Re-running CI.

PR #1472, opened 2026-07-10
fix(cloudflare-worker): track known-good pointer in DMG test

trieloff:
## Problem

The `GET /download/slicc.dmg` worker tests hardcoded the release version `5.37.0` (as `OLDER_DMG_URL` and asset-name literals). A release-bot `[skip ci]` commit bumped the bundled pointer `packages/cloudflare-worker/src/known-good-macos.json` from `5.37.0` to `5.54.2`.

Because the bump landed with `[skip ci]`, it never ran CI on its own — but the merge-ref CI (each open PR's branch merged into `main`) then evaluated the worker against the new pointer, and the hardcoded `5.37.0` expectations no longer matched the known-good redirect. This broke the DMG suite for **every open PR**, including #1470.

## Fix

Derive the expected version from the JSON pointer instead of hardcoding it:

- Import `../src/known-good-macos.json` and read `knownGoodMacos.version`.
- Build `KNOWN_GOOD_DMG_URL` / `KNOWN_GOOD_DMG_ASSET` from that version.
- Express the pagination-floor fixtures relative to the pointer (`newerThanPointer` / `olderThanPointer`) so the scenarios stay meaningful regardless of the current pointer value.

This makes the tests drift-proof: future release-bot pointer bumps can no longer break the DMG suite.

## Verification

15 tests green:

    npx vitest run --project cloudflare-worker -t "GET /download/slicc.dmg"

### DIFF
--- a/packages/cloudflare-worker/tests/index.test.ts
+++ b/packages/cloudflare-worker/tests/index.test.ts
@@ -9,6 +9,7 @@ import worker, {
   parseAllowedCapabilityOrigins,
   resolveCherryFrameAncestors,
 } from '../src/index.js';
+import knownGoodMacos from '../src/known-good-macos.json';
 import { SessionTrayDurableObject } from '../src/session-tray.js';
 import {
   type CreateTrayRequest,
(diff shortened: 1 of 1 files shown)


trieloff:
dupe #1473

PR #1480, opened 2026-07-10
Reduce and harden Cloudflare worker deploys (gate + retry)

trieloff:
## Why

Semantic-release has been creating git tags without the corresponding GitHub Release. Root cause: `publish-worker.sh` runs on **every** release via `@semantic-release/exec`, and its `wrangler deploy` intermittently fails on the `POST /zones/{zone}/workers/routes` trigger call ("Some triggers failed to deploy"). Under `set -euo pipefail` that aborts the release **after** the tag was pushed but **before** the GitHub Release is created.

This PR reduces how often the worker deploys and makes the deploys that do run resilient to that transient failure.

## What

**1. Gate the worker deploy on relevant changes**
- New `--gate=worker` / `decideWorkerGating` in `release-native.mjs`, reusing the existing `getChangedFiles` / `matchesAnyPrefix` / `isFirstRelease` helpers.
- `publish-worker.sh` runs the gate first; if nothing worker/UI-relevant changed since the last release tag it **skips** both `wrangler deploy`s and the smoke test.
- On skip it **still refreshes the 14-day R2 asset archive** so GC retention is preserved.
- Prefix set: `cloudflare-worker`, `webapp`, `webcomponents`, `spoon`, `cherry`, `shared-ts`, `cloud-core`, `e2b-template`. First release / empty last tag always deploys.
- `.releaserc.json` wires `SLICC_LAST_RELEASE_TAG` into `publish:worker`.

**2. Retry the wrangler deploy with backoff**
- New `deploy_with_retry` helper wraps both the hub and preview deploys: 6 attempts, 15s backoff, `WRANGLER_LOG=debug`.
- Re-running after a partial success (script deployed, routes failed) is safe — wrangler reconciles.
- Final failure surfaces the wrangler debug log and exits non-zero.

## Not in this PR (follow-up)

Reordering the semantic-release plugins so `@semantic-release/github` runs before `@semantic-release/exec` — that is what would *guarantee* a Release exists even if a deploy ultimately fails. The gate + retry reduce the chance of hitting the flaky deploy; the reorder is the belt-and-suspenders fix and is tracked separately.

## Verification

- `release-native` Vitest: 48/48 pass (includes new gate cases)
- `bash -n packages/cloudflare-worker/scripts/publish-worker.sh`
- `npm run lint`
- `npm run typecheck`

Docs updated in `packages/cloudflare-worker/CLAUDE.md`.

---

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -15,7 +15,7 @@
       "@semantic-release/exec",
       {
         "prepareCmd": "SLICC_RELEASED_AT=\"$(date -u +'%Y-%m-%dT%H:%M:%SZ')\" npm run build && npm run package:release && node packages/dev-tools/tools/release-native.mjs --last='${lastRelease.gitTag}' --next='${nextRelease.version}'",
-        "publishCmd": "npm run publish:worker && node packages/dev-tools/tools/release-native.mjs --gate=chrome --last='${lastRelease.gitTag}'"
+        "publishCmd": "SLICC_LAST_RELEASE_TAG='${lastRelease.gitTag}' npm run publish:worker && node packages/dev-tools/tools/release-native.mjs --gate=chrome --last='${lastRelease.gitTag}'"
       }
     ],
     [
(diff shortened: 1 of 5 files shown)


trieloff:
Thanks @chatgpt-codex-connector — confirmed and correct. The worker gate was a no-op because the `chore(release):` version bump lands on `HEAD` (via `@semantic-release/git` in *prepare*) before the gate runs in *publish*, so `git diff <lastTag> HEAD` always saw `package.json` change.

Since this PR was already merged, the fix ships as a standalone follow-up: **#1482**. It adds a `resolveDiffRef` helper that diffs against `HEAD^` when the HEAD subject is `chore(release):`, so the auto version bump is ignored while genuine dependency changes still trigger a deploy. 54/54 gate tests (6 new regression cases).

PR #1519, opened 2026-07-15
feat(realm): robust synchronous fs in browser realms via SW sync-XHR bridge (no SAB/COOP-COEP)

karlpauls:
## What this is

Makes **synchronous filesystem APIs** (`readFileSync` / `writeFileSync` / `existsSync` / `statSync` / `readdirSync` / `rmSync` / …) in the kernel-worker **JS realm** robust and effectively **unbounded**, without `SharedArrayBuffer` or COOP/COEP.

Today those APIs are backed only by a bounded (500-file / 1 MB-per-file / 10 MB-total) point-in-time snapshot: anything outside it silently reads as empty/absent. This PR keeps that snapshot as a zero-cost fast path and adds a **synchronous-XHR bridge** for the misses, over-caps, and every write.

## How it works (Approach D)

`readFileSync`/`writeFileSync` first try the in-memory snapshot. On a cache miss (`ENOENT`), over-cap (`ENOSYNC`), or any write, the realm issues a **synchronous `XMLHttpRequest`** to `/__slicc/fs-sync/<path>` (permitted off the main thread — realm code runs in a `DedicatedWorker`):

```
realm worker ──sync XHR──▶ controlling SW (llm-proxy-sw)
   (blocks this           │  own fetch listener for /__slicc/fs-sync/*
    worker thread)         ▼
                    per-session NONCE'd BroadcastChannel
                           │
                           ▼
              kernel-worker sync-fs responder ──▶ CALLING REALM's own ctx.fs
                                                   (RestrictedFS + sudo-fs)
```

The responder runs in the kernel worker (co-located with the token registry and every realm's `ctx.fs`), so there is **no page hop** and **no deadlock** — the blocked realm worker is a different thread from the VFS owner. Reads/writes resolve to the **calling realm's own `ctx.fs`**, addressed by an unguessable **per-realm capability token** (minted in `attachRealmHost`, revoked on dispose), so ACL + sudo + POSIX errno are all preserved exactly as the async path.

Gated on a confirmed controlling SW (`syncFsBridgeEnabled`, threaded page→worker); off → the bounded snapshot (and the in-process test factory, which would deadlock on a sync XHR, stays snapshot-only).

## Security model (load-bearing)

- **Per-session nonce'd channel** (`slicc-sync-fs-<uuid>`): the channel name is minted in `wc-live.ts`, delivered page→kernel via `postMessage` (**never** reaches a realm) and page→SW via `controller.postMessage`, with an MV3-eviction recovery handshake. A realm is a `worker` SW-client and can neither learn nor set the nonce, so it cannot join the channel to steal another realm's token or spoof responses.
- **Nonce-set gate** (`maySetSyncFsNonce`): only a **top-level/auxiliary window** client may set the SW's channel nonce — rejects `worker` clients *and* `nested` window clients (a same-origin `allow-same-origin` srcdoc sprinkle/dip iframe is a `window` client that would otherwise repoint the module-global nonce and laterally harvest/spoof other realms).
- **SPA-fallback masquerade guard**: every sync-fs response carries an `x-slicc-fs` marker; a marker-less 2xx (a stale/absent SW letting the request hit the network → `index.html`) is rejected as `EIO`, never read as file bytes. Non-2xx errno headers are trusted only when the marker is present.
- Route stays a VFS read/write surface only; the raw XHR is CORS-confined to same-origin.

## Review history folded in

- **Deep review** closed a **CRITICAL sandbox escape** — the original fixed-name BroadcastChannel was joinable by any same-origin realm (token theft + response spoofing). Fixed via the nonce'd channel above.
- **Adversarial re-review** confirmed the escape is closed and flagged **Finding 1** (nested-frame window client repointing the nonce) → fixed via `maySetSyncFsNonce`.
- **Coherence:** read-your-deletes tombstones (Coh#1); `statSync().size` reports a truncated file's real size (Coh#2); the snapshot walk now separates a CONTENT budget (stop reading bytes at 500 files / 10 MB) from an ENTRY budget (`SYNC_FS_MAX_ENTRIES = 20000`, stop walking) so `existsSync`/`statSync` stay correct past the content cap while `readFileSync` bridges (Coh#3).

## Accepted limitations (documented, not bugs)

- A sync-write `EIO` is **at-least-once** — not a guarantee of non-commit (a timeout can fire after the responder already committed). Callers must re-read to confirm. (spec §11)
- Files beyond the 20 000-entry metadata cap report absent to `existsSync`/`statSync` (bounded residual of the snapshot-backed metadata model; `readFileSync` still bridges).

## Deferred (tracked separately)

- **#1522** — the same SW `message` listener sets `bridge-config` / `extension-delegate` from any client-with-id without the window gate (re-review Finding 2). LOW, pre-existing, bounded by the poster's own `source.id`; deliberately not bundled into a filesystem PR.

## Testing / verification

- ~110 sync-fs tests (token registry, dispatch + ACL-escape guard, responder + dedup + unowned-token silence, XHR bridge + encode/decode + masquerade guard, SW handler, cache incl. size/tombstones, shim bridge, acceptance gates, snapshot-walk budgets, `maySetSyncFsNonce` matrix).
- Browser-bundle + CLI + kernel-worker-guard typecheck exit 0; biome/lint:ci/deadcode clean; `test:coverage:webapp` above floors; webapp + extension builds; linear history.

**Manual browser smoke still pending** (real SW + DedicatedWorker: live sync XHR, sudo-under-sync-write reentrancy with no deadlock, cross-float) — the paths that cannot be node-tested. Requested to run after external review lands.

## Docs

Spec `docs/superpowers/specs/2026-07-15-sync-fs-browser-realm-design.md` + plan `docs/superpowers/plans/2026-07-15-sync-fs-browser-realm-plan.md` (branch-only, scrubbed from main by the planning-artifact cleanup). `packages/webapp/CLAUDE.md` "Synchronous fs bridge" describes the runtime.


karlpauls:
## ⚠️ Deep review: DO NOT MERGE — one CRITICAL sandbox escape (fix in progress)

A 5-agent adversarial review (ACL/security, fail-closed, coherence, concurrency, lifecycle) + a manual pass found a **critical sandbox escape introduced by this branch**, plus two correctness HIGHs. The per-realm capability *model* is correct and the plumbing/lifecycle is sound, but the **transport voids the model**.

### 🔴 CRITICAL — sandbox escape via the shared `slicc-sync-fs` BroadcastChannel
The SW↔kernel transport is a **named, origin-scoped `BroadcastChannel`** that realm workers can join, and realm user code has unrestricted `BroadcastChannel` access (the realm scrubs no ambient globals; `new AsyncFn(...names, code)` only shadows the bridge params).
- **Token theft**: a malicious scoop realm subscribes to `slicc-sync-fs`, harvests any other realm's capability token (incl. the **cone's full-VFS token**) off the channel, and reads/writes outside its sandbox — full escape + privilege escalation. The responder authorizes on token possession only, with no source binding.
- **Response spoofing** (no token needed): observe a victim's request `id` on the channel, post a forged `sync-fs-res` with attacker bytes → poison the victim's `readFileSync`.
- **Destructive ops**: the responder implements `rm -r`/`rename` though the SW only issues read/write; direct channel posting reaches them → cross-scope destruction.

**Fix (in progress):** move the SW↔kernel transport off the realm-joinable BroadcastChannel onto a **private channel** (SW→page→kernel over the existing private page↔worker RPC), + a responder op-allowlist + source binding, + defense-in-depth scrub of `BroadcastChannel` from realm globals.

### 🟠 HIGH — correctness
- **`writeFileSync` reported `EIO` can still commit**: the responder's `fs.writeFile` has no timeout; a sudo-gated sync write approved after the 30 s cap returns `EIO` to the realm but lands on the VFS. Fix: non-interactive sudo gate on the sync path (needs-approval → immediate `EACCES`).
- **read-after-delete returns stale bytes**: deletes are cache-only; `readFileSync` falls back to the live bridge on ENOENT with no tombstone → returns the deleted file's old content. Fix: tombstone.
- **responder non-idempotent**: a cold-start re-post under load runs the op twice (double sudo prompt / double write / stale clobber). Fix: id-dedup.

### 🟡 Pre-existing (separate follow-up)
A realm can `XHR /preview/<any-path>` and read the **global VFS** (unscoped `getReader()`), bypassing `RestrictedFS` — read-only, predates this PR, but a real scoop-sandbox read hole. Tracking separately.

Plus several LOW hardening items (`..` normalization, over-cap `statSync().size`, `decodeURIComponent` guard, timeout margin, errno validation, non-2xx marker).

**Status:** fixes being implemented on this branch; will re-run the adversarial reviewers + the live browser smoke before this is merge-ready. Holding merge until then.

karlpauls:
## Update — everything known now addressed (awaiting external review + live smoke)

Since the deep-review do-not-merge verdict, all known findings from that pass **and** the adversarial re-review are now fixed. Latest: `4c31e1861`.

### Adversarial re-review verdict (nonce'd-channel fix)
The re-review confirmed the **original sandbox escape is CLOSED** — a malicious scoop realm can neither learn nor set the channel nonce (nonce is transferred page→kernel via `postMessage`, never reaches a realm; the channel-rename primitive is authenticated). One residual was flagged:

- **Finding 1 (LOW/MED, PLAUSIBLE) — now FIXED.** The nonce gate accepted *any* same-origin `window` client, but a full-document sprinkle/dip renders as an `allow-same-origin` srcdoc iframe (also a `window` client, `frameType === 'nested'`) that could repoint the module-global nonce and *laterally* harvest/spoof **other** realms' sync-fs traffic. Tightened to require a non-nested browsing context (`top-level`/`auxiliary`) via the new testable `maySetSyncFsNonce` predicate. The sole publisher (`wc-live.ts`) always runs in the top-level leader tab.

### Coherence fixes
- **Coh#2** — `statSync().size` now reports a truncated file's **real** size (was 0). Snapshot walk records real byte size on every metadata placeholder.
- **Coh#3** — `existsSync`/`statSync` stay correct for files **past** the content budget. The walk now splits a CONTENT budget (stop reading bytes at 500 files / 10 MB) from an ENTRY budget (`SYNC_FS_MAX_ENTRIES = 20000`, stop walking). Over-budget files become metadata-only placeholders instead of vanishing; `readFileSync` bridges for them.

### Documented (accepted, not fixed)
- **FC#1** — a sync-write `EIO` is **at-least-once**, not a guarantee of non-commit (a timeout can fire after the responder committed). Documented in spec §11 + the XHR write site; callers must re-read to confirm.

### Known pre-existing, deliberately NOT bundled here
- **Re-review Finding 2 (LOW, pre-existing)** — the same SW `message` listener sets `bridge-config` / `extension-delegate` from any client-with-id without the `window` gate. Bounded (keyed by the poster's own `source.id`; a realm can already fetch arbitrary origins), not introduced by this PR, and touching the proxy-config path risks the LLM-proxy flow. Tracked for a separate change rather than widened into this sync-fs security fix.

### Verification
730 realm tests + 62 sw-config tests green; browser-bundle + kernel-worker-guard typecheck exit 0; biome clean.

**Still do-not-merge** until the external review lands and the live browser smoke passes.

karlpauls:
## Second full review pass → applied all findings (9 commits, `4dd63d33`..`6fb12738`)

Ran another comprehensive 5-agent review over the current (nonce-hardened) state and implemented **every** finding — Important, Suggestions, and the test gaps. Each commit is self-contained with its own tests and passed the local pre-push gate. Net: 23 files, +512/−141.

### Important
- **`faaece95d` — multi-tab nonce break.** The SW kept a *single* last-writer-wins `syncFsNonce`, so two same-origin leader tabs (each minting its own nonce) collided: the SW bound one tab's channel and the other tab's realms stalled the full 25s → `EIO` on every sync-fs op. It failed *closed* (never corruption/leak), but it's a real break in a topology we support (duplicate leader tab; `main.ts` routes every standalone/hosted tab to `mountWcUiLive`). The per-session nonce had also silently falsified the responder's own "both workers receive every request" comment. **Fix:** the SW now holds a channel **per live nonce** and fans every request out to all of them — the responder's existing token-ownership silence guarantees only the owning worker answers (mirrors the per-client `BridgeConfigCache` used for fetch-proxy). Comment corrected.
- **`608c7ea3e` — silent snapshot-init failure.** `initSyncFsCache` swallowed a `vfs.snapshot` rejection → empty cache → `existsSync`/`statSync` report **absent for files that exist** (metadata is cache-only in phase-1), with no breadcrumb. Now a bridge-enabled realm emits a `[sync-fs] snapshot failed …` line (test/no-bridge hosts stay quiet).
- **`4a16d2e35` — security-narrative comments.** The live channel was called `slicc-sync-fs` (the *fixed name the design rejects as insecure*) in three places incl. `CLAUDE.md`; the `maySetSyncFsNonce` doc claimed "top-level only" while the code accepts non-nested (top-level **or** auxiliary). Corrected + documented why `auxiliary` is safe. Behavior deliberately unchanged (no pre-smoke regression risk).

### Suggestions / hardening
- **`f5ca06631`** — `SyncFsResult` success arm is now a discriminated union on `kind` (`bytes`|`json`|`void`); `buildResponse` can no longer silently drop a phase-2 `json` result. Dropped the bespoke all-optional `SyncFsResEnvelope` and narrow against the shared wire union instead.
- **`b21d34ff8`** — channel discriminants (`sync-fs-req/ack/res`) promoted to `typeof` constants (a typo now fails at compile time, not runtime).
- **`166c70f91`** — branded `SyncFsToken` + `SyncFsNonce` (compile-time only; erases to string, so boundaries are transparent).
- **`85f8e7e7a`** — `DEDUPE_TTL` is now derived from a shared `SYNC_FS_REQUEST_TIMEOUT_MS` (≥ the SW re-post window), closing the 15s<25s margin where a very-late re-post could be re-dispatched (double write); flush failure is now a loud `[sync-fs] ERROR: … were NOT persisted` breadcrumb with counts.
- **`b095234bc` / `6fb127383`** — 6 new fail-closed tests (unmarked-non-2xx→EIO, malformed-encoding→EINVAL, malformed-`.code`→EIO, replay-after-settle, write-throw-no-commit, binary-write round-trip) + gate fixups.

**Two conscious calls:** kept the `maySetSyncFsNonce` `auxiliary` allowance (fixed the comment instead of tightening — zero regression risk before the smoke); and surfaced the flush failure as a loud ERROR line rather than overriding the exit code (a separate behavior decision for post-run durability).

**Gates:** `typecheck` (all 9), `lint:ci`, `deadcode`, webapp build, extension build, `test:coverage:webapp` — all green (9056 tests, coverage floors met).

---

## Proposed live-smoke checklist

Node/vitest genuinely can't exercise these — they need a real browser. Ordered by importance:

1. **Sync XHR actually blocks the realm worker without deadlock** — the whole premise (sync `XMLHttpRequest` + `responseType='arraybuffer'` + `timeout`, off the main thread, resolved by a *different* thread that owns the VFS). Confirm `readFileSync`/`writeFileSync` work inside a real `node -e` / `.jsh` / `workflow` realm.
2. **Real SW interception of `/__slicc/fs-sync/*`** — incl. the two IIFE-only fail-closed paths: malformed-path → 400 `EINVAL`, and no-nonce → 503 `EIO` + `requestSyncFsNonce`. And the marker guard's real trigger: a stale/absent SW letting the XHR hit the network → SPA `index.html` 200 must be rejected as `EIO` (never read as file bytes / write-success).
3. **Multi-tab fan-out (the new fix)** — open the app in **two same-origin tabs**, run sync-fs in both, confirm neither stalls to `EIO` (this is the exact scenario `faaece95d` fixes; single-tab won't exercise it).
4. **Nonce delivery + MV3 eviction self-heal** — kill/evict the SW, confirm it re-requests the nonce from the page(s) and the *next* op succeeds while the triggering one fails closed.
5. **Real client-type classification of the gate** — confirm a real **realm worker** (`type:'worker'`) and a real **srcdoc sprinkle/dip iframe** (`frameType:'nested'`) are both rejected by `maySetSyncFsNonce` and cannot repoint the channel / harvest a token.
6. **Sudo-under-sync-write reentrancy** — a sync write that trips a sudo prompt: confirm the pending modal doesn't deadlock the blocked realm, and a broker timeout → `EACCES` (not a hang).
7. **Extension-float parity** — run all of the above in the thin-extension hosted leader tab (is there a *controlling* SW at the sliccy.ai origin there?).
8. **`existsSync`/`statSync` past the content budget** — a workspace with >500 files / >10 MB: files past the content cap must still report present + correct `size` (Coh#2/#3); and the 20k ENTRY cap (the one path with no unit coverage) behaves.
9. **Large / binary transfer** — a multi-MB binary `readFileSync`/`writeFileSync` round-trips intact through the real XHR arraybuffer + BroadcastChannel structured clone.

karlpauls:
## ✅ Live browser smoke — sync-fs bridge (HEAD `6fb127383`)

Reviewed the second-pass rework (9 commits, `4dd63d33..6fb12738`) with **two independent adversarial passes** (mine + a fresh subagent reading the current files, not just the diff). Verdict: **the original sandbox escape stays CLOSED and every rework commit is sound** — the multi-tab fan-out only ever reaches legitimate nonce channels (still gated by the unchanged `maySetSyncFsNonce`), response correlation is by per-request UUID, the nonce never reaches a realm (`RealmInitMsg` carries only the token), and the `DEDUPE_TTL = timeout+5s` fix genuinely closes the old 15s<25s double-dispatch window. Only nit: `syncFsChannels` grows monotonically per SW lifetime (MV3 eviction resets it) — negligible.

Then drove a real browser smoke over CDP against a locally-built bundle (`dist/ui` with the reworked `llm-proxy-sw.js` as the **controlling** SW + a real kernel-worker `DedicatedWorker`), in **both** floats: standalone (`dev:standalone:fresh`, wrangler :8787 + node bridge) and the **extension hosted-leader tab** (`dev:extension:fresh`, CDP :9333). Realm code driven via the shell exec bridge (`node -e` / `.jsh`).

### Live-verified — all pass

| # | Scenario | Result |
|---|----------|--------|
| 1 | Sync `readFileSync`/`writeFileSync` round-trip through the real XHR→SW→responder, no deadlock | ✅ write+read coherent; page-side async `vfs cat` confirms the write hit the **real VFS** |
| 9 | Large **binary** round-trip (2 MB) through XHR arraybuffer + BroadcastChannel structured clone | ✅ `patternOk`, checksum exact (`267386880` = 8192×32640), `byte0=0`/`byteLast=255`, len 2097152 |
| — | Bridge **read-miss** (over-cap file → `ENOSYNC` → bridge GET) | ✅ 2 MB read back intact |
| 8 (Coh#3) | `existsSync`/`statSync` correct **past** the 500-file content budget | ✅ 520 files → **all 520 present**, `absentCount:0` (old bug dropped ~20); beyond-budget file `statSync().size=1` |
| Coh#2 | `statSync().size` on over-cap/truncated placeholders | ✅ 2 MB file reports `2097152`, not 0 |
| Coh#1 | Read-your-deletes: sync `unlinkSync` then `readFileSync` | ✅ `ENOENT` — not resurrected via the bridge |
| — | Errno fidelity: read nonexistent | ✅ `ENOENT` (not a bare `EIO`) |
| — | Dir ops: `mkdirSync -p` / `readdirSync` / `statSync().isDirectory` | ✅ `readdir: a.txt,sub`, `isdir: true` |
| — | **async↔sync coherence**: sync-write→async-read; async-write-**new-file**→sync-read | ✅ both coherent (`SYNC1`; `ASYNC2` via bridge miss) |
| 3 | **Multi-tab fan-out** (the headline rework) | ✅ opened a 2nd leader tab; the **original** tab's sync-fs still works after tab2 published its own nonce (under the old last-writer-wins bug tab1 would `EIO` forever) |
| 4 | MV3 SW-eviction self-heal | ✅ observed repeatedly: first op after SW eviction fails **closed** (`EIO` + `requestSyncFsNonce`), the page republishes, next op succeeds — exactly the documented behavior |
| 2 | SW own-listener fail-closed: malformed path | ✅ `%ZZ` → **400 `EINVAL`** + `x-slicc-fs:1` marker; no-nonce → 503 `EIO` path exercised via the cold-start above |
| 5 | Client-type gate under live fire (the cherry/sprinkle/dip escape) | ✅ a **nested `allow-same-origin` srcdoc iframe** posted a nonce to the SW and listened on that channel; after 5 real sync-fs writes carrying tokens, the attacker channel received **0** messages — the gate kept it out of the fan-out |
| 7 | **Extension hosted-leader-tab float** | ✅ controlling SW = `llm-proxy-sw.js` (activated); sync write+read round-trips; async read confirms real-VFS persistence — identical to standalone |
| — | Regression | ✅ normal shell (`ls`, async `fs.promises`), non-sync-fs compute realm — all unaffected (fast path) |

### Performance sanity — no regression on the hot path
- **Per sync op: ~3 ms** (60 sequential `writeFileSync` = 182 ms) — a healthy XHR→SW→BroadcastChannel→`ctx.fs` round-trip.
- Realm boot (snapshot walk): **0.35 s @ ~520 files → 0.71 s @ ~5 000 files**. This is the one thing to note: Coh#3 makes the boot walk emit metadata past the content budget up to the 20 000-entry cap (was 500), so on a **large** workspace every realm spawn pays a bounded, ~linear (~0.08 ms/file, ≤~1.6 s at the cap) snapshot cost it didn't before. It's the price of correct `existsSync`/`statSync`; typical workspaces (<1 000 files) stay <0.5 s. A lazy snapshot (build on first `*Sync`) or the phase-2 bridged `existsSync`/`statSync` would remove it — worth a follow-up, not a blocker.

### Covered by node tests, not live-drivable in a headless smoke (stated honestly)
- **Sudo-under-sync-write (item 6):** the shell exec bridge I drive is intentionally **ungated** (the human at the terminal is the approver), and triggering the *agent's* sudo-wrapped fs needs an LLM turn the fresh smoke profile has no provider for. The sudo **decision** path is locked by the acceptance gate with a **real `sudoFs` + deny broker**: sync write on deny → **`EACCES`** (broker consulted), non-gated write succeeds, out-of-sandbox read/write/`../` denied, revoked token → `EACCES`. The only unproven bit is the live page-*reentrancy* (no deadlock while a modal is up) — architecturally sound (realm worker, SW, and page are separate threads) but not live-driven.
- **SPA-fallback marker guard:** needs a broken/absent SW serving `index.html` for `/__slicc/fs-sync/*`; unit-tested (marker-less 2xx → `EIO` for both read and write).
- **Exec-boundary coherence** (flush-before / re-snapshot-after): the realm's `exec` is a sliccy module, not callable from an injected one-liner; covered by `realm-rpc` ("flushes-before then re-snapshots-after") + `sync-fs-coherence` unit tests. (The underlying `vfs.snapshot` RPC is exercised live on every op.)

### One characterized behavior (pre-existing, documented — not a regression)
Within a single realm, an `fs.promises` (async) overwrite of a file **already in the boot snapshot** is not visible to a later `readFileSync` (returns the stale cached bytes; the async read sees the new content). This is inherent to the point-in-time snapshot and matches the documented coherence model (§4: external-writer/exec-boundary coherence) — it heals at the next exec boundary and was equally true before this PR (new files and sync-written files are always coherent). Flagging only because the interaction was explicitly in scope.

**Bottom line:** the reworked bridge works end-to-end in a real browser across both floats, the sandbox escape is closed under live fire, coherence/deletes/binary/multi-tab all behave, and there's no hot-path perf regression. Remaining items are node-covered or the known large-workspace boot cost. IMO green for merge once a human eyeballs the boot-walk-cost note.

karlpauls:
## ✅ Sudo-under-sync-write — now verified LIVE (was node-covered only)

Followed up the earlier smoke with a real login (Adobe) so the agent's **sudo-gated** fs was in play. Seeded a deterministic rule `/etc/sudoers.d/smoke-guard` → `Write /workspace/guarded/**` (live-reloaded), then had the **cone** run `node -e 'fs.writeFileSync("/workspace/guarded/test.txt", …)'` — a real `writeFileSync` through the bridge → kernel responder → `ctx.fs` (sudo-fs) → rule match → approval prompt.

Findings:
- **The prompt fires from a sync write.** The full chain reaches the sudo gate live. Note: in **standalone** the cone approval is the **HTTP broker** → node-server **native macOS `osascript` dialog** ("SLICC sudo", Deny / Allow Once / Always) via `/api/sudo-approve` — not the panel-RPC `window.confirm` path (that's the extension-panel responder). Worth knowing for anyone testing this.
- **Deny → `EACCES`, fail-closed.** Clicked *Deny* → the realm's `writeFileSync` threw `Error: EACCES: sync-fs bridge, '/workspace/guarded/test.txt'` and the file stayed absent (`ENOENT`). (`POST /api/sudo-approve` 11.7s.)
- **Allow → write commits.** Clicked *Allow Once* → cone reported `WRITE_RETURNED_OK`, exit 0; the file now reads `ALLOWED-WRITE` on the real VFS. (`POST /api/sudo-approve` 7.9s.)
- **No deadlock (the §11 reentrancy concern).** While the sync write was pending (realm worker blocked on the sync XHR) and the prompt was up, a page `Runtime.evaluate` returned in **5.6 ms** — the page thread was never blocked. The channels are disjoint by construction (sync-fs = SW fetch + BroadcastChannel; sudo = HTTP broker / panel-RPC), so there's no shared chokepoint.
- **Ungated surfaces stay ungated.** The panel terminal / shell exec bridge write to `/workspace/guarded/` **without** a prompt — gating is correctly scoped to the agent's fs.

This closes the one item the prior smoke could only cover with node tests (`sync-fs-acceptance.test.ts`: real `sudoFs`, deny→`EACCES`). The live behavior matches. The sudo-under-sync-write path is now verified end-to-end in a real browser.

karlpauls:
## 📋 Consolidated test report — full picture for reviewers (supersedes the two smoke comments above)

This is the authoritative summary of everything verified for the sync-fs bridge at HEAD `6fb127383`. The two earlier comments (live-smoke + sudo) are folded in here; where they differ, this one wins (notably: **sudo is now LIVE-verified**, not node-covered-only).

### How it was tested
Real browser over CDP against a locally-built bundle (`dist/ui` with the reworked `llm-proxy-sw.js` as the **controlling** SW + a real kernel-worker `DedicatedWorker`), driving realm code through the shell exec bridge and the WC composer. Run in **both floats**: standalone (`dev:standalone:fresh`) and the **extension hosted-leader tab** (`dev:extension:fresh`). Sudo run with a real Adobe login so the agent's gated fs was in play.

### Rework soundness (9 commits `4dd63d33..6fb12738`)
Two independent adversarial reviews (mine + a fresh subagent reading current files) concurred: **the sandbox escape stays closed** and all commits are sound. Fan-out only reaches legitimate nonce channels (still gated by `maySetSyncFsNonce`); response correlation is per-request UUID; the nonce is never in `RealmInitMsg`; `DEDUPE_TTL = timeout+5s` closes the old 15s<25s double-dispatch window. Only nit: `syncFsChannels` grows monotonically per SW lifetime (MV3 eviction resets it) — negligible.

### Live-verified — all pass

| Area | Scenario | Evidence |
|---|---|---|
| Core | `writeFileSync`+`readFileSync` round-trip, no deadlock | coherent read-back; page-side async `vfs cat` confirms the write hit the **real VFS** |
| Core | Large **binary** (2 MB) round-trip | checksum exact (`267386880`), `byte0=0`/`byteLast=255`, len 2097152 |
| Core | Bridge **read-miss** (over-cap → `ENOSYNC` → bridge GET) | 2 MB read back intact |
| Coherence | **Coh#1** read-your-deletes (sync `unlinkSync`→`readFileSync`) | `ENOENT` — not resurrected via bridge |
| Coherence | **Coh#2** `statSync().size` on over-cap placeholder | reports real `2097152`, not 0 |
| Coherence | **Coh#3** `existsSync`/`statSync` past the 500-file content budget | 520 files → **all present**, `absentCount:0`; beyond-budget file size correct |
| Coherence | **async↔sync**: sync-write→async-read; async-write-new-file→sync-read | coherent (`SYNC1`; `ASYNC2` via bridge miss) |
| Coherence | **End-of-realm flush** of cache-only ops (mkdir/rename/unlink/copy) | fresh realm sees `flush-dir` = `[sub, renamed.txt, copy.txt]`; renamed src gone, deleted file gone, empty `sub` dir persisted |
| Errno | read nonexistent | `ENOENT` (not bare `EIO`) |
| Dir ops | `mkdirSync -p` / `readdirSync` / `statSync().isDirectory` | correct |
| Transport | **Multi-tab fan-out** | 2nd leader tab published its own nonce; the **original** tab's sync-fs still works (old last-writer-wins bug would `EIO` it forever) |
| Transport | MV3 SW-eviction self-heal | first op after eviction fails **closed** (`EIO` + `requestSyncFsNonce`); next op succeeds |
| Transport | SW own-listener fail-closed: malformed path | `%ZZ` → **400 `EINVAL`** + `x-slicc-fs:1` marker; no-nonce → 503 `EIO` (via cold-start) |
| Security | **Nested-iframe nonce hijack** (cherry/sprinkle/dip escape) | a same-origin `allow-same-origin` srcdoc iframe posted a nonce + listened; after 5 real token-bearing writes it received **0** — the gate kept it out of the fan-out |
| Float | **Extension hosted-leader tab** | controlling SW = `llm-proxy-sw.js`; sync write+read round-trips; async read confirms real-VFS persistence — identical to standalone |
| **Sudo** | **Deny → fail-closed** | native "SLICC sudo" `osascript` dialog → *Deny* → realm threw `EACCES: sync-fs bridge, '/workspace/guarded/test.txt'`; file absent (`POST /api/sudo-approve` 11.7s) |
| **Sudo** | **Allow → commits** | *Allow Once* → `WRITE_RETURNED_OK` exit 0; file = `ALLOWED-WRITE` on the real VFS (`POST /api/sudo-approve` 7.9s) |
| **Sudo** | **No deadlock** (§11 reentrancy) | page `Runtime.evaluate` returned in **5.6 ms** while the sync write was pending + prompt up; sync-fs (SW+BroadcastChannel) and sudo (HTTP broker) are disjoint channels |
| **Sudo** | Gating is agent-scoped | ungated panel terminal / shell exec bridge write to the gated path with **no** prompt |
| Regression | normal shell (`ls`), async `fs.promises`, non-sync-fs compute realm | all unaffected (fast path) |

**Note on the standalone sudo prompt:** it's a **native macOS `osascript` dialog** spawned by node-server (HTTP broker → `/api/sudo-approve`), not the panel-RPC `window.confirm` (that's the extension-panel responder). So it's clickable only by a human — the intended security property.

### Performance
- Per sync op **~3 ms** (60 sequential writes = 182 ms).
- Realm boot (snapshot walk): **0.35 s @ ~520 files → 0.71 s @ ~5 000 files**. Coh#3 walks metadata past the content budget to the 20 000-entry cap (was 500), so a **large** workspace pays a bounded, ~linear (~0.08 ms/file, ≤~1.6 s at cap) per-spawn cost it didn't before. Price of correct `existsSync`/`statSync`; typical workspaces (<1 000 files) stay <0.5 s. A lazy snapshot / phase-2 bridged metadata would remove it — follow-up, not a blocker.

### Covered by node tests, not live-drivable in a headless smoke
- **SPA-fallback marker guard** — needs a broken/absent SW serving `index.html` for `/__slicc/fs-sync/*`; unit-tested (marker-less 2xx → `EIO`, read + write).
- **Exec-boundary coherence** (flush-before / re-snapshot-after) — the realm's `exec` is a sliccy module, not callable from an injected one-liner; covered by `realm-rpc` ("flushes-before then re-snapshots-after") + `sync-fs-coherence`. The underlying `vfs.snapshot` RPC is exercised live on every op.
- **Broker-unavailable/timeout → `EACCES`** — the live deny path (`EACCES`) is verified above; the *timeout*→`EACCES` variant is unit-tested (`sync-fs-acceptance.test.ts` with a real `sudoFs`).

### Characterized behavior (pre-existing, documented — not a regression)
Within a single realm, an `fs.promises` (async) overwrite of a file **already in the boot snapshot** is not visible to a later `readFileSync` (returns stale cached bytes; the async read sees the new content). Inherent to the point-in-time snapshot, matches the §4 coherence model, heals at the next exec boundary, and was equally true before this PR. New files and sync-written files are always coherent.

### Deferred (tracked)
- **#1522** — the SW `message` listener sets `bridge-config`/`extension-delegate` without the `window` gate (re-review Finding 2). LOW, pre-existing, bounded by the poster's own `source.id`; intentionally not bundled into a filesystem PR.

**Bottom line:** the reworked bridge works end-to-end in a real browser across both floats; the sandbox escape is closed under live fire; coherence/deletes/binary/multi-tab/flush all behave; sudo deny/allow/no-deadlock are live-verified; no hot-path perf regression. Remaining items are node-covered or the documented large-workspace boot cost. Still do-not-merge pending external review.

karlpauls:
## Follow-up: review responses + three commits (parity, cold-start, gate-tighten)

Thanks for the deep review — it caught a real hole. All four points verified against the code **and** the live rig (standalone, real login); one was a genuine security bug, one a real-but-inherent tradeoff, one off. Also landed two feature fixes surfaced by an agent sync-fs experiment. New commits on the branch: `0eda55c70`, `1ca4aa1b3`, `20932fd74` (on top of phase-2 `5ee517c69`).

### Responses to the review

**#1 — `auxiliary`-window nonce bypass → CONFIRMED, FIXED (`20932fd74`).** You're right. Sprinkle iframes render with `allow-popups` (`sprinkle-renderer.ts`), so sprinkle JS could `window.open` a same-origin scriptable auxiliary window, post an attacker nonce (the old gate accepted `auxiliary`), and — because the SW fans every request to **all** channels — that channel would receive every realm's capability token. `maySetSyncFsNonce` now admits **`top-level` only**. The sole publisher (`wc-live`) is always top-level (verified live: `opener===null`, `top===true`) across standalone / the extension hosted-leader tab / the electron overlay; `auxiliary` bought nothing legitimate (a manually `window.open`'d leader just falls back to the bounded snapshot). A `top-level` context isn't reachable with attacker script from a sprinkle (popups are `auxiliary`), so this is the tightest gate that still admits the real leader. **Verified live on the fixed build:** sync-fs still works (leader accepted); a nested `allow-popups` sprinkle iframe posting an attacker nonce received **0 of 4** token-bearing writes; and cold-start recovery still succeeds *through* the tightened gate.

**#2 — sudo approval latency vs. the sync timeout → CONFIRMED real & inherent; decided + documented (spec §11).** The sync path is hard-capped by the 30s XHR `timeout` (set before `send()`, not extendable mid-flight — a fundamental sync-XHR limit) and the SW's ~25s budget, both well under human approval latency. So a slow-to-approve sudo-gated **sync** write returns `EIO` but still commits — the at-least-once behavior, most acute exactly on the sudo path, as you noted. This is separate from deadlock and can't be fixed on the sync primitive. Verified the escape hatch: the **async** realm RPC has **no timeout** (`realm-rpc.ts`), so `fs.promises.writeFile` waits the broker's full 600s and returns the true approve/deny. Decision: sync stays best-effort; **writes needing human approval should use the async fs**; a sync-write throw ≠ "did not write" — re-read to confirm.

**#3 — cross-float parity → OFF (verified in code + live).** The extension **leader tab** is served from the hosted origin where `chrome.runtime.id` is undefined, so `isExtension` is false and `main.ts` routes it to **`mountWcUiLive`**, not `mountWcUiExtension` — the bridge IS enabled there (proven live earlier: controlling `llm-proxy-sw` + a sync-fs round-trip in the extension hosted-leader tab). `mountWcUiExtension` (side panel / detached popout) runs **no realms** — the kernel worker lives in the leader tab. So there's no standalone-vs-extension `readFileSync` divergence: the bridge is available wherever a realm can execute. Your grep saw the truth (`syncFsBridgeEnabled` only in `wc-live`) — that IS the extension leader. Documented in spec §7 to preempt the confusion.

**#4 — security paths only tested against fakes → added live coverage.** This session's real-browser smoke (both floats) covered the escape closure by execution, not just spec-reading: the nested `allow-popups` sprinkle-iframe "add a channel" case (0 tokens), cold-start recovery, and sudo deny→`EACCES` / allow→commit. The `auxiliary` case is a unit test + closed by the fix.

**"What's actually breaking today"** — the strong justification you asked for: the cone hit the **worst failure mode** in a real experiment — `existsSync`/`statSync`/`readFileSync` silently returning absent/empty for real files created after the boot snapshot. Silent wrong data, not an error. Phase-2 + this work fixes exactly that.

### The two feature commits (also new since the last PR update)

Both surfaced by the same agent experiment (a `synctest.mjs` exercising `node:fs`):

- **`0eda55c70` — Node-parity sync methods.** The realm shim was missing common ones; a ported script hit "not a function" on `appendFileSync` (and `rmdirSync`). Added `appendFileSync`/`truncateSync`/`rmdirSync`/`accessSync`/`lstatSync`/`realpathSync`/`cpSync`/`chmodSync`, reusing extracted helpers so they share the exact cache→bridge fallback + read-your-deletes tombstone logic. fd-based (`openSync`/…) and symlink ops stay unsupported by design. Verified live: all 8 functional (`appendFileSync`→"ab", `cpSync` copies a tree).
- **`1ca4aa1b3` — cold-start `EIO` fix.** An MV3 SW eviction+respawn drops the in-memory nonce and `controllerchange` doesn't re-fire, so the first sync-fs op of a resuming turn failed closed with a spurious `EIO` (worse for `existsSync`→false). Fix **C**: on a cold op the SW asks the page to re-publish and awaits a bounded `createNonceWaiter` (≤2s, well under the XHR timeout), then succeeds instead of `EIO` (warm path never waits; still fail-closed on timeout). Fix **K**: `wc-live` re-publishes the nonce on tab focus/visibility so C rarely fires. Verified live: after a 65s idle (SW evicted), a single cold `writeFileSync` succeeded with `need-nonce` firing *during* the op — the exact pattern that reliably `EIO`'d before.

### Verification
854 node tests green; browser + kernel-worker-guard typecheck 0; biome clean. Live smoke on the final gate-tightened bundle covered #1's closure, cold-start recovery, and warm read/write — plus the earlier full smoke (multi-tab fan-out, MV3 self-heal, marker guard, extension float, sudo deny/allow) still holds (untouched paths).

PR #1579, opened 2026-07-20
test(webapp): cover shell commands and playwright handlers

catalan-adobe:
## What

Adds unit tests for 11 previously under-tested `packages/webapp` shell files, plus a reusable test harness for the playwright-cli handlers. Test-only change — no production code is modified.

### Shell commands
| File | Lines before → after |
|------|----------------------|
| `shell/supplemental-commands/fswatch-command.ts` | 4.6% → 100% |
| `shell/supplemental-commands/theme-command.ts` | 3.4% → 100% |
| `shell/supplemental-commands/pdftk-command.ts` | 39% → 97% |
| `shell/supplemental-commands/esbuild-command.ts` | 45% → 93% |

### Playwright handlers
| File | Lines before → after |
|------|----------------------|
| `playwright/handlers/network-requests.ts` | 55% → 85% |
| `playwright/handlers/state.ts` | 4% → 100% |
| `playwright/handlers/interaction.ts` | 71% → 80% |
| `playwright/handlers/mouse.ts` | 79% → 96% |
| `playwright/handlers/routing.ts` | 82% → 93% |
| `playwright/handlers/devtools.ts` | 77% → 96% |
| `playwright/handlers/snapshot.ts` | 69% → 97% |

### Test harness
`tests/shell/helpers/playwright-harness.ts` provides a mock `BrowserAPI` + `CDPTransport` (with an event-emit driver for CDP event handlers), an empty `PlaywrightState` factory, and a `createHandlerCtx` builder. The seven handler suites share it.

## How it's tested

- `fswatch`/`theme` had no test file; the rest only exercised argument parsing or happy paths. New tests cover operation bodies and error branches: `fswatch` via the `globalThis` watcher hook, `theme` via a mocked panel-RPC client, `pdftk` via mocked `@cantoo/pdf-lib`/`unpdf`, `esbuild` via a mocked `getEsbuild` wasm loader plus the VFS resolver plugin.
- Playwright handlers run against the mock browser/transport; the network-requests and routing suites drive captured CDP events (`Network.*`, `Fetch.requestPaused`) through the harness emitter.

## Notes

- 195 tests pass across the 11 files (2 heavy-wasm esbuild tests stay skipped behind `SLICC_TEST_HEAVY_WASM`, matching the existing gate).
- Commits are one-file-each for review and rollback.
- The `webapp` per-package coverage floor will rise via the nightly ratchet once CI measures these files.

### DIFF
--- a/packages/webapp/tests/shell/helpers/esbuild-mock-ctx.ts
+++ b/packages/webapp/tests/shell/helpers/esbuild-mock-ctx.ts
@@ -1,0 +1,39 @@
+import type { CommandContext, IFileSystem } from 'just-bash';
+import { vi } from 'vitest';
+
+/**
+ * File-store-backed {@link CommandContext} for esbuild command tests. Reads and
+ * writes go through an in-memory map so bundle/transform paths resolve VFS
+ * files without a real filesystem. Shared by the mocked-loader unit suite and
+ * the opt-in live-wasm suite so the ctx shape stays identical across both.
+ */
+export function createEsbuildMockCtx(
+  overrides: Partial<{ fs: Partial<IFileSystem>; cwd: string; stdin: string }> = {}
+): CommandContext {
+  const fileStore = new Map<string, string>();
+  const fs: Partial<IFileSystem> = {
+    resolvePath: (base: string, path: string) =>
+      path.startsWith('/') ? path : `${base.replace(/\/$/, '')}/${path}`,
+    exists: vi.fn().mockImplementation(async (p: string) => fileStore.has(p)),
+    readFile: vi.fn().mockImplementation(async (p: string) => {
+      const v = fileStore.get(p);
+      if (v === undefined) throw new Error(`ENOENT: ${p}`);
+      return v;
+    }),
+    writeFile: vi.fn().mockImplementation(async (p: string, content: string | Uint8Array) => {
+      fileStore.set(p, typeof content === 'string' ? content : new TextDecoder().decode(content));
+    }),
+    stat: vi.fn().mockImplementation(async (p: string) => {
+      if (!fileStore.has(p)) throw new Error(`ENOENT: ${p}`);
+      return { isFile: true, isDirectory: false, size: fileStore.get(p)!.length };
+    }),
+    readFileBuffer: vi.fn().mockImplementation(async () => new Uint8Array()),
+    ...overrides.fs,
+  };
+  return {
+    fs: fs as IFileSystem,
+    cwd: overrides.cwd ?? '/workspace',
+    env: new Map<string, string>(),
+    stdin: overrides.stdin ?? '',
+  } as unknown as CommandContext;
+}
--- a/packages/webapp/tests/shell/helpers/playwright-harness.ts
+++ b/packages/webapp/tests/shell/helpers/playwright-harness.ts
@@ -7,7 +7,7 @@ import type {
   PlaywrightState,
 } from '../../../src/shell/supplemental-commands/playwright/types.js';
 
-type EventListener = (params: Record<string, unknown>) => void;
+type EventListener = (params: Record<string, unknown>) => unknown;
 
 /** Fresh, empty {@link PlaywrightState} for handler tests. */
 export function createPlaywrightState(): PlaywrightState {
@@ -31,8 +31,12 @@ export function createPlaywrightState(): PlaywrightState {
 export interface MockTransport {
   transport: CDPTransport;
   send: ReturnType<typeof vi.fn>;
-  /** Fire a CDP event to every registered listener. */
-  emit: (event: string, params: Record<string, unknown>) => void;
+  /**
+   * Fire a CDP event to every registered listener and await async handlers, so
+   * a handler that issues follow-up `transport.send` calls has settled before
+   * the returned promise resolves (no timer flushing needed).
+   */
+  emit: (event: string, params: Record<string, unknown>) => Promise<void>;
   /** True once a listener for `event` is registered. */
   hasListener: (event: string) => boolean;
 }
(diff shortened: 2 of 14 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/tests/shell/helpers/playwright-harness.ts:13-28]** — The fresh `PlaywrightState` factory centralizes a fairly large state shape and prevents suites from accidentally sharing mutable maps.
- **[packages/webapp/tests/shell/helpers/playwright-harness.ts:44-69]** — The transport harness models listener registration/removal and allows tests to drive actual CDP event handlers rather than calling internal implementation functions directly.
- **[packages/webapp/tests/shell/supplemental-commands/playwright/handlers/network-requests.test.ts:59-90]** — The request test exercises the complete request/response/loading event sequence, including session filtering and asynchronous response-body retrieval.
- **[packages/webapp/tests/shell/supplemental-commands/fswatch-command.test.ts:143-189]** — The fswatch suite verifies glob filtering, event payload routing, registry listing, deletion, and actual unsubscribe behavior. Resetting the module graph appropriately isolates the module-level registry.
- **[packages/webapp/tests/shell/supplemental-commands/theme-command.test.ts:105-152]** — Theme tests distinguish missing files, malformed JSON, structurally invalid themes, and successful RPC payloads rather than merely checking exit codes.
- The change is correctly test-only, introduces no dependencies, and broadly aligns with the stated set of shell commands and Playwright handlers.

### Issues

*(none found)*

- **[packages/webapp/tests/shell/supplemental-commands/esbuild-command.test.ts:22-34,397-415]** — The file-wide `vi.mock` also replaces the implementation used by the pre-existing “live wasm” tests.
  - Vitest hoists this mock, so `resetEsbuildForTests()` is a no-op and `getEsbuild()` always returns the fake module. The version test therefore passes against the hard-coded `0.25.0` without loading wasm. With `SLICC_TEST_HEAVY_WASM=1`, the transform test will use the unconfigured `vi.fn()` rather than real esbuild and will likely fail with exit code 1. More importantly, the opt-in test can no longer detect loader or wasm integration regressions.
  - Move the live tests into a separate unmocked test file, or isolate the mocked tests behind dynamic imports and `vi.doMock`/`vi.unmock` with module resets.

- **[packages/webapp/tests/shell/supplemental-commands/pdftk-command.test.ts:32-33,307-345,376-383]** — The PDF mock and assertions verify page counts, but not which pages or rotations are produced.
  - `copyPages` discards the supplied indices, while the tests only count `addPage` calls. An off-by-one implementation that copied pages 2–3 for `1-2`, reversed handle order, or applied `left` instead of `right` would still pass. Likewise, `setRotation` is only checked for invocation, not the selected page or angle. These are the core correctness risks in `cat` and `rotate`.
  - Make `copyPages` a spy that returns page identities tied to source documents and indices, then assert ordered indices such as `[0, 1]`, handle order, and exact rotation arguments such as `{ angle: 90 }`.

#### 💡 Minor (Nice to Have)

- **[packages/webapp/tests/shell/helpers/playwright-harness.ts:10,35,65-67]** — `emit` discards listener return values even though routing listeners are asynchronous.
  - Tests consequently use timer flushing and an async listener rejection may surface as an unhandled rejection rather than a deterministic failure attributed to `emit`.
  - Type listeners as returning `unknown` and make `emit` return `Promise<void>` using `await Promise.all(...)`.

- **[packages/webapp/tests/shell/supplemental-commands/playwright/handlers/interaction.test.ts:186-201,222-237]** — Some descriptions promise behavior that is not fully asserted.
  - The type test checks only Enter `keyDown`, not `keyUp`; the fill test named “with the React fallback and submits” checks neither the fallback `Runtime.callFunctionOn` nor the Enter events. Removing those behaviors would leave the tests green.
  - Assert both key events and inspect the relevant `Runtime.callFunctionOn` invocation/function declaration.

### Recommendations

- Separate mocked unit suites from opt-in integration suites whenever a module-wide mock would intercept both.
- For transformation-heavy commands such as `pdftk`, assert ordered semantic outputs—source page, destination order, and rotation—not only invocation counts.
- Keep the event harness deterministic by allowing callers to await asynchronous listeners.

### Assessment

**Reasoning:** The added coverage is broad and generally well structured, but the esbuild mock breaks the intended live-wasm validation, and the PDF tests do not yet protect the most failure-prone page-selection and rotation semantics.

---
*Review generated by pr-review custom script · model: github-copilot/gpt-5.6-sol*

catalan-adobe:
Thanks for the thorough review — addressed all four points in follow-up commits:

**Should Fix**
- **esbuild mock intercepting the live-wasm suite** (`76f13a4d7`) — the file-wide `vi.mock` did defeat the `SLICC_TEST_HEAVY_WASM` tests (version passed against the hard-coded `0.25.0`; transform would hit an unconfigured `vi.fn`). Moved the two live tests into a new unmocked file `esbuild-command.live.test.ts` (vitest scopes `vi.mock` per-file) and extracted the shared ctx factory to `tests/shell/helpers/esbuild-mock-ctx.ts`. Verified: `SLICC_TEST_HEAVY_WASM=1` now runs the live tests against the real loader and they pass.
- **pdftk page order / rotation** (`a43d29ea0`) — the mock now records each `copyPages(src, indices)` call. Tests assert the exact 0-based indices in order (`1-2` → `[0,1]`, `2-end 1` → `[[1,2],[0]]`), handle-merge order (`A`→`in1` then `B`→`in2`, each `[0,1,2]`), and the exact rotation (`setRotation` called twice with `{ angle: 90 }`). An off-by-one or wrong-angle regression now fails.

**Nice to Have**
- **interaction assertions** (`64e5c329e`) — the type/fill submit tests now assert the Enter `keyUp` (not just `keyDown`) and the React native-setter fallback `Runtime.callFunctionOn` invocation.
- **async listeners in the harness** (`586deac02`) — `emit()` now awaits listener promises; the routing/network-requests suites drop the `setTimeout` flush and `await transport.emit(...)` directly.

195 pass / 2 skipped locally (the 2 live-wasm tests behind the flag), all lint- and typecheck-clean.

PR #1584, opened 2026-07-20
Fix terminal panel erasing command output that lacks a trailing newline

### LINKED ISSUE #1583: Terminal panel: command output without trailing newline is invisible (erased by readline prompt redraw)
## Repro (interactive Terminal panel, kernel shell)

```
/ $ echo -n ABC123
/ $
```

Expected: `ABC123` visible (bash renders `ABC123` + prompt on next line; zsh adds the reverse-video `%` marker).
Actual: nothing — the output row is erased.

Also reproduces with `cat` on any file lacking a trailing newline (e.g. an 8-byte `scoop-ok` file written by a scoop): the terminal shows the command echo and then a fresh prompt, no content.

Proof the bytes are emitted: `cat file; echo TAIL` renders `scoop-okTAIL` — the stdout arrives, but the final unterminated line is clobbered.

## Root cause

`packages/webapp/src/kernel/remote-terminal-view.ts` — `runPromptLoop()` → `readNextLine()` → `readline.read(PROMPT)` (xterm-readline). readline redraws the prompt with a carriage-return/line-clear at the current row, erasing whatever unterminated output the preceding `terminal-output` event left there (`handleEvent` writes `event.data` raw).

The chat-panel shell tool is unaffected — it consumes captured stdout from `client.exec` rather than the rendered terminal.

## Suggested fix

Before re-entering `readline.read(PROMPT)`, check `terminal.buffer.active.cursorX > 0` and write `\r\n` first (or the zsh-style reverse-video `%` + newline) so partial lines survive the prompt redraw.

Found while adding VFS smoke checks to the cdp-smoke-test skill (PR #1582).

catalan-adobe:
Fixes #1583

## Problem

`echo -n ABC123` or `cat` of a file without a final newline showed **nothing** in the interactive Terminal panel: the output bytes arrived (`cat file; echo TAIL` proved it), but xterm-readline's prompt redraw starts with a carriage-return/line-clear and erased the unterminated row. The chat-panel shell tool was unaffected (captured stdout, not the rendered terminal).

## Fix

`runPromptLoop` → `readNextLine` now goes through `readAfterLineGuard`, which calls the new exported `ensurePromptLineStart(terminal)`:

1. zero-byte `write('')` as a flush barrier — its callback fires only after xterm parsed all queued output, so `buffer.active.cursorX` is trustworthy;
2. writes `\r\n` **only** when the cursor sits mid-row, so empty prompt iterations stay single-spaced.

The dispose-abort race in `readNextLine` is preserved (abort promise registered before the guard runs).

## Tests

`packages/webapp/tests/kernel/remote-terminal-prompt-line-guard.test.ts` — 3 cases against a fake terminal with xterm-style async writes: mid-row → `\r\n`, column-0 → untouched, and cursorX read **only after** the flush barrier (queued-output simulation). Mutation-checked (breaking the condition fails 2/3). Full kernel suite: 78 files / 1267 tests pass; `tsc --noEmit` clean; touched-file exemption gate OK.

## Live verification

Rebuilt the webapp, hot-swapped `dist/ui` under the running `dev:standalone:fresh` harness, reloaded the leader tab: `echo -n ABC123` → `ABC123` visible, `cat /shared/scoop-note.txt` (8-byte scoop artifact) → `scoop-ok` visible, prompt on the next row, zero console errors.

### DIFF
--- a/packages/webapp/src/kernel/remote-terminal-view.ts
+++ b/packages/webapp/src/kernel/remote-terminal-view.ts
@@ -420,7 +420,23 @@ export class RemoteTerminalView {
     const aborted = new Promise<never>((_resolve, reject) => {
       this.abortPromptLoop = reject;
     });
-    return Promise.race([this.readline.read(PROMPT), aborted]);
+    return Promise.race([this.readAfterLineGuard(), aborted]);
+  }
+
+  /**
+   * Guard the partial output line, then hand the row to readline.
+   *
+   * Command output that does not end in a newline (`echo -n foo`, `cat`
+   * of a file lacking a trailing `\n`) leaves the cursor mid-row, and
+   * readline's prompt redraw erases that row — the output became
+   * invisible (#1583). Flush pending writes first so `cursorX` reflects
+   * everything `handleEvent` streamed, then start the prompt on a fresh
+   * row only when needed.
+   */
+  private async readAfterLineGuard(): Promise<string> {
+    if (this.terminal) await ensurePromptLineStart(this.terminal);
+    if (!this.readline) throw new Error('readline not mounted');
+    return this.readline.read(PROMPT);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/src/kernel/remote-terminal-view.ts:418-439]** — The implementation matches the PR plan: the dispose-abort promise is registered before the asynchronous line guard, and `readline.read()` starts only after the guard completes.
- **[packages/webapp/src/kernel/remote-terminal-view.ts:1102-1121]** — `PromptLineGuardTerminal` is a narrow structural interface that keeps the behavior independently testable. The zero-byte write correctly acts as an xterm parse barrier, and `\r\n` is emitted only when `cursorX > 0`, avoiding extra blank lines.
- **[packages/webapp/src/kernel/remote-terminal-view.ts:1107-1115]** — Comments clearly document why the barrier and conditional newline are necessary, including the asynchronous xterm behavior.
- **[packages/webapp/tests/kernel/remote-terminal-prompt-line-guard.test.ts:31-52]** — Tests cover the important branches: partial output, an already-complete line, and queued output that changes the cursor only after the flush callback.
- The change is tightly scoped, introduces no dependencies, and preserves existing terminal/session architecture.

### Issues

*(none found)*

*(none found)*

#### 💡 Minor (Nice to Have)

*(none found)*

### Recommendations

If a practical browser-level harness is available, consider adding an integration regression test using the real xterm/readline combination. The current focused tests validate the guard well, but an integration test would also ensure the prompt loop remains wired through it during future refactoring.

### Assessment

**Ready to merge?** Yes

**Reasoning:** The implementation aligns with the stated plan, correctly accounts for xterm’s asynchronous write queue, preserves the disposal race behavior, and has focused coverage for all planned cases. No merge-blocking correctness or architectural issues were found.

---
*Review generated by pr-review custom script · model: github-copilot/gpt-5.6-sol*

PR #1591, opened 2026-07-21
fix(deps): pin e2b <2.33.0 (workerd createRequire) + split Renovate updates one-per-PR

karlpauls:
## Problem

The grouped Renovate PR **#1585** went red on `webapp` (E2E) and `Deploy staging + smoke test`. Root cause: a transitive bump of **`e2b` 2.32.0 → 2.33.0** (bundled into the tray-hub worker `slicc-tray-hub` via `@slicc/cloud-core`, whose `CloudSessionsDurableObject` drives the cloud-cone lifecycle at runtime).

e2b **2.33.0** started emitting an ESM-interop shim (from its bundler, Rolldown) at the top of `dist/index.mjs`:
```js
import { createRequire } from "node:module";
var __require = /* #__PURE__ */ (() => createRequire(import.meta.url))();
```
It runs at **module-eval time**, and under Cloudflare **workerd** `import.meta.url` is `undefined`, so `createRequire(undefined)` throws the instant the worker is instantiated — before any request. That breaks `wrangler dev` (webapp E2E web-server can't start) and `wrangler deploy` (Cloudflare validation error 10021).

e2b **2.32.x** has no such shim (a clean worker bundle has **zero** `createRequire`), and the SDK is otherwise fetch-based (`openapi-fetch` + `@connectrpc/connect-web`), so it runs in workerd fine. This is purely an upstream build-artifact regression — and it **persists through the latest** e2b (verified 2.33.1 / 2.34.0 / 2.35.0), so upgrading forward does not fix it.

Full trace: this is why the *same* `renovate/non-major` branch passed staging on 07-18/19/20 and only failed 07-21 (the day e2b 2.33.0 entered the re-resolved lockfile).

## Fix (option #1: pin)

- Cap `e2b` `>=2.23.0 <2.33.0` in **root**, **cloud-core**, and **node-server** `package.json` → single hoisted `2.32.0`, no version churn in the lockfile (only the range strings change).
- Add a `renovate.json` `allowedVersions: "<2.33.0"` rule for `e2b` (with rationale) so Renovate stops proposing `>=2.33.0`.
- **Regression guard**: `packages/cloud-core/tests/e2b-workerd-pin.test.ts` asserts the resolved e2b stays `<2.33.0` (proven to catch 2.33.0/2.33.1/2.34.0/2.35.0).

We deliberately **do not** un-bundle e2b from the worker: the worker genuinely calls the SDK at runtime, so that would mean relocating the whole cloud-cone lifecycle off workerd (architectural), not a bundling tweak. Node runtimes (`node-server --cloud`) don't hit the crash but are capped too, to keep one hoisted version and avoid churn. Drop the cap once e2b makes the shim lazy or ships an edge/browser export (worth filing upstream).

## Renovate: one update per PR

The `#1585` confusion came from the broad **"non-major dependencies"** catch-all group bundling a dozen updates into one PR. This PR removes that group so **each dependency updates in its own PR** — a breaking transitive bump then fails *in isolation* and is trivial to identify, and can't block unrelated safe updates. The purposeful cohesion groups are kept (`semantic-release`, `xterm`, `@earendil-works/*`, `just-bash`, patched dependencies). We add an explicit `prHourlyLimit: 4`, and the 7-day `minimumReleaseAge` cooldown + `automerge` keep the PR volume in check.

Once this lands, Renovate closes the now-obsolete grouped #1585 and reopens isolated per-dependency PRs; the e2b one won't propose 2.33.0 (capped).

## Docs
`docs/pitfalls.md` → new **"e2b SDK in the Worker: `createRequire` Breaks workerd"** section.

## Verification
| Check | Result |
| --- | --- |
| Worker dry-run bundle `createRequire` count | ✅ 0 (was throwing on 2.33.0) |
| `wrangler dev` boot | ✅ `Ready on http://127.0.0.1:8798`, no throw |
| Regression guard test | ✅ passes; catches ≥2.33.0 |
| cloud-core tests + coverage floor | ✅ exit 0 (103 tests) |
| `renovate-config-validator` | ✅ "Config validated successfully" |
| lint:docs / biome / prettier / cloud-core typecheck | ✅ |

### DIFF
--- a/docs/pitfalls.md
+++ b/docs/pitfalls.md
@@ -1158,8 +1158,8 @@ The tray-hub worker (`slicc-tray-hub`) bundles the **e2b SDK** — its
 and e2b enters the worker bundle transitively through `@slicc/cloud-core`
 (`packages/cloud-core` depends on `e2b`).
 
-**e2b `2.33.0` broke this.** Its build began emitting an esbuild ESM-interop
-shim at the top of `dist/index.mjs`:
+**e2b `2.33.0` broke this.** Its build began emitting an ESM-interop shim
+(from its bundler, Rolldown) at the top of `dist/index.mjs`:
 
 ```js
 import { createRequire } from 'node:module';
--- a/docs/pitfalls.md
+++ b/docs/pitfalls.md
@@ -1150,6 +1150,44 @@ and Vite injects `vite:preloadError` only into the PAGE bundle — so a `window`
 listener alone can't catch it. Recovery is the four-trigger guarded reload in
 `core/stale-asset-channel.ts` + `ui/boot/setup-preload-error-reload.ts`.
 
+## e2b SDK in the Worker: `createRequire` Breaks workerd (pin < 2.33.0)
+
+The tray-hub worker (`slicc-tray-hub`) bundles the **e2b SDK** — its
+`CloudSessionsDurableObject` drives the cloud-cone lifecycle
+(`createSubstrate('e2b', …)` → `startCone`/`resumeCone`/`pauseCone`/`killCone`),
+and e2b enters the worker bundle transitively through `@slicc/cloud-core`
+(`packages/cloud-core` depends on `e2b`).
+
+**e2b `2.33.0` broke this.** Its build began emitting an esbuild ESM-interop
+shim at the top of `dist/index.mjs`:
+
+```js
+import { createRequire } from 'node:module';
+var __require = /* #__PURE__ */ (() => createRequire(import.meta.url))();
+```
+
+This runs at **module-eval time**. Under Cloudflare **workerd**, `import.meta.url`
+is `undefined`, so `createRequire(undefined)` throws
+`TypeError: The argument 'path' … Received 'undefined'` the instant the worker is
+instantiated — before any request. That kills both `wrangler dev` (the webapp E2E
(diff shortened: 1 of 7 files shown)


karlpauls:
Thanks for the review — addressed the three observations in `656ad36`:

1. **`runtime-package.json` (e2b-template)** — left uncapped by design. It's not a workspace, not in our lockfile, and not scanned by Renovate (no custom manager for that filename), and it lists deps installed *inside the e2b sandbox* — a Node runtime where `createRequire(import.meta.url)` works. So it neither has the workerd bug nor affects the single hoisted version; capping it would only hold the sandbox back from e2b updates it can safely take.
2. **Guard scope** — hardened: the test now asserts on *every* e2b entry in the lockfile (hoisted **and** any nested `packages/*/node_modules/e2b`), so a future uncapped consumer forcing a nested `>=2.33.0` can't slip past.
3. **PR-body/`prHourlyLimit` nit** — added an explicit `prHourlyLimit: 4` to throttle the now-ungrouped per-dependency PRs, and corrected the PR body to match.

Also corrected the bundler attribution across `docs/pitfalls.md`, the Renovate rule, and the guard-test comment: the top-level `createRequire(import.meta.url)` shim is emitted by e2b's bundler **Rolldown**, not esbuild.

PR #1606, opened 2026-07-22
ci(renovate): auto-reflow formatter/linter bumps so they self-heal

karlpauls:
## Problem

A Renovate bump of a **formatter/linter** (`@biomejs/biome`, `prettier`) reflows existing source, so the CI `lint` gate (`biome check .` / `prettier --check .`) fails on every file the new version would reformat. Concretely, **#1602** (biome 2.5.3 → 2.5.4) fails `lint` → `ci` because 2.5.4's formatter reflows a curried-call construct (`it.each([...] as const)(...)`) in 12 test files. Renovate opens the bump but never runs the formatter, so the PR can't go green on its own.

The obvious fix — Renovate `postUpgradeTasks` running `npm run lint` — **isn't available**: this repo uses the **Mend-hosted Renovate app**, which disables arbitrary command execution (`allowedPostUpgradeCommands` is self-hosted-only).

## Solution

A companion workflow that does the reconcile, mirroring the existing **`renovate-patch-reconcile.yml`** pattern:

- **`renovate.json`** — label `@biomejs/biome` + `prettier` bumps `formatter-bump` (kept as separate PRs, per the one-dep-per-PR philosophy).
- **`.github/workflows/renovate-format-reconcile.yml`** — on a `formatter-bump`-labelled Renovate PR: `npm ci --ignore-scripts`, run `biome check --write . && prettier --write .` (the formatting half of `npm run lint`, same order the repo uses), and if the tree changed, commit + push the reflow. The push re-runs the check suite (now green) → Renovate automerges.

### Security posture (identical to `renovate-patch-reconcile`)
- Gated to **Renovate's own branch PRs in this repo** (`head.repo == repository && actor == renovate[bot]`) carrying the label — never forks.
- **`npm ci --ignore-scripts`** so the bumped dependency's lifecycle code doesn't run (`minimumReleaseAge` bounds the malicious-version window).
- Pushes with **`BOT_PAT`** (already present) — a `GITHUB_TOKEN` push is anti-recursion-suppressed and wouldn't re-run CI.
- Third-party actions pinned to the **same immutable commit SHAs** as `renovate-patch-reconcile`.
- No untrusted input reaches a shell: `head.ref` is only `actions/checkout`'s `ref:` input; the sole `run:` interpolation is the integer PR number.
- Formatters run with `|| true` so a bump that *also* adds an unfixable lint error still gets its auto-fixable reflow pushed — the PR's own `lint`/`ci` checks stay the gate (red → no automerge → human handles the remainder). This workflow only applies + pushes fixes; it is not itself the gate.

## Unblocking #1602
Once this lands on `main`, add the `formatter-bump` label to **#1602** (or let Renovate re-label it on its next scan) — the `labeled` trigger fires the workflow, it reflows the 12 files, pushes, and #1602 goes green.

### DIFF
--- a/.github/workflows/renovate-format-reconcile.yml
+++ b/.github/workflows/renovate-format-reconcile.yml
@@ -50,10 +50,13 @@ jobs:
   reconcile:
     runs-on: ubuntu-latest
     # Only Renovate's own branch PRs (head in this repo → BOT_PAT push works)
-    # that carry the formatter-bump label added by renovate.json.
+    # that carry the formatter-bump label. Gate on the PR AUTHOR
+    # (pull_request.user.login), NOT github.actor: on a `labeled` event the actor
+    # is the human who added the label, so an actor gate would skip the manual
+    # "label an already-open Renovate PR to unblock it" path (e.g. #1602).
     if: >-
       github.event.pull_request.head.repo.full_name == github.repository &&
-      github.actor == 'renovate[bot]' &&
+      github.event.pull_request.user.login == 'renovate[bot]' &&
       contains(github.event.pull_request.labels.*.name, 'formatter-bump')
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
--- a/.github/workflows/renovate-format-reconcile.yml
+++ b/.github/workflows/renovate-format-reconcile.yml
@@ -1,0 +1,105 @@
+name: Renovate Format Reconcile
+
+# A Renovate bump of a formatter/linter (biome, prettier) can reflow existing
+# source, so the CI `lint` gate (`biome check .` / `prettier --check .`) fails
+# on every file the new version would reformat — e.g. biome 2.5.4's curried-call
+# reflow (#1602). The Mend-hosted Renovate app cannot run `postUpgradeTasks`, so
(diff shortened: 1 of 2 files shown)


karlpauls:
Thanks — good catch on the 🟡.

**🟡 (fixed, `f45138dab`):** you're right that on a `labeled` event `github.actor` is the human who added the label, which would silently skip the workflow and kill the "manually label #1602 to unblock it" path. Switched the guard to `github.event.pull_request.user.login == 'renovate[bot]'` (robust across `opened`/`synchronize`/`reopened`/`labeled`, still fork-restricted via `head.repo`), and added an inline note so it doesn't get reverted to `github.actor`.

**🔵 (acknowledged, deferring):** auto-maintaining `.git-blame-ignore-revs` would need the workflow to commit the reflow, read back its own SHA, append it, then commit/push again — a two-commit/amend dance that's more fragile than it's worth for the occasional formatter-bump reflow (formatting-only lines, and `.git-blame-ignore-revs` needs per-dev `blame.ignoreRevsFile` opt-in anyway). If these reflows ever get noisy in blame, a maintainer can add the file by hand. Left out of this PR.

Note: `renovate-patch-reconcile.yml` has the same latent `github.actor` limitation — harmless today since Renovate labels patched deps on `opened`, but worth the same one-line change in a separate cleanup if we want it robust to manual labeling.

PR #1640, opened 2026-07-23
chore: clear all lint warnings with granular commits

catalan-adobe:
Summary
- clears all existing lint warnings in the repo with focused, review-friendly commits
- removes remaining noExplicitAny, naming, comma-operator, and unused-suppression warnings
- tightens several typing surfaces across providers, bridge/runtime hooks, tool adapters, and Bedrock CAMP payload shaping

Validation
- npm run lint passes
- npm test passes

Notes
- commit history is intentionally split into many small commits to simplify review and selective reverts

### DIFF
--- a/docs/pitfalls.md
+++ b/docs/pitfalls.md
@@ -877,6 +877,30 @@ The sprinkle subsystem is the canonical reference for full bidirectional dispatc
 
 Historical note: prior to the thin-bridge release the equivalent split was the chrome-extension side panel (page) vs the offscreen document (agent), bridged through `chrome.runtime.sendMessage` routed by the service worker. The realms changed but the page/worker idea is the same.
 
+## `any` → `unknown` Refactors at External Boundaries
+
+Replacing `any` with `unknown` is good, but boundary code must preserve runtime
+null-safety and shape checks. The common regression is casting and then reading
+fields directly:
+
+```ts
+const envelope = message as { source?: string };
+if (envelope.source !== 'service-worker') return; // crashes when message === null
+```
+
+Use a guard first:
+
+```ts
+if (typeof message !== 'object' || message === null) return;
+const envelope = message as { source?: string };
+```
+
+Apply this rule to external inputs (`chrome.runtime.onMessage`, parsed JSON,
+postMessage payloads, fetch responses) and environment probes. For optional
+globals such as `indexedDB`, check the **value** (`typeof ... !== 'undefined'`)
+instead of only property presence (`'indexedDB' in globalThis`), because test
+and SSR shims may define the property as `undefined`.
+
 ## Dual-Mode Testing Checklist
 
 When adding a feature that touches:
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -331,6 +331,25 @@ Key pattern: Helper functions reduce boilerplate and make tests more readable.
 
 For skipped categories, ensure **manual verification in both CLI and extension modes** before committing.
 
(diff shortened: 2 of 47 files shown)


catalan-adobe:
### Strengths

- **`packages/webapp/src/shell/supplemental-commands/models-command.ts:42-55`** — Artificial Analysis responses are normalized with explicit runtime checks rather than allowing arbitrary JSON values into numeric model metadata.
- **`packages/webapp/src/providers/account-store.ts:1008-1033`** — The service-worker response is treated as `unknown`, narrowed before use, and existing transport/service error logging is preserved.
- **`packages/webapp/src/providers/built-in/bedrock-camp.ts:296-311`** — Tool-result JSON values are now serialized before sanitization. This avoids calling string operations on object-valued `json` content while retaining support for text and images.
- **`packages/chrome-extension/src/service-worker.ts:1388-1401`** — Replacing repeated storage casts with a single structural `StorageArea` binding improves clarity without changing the secrets pipeline.
- **`packages/webapp/src/core/tool-adapter.ts:122-137`** and **`packages/webapp/src/tools/tool-ui.ts:219-227`** — Tool generics and update payloads now communicate the actual unknown/detail surfaces instead of propagating `any`.
- The comma-operator disposer rewrites retain disposal order, test-helper renames were updated consistently across in-repo callers, and no dependencies or lockfile changes were introduced.

### Issues

*(none found)*

- **`packages/webapp/src/shell/mcp/provider.ts:45-46`** — The IndexedDB availability check now tests property presence rather than whether IndexedDB is usable or defined.
  - `'indexedDB' in globalThis` is `true` when an SSR/test shim declares `globalThis.indexedDB = undefined`. In that environment, `ensureMcpProviderRegistered()` proceeds into LightningFS even though this guard exists specifically to prevent its deferred unhandled rejection.
  - Restore the value check without `any`, and add coverage for a present-but-undefined property:
    ```ts
    function hasIndexedDB(): boolean {
      return typeof (globalThis as { indexedDB?: unknown }).indexedDB !== 'undefined';
    }
    ```

#### 💡 Minor (Nice to Have)

- **`packages/webapp/src/providers/oauth-service.ts:385-390`** — Casting an `unknown` runtime message does not make property access null-safe.
  - `chrome.runtime.onMessage` is a shared broadcast boundary and accepts arbitrary serializable payloads. If another extension context broadcasts `null`, `envelope.source` throws; the old optional-chain implementation ignored it safely.
  - Validate before casting:
    ```ts
    const handler = (message: unknown) => {
      if (typeof message !== 'object' || message === null) return;
      const envelope = message as {
        source?: string;
        payload?: { type?: string; error?: string; redirectUrl?: string };
      };
      // ...
    };
    ```
    A test should invoke the captured listener with `null`, then with a valid OAuth result, confirming the flow remains operational.

### Recommendations

- Add focused boundary tests wherever `unknown` replaces `any`, especially Chrome message listeners and external JSON payloads. These are where assertions can accidentally remove prior optional-chain/null behavior.
- Keep lint-only commits behavior-preserving where possible; use narrow runtime guards instead of assertions or fallback values when handling external data.
- The granular commit structure and lack of dependency churn make selective review and reverts straightforward.

### Assessment

**Reasoning:** The PR is well-aligned with its lint-cleanup goal and most changes are disciplined type refinements. The IndexedDB guard introduces a concrete SSR/test-environment regression, while the OAuth listener has a smaller null-boundary regression that is inexpensive to harden before merge.

---

PR #1655, opened 2026-07-24
Add pre-push lint gate mirroring CI lint job

### LINKED ISSUE #1577: Lint failures caught only in CI: local hooks don't mirror the CI lint gate
## Summary

Lint failures are frequently caught **only in CI** (the PR's GitHub Actions `lint` job), forcing an extra local-fix → push → re-run round trip. The root cause: **our local git hooks and the CI lint gate check materially different things.** Nothing local mirrors the CI `lint` job, so several classes of failure can only surface after a PR is opened.

This issue documents the gap so we can decide how to close it. No code change is proposed yet — comments welcome on the preferred direction (see **Options** below).

## What runs locally today

- **`.husky/pre-commit`** → `npx lint-staged`
  - `biome check --write` on **staged** `*.{ts,tsx,js,jsx,mjs,cjs,json,jsonc,css}`
  - `prettier --write` on **staged** `*.{md,html,yaml,yml}`
  - plus `check-doc-sizes.mjs` when a budgeted doc is staged
  - → effectively **auto-fix formatting on staged files only**
- **`.husky/pre-push`** → linear-history check **only** (`check-linear-history.sh`)
  - → **no lint runs on push at all**

## What CI's `lint` job runs

`npm run lint:ci` plus extra steps (`.github/workflows/ci.yml`):

1. `biome check .` — whole repo, **no** `--write`
2. `prettier --check .` — whole repo
3. `lint:docs`, `lint:skills --strict`, `lint:no-innerhtml`, `lint:no-ui-in-providers`, `lint:patches`, `lint:no-raw-chrome-runtime-id`, `lint:hosted-origin`
4. `check-touched-exemptions.mjs` — function-size + cognitive-complexity boy-scout gate
5. `check-manifest-justifications.sh` — Chrome Web Store manifest justifications
6. `knip` dead-code — default gate **and** `--production` files gate

## The concrete holes

Things CI fails on that **nothing local catches**:

| Gap | Local coverage | CI coverage |
| --- | --- | --- |
| Custom `lint:*` scripts (innerHTML, skills, patches, ui-in-providers, raw-chrome-runtime-id, hosted-origin) | ❌ none | ✅ `lint:ci` |
| Complexity boy-scout gate (`check-touched-exemptions.mjs`) | ❌ none (docs call it "easy to miss locally") | ✅ dedicated step |
| knip dead-code (default + `--production`) | ❌ none | ✅ dedicated steps |
| CWS manifest justifications | ❌ none | ✅ dedicated step |
| **Strictness divergence** — `lint:skills` | `npm run lint` uses **non-strict** | CI uses `--strict` |
| **Scope divergence** — biome/prettier | lint-staged touches **staged files only** | CI scans the **whole repo** (`.`) |

Two subtler consequences of the scope/strictness divergence:

- A file committed with a formatting/lint issue via `git commit --no-verify`, or a file type lint-staged doesn't route, is invisible until the CI whole-repo scan.
- Even a diligent contributor running `npm run lint` locally **does not** reproduce CI, because `lint` uses `--write`/non-strict while `lint:ci` uses check-mode/`--strict`, and `lint` omits the complexity gate + knip + manifest steps entirely.

So: **there is no single local step that reproduces the CI `lint` job.** `docs/verification.md` documents the manual pass, but nothing enforces or nudges it before a PR goes red.

## Options to discuss

- **(A) Automatic pre-push hook** — extend `.husky/pre-push` to run the full CI lint gate (`lint:ci` + complexity + knip) before every push. Strongest guarantee; adds seconds-to-minutes per push.
- **(B) One convenient command** — a single `npm run verify` / `lint:local` that exactly mirrors CI's `lint` job, run manually or via a git alias. Zero forced latency; relies on discipline.
- **(C) Both** — the command plus a lightweight pre-push hook with a `--no-verify` escape hatch, scoped to only-changed packages for speed.

Open questions for reviewers:

1. Acceptable added latency on `git push`? (drives A vs B vs C)
2. Should a local gate scope to changed packages/files for speed, or always run whole-repo to match CI exactly?
3. Do we want `npm run lint` itself to converge toward `lint:ci` (single source of truth), or keep the write/check split and add a separate mirror command?

## References

- `.github/workflows/ci.yml` — `lint` job
- `package.json` — `lint`, `lint:ci`, `lint:*`, `deadcode*` scripts
- `.husky/pre-commit`, `.husky/pre-push`
- `docs/verification.md` — documented manual pass

catalan-adobe:
## Summary

Closes #1577. The local git hooks now mirror the CI `lint` job so lint failures surface before the push, not after a red PR.

## What changed

**New: `packages/dev-tools/tools/pre-push-lint-gate.sh`**
Parallelized script running all CI lint checks (~6 s wall-clock vs ~15 s sequential):

| Group | Checks |
|-------|--------|
| `biome` | `biome check .` (whole repo) |
| `prettier` | `prettier --check .` (whole repo) |
| `custom-lints` | lint:docs, lint:skills --strict, lint:skill-router, lint:no-innerhtml, lint:no-ui-in-providers, lint:patches, lint:no-raw-chrome-runtime-id, lint:hosted-origin |
| `complexity` | Boy-scout complexity gate |
| `manifest` | CWS permission justifications |
| `deadcode` | knip default gate |
| `deadcode-prod` | knip production files gate |

**`.husky/pre-push`** — calls the gate for feature branches (skips main).

**`package.json`** — adds `npm run verify` for on-demand use.

**`docs/verification.md`** — documents the automatic gate and updates the standard pass.

Escape hatch: `git push --no-verify`.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -12,10 +12,10 @@ set -e
 # Escape hatch: `git push --no-verify`.
 
 DEFAULT_BRANCH="main"
-BASE_REF="origin/$DEFAULT_BRANCH"
+export BASE_REF="origin/$DEFAULT_BRANCH"
 if ! git rev-parse --verify --quiet "$BASE_REF" >/dev/null 2>&1; then
   # Fresh clone without an origin/main ref — fall back to the local branch.
-  BASE_REF="$DEFAULT_BRANCH"
+  export BASE_REF="$DEFAULT_BRANCH"
 fi
 
 # Track whether we're pushing a feature branch (non-main). The lint gate only
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -18,6 +18,9 @@ if ! git rev-parse --verify --quiet "$BASE_REF" >/dev/null 2>&1; then
   BASE_REF="$DEFAULT_BRANCH"
 fi
 
+# Track whether we're pushing a feature branch (non-main). The lint gate only
+# runs for feature branches — pushes to main itself are gated by the merge queue.
+pushing_feature=false
 status=0
 while read -r local_ref local_oid remote_ref remote_oid; do
   # Skip branch deletions (local oid is all zeros).
@@ -30,9 +33,18 @@ while read -r local_ref local_oid remote_ref remote_oid; do
   if [ "$remote_ref" = "refs/heads/$DEFAULT_BRANCH" ]; then
     continue
   fi
+  pushing_feature=true
   if ! bash packages/dev-tools/tools/check-linear-history.sh "$BASE_REF" "$local_oid"; then
     status=1
   fi
 done
 
-exit "$status"
+[ "$status" -ne 0 ] && exit "$status"
+
+# ── Lint gate ────────────────────────────────────────────────────────────────
+# Mirror the CI `lint` job locally so failures surface before the push, not
+# after a red PR. Only runs for feature branches.
+# Escape hatch: `git push --no-verify`.
+if [ "$pushing_feature" = true ]; then
+  bash packages/dev-tools/tools/pre-push-lint-gate.sh
+fi
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -33,7 +33,10 @@ while read -r local_ref local_oid remote_ref remote_oid; do
   if [ "$remote_ref" = "refs/heads/$DEFAULT_BRANCH" ]; then
     continue
   fi
-  pushing_feature=true
+  # Only feature branches trigger the lint gate — skip tags and other refs.
+  case "$remote_ref" in
+    refs/heads/*) pushing_feature=true ;;
+  esac
   pushed_oid="$local_oid"
   if ! bash packages/dev-tools/tools/check-linear-history.sh "$BASE_REF" "$local_oid"; then
     status=1
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -34,6 +34,7 @@ while read -r local_ref local_oid remote_ref remote_oid; do
     continue
   fi
   pushing_feature=true
+  pushed_oid="$local_oid"
   if ! bash packages/dev-tools/tools/check-linear-history.sh "$BASE_REF" "$local_oid"; then
     status=1
   fi
@@ -46,5 +47,20 @@ done
 # after a red PR. Only runs for feature branches.
 # Escape hatch: `git push --no-verify`.
 if [ "$pushing_feature" = true ]; then
+  # Ensure the working tree matches the pushed commit. Without this, the gate
+  # could pass against uncommitted fixes while the pushed commit still has
+  # lint failures, or fail due to unrelated uncommitted changes.
+  head_oid="$(git rev-parse HEAD)"
+  if [ "$pushed_oid" != "$head_oid" ]; then
+    echo "pre-push: pushed OID ($pushed_oid) differs from HEAD ($head_oid)." >&2
+    echo "Check out the branch you are pushing, or use git push --no-verify." >&2
+    exit 1
+  fi
+  if ! git diff --quiet HEAD; then
+    echo "pre-push: working tree has uncommitted changes." >&2
+    echo "Commit or stash them so the lint gate checks what will actually be pushed." >&2
+    exit 1
+  fi
+
   bash packages/dev-tools/tools/pre-push-lint-gate.sh
 fi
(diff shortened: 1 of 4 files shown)


catalan-adobe:
### Strengths

- **[packages/dev-tools/tools/pre-push-lint-gate.sh:41-64]** — The seven check groups accurately cover the CI lint job, including strict skill linting, both knip passes, manifest validation, and the separate complexity gate.
- **[packages/dev-tools/tools/pre-push-lint-gate.sh:67-91]** — Background failures are explicitly collected and reported rather than being lost under parallel execution. Waiting for every PID also gives developers all failures in one pass.
- **[.husky/pre-push:25-42]** — Existing deletion, main-branch, and linear-history behavior is preserved, with lint only starting after linear-history checks succeed.
- **[package.json:92]** and **[docs/verification.md:7-30]** — The on-demand `verify` entry point and updated standard verification instructions make the gate discoverable without adding dependencies.

### Issues

*(none found)*

- **[.husky/pre-push:23-49; packages/dev-tools/tools/pre-push-lint-gate.sh:12-64]** — The hook discards the pushed `local_oid` and runs every check against the current working tree/current `HEAD`, which may not be the content being pushed.
  - A developer can commit lint-invalid code, fix it without committing, and then push: the gate sees the fixed working tree while the invalid commit is sent. Conversely, unrelated uncommitted changes can block a valid push. `git push origin other-branch` similarly validates the checked-out branch rather than `other-branch`.
  - This undermines the gate’s main purpose and is especially important because the hook already receives the exact pushed OIDs.
  - Validate each distinct feature-branch OID in a temporary detached worktree, or explicitly require one pushed OID equal to `HEAD` and a clean tracked working tree before running the gate. Add tests for dirty worktrees and pushing a non-checked-out branch.

- **[packages/dev-tools/tools/pre-push-lint-gate.sh:42-43]** — Bare `npx` can fetch and execute unpinned registry packages when the local binaries are absent.
  - On a fresh or partially installed checkout, the automatic hook may prompt, access the network, or execute a package version that is not the lockfile-pinned CI version. In particular, `npx biome` may resolve the registry package named `biome`, rather than the repository’s `@biomejs/biome` dependency.
  - Fail clearly when dependencies are missing and execute the pinned binaries directly:
    ```bash
    [[ -x node_modules/.bin/biome && -x node_modules/.bin/prettier ]] || {
      echo "Run npm ci before npm run verify" >&2
      exit 2
    }
    run_check biome ./node_modules/.bin/biome check .
    run_check prettier ./node_modules/.bin/prettier --check .
    ```

- **[.husky/pre-push:15-18,49; packages/dev-tools/tools/pre-push-lint-gate.sh:59]** — The hook’s base-ref fallback is not propagated to the complexity check.
  - `BASE_REF` is only a shell variable and the gate is invoked without an argument. Consequently, `check-touched-exemptions.mjs:64-68` falls back independently to `origin/main`. In the exact fresh/custom-remote scenario handled by lines 15-18, linear-history can use local `main`, but the subsequent complexity check fails because `origin/main` is absent.
  - Pass the selected base explicitly or export it for the child:
    ```sh
    BASE_REF="$BASE_REF" bash packages/dev-tools/tools/pre-push-lint-gate.sh
    ```
    Preferably, make the gate accept an explicit base argument and pass that to `check-touched-exemptions.mjs`.

#### 💡 Minor (Nice to Have)

- **[.husky/pre-push:31-36]** — Every non-main ref, including tags, sets `pushing_feature=true`.
  - `git push --tags` therefore runs the full gate even though no feature branch is being pushed, contrary to the documented behavior.
  - Only set the flag for `refs/heads/*` refs.

- **[packages/dev-tools/tools/pre-push-lint-gate.sh:17,35]** — The exit trap removes logs but does not terminate outstanding background checks.
  - Interrupting the hook can leave expensive scanners running after the push is cancelled, particularly because non-interactive background jobs may ignore terminal `SIGINT`.
  - Add `INT`/`TERM` cleanup that kills and waits for recorded PIDs before removing the temporary directory.

### Recommendations

Add a lightweight integration test using a temporary Git repository and stub executables. It should cover successful and failed aggregation, main/deletion/tag handling, missing dependencies, dirty working trees, alternate pushed refs, and base-ref fallback. This would catch the ref-context bugs without running the real scanners.

### Assessment

**Ready to merge?** No

**Reasoning:** The check coverage and failure aggregation are solid, but the gate currently validates the working tree rather than the commits being pushed, allowing the exact lint failures it is intended to prevent to reach CI. The unpinned `npx` execution and lost base-ref fallback should also be corrected before enabling this automatically for contributors.

---

PR #1692, opened 2026-07-27
fix(transcript-export): delegate approval when the leader is headless

catalan-adobe:
## Problem

Follower transcript export never completed against a cloud (hosted-leader) cone.

The approval gate renders a `<slicc-dialog>` into the **leader tab's** DOM. In the hosted-leader float that tab is headless Chromium inside an e2b sandbox — no human, no VNC, no screencast (the cloud dashboard is session management only; the user's own browser joins as a *follower*). The approval promise never settled, so `handleTranscriptExportRequest` sent `transcript.export.pending` and then stopped: it never denied and never built the ZIP.

The failure was sticky, so a retry behaved differently from the first attempt:

- the follower's `exportInFlight` only resets in a `.finally()` that never ran → menu stuck on "Exporting…" until reload;
- the leader's per-follower in-flight cap then auto-denied every subsequent request.

Neither side had a timeout, so nothing ever surfaced an error.

## Fix

A leader started with `kind: 'hosted'` delegates the approval prompt to the **requesting follower**, whose user is the only human in the session:

```
follower  ──transcript.export.request──────────▶  headless leader
follower  ◀─transcript.export.approve.request──  headless leader
  (renders the same dialog locally)
follower  ──transcript.export.approve.response─▶  headless leader
follower  ◀─transcript.export.start / chunks───  headless leader
```

The gate is **delegated, not skipped** — the same one-time `openTranscriptExportApproval()` dialog runs, just in the realm where a human can answer it. It still needs a genuine click; neither the agent nor a Cherry host page can forge it, since SLICC owns that iframe's DOM.

Every failure path denies: no reply within 120 s, follower disconnect, cancel, an unwired handler, a dialog that throws, or a prompt that could not be sent. The leader therefore **always** sends a terminal response, so a follower can no longer hang.

**All other floats are unchanged** and keep the leader-side dialog — covered by explicit regression tests.

## Protocol

Adds `transcript.export.approve.request` (L→F) and `transcript.export.approve.response` (F→L); `TRAY_SYNC_PROTOCOL_VERSION` 3 → 4. Mirrored in the golden-fixture corpus, `SyncProtocol.swift`, and the Go CLI constant. iOS never originates an export, so it is never asked to approve one (`.unknown` / `undecodable`).

Also fixes `buildCorpusDocument()` pinning `traySyncProtocolVersion` to the literal `3` instead of the constant — the Swift suite asserts these match, so a hand-pinned copy would have passed a stale version through any bump.

## UX detail

The dialog names the actual recipient of the bytes, which differs by realm: "The follower will receive" (leader-side), "This device will download" (delegated), or "The embedding page will receive" (delegated + Cherry, where the ZIP leaves SLICC for the host page).

## Tests

16 new tests in `tray-transcript-export-delegated-approval.test.ts`: delegation vs. local dialog, approve/deny, timeout, in-flight slot release, unknown/late `requestId`, disconnect mid-approval, unsendable prompt, follower fail-closed paths, and a cross-wired end-to-end cloud run producing a verified ZIP Blob.

Both directions were mutation-checked: restoring the original bug fails 9 tests, and flipping the follower's fail-closed check to fail-open fails 1.

## Verification

`npm run verify` (7/7) · `check-touched-exemptions` · `typecheck` · `test` (12384 passed) · `test:coverage` · `build` · extension build · `go test ./internal/protocol`.

## Note

`session export` in the cone's shell has no approval gate and already worked in cloud — that remains the agent-side path.


catalan-adobe:
### Strengths

- **[packages/webapp/src/scoops/tray-leader-sync.ts:1763]** — Delegated approvals are scoped by follower identity plus `requestId`, preventing one follower from approving another follower’s export.
- **[packages/webapp/src/scoops/tray-leader-sync.ts:1765]** — The 120-second timeout and immediate handling of failed sends provide a bounded, fail-closed replacement for the original indefinite hang.
- **[packages/webapp/src/scoops/tray-leader-sync.ts:1732]** — Cancellation, disconnect cleanup, and late-response handling are idempotent and correctly release the per-follower export slot.
- **[packages/webapp/src/scoops/tray-follower-sync.ts:1457]** — Missing or throwing approval handlers default to denial rather than silently authorizing an export.
- **[packages/webapp/src/scoops/tray-sync-protocol-corpus.ts:511]** — Deriving the corpus protocol version from `TRAY_SYNC_PROTOCOL_VERSION` eliminates the stale hand-pinned version problem.
- **[packages/webapp/tests/scoops/tray-transcript-export-delegated-approval.test.ts:133]** — The new suite covers delegation, local-dialog regression behavior, timeout, slot release, owner scoping, send failure, disconnect, and a cross-wired verified-Blob flow.
- **[packages/webapp/src/ui/wc/wc-transcript-export.ts:143]** — Recipient-specific consent language is a good privacy improvement and clearly distinguishes local downloads from Cherry host-page delivery.

### Issues

*(none found)*

- **[packages/webapp/src/scoops/tray-follower-sync.ts:1457]** — A delegated approval dialog is not tied to the export request’s lifecycle.
  - The handler invokes the dialog without first verifying that `requestId` is still in `activeExportRequests`, and it has no cancellation signal. If Cherry aborts while the dialog is open, or the leader’s 120-second timeout sends `transcript.export.denied`, the export Promise settles but the dialog remains visible. A retry can then open another dialog; clicking **Allow** on the stale dialog sends a late response that the leader ignores, making the user believe the retry was approved.
  - This also permits a prompt already in transit to render after the local request was cancelled.
  - Suggested fix: correlate prompts with `activeExportRequests`, track a per-request approval `AbortController`, and abort it from denial/error/disconnect/cancel cleanup. Extend `openTranscriptExportApproval` to accept a signal that hides/removes the dialog and resolves `false` on abort. Add tests for timeout/cancel while the dialog Promise remains unresolved and for retry without stacked prompts.

- **[packages/webapp/src/ui/wc/wc-follower.ts:501]** — Cherry approval copy can misidentify the recipient outside Chromium/WebKit.
  - `ancestorOrigin` comes directly from the non-standard `location.ancestorOrigins`. In Firefox it is absent, while Cherry can still connect using the `document.referrer` fallback in `main-cherry.ts`. The approval then omits `hostOrigin`, causing `wc-transcript-export.ts:147-151` to say **“This device will download”**, even though the verified Blob is actually returned to the embedding page.
  - This weakens the explicit recipient disclosure the PR intends at the consent boundary.
  - Suggested fix: expose the trusted parent origin already resolved by `setupCherryFollower()` through the prelude and use it here. Alternatively, pass an explicit recipient kind such as `'embedding-page'` independently of whether an origin string is available.

#### 💡 Minor (Nice to Have)

*(none found)*

### Recommendations

- Add UI-level tests around delegated-dialog lifecycle, not only protocol verdicts: leader timeout, Cherry abort, disconnect, late click, and immediate retry.
- Reuse one canonical parent-origin result for Cherry handshake validation, target advertisement, and approval disclosure instead of independently reading `ancestorOrigins`.

### Assessment

**Reasoning:** The protocol and leader-side fail-closed behavior address the production hang well, with strong ownership and timeout coverage. The remaining lifecycle and recipient-disclosure issues should be fixed so stale consent dialogs cannot survive terminal outcomes and Cherry users are always told where the exported bytes go.

---

catalan-adobe:
Both Important findings addressed in 6271b58a4. Verified each against the codebase before implementing.

### 1. Delegated prompt not tied to the export lifecycle — confirmed, fixed

Reproduced the reasoning end-to-end: `handleExportDenied` (which is what the leader's 120 s timeout ultimately triggers), the local `onAbort`, and `handleDisconnect` all deleted the entry and settled the Promise while leaving the dialog on screen. A later **Allow** sent a verdict into `settleApproval`, whose waiter was already gone — a no-op that looks like consent.

Three changes:

- **Prompt only for a live request.** `handleExportApprovalRequest` now looks up `activeExportRequests` first; an unknown `requestId` is denied without rendering anything. This also covers @claude's non-blocking observation about unsolicited prompts from a misbehaving or version-skewed leader.
- **`AbortSignal` on the dialog.** A per-request `approvalAbort` is stored on the entry and fired by every terminal path (`handleExportDenied`, `handleExportError`, `onAbort`, `handleDisconnect`). `openTranscriptExportApproval` accepts the signal, and on abort hides *and removes* the element so no later click can reach it, resolving `false`. An already-aborted signal never opens a dialog at all.
- **Late verdicts are denials.** After the dialog resolves, the handler re-checks the abort state and the entry before replying, so a post-settlement click is reported as `approved: false`.

### 2. Cherry recipient disclosure — confirmed, fixed

`main-cherry.ts` `resolveParentOrigin()` is a 3-tier resolve (`ancestorOrigins` → `document.referrer` → same-origin), so reading `location.ancestorOrigins` again in `wc-follower.ts` was strictly weaker: in Firefox it is absent, Cherry still boots via the referrer, and the dialog would have said "This device will download" while the Blob went to the embedding page.

Took the "reuse one canonical result" recommendation: added `CherryHostTransport.hostOrigin`, which returns the `targetOrigin` the transport was constructed with at boot — the same value used for the handshake and every `postMessage`. The doc comment warns against re-deriving it.

### Tests

+6 protocol/lifecycle tests and +6 UI-level dialog tests, per the recommendation to cover the dialog lifecycle and not just protocol verdicts: leader timeout, Cherry abort, late click after abort, unsolicited prompt, already-aborted signal, and recipient copy across all three realms.

Mutation-checked all three new guards — removing the in-flight guard, the denial-path abort, or the dialog's signal listener each fails tests.

One thing worth flagging: adding the in-flight guard revealed that **four existing follower tests had become vacuous** — they simulated `approve.request` for requests that were never made, so they started passing via the new deny-early path without ever reaching the handler. They now perform a real request/prompt handshake and assert the callback was actually invoked.

Full pass green: `verify` (7/7) · `check-touched-exemptions` · `typecheck` · `test` (12395) · `test:coverage` · both builds.

PR #1696, opened 2026-07-27
fix(tray): treat an unresponsive peer as stalled, not dead

catalan-adobe:
## What this fixes

A follower connected to a cloud (hosted-leader) cone disconnects repeatedly, recovers, and
disconnects again. The user sees one of two cards, seemingly at random:

- `Connection to leader lost: Data channel closed`
- `Connection to leader lost: Keepalive timeout — leader not responding`

They are the same event. Both ends run `DataChannelKeepalive` (10 s ping, 3 missed = dead) on
the **same thread that serves the agent**, so a peer that is merely CPU-starved stops answering
long before its transport dies. Whichever watchdog fired first decided the wording:

| Fires first | Action | Card shown to the user |
| --- | --- | --- |
| Leader's | `removeFollower()` → `sync.close()` | `Data channel closed` |
| Follower's | `handleDisconnect()` → `sync.close()` | `Keepalive timeout` |

Both paths **close a healthy data channel** and force a full ICE/DTLS renegotiation. The
disconnect was self-inflicted: the transport was fine the whole time.

The hosted-leader float makes this routine. Chromium (browser + renderer + network/storage
utilities), the SLICC kernel worker (WASM shell, realms, OPFS) and node-server share one e2b
sandbox, and the template never set a CPU count — so it ran on e2b's default of 2 vCPUs. An
ordinary busy turn starves the leader's main thread well past 30 s.

## Changes

**1. `DataChannelKeepalive` — stall ≠ death.** Two thresholds instead of one:

- `maxMissed` (3, ~30 s) while `isTransportOpen()` holds → **stalled**. Logged once, probing
  continues, `onRecovered` fires when the peer answers. Nothing is torn down.
- `hardMaxMissed` (30, ~5 min) **or** a transport that is actually gone → **dead**, exactly as
  before.

`isTransportOpen` defaults to `() => false`, so any caller that omits it keeps the original
behavior.

**2. Both tray managers pass `isTransportOpen: () => sync.isOpen`** — the follower
(`tray-follower-sync.ts`) and the leader (`tray-leader-sync.ts`), since eviction closes the
channel on that side too.

**3. `cpuCount: 4` on the e2b template** (`SLICC_E2B_CPU_COUNT` to override). CPU allocation is
a template-build option in the e2b SDK (`BasicBuildOptions`, default 2) — there is no
`Sandbox.create` override, so it has to be set at build time. Takes effect on the next template
build.

## Tests

`npm run test` — 12381 passed. New coverage in `data-channel-keepalive.test.ts`:
stall instead of death, probing continues while stalled, stall reported once, recovery via pong
and via inbound ping, death at the hard deadline, death at `maxMissed` when the transport is
closed, and death when an open transport closes mid-stall.

The two existing tests that encoded the old behavior (`removes follower when keepalive declares
dead`, `calls onDisconnect when keepalive declares dead`) now assert the new contract, with the
death path retained against the hard deadline.

Mutation-checked: reverting the guard in `declareUnreachable` fails 10 of the new/updated tests;
restoring it passes all of them.

## Cross-runtime parity

`packages/ios-app/SliccFollower/Sync/Keepalive.swift` is a port of the same class and carries the
same behavior. **Deliberately excluded from this PR**: the wire protocol is untouched (this is
local timing policy only), so there is no protocol drift, and the Swift port belongs in a change
the iOS CI job can validate. Follow-up.

## Follow-ups (not in this PR)

- **UI split.** A stall is now invisible — the chat simply pauses until the leader catches up.
  It should show a transient "leader is busy" banner, distinct from the permanent
  `slicc-error-card` that a real disconnect appends to the transcript.
- **Follower diagnosability.** `slicc.followerTrayStatus` (attach attempts, last attach code,
  reconnect attempts) is mirrored to `localStorage` only by `wc-tray.ts`; the `/join/<token>`
  no-kernel mount in `wc-follower.ts` never writes it, so the float most users run has no
  connection telemetry. Combined with the prod log level being pinned to `ERROR`, this class of
  failure leaves no trace.
- **Unbounded tray messages.** `agent_event` (screenshots via `open --view`, untruncated shell
  tool results), `user_message` attachments, `cdp.event` and `sprinkle.update` are sent
  unchunked. Chrome 152 throws `TypeError: Trying to send message larger than max-message-size`
  above 262144 bytes; `TraySyncChannel.send` swallows it and returns `false`, which every
  non-chunked caller ignores — so followers silently miss those events. Only `snapshot`,
  `cdp.response`, `sprinkle.content` and `fs.response` chunk today.

## Verification

`npm run verify` · `check-touched-exemptions` · `npm run typecheck` · `npm run test` ·
`npm run test:coverage` · `npm run build` · `npm run build -w @slicc/chrome-extension` — all pass.


catalan-adobe:
### Strengths

- **`packages/webapp/src/scoops/data-channel-keepalive.ts:83-91,141-175`** — The two-threshold state machine is small and understandable. The `isTransportOpen` default preserves legacy behavior for callers that do not opt in.
- **`packages/webapp/src/scoops/tray-follower-sync.ts:264-283`** and **`packages/webapp/src/scoops/tray-leader-sync.ts:559-578`** — Both sides correctly provide live `sync.isOpen` callbacks, preventing either watchdog from tearing down a transport during a transient stall.
- **`packages/webapp/tests/scoops/data-channel-keepalive.test.ts:256-376`** — Tests cover the important transitions: single stall notification, continued probing, ping/pong recovery, hard deadline, closed transport, and closure during a stall.
- **`packages/webapp/tests/scoops/tray-follower-sync.test.ts:1304-1386`** and **`packages/webapp/tests/scoops/tray-leader-sync.test.ts:1299-1360`** — Manager-level tests verify the real externally visible contract rather than only internal counters.
- **`packages/dev-tools/e2b-template/template.ts:30-42,88-92`** — CPU allocation is applied at the correct template-build layer, with a conservative default and input validation.
- **`docs/architecture.md:408`** — The changed liveness policy and its rationale are documented clearly without changing the wire protocol.

### Issues

*(none found)*

*(none found)*

#### 💡 Minor (Nice to Have)

- **`packages/webapp/src/scoops/data-channel-keepalive.ts:147,161`** — `hardMaxMissed` is not independently enforced when configured below `maxMissed`.
  - `declareUnreachable()` is only invoked after `missedPongs >= maxMissed`. For example, `maxMissed: 60` and `hardMaxMissed: 30` declares death at 60 misses, contrary to the documented hard deadline.
  - Validate positive integer thresholds and require `hardMaxMissed >= maxMissed`, or check the hard threshold independently before the soft threshold.

- **`packages/webapp/src/scoops/data-channel-keepalive.ts:100-121,172-175`** — A stopped/dead keepalive can remain marked stalled and later emit a recovery callback.
  - At the hard deadline, `stop()` does not clear `stalled`. A subsequent `receivePing()` or `receivePong()` can therefore call `onRecovered` even though the state machine is terminal. Current tray cleanup makes this unlikely, but it is inconsistent with the reusable class contract and the `isStalled` getter.
  - Clear the stalled flag on terminal stop and ignore inbound liveness calls after `stopped` becomes true.

- **`packages/dev-tools/e2b-template/template.ts:38`** — The new `SLICC_E2B_CPU_COUNT` build control is not documented in the template README.
  - Operators reading `packages/dev-tools/e2b-template/README.md` can discover the template-name override but not the CPU override.
  - Add the default, valid input, cost implication, and an example build command to that README.

### Recommendations

- Prioritize the acknowledged stall-visibility follow-up. The new stall notices use `warn`/`info`, while production logging defaults to `ERROR`, so a multi-minute production stall remains operationally invisible. A bounded metric or follower runtime-status transition would provide observability without turning expected stalls into error logs.
- Add small boundary tests for contradictory threshold configuration and for inbound ping/pong after terminal stop.

### Assessment

**Ready to merge?** Yes

**Reasoning:** The implementation matches the planned liveness policy, preserves compatibility, and has strong unit and manager-level coverage. The remaining findings are defensive API-state and documentation improvements that do not affect the default production path.

---

catalan-adobe:
Thanks both — all three minor findings were real. Verified each against the code and fixed them in 716b586e9.

**1. `hardMaxMissed` below `maxMissed` silently never applied** — confirmed. `tick` only calls `declareUnreachable()` once `missedPongs >= maxMissed`, so with `maxMissed: 60, hardMaxMissed: 30` death lands at 60 — *later* than the configured hard deadline promises, not earlier. Fixed by failing fast at construction: `intervalMs`, `maxMissed` and `hardMaxMissed` must be positive integers and `hardMaxMissed >= maxMissed`, otherwise `RangeError`. Matches the existing precedent in `FollowerSyncManager`'s constructor (`sprinkleFetchTimeoutMs`). I chose validation over "check the hard threshold independently" because a hard deadline earlier than the soft one has no coherent meaning — better to reject it than to define it.

**2. Terminal state could still report a recovery** — confirmed. `stop()` left `stalled` set and `receivePong`/`receivePing` didn't check `stopped`, so a late pong after death fired `onRecovered` on a dead machine. `stop()` now clears `stalled` (without notifying) and both receive paths early-return once stopped. Latent rather than live today (both call sites `unsubscribe()` before/with `stop()`), but the class is reusable and exposes `isStalled`, so the contract should hold on its own.

**3. `SLICC_E2B_CPU_COUNT` undocumented** — fixed. New "Sandbox size" section in `packages/dev-tools/e2b-template/README.md` covering the 4 vCPU / 8192 MB pin, why 2 isn't enough, that `cpuCount` is **build-time only** (`BasicBuildOptions` — no `Sandbox.create` override, so changing it needs a republish), the positive-integer contract, an example command, and the per-cone cost implication.

**Boundary tests added** as recommended: contradictory thresholds, each non-positive-integer threshold (`it.each`), and inbound ping/pong after terminal stop. 23 tests in `data-channel-keepalive.test.ts`, 12386 across the suite.

---

One correction on the buffering note, since the conclusion is right but the mechanism isn't:

> This is already defended by `broadcastToAllFollowers`'s throttled error log and the SCTP `OperationError`/`InvalidStateError` swallow

`broadcastToAllFollowers`'s `try/catch` is actually **unreachable**: `TraySyncChannel.send` already catches every throw internally, logs, and returns `false` — it never propagates. So `followerBroadcastErrorLogAt` and `BROADCAST_ERROR_THROTTLE_MS` are dead state, and broadcast failures surface through the generic path without the `bootstrapId`. Pre-existing, not a regression from this PR, and the "conscious trade-off, not a regression" conclusion still stands — the swallow does absorb it, just one layer lower than described. Filed with the related size-limit work.

**Follow-ups filed** so they don't get lost, as both reviews asked:
- #1698 — a multi-minute stall is operationally invisible (the sharpest point in the review: this PR *introduces* that silence, and raising the stall log to `error` is the wrong fix)
- #1699 — the `/join/<token>` no-kernel follower writes no connection telemetry
- #1700 — oversize tray messages silently dropped, incl. the dead-code note above

Full pass re-run green: `verify` · `check-touched-exemptions` · `typecheck` · `test` (12386) · `test:coverage` · `build` · `build -w @slicc/chrome-extension`.

catalan-adobe:
Reconsidered — you're right on two of the three. Pushed in 2e783cd1f.

**#1698 (stall is invisible) — now fixed here.** This one I introduced: before the split you got a wrong red card, after it you got silence. Deferring a regression I created was the wrong call.

- `FollowerSyncManager` gains `onLeaderStalled(stalled)` and marks a `stalled` overlay on the follower runtime status. Overlay, not a `state` variant — the channel never dropped, so the follower is still `connected`, and `getFollowerStatusWithFallback`'s `!== 'inactive'` check keeps working untouched. Cleared when the connection really ends.
- `page-follower-tray` forwards it as its own option rather than reusing `onConnectionChange`, so a cherry host never receives `slicc.follower.disconnected` for a peer that is merely working.
- `wc-follower` renders `The leader is busy — hang on…`, disabled so a message typed mid-stall can't be silently dropped.

Per the review's own guidance, the stall log stays at `warn`/`info` — the visibility comes from status + UI, not from promoting an expected event to `error`.

**#1699 (no follower telemetry) — also fixed here.** `wc-follower` now mirrors the tray status into `localStorage` exactly as `wc-tray.ts:593` does. Marginal cost was ~10 lines in a file this commit already touches, and it's the difference between a diagnosable disconnect and a silent one on the float most people actually run. Both issues closed by the commit.

**#1700 (oversize messages silently dropped) — still deferred, deliberately.** Different call, and I'd rather defend it than be consistent for its own sake:

- It isn't caused or worsened by this PR. It predates it and is orthogonal to liveness.
- The fix belongs at `TraySyncChannel.send` and changes the send path for **every** message type on both directions — generic chunking (a reassembly buffer, ordering, and a cancellation story per type) or fail-loud semantics that ~15 currently-ignoring call sites would have to start handling.
- That needs its own test matrix and its own review. Bolting it onto a keepalive PR would mean a reviewer signing off on two unrelated risk profiles at once, and if it regresses the tray, bisecting lands on a commit that also changed disconnect behavior.

It's specified in #1700 with the measured Chrome 152 numbers and the dead-code finding, and I'm happy to take it next as its own PR.

Re-verified green: `verify` · `check-touched-exemptions` · `typecheck` · `test` (12391) · `test:coverage` · `build` · `build -w @slicc/chrome-extension`.

PR #1709, opened 2026-07-27
fix(tray): chunk oversize messages instead of dropping them silently

### LINKED ISSUE #1700: Tray: oversize messages are silently dropped (agent_event, attachments, cdp.event)
Follow-up to #1696.

Measured in Chrome 152 (loopback `RTCDataChannel`, the browser the floats actually run):

| case | result |
| --- | --- |
| `sctp.maxMessageSize` | 262144 |
| `send()` 262145 B | `TypeError: Trying to send message larger than max-message-size` — **channel stays open**, message dropped |
| ~16.6 MB buffered | `OperationError: RTCDataChannel send queue is full` — also non-fatal, dropped |

`TraySyncChannel.send` (`tray-sync-protocol.ts`) catches that throw, logs, and returns `false`. **Every non-chunked caller ignores the return value**, so the message is simply lost.

Only 4 message types chunk today: `snapshot`, `cdp.response` (`tray-sync-protocol.ts`), `sprinkle.content` (32 KB), `fs.response` (64 KB). Unbounded and unchunked:

- **`agent_event`** (`tray-leader-sync.ts` `broadcastEvent`) — carries `tool_result.result` (untruncated: `tools/bash-tool.ts` returns `stdout + stderr` in full), `tool_use_start.toolInput`, `tool_ui.html`. `open --view --size high` inlines a base64 image into shell stdout (`open-command.ts`, 1536 px), so it routinely exceeds 256 KiB — **followers silently never see screenshots or large tool results**.
- `user_message` / `user_message_echo` with inline attachments
- `cdp.event`, `sprinkle.update`, `lick`

Fix belongs at the chokepoint (`TraySyncChannel.send`): consult `RTCSctpTransport.maxMessageSize` (nothing in the repo reads it today) and either chunk generically or fail loudly instead of silently.

Related dead code found while checking this: `broadcastToAllFollowers`'s `try/catch` is unreachable, because `TraySyncChannel.send` already swallows every throw and returns `false`. So `followerBroadcastErrorLogAt` and `BROADCAST_ERROR_THROTTLE_MS` are dead state, and broadcast failures are logged by the generic path without the `bootstrapId`.

catalan-adobe:
Closes #1700.

## What this fixes

`TraySyncChannel.send` handed every message straight to `RTCDataChannel.send()`. Chrome throws
`TypeError: Trying to send message larger than max-message-size` above `sctp.maxMessageSize`
(262144 in Chrome 152) and `OperationError: send queue is full` at ~16 MB buffered. `send` caught
both, logged, and returned `false` — **which every non-chunked caller ignored**.

Only 4 message types chunked: `snapshot`, `cdp.response`, `sprinkle.content`, `fs.response`.
Everything else was unbounded, including `agent_event`, which carries untruncated `tool_result`
output and the base64 image `open --view --size high` inlines into shell stdout. Followers
silently never saw screenshots or large tool results.

Nothing in the repo read `RTCSctpTransport.maxMessageSize`.

## Approach

A transport frame, `__chunk` (`TrayChunkFrame`), that is deliberately **not** a member of
`LeaderToFollowerMessage` / `FollowerToLeaderMessage`. It sits one layer below them: senders split
an oversize serialized message into frames, receivers reassemble and only then decode the union.

Each runtime already had exactly one inbound and one outbound chokepoint, so interception is
small and local:

| runtime | outbound | inbound |
| --- | --- | --- |
| TS | `TraySyncChannel.send` | `TraySyncChannel` ctor listener |
| Go | `Conn.SendJSON` | `Conn.dispatch` |
| Swift | `AppState.sendToLeader` | `AppState.handleDataChannelMessage` |

Because a frame is not a union variant it needs **no golden-corpus fixture and no per-variant iOS
decode expectation**, handler switches never see it, and every message type is covered — including
ones not written yet. The bug was a *class* of bug, not five instances of one.

The real limit is now consulted: `sctp.maxMessageSize` is plumbed onto the channel at the two
points one enters `tray-webrtc.ts`, falling back to the RFC 8831 §6.6 floor (65536) when the
transport hasn't negotiated yet.

### The loud floor

Chunking doesn't fix the second failure mode, so `send` refuses **before** it can throw:

| Bound | Value | Behaviour |
| --- | --- | --- |
| `TRAY_MAX_MESSAGE_BYTES` | 8 MiB | Refuse rather than queue hundreds of frames |
| `TRAY_SEND_HIGH_WATER_BYTES` | 8 MiB queued | Refuse a chunked send on a congested channel |

The high-water check is deliberately **not** applied to small messages, so keepalive ping/pong
still flows under congestion — otherwise this would re-create the "busy peer looks dead" failure
#1696 just fixed.

A refused `agent_event` is retried as a marker event (`degradeOversizeAgentEvent`), so the
follower's transcript shows the gap instead of hiding it. `screenshot` degrades to an `error`
event rather than a stubbed `base64` field, which would render as a broken image.

### Dead code

`broadcastToAllFollowers` wrapped `follower.sync.send` in a try/catch that could never run —
`send` already caught everything and reported through its return value. The throttle state it
guarded (`followerBroadcastErrorLogAt`, `BROADCAST_ERROR_THROTTLE_MS`) was dead, and real failures
logged through the generic path without a `bootstrapId`. It now reads the return value, so
failures are both actionable and attributable.

## Notes for review

- **`FollowerSyncManager.swift` is documented dead code** (`packages/ios-app/CLAUDE.md`, zero
  instantiation sites). The live path is `AppState`, which is where this landed. The framing and
  eviction rules were extracted to `TrayChunkReassembler` so they're testable without standing up
  an app, a peer connection, or a data channel.
- **Byte-accurate sizing.** The four pre-existing chunkers compare `.length` (UTF-16 units)
  against a byte budget, under-counting CJK 3×. The new code measures UTF-8 bytes; Go and Swift
  slice at rune/character boundaries.
- **`AppState`'s big switch moved to `routeLeaderMessage` untouched**, keeping its cyclomatic
  complexity (32) and body length (130) at their pre-existing baseline rather than pushing them
  higher. SwiftLint warning count is unchanged at 39.
- **The four bespoke chunkers are left in place.** They're correct and tested, and removing
  `snapshot_chunk` et al. from the wire is a second breaking protocol change across three
  runtimes — out of scope here. Happy to file the consolidation.

## Tests

`packages/webapp/tests/scoops/tray-sync-chunking.test.ts` (21) — framing size guarantees against a
channel that enforces the real limits, byte-vs-UTF-16 sizing, round-trip fidelity, reassembly
(out-of-order, duplicate, partial, concurrent, evicted, malformed, close).
`tray-leader-sync-oversize.test.ts` (8) — degradation per variant, per-follower degradation, intact
delivery via framing. `packages/slicc-cli/internal/tray/conn_test.go` (12) and
`packages/ios-app/.../ChunkFramingTests.swift` (12) mirror the same contracts.

Mutation-checked: reverting the chunking path fails 8 tests; `utf8ByteLength → .length` fails the
UTF-8 sizing test; dropping Go's rune-boundary walk-back fails the boundary test.

A test caught a genuine cross-runtime inconsistency during development: TS delivered a malformed
`__chunk` to handlers as a message while Go and Swift dropped it. TS now drops it too.

## Verification

`npm run verify` (7/7) · `check-touched-exemptions` (0 on debt lists) · `npm run typecheck` ·
`npm run test` (12457 passed) · `npm run test:coverage` (exit 0) · `npm run build` ·
`npm run build -w @slicc/chrome-extension` · `go build` / `vet` / `test` · `gofmt` clean ·
SwiftLint 39 = baseline. Re-run after rebasing onto current `main` (which now includes #1696).

**iOS tests have not been executed locally** — this machine has no installed simulator runtimes
(`xcrun simctl list runtimes` is empty), so CI's `xcodebuild test` job is their first real run.
They were verified with `swiftc -typecheck` and `-parse`.


catalan-adobe:
Both review findings fixed in `23f5a0ec` (rebased onto current `main`, which now includes the `oauth.ts` PKCE refactor).

**1. TS reassembler per-message cap** — confirmed: Go (`conn.go:472`) and Swift (`TrayChunkFraming.swift:137`) capped per message, TS bounded only the 32 MiB aggregate. Worse than asymmetric — the TS *receiver* would accept four times what the TS *sender* will emit. TS now caps at `TRAY_MAX_MESSAGE_BYTES` too.

**2. Go duplicate-frame sentinel** — confirmed and slightly worse than "theoretical". The empty-string sentinel doesn't stall the message; it *over-counts*: a duplicated empty frame advances the arrival count twice, so an N-frame message can complete with a hole in it and emit truncated JSON. Now tracked with an explicit `seen []bool`, matching TS/Swift.

Both fixes have regression tests, and both were mutation-checked — which caught that my **first attempt at each test passed for the wrong reason**:

- the TS test flooded 400 frames of `'x'`, which never forms valid JSON, so it was failing at `JSON.parse` rather than at the cap. Rewritten to frame a *valid* >8 MiB message the way a foreign peer could send it.
- the Go test sent a single empty frame, which works fine under either implementation. Rewritten to send a *duplicated* empty frame, which is where the sentinel actually breaks.

With the fixes reverted, each test now fails; with them restored, all pass.

Re-verified on the new base: `verify` 7/7 · `typecheck` · `test` (12506) · `test:coverage` (exit 0) · both builds · `go build`/`vet`/`gofmt` · `check-touched-exemptions` OK.

catalan-adobe:
### Strengths

- **[packages/shared-ts/src/tray-sync-protocol.ts:115-148]** — Modeling `__chunk` below the semantic message unions is a sound architectural choice. It centralizes the fix at transport chokepoints and automatically covers future message variants.
- **[packages/webapp/src/scoops/tray-sync-protocol.ts:491-526]** — The TypeScript path measures serialized UTF-8 bytes and consults the negotiated SCTP limit, with a conservative fallback.
- **[packages/webapp/src/scoops/tray-sync-protocol.ts:556-588]** — The 8 MiB hard cap, congestion refusal, and exemption for small keepalive messages directly address both reported failure modes without recreating the busy-peer disconnect problem.
- **[packages/webapp/src/scoops/tray-leader-sync.ts:773-826]** — Broadcast failures now use the actual `send()` result, include `bootstrapId`, and selectively retry degradable agent events. This is substantially more observable than the previous ineffective `try/catch`.
- **[packages/slicc-cli/internal/tray/conn.go:533-555]** — Go propagates framing and transport failures with the failed frame index instead of silently continuing.
- **[packages/webapp/tests/scoops/tray-sync-chunking.test.ts:242-347]** — The tests cover meaningful protocol behavior: duplicates, out-of-order delivery, concurrent streams, incomplete messages, eviction, oversize input, and malformed frames.

### Issues

*(none found)*

- **[packages/webapp/src/scoops/tray-sync-protocol.ts:327-346]** — TypeScript can split an astral Unicode character between its UTF-16 surrogates, corrupting messages received by Go or Swift.
  - At the default limit, the boundary is 16,256 UTF-16 units. If an emoji such as `🍦` straddles that boundary, the two frame JSON documents contain separate `\ud83c` and `\udf66` escapes. JavaScript preserves and rejoins those lone surrogates, so the local round-trip test passes, but Go’s `encoding/json` decodes each unmatched surrogate as U+FFFD; Swift may reject it. A valid oversized tool result can therefore arrive as corrupted text or never complete.
  - Build slices first while ensuring the cut never falls between a high and low surrogate, then assign `totalChunks`. Add a test with an emoji deliberately positioned across the exact frame boundary and pass each encoded frame through a non-JavaScript decoder.

- **[packages/slicc-cli/internal/tray/conn.go:453-477]** — Frame metadata is not bounded or kept consistent for an existing `chunkId`, allowing a peer to crash or exhaust a receiver.
  - A first frame with `totalChunks: 2, chunkIndex: 0`, followed by the same `chunkId` with `totalChunks: 100, chunkIndex: 99`, passes validation and panics when indexing `entry.seen[99]`. A single frame with `totalChunks: 1_000_000_000` attempts enormous allocations at lines 466-467. Swift has the same eager-allocation risk at **[packages/ios-app/SliccFollower/Models/TrayChunkFraming.swift:109-126]**, while the shared TS guard at **[packages/shared-ts/src/tray-sync-protocol.ts:186-200]** also accepts arbitrary counts.
  - Define a conservative maximum frame count derived from the 8 MiB cap, reject it before allocation, and require every subsequent frame’s `totalChunks` to equal the buffer’s original value before indexing. Add mismatch and excessive-count tests in all three runtimes.

- **[packages/ios-app/SliccFollower/App/AppState.swift:949-963]** — The iOS sender ignores every `sendData` result, including failures partway through a framed message.
  - `WebRTCManager.sendData` returns `false` when the channel is unavailable or rejects a write. If the send queue fills at frame N, `sendToLeader` silently continues with later frames; the leader retains an incomplete reassembly and the user receives no indication that their large pasted message or response was lost. This recreates the silent-drop behavior the PR is intended to remove.
  - Check every return value, stop after the first failure, and log/surface the message type and frame index. Encoding failures currently hidden by `try?` should also be logged. Ideally return a structured send result so user-facing sends can update `lastError`.

#### 💡 Minor (Nice to Have)

- **[packages/ios-app/SliccFollower/Models/TrayChunkFraming.swift:51-59]** — A Swift `Character` can itself exceed the frame budget.
  - Extended grapheme clusters can contain arbitrarily many combining scalars. When such a character is larger than `budget` and `current` is empty, it is appended whole, producing a frame over 65,536 bytes.
  - Iterate Unicode scalars rather than grapheme clusters—each scalar is at most four UTF-8 bytes—and retain the encoded-frame size assertion in tests.

- **[packages/ios-app/SliccFollower/Models/SyncProtocol.swift:153-156]** — The framing comment says dead `FollowerSyncManager.handleMessage` performs reassembly, while the live implementation is `AppState.handleDataChannelMessage`.
  - Update the reference so future protocol work is directed to the actual chokepoint.

### Recommendations

- Add cross-runtime wire vectors that encode individual frames in TS and decode/reassemble them in Go and Swift. Concatenating `chunkData` before frame serialization cannot detect surrogate-handling differences.
- Centralize protocol invariants across implementations: maximum frame count, stable `totalChunks`, per-message bytes, aggregate in-flight bytes, and malformed-frame rejection.
- Add explicit tests for receiver metadata mismatch, excessive `totalChunks`, Swift transport refusal, and Unicode exactly at frame boundaries.

### Assessment

**Ready to merge?** No

**Reasoning:** The overall transport-layer design is strong, but valid Unicode can be corrupted across runtimes, malformed frame metadata can crash or exhaust followers, and iOS still silently ignores transport failures. These should be corrected before relying on framing for lossless delivery.

---

catalan-adobe:
Rebased onto current `main` (64 commits, including the `decompose-tray-leader-sync` refactor) and addressed all five review findings. Each was verified before fixing — all five held up, and two were worse than reported.

## Conflicts

#1708 decomposed `tray-leader-sync.ts` while this branch was open, moving `broadcastToAllFollowers` to `tray-leader/follower-registry.ts` and `broadcastEvent` to `tray-leader/broadcast.ts` — **carrying the unreachable try/catch across verbatim**. Resolved by taking main's decomposition and re-applying the fix at the new homes; the oversize test moved to `tests/scoops/tray-leader/broadcast-oversize.test.ts` to match the mirrored layout.

One deliberate change of course: main has a test asserting the *throwing* `send` path (`follower-registry.test.ts`). Rather than delete it, the catch is retained as an explicit backstop — one hostile or stubbed channel must not abort the fan-out loop and strand the followers behind it — while the return value is what actually drives failure reporting. The issue's substance (bootstrapId attribution, live throttle state) is unaffected.

## ⚠️ Surrogate splitting — confirmed, cross-runtime corruption

Verified end to end against a real Go decoder:

```
frame1: {"chunkData":"aaaaaaaaa\ud83c"}
frame2: {"chunkData":"\udf66bbbb"}
JS rejoin  -> identical to original    (my round-trip test was blind to this)
Go rejoin  -> "aaaaaaaaa\ufffd\ufffdbbbb"          (emoji destroyed, unrecoverable)
```

Fixed by never cutting between a high and low surrogate, building slices before counting them. The new test asserts per-frame well-formedness — the property a non-JS decoder actually needs — and fails on both new cases when the guard is removed.

## ⚠️ Frame metadata — confirmed, this is a remote crash

Stronger than "can crash or exhaust": two frames panic the Go follower outright.

```
panic: runtime error: index out of range [99] with length 2
  tray.(*Conn).acceptChunkFrame  conn.go:473
```

It runs on the data-channel read goroutine, so an unrecovered panic takes the process down — remotely triggerable by any peer. `totalChunks` is now pinned for a chunkId's lifetime and bounded (`TRAY_MAX_CHUNK_COUNT`) before allocation, in all three runtimes plus the shared guard; senders refuse to exceed the same bound so both ends stay symmetric. The Go regression test reproduces the panic exactly when the guard is removed.

## ⚠️ iOS ignoring `sendData` — confirmed

`WebRTCManager.sendData` returns `Bool`; the framed loop discarded it, so a failure at frame N kept pushing N+1. Every write is now checked and the loop stops at the first failure, with the frame index logged. `try?` on encoding was hiding errors for the same reason and now logs.

## 💡 Both minors — confirmed

Grapheme clusters measured: `"a"` + 60k combining marks is **one** `Character` producing a **120,001-byte** frame against the 64 KiB limit. Now sliced by unicode scalar (capped at exactly the budget, verified). The stale `FollowerSyncManager` doc reference now points at `AppState`.

## Not adopted

The suggested cross-runtime wire-vector fixtures are a good idea but belong in their own change — they need a corpus-style generator and a decode harness in two languages. The surrogate property test covers the specific defect found here. Happy to file it.

## Verification

`verify` 7/7 · `typecheck` · `test` (12659) · `test:coverage` (exit 0) · both builds · `go build`/`vet`/`gofmt` · `check-touched-exemptions` · SwiftLint 33 (baseline 39).

Two notes: the `ios-app/CLAUDE.md` doc-size gate left only 68 chars of headroom, so the iOS docs now live inside existing table padding at **zero net growth**. And `test:coverage` failed once on an unrelated flaky playwright iframe test — it also fails on unmodified `main`, passes 3/3 in isolation, and passed on a clean re-run.

PR #1710, opened 2026-07-27
fix(webapp): restore follower Browser dock item + stop hosted-tab CDP loop

### LINKED ISSUE #1706: Follower: Browser · CDP surface is broken end-to-end on a hosted-tab follower
Observed on a live hosted cone (`sliccy.ai/cloud` → **Open ↗** → `/join/<sessionId>.<token>`) opened as a browser-tab follower in Chrome Canary, against a local build of `main` @ `5bd2a90b9` (5.78.0).

A hosted-tab follower has no local Chrome bridge, but the follower boot path still treats it as if it does. Two halves of the same wrong assumption:

## 1. The follower polls a `/cdp` endpoint that cannot exist

`refreshTargets()` in `packages/webapp/src/ui/page-follower-tray.ts` is gated only on `options.uiOnly`:

```ts
// page-follower-tray.ts:334
if (options.uiOnly) return; // UI-only follower advertises no CDP target
...
// page-follower-tray.ts:366
if (!options.uiOnly) {
  targetRefreshInterval = setInterval(() => void refreshTargets(), refreshIntervalMs);
  void refreshTargets();
}
```

and `uiOnly` is only ever true for a cherry embed:

```ts
// wc-follower.ts:227
const uiOnly = isCherry && new URLSearchParams(window.location.search).get('ui-only') === '1';
```

A `/join/...` tab on the hosted origin is not cherry, so `uiOnly === false` and it polls `browserAPI.listPages()` forever. That dials `getDefaultCdpUrl()` (`browser-api.ts:61-64`) = `wss://<current-origin>/cdp`. On `www.sliccy.ai` that path hits the SPA fallback and returns HTML with a `200`, so the WebSocket upgrade can never succeed.

Measured: **99 failed connections in 13 minutes**, roughly one per 5s, no backoff ceiling, never gives up.

```
2026-07-27T15:30:49.457Z [log.error] WebSocket connection to 'wss://www.sliccy.ai/cdp' failed:
...
2026-07-27T15:39:23.473Z [log.error] WebSocket connection to 'wss://www.sliccy.ai/cdp' failed:
[cdp] (suppressed 15 similar: "Connection failed")
[page-follower-tray] Follower CDP target listing failed (best-effort, throttled)
```

`ThrottledErrorTracker` keeps the console readable but does not stop the dialing.

`browser-api.ts:80-89` already documents this exact hazard for the thin-bridge/cherry case. The plain-tab follower is not covered by it.

## 2. The follower's Browser · CDP dock button does nothing

Clicking the globe sets `active=browser` on `slicc-dock` and then nothing happens — no pane, no overlay, no error.

`wireDockToWorkbench` returns early for `browser` on purpose, because the globe is meant to open the full-screen tab overlay rather than a workbench pane:

```ts
// wc-shell.ts:295-297
// The Browser globe opens the FULL-SCREEN tab overlay (wc-browser.ts) —
// a workspace pane underneath it would just be dead chrome.
if (id === 'browser') return;
```

That overlay is wired in `wireWcBrowserOverlay` (`wc-live.ts:1168-1180`), which is leader-only:

```ts
const standalone = options.standalone;
if (!standalone) return;
void import('./wc-browser.js').then(({ wireWcBrowser }) => ...);
```

`wc-follower.ts` has zero references to `wc-browser`. So on a follower the dock event is ignored by the pane handler by design and no other listener exists. Dead button.

This matters because the follower's own placeholder copy advertises the capability:

> Files live on the leader. A follower mirrors the leader's chat, sprinkles, **and browser tabs** — not its filesystem.

For contrast, on the leader the same button opens a working `Browser · open tabs` overlay listing all 3 tabs. On the follower, `Files · VFS` opens correctly (pane 217px), so the workbench itself is fine — this is specific to the browser surface.

This is checklist item 8 in `docs/review-patterns.md` (follower wiring parity — check all boot paths).

## Why these are one issue

Different mechanisms, but one design decision and one fix: **a hosted-tab follower has no local browser, so it should render the leader's tabs and stop advertising its own.** Fixing either half alone leaves it half-wrong — wire the overlay and it opens empty; stop the polling and the button is still dead.

## Suggested direction

- Gate `refreshTargets()` on whether a local `/cdp` bridge actually exists, not on `isCherry && ui-only`. `uiOnly` is describing the wrong property.
- Either wire `wc-browser` into the follower boot path so the globe shows the leader's advertised tabs, or hide the dock item on followers that cannot serve it. A dead button with no error is the worst of the three options.

## Also in scope (same file)

`wc-shell.ts:270-274` builds a `slicc-surface[surface-id=browser]` with a placeholder note that line 297 makes permanently unreachable — the dock is the only way to activate a surface, and it returns early for `browser`. Dead chrome; delete it or make it reachable.

## Verification

- Repro needs a running hosted cone and its `/join/` link; `.agents/skills/cdp-smoke-test/` drives it (`slicc-cdp watch` catches the loop; note `SLICC_UI_ORIGIN` must point at the follower origin).
- Regression test belongs in `packages/webapp/tests/ui/` — assert a non-cherry follower mount starts no target-refresh interval, and that a follower boot wires a browser-dock handler.

catalan-adobe:
Fixes #1706.

Two halves of the same wrong assumption — that a follower has a local Chrome to drive. Both are invisible on a leader, which is why they survived four months.

## 1. The Browser dock item was inert on followers

`wireDockToWorkbench` hardcoded a skip for the browser surface:

```ts
// The Browser globe opens the FULL-SCREEN tab overlay (wc-browser.ts) —
// a workspace pane underneath it would just be dead chrome.
if (id === 'browser') return;
```

That helper is shared by every boot path. The overlay it defers to (`wireWcBrowserOverlay`) is wired only when `options.standalone` is set. So leaders traded a pane for an overlay; followers lost the pane and got nothing — an inert globe, no error, and a `slicc-surface[surface-id=browser]` that became permanently unreachable.

The suppression is now a **claim** rather than a hardcoded id:

```ts
// wc-shell.ts — read at click time; the overlay wiring runs after mountWcShell
if (overlaySurfaces.has(id)) return;

// wc-browser.ts — claimed as the overlay wires itself
refs.overlaySurfaces.add('browser');
```

Floats that never reach that wiring keep the pane placeholder, and a leader whose lazy `import('./wc-browser.js')` fails now degrades to the same fallback instead of a dead button — a hole the old code had too.

## 2. Hosted-tab followers dialed a `/cdp` that cannot exist

The advertisement gate was `isCherry && ui-only=1`, which describes cherry rather than the property that matters. A follower opened from a hosted `/join/…` link matches neither branch, so it polled `listPages()` every 5s against `getDefaultCdpUrl()` = `wss://<hosted-origin>/cdp` — a path the SPA fallback answers with HTML, so the upgrade can never succeed.

Measured on a live cone: **99 failed connections in 13 minutes**, no backoff ceiling, never gives up.

Replaced with a predicate keyed on the local CDP surface *existing*:

```ts
export function followerAdvertisesCdpTargets(isCherry: boolean, search: string): boolean {
  if (isCherry) return new URLSearchParams(search).get('ui-only') !== '1';
  return parseBridgeLaunchParams(search) !== null;
}
```

`StartPageFollowerTrayOptions.uiOnly` becomes `advertisesCdpTargets` so the option names what it gates — the old name is what let the drift happen. The cherry SDK's own `mountSlicc({ uiOnly })` option is a different thing and is untouched.

## Tests

Both fixes verified by mutation — reverting either implementation fails the new assertions:

| mutation | failing tests |
|---|---|
| restore `if (id === 'browser') return;` | 1 (`an UNCLAIMED browser dock item opens the workspace pane`) |
| restore `isCherry && ui-only=1` | 3 (hosted-tab mount + 2 predicate cases) |

One assertion worth calling out: the suppression test asserts `listPages` is **never called**, not merely that `advertiseTargets` isn't. The dialing loop was the actual user-visible bug, and an advertise-only assertion passes while it runs.

Also added: the omitted-flag default, a bridge-param-without-token case, and a claim-registered-after-mount case (the claim lands after `mountWcShell`, so the handler must read the set at click time).

## Verification

`npm run verify` · `check-touched-exemptions` · `typecheck` · `test` (12,439 passed) · `test:coverage` (floors pass) · `npm run build` · `npm run build -w @slicc/chrome-extension` — all green.

Live-checked the leader path against a local harness after the change: the globe still opens `Browser · open tabs` with all 3 tabs and no pane behind it.

**Not live-verified:** the follower half. The follower tab loads from the deployed `www.sliccy.ai`, so it can't exercise a local build — that half rests on the unit tests and the mutation check above. Worth a look on staging.

## Docs

`docs/review-patterns.md` category 8 gains the two trigger shapes this bug had (a fallback removed from a shared helper for a leader-only replacement; a gate keyed on a float name where a capability check belongs) plus the claim pattern as remediation. `.github/copilot-instructions.md` synced within its 4,000-char cap.

`packages/webapp/CLAUDE.md` was left alone deliberately — it sits at 19,954 of its 20,000-char limit, so there was no room. The detail lives in the code comments and in `review-patterns.md` instead.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -54,6 +54,8 @@ Every leader broadcast (`LeaderToFollowerMessage` / `broadcast*` in `tray-leader
 needs a matching follower handler in `tray-follower-sync.ts` AND a UI action wired in
 `wc-follower.ts`. New interactive elements on leader surfaces need follower counterparts.
 Check all three boot paths (`mountWcUiLive` / `mountWcUiFollower` / `mountWcUiExtension`).
+Also flag: a fallback removed from a shared helper for a leader-only replacement, or a
+gate keyed on a float name (`isCherry`) where a capability check belongs (#1706).
 The largest empirical failure class (~30–40 commits since 2026-03).
 
 ## 8. Origin / bridge routing contract (often Major)
--- a/docs/review-patterns.md
+++ b/docs/review-patterns.md
@@ -196,6 +196,12 @@ regression test for each bug fix, and keep coverage at or above the package floo
   UI surface.
 - A new follower-side handler in `tray-follower-sync.ts` with no corresponding UI action
   wired in `ui/wc/wc-follower.ts` (or vice versa).
+- **Removing a fallback from a shared helper** (e.g. `wc-shell.ts`) because a
+  leader-only replacement now covers it. `wireDockToWorkbench`, `prepareWcShell` and
+  friends run on every boot path; the replacement usually does not.
+- **Gating behavior on a float name** (`isCherry`, `isExtensionRealm()`, `ui-only=1`)
+  where the real condition is a capability. Float lists go stale the moment a new float
+  ships; capability checks do not.
 
 **Historical precedents**
 
@@ -205,6 +211,15 @@ regression test for each bug fix, and keep coverage at or above the package floo
   buttons had no action handler, so clicking them silently no-oped.
 - **PR #1261**: a fix landed on the non-primary boot path first, then caused a 5-second
   pill-reset regression that had to be fixed the next day.
+- **Issue #1706** (both shapes above, four months undetected). `a99a5faa4` added
+  `if (id === 'browser') return;` to `wireDockToWorkbench` so the leader's new
+  full-screen tab overlay had no dead pane behind it — but that helper is shared, and
+  the overlay is wired only when `options.standalone` is set. Leaders traded a pane for
+  an overlay; followers lost the pane and got nothing, leaving an inert dock item and an
+  unreachable `slicc-surface`. Separately, the follower's CDP advertisement was gated on
+  `isCherry && ui-only=1` rather than "has a local CDP bridge", so hosted-tab followers
+  — a float that shipped nine days after the gate was written — dialed
+  `wss://<hosted-origin>/cdp` every 5s indefinitely.
 
 **Class size** — ~30–40 commits since 2026-03; the largest empirical failure class in the
 repo.
@@ -213,6 +228,12 @@ repo.
 _and_ that the UI surface wires the user action back to the leader. Check all three boot
 paths (`mountWcUiLive` / `mountWcUiFollower` / `mountWcUiExtension`). Record the iOS
 mirror decision in the corpus (`packages/ios-app/`).
+
+When a leader-only feature supersedes shared behavior, let the feature **claim** what it
+replaces at wiring time rather than teaching the shared helper about it (see
+`WcShellRefs.overlaySurfaces`). The fallback then survives by default on every path that
+does not wire the feature — including a leader whose lazy import fails — instead of
+depending on two files agreeing forever.
 
 ### 9. Origin / bridge routing contract
 
@@ -295,7 +316,7 @@ follower (tray) and Cherry paths. See [docs/transcript-export.md](transcript-exp
 
 ---
 
-## Approval gate changes in `wc-tray.ts` or `wc-transcript-export.ts`.
+## Approval gate changes in `wc-tray.ts` or `wc-transcript-export.ts`
 
 When the approval dialog is changed: verify that Deny still resolves with `permission-denied`,
 that no approval escapes require a new prompt, and that the binary-attachment warning is
--- a/packages/webapp/src/ui/boot/setup-standalone-prelude.ts
+++ b/packages/webapp/src/ui/boot/setup-standalone-prelude.ts
@@ -62,6 +62,22 @@ export interface LickForwardingClient {
 export interface StandalonePreludeResult {
   browser: BrowserAPI;
   realCdpTransport: CDPTransport;
+  /**
+   * Whether this page realm reached a real Chrome — the single answer to
+   * "can `browser.listPages()` succeed here?", decided by whichever transport
+   * branch below actually ran.
+   *
+   * `false` means the page has a `BrowserAPI` with no route to Chrome: calling
+   * into it dials `getDefaultCdpUrl()` (= `wss://<page-origin>/cdp`), which on
+   * a hosted origin the SPA fallback answers with HTML, so the upgrade never
+   * succeeds. A follower that polls on that route retries forever (#1706).
+   *
+   * Callers must read this rather than re-deriving it from URL params: the
+   * three branches disagree about which params matter (the extension-leader
+   * branch deliberately ignores bridge params), and duplicating that logic is
+   * exactly the drift #1706 was.
+   */
+  hasLocalCdpSurface: boolean;
   cherryJoinUrl?: string;
   cherryTransport?: CherryHostTransport;
   instanceId: string;
@@ -324,6 +340,10 @@ export async function setupStandalonePrelude(
     localLickWsUrl = bridge.lickWsUrl;
   }
 
+  // Mirrors the three transport branches below, in order: cherry's synthetic
+  // target, the extension bridge's `chrome.debugger`, or the node-server bridge.
+  const hasLocalCdpSurface = runtimeMode === 'cherry' || useExtensionBridge || bridge !== null;
+
   const runtimeConfig = await fetchRuntimeConfig();
   const runtimeDefaultWorkerBaseUrl = shouldUseRuntimeModeTrayDefaults(
     runtimeMode,
@@ -412,6 +432,7 @@ export async function setupStandalonePrelude(
   return {
     browser,
     realCdpTransport,
+    hasLocalCdpSurface,
     cherryJoinUrl,
     cherryTransport,
     instanceId: mintInstanceId(),
--- a/packages/webapp/src/ui/page-follower-tray.ts
+++ b/packages/webapp/src/ui/page-follower-tray.ts
@@ -91,11 +91,19 @@ export interface StartPageFollowerTrayOptions {
   runtime?: string;
 
   /**
-   * When true, the follower keeps handshake + transport + chat sync but suppresses CDP target
-   * advertisement. Used by the extension's per-page cherry sidebar (`?cherry=1&ui-only=1`), where
-   * the extension controls the tab via real `chrome.debugger` CDP instead of a weak synthetic cherry target.
+   * Whether this follower has a CDP surface worth advertising to the leader.
+   * When false the follower keeps handshake + transport + chat sync but never
+   * enumerates or advertises targets.
+   *
+   * Must reflect whether a local CDP surface actually EXISTS — not which float
+   * is running. A follower without one that still polls will dial
+   * `getDefaultCdpUrl()` (= `wss://<page-origin>/cdp`) on a 5s loop forever;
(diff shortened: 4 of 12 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/src/ui/page-follower-tray.ts:342,374]** — CDP suppression guards both event-driven refreshes and interval creation, ensuring `listPages()` is never reached when no CDP surface exists.
- **[packages/webapp/src/ui/wc/wc-shell.ts:307-316]** and **[packages/webapp/src/ui/wc/wc-browser.ts:97-109]** — The claim-based overlay design cleanly preserves the fallback and correctly reads claims at click time, after lazy wiring.
- **[packages/webapp/tests/ui/page-follower-tray-advertise.test.ts:150-166]** — The regression test asserts the user-visible failure mechanism—CDP enumeration itself—rather than merely checking that advertisement was skipped.
- **[packages/webapp/tests/ui/wc/wc-shell.test.ts:164-238]** — Tests cover unclaimed, claimed, and post-mount claims, including the key lazy-import ordering case.
- **[packages/webapp/src/ui/wc/wc-follower.ts:240-242]** — Partial and malformed bridge configuration is delegated to the existing canonical parser rather than reimplemented.

### Issues

*(none found)*

- **[packages/webapp/src/ui/wc/wc-follower.ts:240-242]** — Bridge query parameters are not the only non-Cherry way this page can have a local CDP surface.
  - A pinned thin-extension leader uses `?slicc=leader&ext=…` and receives a real `ExtensionBridgeTransport` at `setup-standalone-prelude.ts:350-360`, without node bridge parameters. `host join` explicitly persists the follower URL for reload (`setup-standalone-panel-rpc.ts:106-112`), and the stored URL makes the next boot enter follower mode (`runtime-mode.ts:55-66`). On that supported reload path, this predicate returns `false`, so `startPageFollowerTray` never advertises tabs despite Chrome being available through the extension bridge.
  - This regresses browser federation for an extension leader that joins another tray and reloads.
  - Derive the capability after `setupStandalonePrelude`, ideally from an explicit capability returned by the prelude. At minimum account for its selected transport:
    ```ts
    const advertisesCdpTargets = isCherry
      ? !uiOnly
      : prelude.realCdpTransport.isExtensionBridge === true ||
        parseBridgeLaunchParams(window.location.search) !== null;
    ```
    Add a test for `?slicc=leader&ext=<id>` plus a stored follower join URL.

#### 💡 Minor (Nice to Have)

- **[packages/webapp/src/ui/wc/wc-shell.ts:284-287]** — The newly reachable fallback tells users that the dock item “opens the full-screen tab switcher” and instructs them to click cards, although followers and failed lazy imports show only this static pane and no cards.
  - The restored button is no longer inert, but its only content incorrectly describes behavior that did not occur.
  - Use fallback-specific copy such as “The browser tab switcher is unavailable in this float,” optionally explaining that browser control remains on the leader.

### Recommendations

- Make local-CDP availability an explicit result of `setupStandalonePrelude` rather than reconstructing it from URL parameters in `wc-follower.ts`. This keeps node bridge, extension bridge, Cherry, and future transports under one capability decision.
- Extend the predicate/mount matrix with the thin-extension persisted-follower reload case, alongside the existing hosted-tab, Cherry, and node-bridge cases.

### Assessment

**Reasoning:** The dock claim implementation and hosted-tab loop suppression are well designed and strongly tested, but the new capability predicate overlooks the existing extension CDP bridge and disables tab advertisement on a supported follower reload path.

---

catalan-adobe:
Thanks both — the gpt-5.6 review found a real regression I introduced. Fixed in 5e7fce205, plus a rebase onto main (the branch was 73 commits behind, which was making `check-touched-exemptions` report main's own debt paydown as if this PR added it).

## ⚠️ Extension bridge — valid, fixed

Confirmed the whole chain before acting on it, because the suggested code referenced an API that doesn't exist (`realCdpTransport.isExtensionBridge` — no such property anywhere in the repo). The *reasoning* was right even though the snippet wasn't:

1. The pinned thin-extension leader is a normal web page, not the extension realm — `resolveUiRuntimeMode`'s extension check keys on the separate `isExtension` argument (`isExtensionRealm()`, `main.ts:57`), which is false there.
2. So the follower fast-path at `runtime-mode.ts:66` is reachable, and a stored join URL from `host join` routes the reload to `'follower'`.
3. `setupStandalonePrelude` then takes the `extLeader && hasChromeRuntimeConnect()` branch → real Chrome, **no bridge launch params** (that branch explicitly sets `bridge = null`).
4. My predicate sniffed for bridge params → `false` → advertisement silently off, where main had it on.

Took the *recommendation* rather than the snippet: the capability is now decided where the branch is taken.

```ts
// setup-standalone-prelude.ts — mirrors the three transport branches
const hasLocalCdpSurface = runtimeMode === 'cherry' || useExtensionBridge || bridge !== null;
```

and the predicate drops to the two questions it should have been asking:

```ts
export function followerAdvertisesCdpTargets(hasLocalCdpSurface: boolean, uiOnly: boolean): boolean {
  return hasLocalCdpSurface && !uiOnly;
}
```

Capability from the prelude; policy (`uiOnly`) separate, because the side panel *has* a surface and deliberately withholds it. No caller re-derives transport semantics from query params now — which was the actual lesson of #1706, and I'd reintroduced a smaller version of it.

**Tests**: the existing extension-leader prelude test now also asserts `hasLocalCdpSurface === true`, with a comment naming this path. New `hasLocalCdpSurface` suite covers no-params, bridge-params, and bridge-without-token. Mutation-checked — dropping `useExtensionBridge` from the expression fails the extension-leader test. The URL-shape cases moved from the predicate suite to the prelude suite, since that's where the decision now lives.

## 💡 Fallback copy — valid, fixed

Right, and it was worse than "not describing what happened": it told the reader to click cards that aren't rendered. Now:

> The tab switcher runs on the leader. This float has no browser of its own to show — ask the leader to open, focus, or close tabs through chat.

Matches the sibling placeholders' phrasing (`wc-follower.ts:123,133`). The test asserts the fallback framing and negatively asserts the stale "click a card" instruction.

## `wc-tray.ts::buildFollowerOptions` (Claude's non-blocking note)

Agreed on all counts, and leaving it. It omits `advertisesCdpTargets`, so the tray-join / role-switch path defaults to advertising — correct for standalone, unchanged from main. The hosted-leader-becomes-follower case is real but hypothetical, and folding it in means threading the prelude capability through a second call site that doesn't currently receive it. Out of scope for #1706; worth its own issue if that transition ships.

## Verification after rebase

`verify` · `check-touched-exemptions` · `typecheck` · `test` (12,637 passed) · `test:coverage` · both builds — all green.

Also dropped two artifacts that snuck in: a macOS-only `package-lock.json` diff (`npm install` stripping `libc: glibc` entries, which would have hurt Linux CI) and a `swift-server/Package.resolved` churn from `npm run build`.

**Still not live-verified:** the follower half, for the same reason as before — the follower tab loads from deployed `www.sliccy.ai`, so a local build can't exercise it. Unchanged ask for a staging look.

Drive-by: removed an unused `header` binding in `wc-shell.test.ts` (pre-existing on main, in a file this PR already touches).

PR #1714, opened 2026-07-28
fix(cherry): negotiate protocol version so vendored older SDKs connect

catalan-adobe:
## Problem

Since 2026-07-27, every production Cherry embed running a vendored pre-v2 SDK fails with:

> **Could not start follower** — Cherry handshake timed out after 30000ms

(P1 incident on `labs.of1.live/liftoff`.)

## Root cause

e1865d39d (2026-07-23) bumped `CHERRY_PROTOCOL_VERSION` 1 → 2 for transcript exports — an additive feature — while `isCherryEnvelope` on both sides validates the version with strict equality. Embedders vendor the host SDK, so a v1 host receives the v2 `handshake.hello`, rejects it with only a console warning in the host page, and never replies. The follower iframe's fast-fail skew detection needs to *receive* a mismatched envelope, so it never fires and the connect eats the full 30s timeout. The error surfaces inside the iframe as "Could not start follower", misdirecting diagnosis toward the cone/joinUrl (which are fine — the failure is in the host↔iframe postMessage handshake, before any cone connection).

## Change

**Follower (webapp, `CherryHostTransport`):**

- Posts one `handshake.hello` per entry in the new `SUPPORTED_CHERRY_PROTOCOL_VERSIONS` ([2, 1], same `channelId` so a host's re-hello logic doesn't misread it), pins the negotiated version from whichever `handshake.welcome` the host answers, and stamps all subsequent outbound envelopes with it. A vendored v1 host now completes the handshake against the current webapp with no re-vendor.
- Ignores v2-only `session.export.*` envelope kinds on a v1-negotiated channel.
- The timeout error message now names the plausible causes (host SDK missing, not listening, or version-skewed).

**Host SDK (`mount.ts`):**

- Replies `handshake.version-mismatch` (new envelope) to a version-skewed hello — only when origin + source prove it is our iframe, and never for the companion lower-version hello of a completed handshake — so a future no-overlap skew fails `connect()` in seconds with "update the older side" instead of the opaque timeout.
- New `hooks.onProtocolMismatch(peerVersion, sdkVersion)` lets embedders surface the skew.

**Protocol mirrors:** `cherry-host-protocol.ts` and `cherry/src/protocol.ts` updated identically (enforced by `protocol-mirror.test.ts`); envelope `cherry` fields are now `number` with runtime validation against an accepted-versions set.

## Tests

- 6 new transport tests: dual-hello shape, v1-host handshake (regression for the incident), negotiated-version stamping in/out, v2-only kind gating, unsupported-version fast-fail.
- 4 new mount tests: mismatch reply + hook, companion-hello suppression, untrusted origin/source, non-handshake skewed envelopes.
- Mutation-verified: removing v1 support fails 4 tests; disabling the mismatch reply fails 1.

## Docs

`packages/cherry/CLAUDE.md` gains a "Protocol version negotiation and bump rules" section: keep the previous version in the supported set when bumping, gate new envelope kinds on the negotiated version, and only drop a version for a genuinely breaking wire change.

### DIFF
--- a/packages/cherry/CLAUDE.md
+++ b/packages/cherry/CLAUDE.md
@@ -194,6 +194,31 @@ same `CherryEnvelope` union, `CHERRY_PROTOCOL_VERSION`, `isCherryEnvelope`,
 difference is the package location (the SDK copy carries no webapp import). When
 you change one, change the other.
 
+## Protocol version negotiation and bump rules
+
+Embedders **vendor** this SDK, so the two sides of the handshake routinely run
+different builds. The contract that keeps that safe:
+
+- The follower iframe posts one `handshake.hello` per entry in
+  `SUPPORTED_CHERRY_PROTOCOL_VERSIONS` (newest first, same `channelId`) and pins
+  the negotiated version from whichever `handshake.welcome` the host answers
+  with. All subsequent envelopes — both directions — are stamped with the
+  negotiated version. Version-gated envelope kinds (e.g. `session.export.*`,
+  v2+) must not be used on a lower-negotiated channel.
+- The host SDK accepts only its own `CHERRY_PROTOCOL_VERSION`. When a
+  handshake attempt arrives at a version it cannot speak (and origin + source
+  prove it is our iframe), it replies `handshake.version-mismatch` so the
+  follower fails `connect()` fast with an actionable error instead of eating
+  the handshake timeout, and fires `hooks.onProtocolMismatch`.
+- **When bumping `CHERRY_PROTOCOL_VERSION`:** keep the previous version in
+  `SUPPORTED_CHERRY_PROTOCOL_VERSIONS` and gate any new envelope kinds on the
+  negotiated version. Removing a version from the supported set hard-breaks
+  every embedder still shipping a vendored SDK at that version — do it only
+  for a genuinely breaking wire change, and say so in the changelog. The
+  2026-07-27 labs incident (P1: every embed failed with an opaque 30s
+  "Cherry handshake timed out") was caused by bumping 1 → 2 for an _additive_
+  feature while both sides still validated with strict equality.
+
 ## Build and test
 
 ```bash
--- a/packages/cherry/src/index.ts
+++ b/packages/cherry/src/index.ts
@@ -60,6 +60,14 @@ export interface HostHooks {
   onPermissionRequest?: (domain: string) => boolean | Promise<boolean>;
   /** Called once the Cherry postMessage handshake completes (welcome sent to follower). */
   onHandshakeComplete?: () => void;
+  /**
+   * Called when the follower iframe attempts a handshake with a cherry
+   * protocol version this SDK build cannot speak (after the follower's own
+   * version fallbacks are exhausted). The mount will not come up until the
+   * older side (this vendored SDK or the SLICC origin) is updated — surface
+   * this to telemetry instead of waiting out the handshake timeout.
+   */
+  onProtocolMismatch?: (peerVersion: number, sdkVersion: number) => void;
 }
 
 /** Effort / thinking level the cone should use. Locked — the UI picker is hidden. */
--- a/packages/cherry/src/index.ts
+++ b/packages/cherry/src/index.ts
@@ -61,11 +61,13 @@ export interface HostHooks {
   /** Called once the Cherry postMessage handshake completes (welcome sent to follower). */
   onHandshakeComplete?: () => void;
   /**
-   * Called when the follower iframe attempts a handshake with a cherry
-   * protocol version this SDK build cannot speak (after the follower's own
-   * version fallbacks are exhausted). The mount will not come up until the
-   * older side (this vendored SDK or the SLICC origin) is updated — surface
-   * this to telemetry instead of waiting out the handshake timeout.
+   * Called at most once per handshake attempt, after a short grace window,
+   * when the follower iframe offered ONLY cherry protocol versions this SDK
+   * build cannot speak. Never fires when a fallback succeeds: a follower that
+   * also offers a version this SDK speaks completes the handshake and cancels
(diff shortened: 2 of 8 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/src/cdp/cherry-host-transport.ts:237-250]** — Sends supported-version hellos newest-first with one shared `channelId`, directly addressing the v1-host regression without requiring re-vendoring.
- **[packages/webapp/src/cdp/cherry-host-transport.ts:357-380, 460-488]** — Correctly pins the welcome’s protocol version, narrows subsequent inbound validation, and gates v2-only export envelopes on v1 channels.
- **[packages/webapp/src/cdp/cherry-host-transport.ts:310-343, 503-549]** — Subsequent outbound CDP, SLICC, and export envelopes consistently use the negotiated version.
- **[packages/cherry/src/mount.ts:324-335]** — Mismatch replies are restricted to the expected iframe origin/source, preventing unrelated frames from probing or influencing negotiation.
- **[packages/cherry/src/protocol.ts:249-310]** — Accepted versions are explicit runtime inputs while the default remains strict, preserving host-SDK behavior.
- **[packages/cherry/tests/protocol-mirror.test.ts:8-29]** — The byte-level mirror invariant is a strong safeguard for the duplicated protocol contract.
- **[packages/webapp/tests/cdp/cherry-host-transport.test.ts:152-252]** — Tests cover the production regression, negotiated stamping, inbound v1 traffic, feature gating, and unsupported-version fast failure.

### Issues

*(none found)*

- **[packages/webapp/src/cdp/cherry-host-transport.ts:426-447]** — The unsupported-version fast-fail path does not verify the envelope’s `channelId`.
  - `acceptEnvelope` rejects the message on version before its nonce check can matter. `diagnoseRejectedMessage` then restores only the origin and source checks. A delayed mismatch response from connection A can therefore arrive while connection B is pending and reject B despite carrying A’s stale channel nonce. Multiple concurrently connecting transports against the same parent have the same cross-channel failure.
  - Require all three factors before rejecting:
    ```ts
    const trustedPeer =
      this._state === 'connecting' &&
      event.data.channelId === this.channelId &&
      this.opts.allowOrigins.includes(event.origin) &&
      event.source === (this.opts.counterpart as unknown as MessageEventSource);
    ```
  - Add a test that sends an unsupported-version reply with a stale channel, verifies the connection remains pending, and then completes it with a matching welcome.

- **[packages/cherry/src/mount.ts:324-335]** — `onProtocolMismatch` fires before protocol fallbacks have been tried, contradicting its public contract.
  - When a future follower advertises `[3, 2]` to this v2 SDK, the v3 hello arrives first and immediately fires the hook. The following v2 hello then succeeds. An embedder following the documentation at **[packages/cherry/src/index.ts:64-68]** may surface a fatal error or tear down a mount that was fully compatible. In a true no-overlap case with multiple hellos, the hook can also fire once per hello.
  - Track a pending mismatch by `channelId`, allow the companion hellos to arrive, cancel it when a supported same-channel hello succeeds, and otherwise emit one reply/hook after a short negotiation grace period. Alternatively, explicitly redefine the hook as non-fatal, per-attempt telemetry and deduplicate it—but the current “fallbacks exhausted / mount will not come up” contract must not remain.
  - Add coverage for a future `[3, 2]` follower connecting to the v2 host without invoking a fatal mismatch hook, plus a no-overlap case that invokes the hook once.

#### 💡 Minor (Nice to Have)

*(none found)*

### Recommendations

- Add an end-to-end protocol compatibility matrix joining the real mount and transport implementations: `[2,1]↔v1`, `[3,2]↔v2`, no-overlap, reconnect with a stale response, and iframe re-handshake.
- Model negotiation as a channel-scoped transaction with one terminal outcome—welcome or mismatch—rather than treating each advertised hello independently.

### Assessment

**Reasoning:** The primary v1 compatibility implementation is sound and well tested, but mismatch handling currently bypasses channel pinning and the new hook can falsely report a fatal incompatibility during a successful fallback.

---

catalan-adobe:
Both Important findings verified against the code and addressed in 6ca01de38:

**1. Fast-fail path missing the channelId factor — fixed.** `diagnoseRejectedMessage` now requires all three factors (origin, source, **and** the channelId nonce) before rejecting a pending `connect()`. The host's genuine `handshake.version-mismatch` reply echoes the current attempt's hello channelId, so the nonce always matches when the rejection is legitimate. New test: a stale-channel mismatch reply arrives during a pending connect → connect stays pending and completes on a matching welcome. Mutation-verified (dropping the nonce factor fails the test).

**2. `onProtocolMismatch` fired before fallbacks were tried — fixed.** The wire reply stays immediate (it's what gives the follower its fast `connect()` error, and a follower that still has a version this SDK speaks ignores it — the reply's version is inside its supported set). The hook is now:
- deduped per handshake-attempt channelId (no-overlap follower: one reply, one report), and
- deferred by a 250 ms grace window that an accepted same-channel hello cancels, so a `[3, 2]` follower against this v2 host completes silently.

`destroy()` clears pending reports; the hook's doc contract in `index.ts` now states "at most once per handshake attempt, never when a fallback succeeds". New tests cover the `[3, 2]` fallback (hook silent), no-overlap dedupe, and destroy cancellation — the fallback test is mutation-verified against the cancel path.

**Recommendations — partially adopted:**
- *Compatibility matrix joining the real mount + transport:* the added unit tests now cover `[2,1]↔v1`, `[3,2]↔v2` (host side), no-overlap, and stale-reply-during-reconnect; the iframe-reload re-handshake suite already exists. A joined end-to-end harness would duplicate what these cover through the same `postMessage` contract, so I've left it out (YAGNI) — happy to add one if a real gap shows up.
- *Channel-scoped negotiation transaction:* the grace-window + per-channelId dedupe implements the "one terminal outcome per attempt" semantics for the host's observable behavior (one reply, one report or a completed handshake) without introducing a state machine.

PR #1718, opened 2026-07-28
docs(webapp): document why the sprinkle sandbox is escapable by design

### LINKED ISSUE #1717: Sprinkle srcdoc iframes log 'can escape its sandboxing' warning on every render
## Symptom

A healthy hosted-follower session (`/join/<sessionId>.<token>` in Chrome) logs this warning repeatedly — once per sprinkle/dip iframe creation, so re-renders make it "numerous and regular":

> An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.

Source: `packages/webapp/src/ui/sprinkle-renderer.ts:583-585` — every sprinkle renders into a `srcdoc` iframe with

```ts
const sandboxTokens = isNestedInAnotherFrame()
  ? 'allow-scripts allow-same-origin allow-popups'
  : 'allow-scripts allow-same-origin';
```

## Analysis

Chrome is right about the mechanics: a same-origin `srcdoc` document with scripts can reach `window.parent.document` and remove its own `sandbox` attribute, so the attribute is not a security boundary here. It is a speed bump against *accidental* misbehavior (forms, top-navigation), nothing more.

This appears to be a known and accepted design property, not a latent hole:

- `allow-same-origin` is required: sprinkles load app-served assets (`/lucide-icons.js`, theme CSS) and go through the VFS service worker, which an opaque-origin (fully sandboxed) srcdoc document cannot use — opaque origins don't inherit the parent's SW.
- The real defense against attacker-authored sprinkle content lives in the SW security gates: `llm-proxy-sw-config.ts` rejects nested and auxiliary browsing contexts for the sync-fs channel nonce precisely because "a same-origin `allow-same-origin` srcdoc sprinkle/dip iframe" is in the threat model there.

## Why file it then

1. **Console noise**: the warning fires per sprinkle render and dominates an otherwise-clean prod console, training users (and RUM triage) to ignore warnings.
2. **The attribute over-promises**: a reader of `sprinkle-renderer.ts` could reasonably believe the sandbox isolates sprinkle content. It doesn't — the isolation story is the SW gate + trust model, and only the `llm-proxy-sw-config.ts` doc block says so.

## Options

- **A. Remove `allow-scripts`-defeating tokens honestly**: drop the `sandbox` attribute (keeping `allow-popups` semantics via other means where needed), and document at the creation site that sprinkles are same-origin trusted-ish content whose containment is the SW gate model. Kills the warning; loses the accidental-misbehavior speed bump (notably `allow-forms` and top-navigation blocking).
- **B. Keep as is, document at the source**: add the rationale comment in `sprinkle-renderer.ts` pointing at `llm-proxy-sw-config.ts`, accept the console noise. Zero behavior change.
- **C. True isolation (larger)**: serve sprinkle documents from a dedicated sandbox origin (like every "user-content" host does), with an asset bridge instead of same-origin SW access. Real security boundary, kills the warning, but a significant architecture change and likely conflicts with the browser-first/no-extra-origin constraints (extension float has no second origin to serve from).

B is the cheapest defensible state; A trades a decorative control for a quiet console; C is the only option that makes the warning genuinely untrue. Worth deciding deliberately rather than letting the warning spam stand unexplained.

## Environment

- Prod `www.sliccy.ai` (v5.81.1, 2026-07-28), hosted-tab follower against a cloud cone, Chrome.
- Session otherwise healthy: no console errors; the only remaining UX noise is the transient connection banner family (#1707/#1698).

catalan-adobe:
## What

Documentation-only change (option B from #1717): a comment block at the sprinkle iframe creation site (`sprinkle-renderer.ts`) recording why `sandbox="allow-scripts allow-same-origin"` is deliberate.

## Why

Chrome logs "can escape its sandboxing" once per sprinkle render, which reads like an oversight. It isn't:

- `allow-same-origin` is required — sprinkles load app-served assets through the VFS service worker, and an opaque-origin (fully sandboxed) srcdoc document cannot use the parent's SW.
- The sandbox attribute is only a speed bump against accidental misbehavior; the actual containment for attacker-authored sprinkle content is the SW security gates (`llm-proxy-sw-config.ts` rejects nested same-origin contexts from the sync-fs channel nonce).

The comment prevents two failure modes: someone "fixing" the warning by dropping `allow-same-origin` (breaks sprinkle asset loading), and someone assuming the sandbox isolates sprinkle content (it does not — the SW gate model does).

Options A (drop the decorative sandbox to silence the warning) and C (true origin isolation) stay recorded in #1717 if the console noise ever needs to go.

## Verification

lint 7/7 · check-touched-exemptions · typecheck · sprinkle-renderer tests (27) — no behavior change.

Closes #1717 (option B confirmed by the completed option-A empirical evaluation — results on the issue)

### DIFF
--- a/packages/webapp/src/ui/sprinkle-renderer.ts
+++ b/packages/webapp/src/ui/sprinkle-renderer.ts
@@ -574,17 +574,26 @@ export class SprinkleRenderer {
     const iframe = document.createElement('iframe');
     // SECURITY NOTE (#1717): `allow-scripts allow-same-origin` together make
     // this sandbox ESCAPABLE — Chrome logs "can escape its sandboxing" once
-    // per iframe, and it is right. That is a known, accepted property, not an
-    // oversight: `allow-same-origin` is required because sprinkles load
-    // app-served assets (`/lucide-icons.js`, theme CSS) through the VFS
-    // service worker, and an opaque-origin (fully sandboxed) srcdoc document
-    // cannot use the parent's SW at all. The attribute is therefore only a
-    // speed bump against ACCIDENTAL misbehavior (forms, top-navigation) — the
-    // real containment for attacker-authored sprinkle content is the service
-    // worker's security gates, which treat this nested same-origin context as
-    // untrusted (see the sync-fs channel-nonce gate in
-    // `llm-proxy-sw-config.ts`). Do not "fix" the warning by dropping
-    // `allow-same-origin`, and do not assume this sandbox isolates anything.
+    // per iframe, and it is right. Do not assume this sandbox isolates
+    // anything: it is only a speed bump against ACCIDENTAL misbehavior
+    // (forms, top-navigation). Sprinkle content is agent-authored code that
+    // runs same-origin with the app — the trust model, not this attribute,
+    // is the boundary. The one hard gate that treats this nested context as
+    // untrusted is scoped to the sync-fs channel nonce (realm capability
+    // tokens) in `llm-proxy-sw-config.ts`; it is NOT general containment.
+    //
+    // What `allow-same-origin` is (and is not) known to affect — verified
+    // during #1718 review: the bridge does NOT need it (pure postMessage
+    // both ways; the parent validates by `event.source` identity), and
+    // neither do the injected tags (`/lucide-icons.js` is a network-served
+    // static, theme CSS is inlined). What WOULD change with an opaque
+    // origin: sprinkle content loses same-origin storage (`localStorage` /
+    // IndexedDB throw), and the document no longer inherits the parent's
+    // service-worker controller, so SW-served VFS subresources
+    // (`/preview/*`, project-serve root-relative paths) inside a sprinkle
+    // would fall through to the network. Dropping the token is option A in
+    // #1717 — do it only with a browser-level regression test over those
+    // behaviors, not as a drive-by "fix" for the console warning.
     //
     // `allow-popups` only for cherry: a sprinkle's own srcdoc iframe sits one
     // level deeper there (host page → cherry iframe → sprinkle iframe), and
@@ -627,7 +636,7 @@ export class SprinkleRenderer {
       );
       iframe.addEventListener(
         'error',
-        (e) => {
+        () => {
           clearTimeout(timer);
           reject(new Error('full-doc iframe failed to load'));
         },
--- a/packages/webapp/src/ui/sprinkle-renderer.ts
+++ b/packages/webapp/src/ui/sprinkle-renderer.ts
@@ -582,18 +582,19 @@ export class SprinkleRenderer {
     // untrusted is scoped to the sync-fs channel nonce (realm capability
     // tokens) in `llm-proxy-sw-config.ts`; it is NOT general containment.
     //
-    // What `allow-same-origin` is (and is not) known to affect — verified
-    // during #1718 review: the bridge does NOT need it (pure postMessage
-    // both ways; the parent validates by `event.source` identity), and
-    // neither do the injected tags (`/lucide-icons.js` is a network-served
-    // static, theme CSS is inlined). What WOULD change with an opaque
-    // origin: sprinkle content loses same-origin storage (`localStorage` /
-    // IndexedDB throw), and the document no longer inherits the parent's
-    // service-worker controller, so SW-served VFS subresources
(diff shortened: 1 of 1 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/src/ui/sprinkle-renderer.ts:575-577]** — Correctly and prominently records that Chrome’s warning is legitimate and that the current sandbox is escapable.
- **[packages/webapp/src/ui/sprinkle-renderer.ts:581-587]** — Clearly warns maintainers not to treat the sandbox as a security boundary and accurately characterizes its remaining restrictions as protection against accidental behavior.
- The change is narrowly scoped, documentation-only, and introduces no behavior, dependency, or test-suite changes.

### Issues

*(none found)*

- **[packages/webapp/src/ui/sprinkle-renderer.ts:578-581]** — The stated reason that `allow-same-origin` is required for app assets to use the parent’s service worker appears incorrect.
  - A `srcdoc` document is not itself controlled by the parent’s service worker merely because `allow-same-origin` is present; in Chromium its `navigator.serviceWorker.controller` remains null. Removing `allow-same-origin` also does not prevent ordinary `<script src="/lucide-icons.js">` loads—they fall through to the network origin.
  - The examples do not support the assertion either: theme CSS is injected inline at lines 537-538, while `/lucide-icons.js` is an app-served script at line 545, not inherently a VFS resource. Consequently, the comment may discourage removing a security-sensitive token based on a dependency that does not exist.
  - This is primarily a problem with the plan’s premise, not its implementation.
  - Suggested fix: verify and document the actual feature that requires same-origin access, ideally with a browser-level regression test demonstrating that dropping the token breaks it. If no such dependency exists, remove this rationale and reconsider option A.

- **[packages/webapp/src/ui/sprinkle-renderer.ts:583-586]** — “The real containment for attacker-authored sprinkle content is the service worker’s security gates” overstates the referenced protection.
  - `maySetSyncFsNonce` is specifically a gate for publishing the sync-fs channel nonce (`llm-proxy-sw-config.ts:425-457`). It protects realm capability tokens; it is not general containment for same-origin sprinkle code, which can access the parent DOM and other same-origin application surfaces.
  - This wording could lead future security reviews to assume broader isolation than the architecture provides, despite the later statement that the sandbox isolates nothing.
  - Suggested fix: narrow the claim, for example: “The sync-fs capability-token boundary is enforced by `maySetSyncFsNonce`; sprinkle code otherwise shares the application origin and must not be considered isolated.”

#### 💡 Minor (Nice to Have)

*(none found)*

### Recommendations

Validate the `allow-same-origin` requirement with a real Chromium integration test before codifying it as architectural guidance. Reference the exact security gate (`maySetSyncFsNonce`) and explicitly distinguish its narrow sync-fs protection from general same-origin trust.

### Assessment

**Ready to merge?** No

**Reasoning:** The change is cleanly scoped, but its central architectural rationale appears factually wrong and its description of SW containment is broader than the referenced gate. Since the PR exists solely to preserve this security rationale for future maintainers, those statements should be corrected before merge.

---

catalan-adobe:
Review response — both Important findings verified against the code and accepted; addressed in f0e1b16ed:

**1. "allow-same-origin required for SW-served assets" — you're right, the original rationale was wrong.** Verified: the bridge is pure postMessage in both directions (`generateBridgeScript` uses `parent.postMessage(..., '*')`; the parent validates by `event.source === iframe.contentWindow` identity, which works from an opaque origin), `/lucide-icons.js` is a network-served static built into `dist/ui/` (vite IIFE entry, not SW-served), and theme CSS is injected inline. The comment now records what was actually verified, and names the two *plausible* dependents of the token — same-origin storage in sprinkle content, and the inherited SW controller for VFS-served subresources (`/preview/*`, project-serve root-relative paths) inside a sprinkle document — as unverified behaviors that need a browser-level regression test before option A ships.

**2. "SW gates are the real containment" — overstated, reworded.** The note now says the sync-fs nonce gate protects realm capability tokens specifically and is NOT general containment; the boundary for sprinkle content is the trust model (agent-authored code running same-origin with the app).

**Consequence for #1717:** since no hard dependency on `allow-same-origin` was demonstrated, option A (drop the token, kill the warning) is more viable than the original analysis suggested. I've switched this PR from `Closes` to `Refs` #1717 so the issue stays open to track the option-A empirical evaluation, and posted the corrected findings there.

Also swept up one unused error-listener parameter in the function being edited (boy-scout; the PR is otherwise still docs-only).

Issue #1698, opened 2026-07-27
Tray: a multi-minute leader stall is operationally invisible


catalan-adobe:
Follow-up to #1696.

#1696 stopped a stalled peer from being torn down, but the new state is silent. `onStalled`/`onRecovered` log at `warn`/`info`, and the prod log level is pinned to `ERROR` (`packages/webapp/src/core/logger.ts` — `setLogLevel` is never called at runtime), so a stall of up to ~5 minutes produces no signal at all: the chat just goes quiet.

Before #1696 the user at least got a (wrong) red card. Now they get nothing, which is better for the connection and worse for the human.

Wanted:
- a transient "leader is busy" affordance, distinct from the permanent `<slicc-error-card>` that `wc-chat-controller.#handleError` appends to the transcript for a real disconnect;
- and/or a follower runtime-status transition so `host` and any telemetry can see it.

Raising the stall log to `error` is explicitly *not* wanted — an expected stall is not an error and would pollute error telemetry.

PR #1719, opened 2026-07-28
fix(webapp): stop rendering transient drops as permanent errors; wire floatbar online

### LINKED ISSUE #1707: Follower: UI misreports leader connection state (transient drop shown as permanent error; floatbar always 'offline')
Observed on a live hosted cone (`sliccy.ai/cloud` → **Open ↗** → `/join/<sessionId>.<token>`) opened as a browser-tab follower in Chrome Canary, against a local build of `main` @ `5bd2a90b9` (5.78.0).

The follower's transport recovers from drops correctly. The UI does not tell the truth about it. Two independent spots:

## 1. A transient drop renders as a permanent failure in the transcript

After ~4 minutes of an otherwise healthy session, the chat thread showed:

> ⚠ **SOMETHING WENT WRONG**
> Connection to leader lost: Data channel closed
> [ ↺ Try again ]

with the composer frozen at `Connecting to leader…`.

Nobody clicked *Try again*. ~60–90s later it reconnected on its own, and a fresh round-trip confirmed the session was fully alive:

```
prompt : "Say only: RECONNECT-OK"
reply  : "RECONNECT-OK"
```

So the recovery path works — `DataChannelKeepalive` is wired at `tray-follower-sync.ts:298` and `startFollowerWithAutoReconnect` did its job. The problem is presentation: a self-healing blip is rendered as a terminal error card, permanently inlined into the conversation history, with a manual retry affordance that is not needed. After recovery the card stays in the transcript.

A transient connection state belongs in a dismissable banner tied to the live connection, not as an immutable message in the thread.

Secondary: the drop produced **no console trace at all**. `tray-follower-sync.ts:326-327` does `log.warn('Data channel closed')`, but nothing reached the page console — a CDP console watcher attached for the whole session captured the CDP-loop noise from #1706 and nothing about the disconnect. Only the UI knew the connection had dropped, which makes this hard to debug from telemetry.

## 2. `slicc-floatbar`'s `online` attribute has no producer

`packages/webcomponents/src/primitives/slicc-floatbar.ts` exposes `online` for the green status dot and the collapsed-badge tooltip:

```ts
// :216-221
get online(): boolean { return this.hasAttribute('online'); }
set online(value: boolean) { this.toggleAttribute('online', !!value); }

// :261
parts.push(this.online ? 'online' : 'offline');

// :278
if (this.online) nodes.push(h('span', { class: 'fdot', part: 'dot' }));
```

Nothing anywhere in `packages/*/src` ever sets it:

```
$ grep -rn "setAttribute('online'\|\.online = \|toggleAttribute('online'" packages/*/src --include=*.ts \
    | grep -v "slicc-floatbar\|\.stories\.\|network.ts"
(no matches)
```

Result: the green dot never lights and the pill reads `follower · $0.00 · offline` while the follower is connected and actively streaming a response. Same on the leader (`npx · live · $0.00 · offline`).

The component contract exists, the storybook exercises it, and the app never drives it. Either wire it to the sync connection state or delete the API.

## Why these are one issue

Both are the follower UI misreporting link state, both live in the presentation layer, and both are fixed by having one source of truth for "am I connected to the leader" and rendering it in the right place. Neither depends on the CDP wiring work.

## Verification

- `.agents/skills/cdp-smoke-test/` reproduces; the drop appeared unprompted after ~4 min idle on a live cone, so it needs a wait rather than a forced disconnect (or force one by killing the leader's channel).
- Tests: `packages/webcomponents/` already covers `slicc-floatbar` rendering — add a case asserting the app-side wiring sets `online`. For the error card, a `packages/webapp/tests/` case asserting a recoverable disconnect does not emit a thread-level error message.

catalan-adobe:
Fixes #1707 (both halves, as scoped in the issue: "both are the follower UI misreporting link state").

## 1. Transient drop no longer renders as a permanent transcript error

`FollowerSyncManager.handleDisconnect` synthesized an agent `error` event ("Connection to leader lost: …"). Because the sync manager doubles as the chat panel's `AgentHandle`, that event rendered a permanent `<slicc-error-card>` in the conversation (with a "Try again" affordance nobody needs — `startFollowerWithAutoReconnect` heals the drop on its own) and fired `trackError('error-card')` into RUM for every self-healing blip.

Now:
- The disconnect updates the follower runtime status and notifies the mount, which owns presentation (`onConnectionChange` → "Connecting to leader…" composer state; `onGaveUp` → terminal "Couldn't reach the leader" state). No transcript card, no RUM error for a transient.
- Leader-**sent** error events (genuine agent errors) still forward and still render cards.
- The disconnect log moves `warn` → `error`, so a real transport drop finally leaves a console trace at the prod log level (`ERROR`) — the issue's secondary finding that the drop was invisible to telemetry. An *expected* stall stays `warn` per #1698's explicit guidance.

## 2. `slicc-floatbar`'s `online` attribute gets its missing producer

New `wc-floatbar-online.ts` — the single source of truth the issue asked for — drives the attribute from the tray runtime statuses:

- lit while this float is **leading** a tray (reachable by followers) or **following** one (connected to a leader);
- a purely local float shows no dot (nothing is linked to it);
- a leader **stall** keeps the dot lit — a stalled follower is still `connected` (the channel is open and recovers on its own).

Installed at both mounts: `wc-tray.ts` (kernel float, covers both roles) and `wc-follower.ts` (the no-kernel `/join/…` follower). In each float the inactive role's status stays `inactive`, so one OR across both statuses is correct everywhere.

## Tests

- `wc-floatbar-online.test.ts` (6): seeding, connect/error/reconnect transitions (the exact regression), stall keeps the dot, leader role, either-role OR, uninstall.
- `wc-follower.test.ts`: mount-level integration test proving the follower mount actually installs the producer — the original bug was precisely an integration gap (component API and app both fine, never connected). Mutation-verified: commenting out the install fails this test.
- `tray-follower-sync.test.ts`: disconnect suite updated to the new contract (status updated, `onDisconnect` fired, **no** agent error event); the existing leader-sent-error forwarding test still passes unchanged.

## Verification

lint 7/7 · check-touched-exemptions · typecheck · full test suite (12,692) · webapp coverage gate · both builds.

### DIFF
--- a/packages/webapp/src/scoops/tray-follower-sync.ts
+++ b/packages/webapp/src/scoops/tray-follower-sync.ts
@@ -321,7 +321,8 @@ export class FollowerSyncManager implements AgentHandle {
       },
     });
     this.keepalive.start();
-    // Emit an error event when the underlying channel drops
+    // Treat an unexpected underlying channel drop as a disconnect (status +
+    // cleanup + onDisconnect; no transcript error event — see handleDisconnect)
     channel.addEventListener('close', () => {
       log.warn('Data channel closed');
       this.handleDisconnect('Data channel closed');
@@ -397,8 +398,8 @@ export class FollowerSyncManager implements AgentHandle {
    * Close the sync channel and clean up. Idempotent — once called, the
    * channel-close event that follows will short-circuit in
    * `handleDisconnect`, so a caller-initiated teardown can't trigger
-   * the error-state side effects (red-error follower status, `error`
-   * event emission, `onDisconnect` callback that drives reconnect
+   * the error-state side effects (red-error follower status, error-level
+   * disconnect log, `onDisconnect` callback that drives reconnect
    * logic) that `handleDisconnect` is designed to fire only on an
    * unexpected channel drop.
    */
@@ -1202,7 +1203,9 @@ export class FollowerSyncManager implements AgentHandle {
    */
   private async executeLocalCDP(
     requestId: string,
-    localTargetId: string,
+    // Unused: the wire message carries the follower-local target id for
+    // symmetry, but the transport routes by sessionId/method alone.
+    _localTargetId: string,
     method: string,
     params: Record<string, unknown> | undefined,
     sessionId: string | undefined
--- a/packages/webapp/src/scoops/tray-follower-sync.ts
+++ b/packages/webapp/src/scoops/tray-follower-sync.ts
@@ -632,11 +632,26 @@ export class FollowerSyncManager implements AgentHandle {
 
   /**
    * Handle a detected disconnect (keepalive dead, channel close/error).
-   * Updates follower status, emits an error event, cleans up, and notifies via onDisconnect.
+   * Updates follower status, cleans up, and notifies via onDisconnect.
+   *
+   * Deliberately does NOT emit an agent 'error' event (#1707): this manager
+   * is installed as the chat panel's AgentHandle, so such an event renders a
+   * permanent `<slicc-error-card>` in the transcript and fires `trackError`
+   * into RUM — for a state that `startFollowerWithAutoReconnect` usually
+   * heals in seconds. Connection state is presented by the mount instead
+   * (`onConnectionChange`/`onGaveUp` → composer placeholders) and recorded in
+   * the follower runtime status for `host`/telemetry. Leader-SENT error
+   * events (genuine agent errors) still forward via `handleLeaderMessage`.
    */
   private handleDisconnect(reason: string): void {
     if (this.disconnected) return; // prevent duplicate cleanup
     this.disconnected = true;
+
+    // `error` (not `warn`): the prod log level is ERROR, and a real transport
+    // drop previously produced no console signal at all — only the UI knew
+    // (#1707). An EXPECTED stall stays `warn` per #1698; an unexpected drop
+    // is precisely the thing an operator console should show.
+    log.error('Follower sync disconnected', { reason });
 
     // Update follower runtime status to error
     const current = getFollowerTrayRuntimeStatus();
@@ -648,9 +663,6 @@ export class FollowerSyncManager implements AgentHandle {
       // thing it described.
       stalled: false,
     });
-
-    // Emit error to UI listeners
-    this.emitEvent({ type: 'error', error: `Connection to leader lost: ${reason}` });
 
     // Clean up keepalive, CDP event forwarding, and sync channel
     this.keepalive.stop();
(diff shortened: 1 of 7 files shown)


catalan-adobe:
### Strengths

- **`packages/webapp/src/scoops/tray-follower-sync.ts:646-675`** cleanly separates transport state from transcript content: disconnects update runtime status, perform cleanup, log at production-visible `error` level, and notify reconnect orchestration without synthesizing an agent error.
- **`packages/webapp/src/scoops/tray-follower-sync.ts:733-747`** preserves genuine leader-sent agent errors, avoiding an overly broad suppression of error cards.
- **`packages/webapp/src/ui/wc/wc-floatbar-online.ts:38-56`** provides a focused, reusable producer that seeds synchronously, tracks both roles with correct OR semantics, and exposes proper subscription cleanup.
- **`packages/webapp/src/ui/wc/wc-floatbar-online.ts:59-64`** correctly treats only active leader and connected follower states as online; follower stalls remain online because they preserve `state: 'connected'`.
- **`packages/webapp/src/ui/wc/wc-follower.ts:467-471`** and **`packages/webapp/src/ui/wc/wc-tray.ts:581-584`** install the producer at both planned integration points.
- **`packages/webapp/tests/scoops/tray-follower-sync.test.ts:1255-1305`** verifies the new disconnect contract for close, error, and duplicate notifications, while **`:224-232`** protects leader-sent error forwarding.
- **`packages/webapp/tests/ui/wc/wc-floatbar-online.test.ts:37-100`** covers seeding, disconnect/reconnect, stalls, leader mode, role OR behavior, and uninstall. The mount-level check at **`packages/webapp/tests/ui/wc/wc-follower.test.ts:118-150`** directly guards the original integration gap.

### Issues

*(none found)*

*(none found)*

#### 💡 Minor (Nice to Have)

- **`packages/webapp/src/scoops/tray-follower-sync.ts:324`** — stale comments still claim channel drops emit an error event; the `close()` documentation at lines 400-401 likewise lists “error event emission” as a disconnect side effect.
  - This directly contradicts the PR’s central behavioral contract and could mislead future maintenance.
  - Update the comments, for example:
    ```ts
    // Handle an unexpected underlying channel drop
    ```
    and remove “`error` event emission” from the `close()` documentation.

### Recommendations

Correct the two stale comments so the documentation consistently reflects the new transport-state-versus-agent-error separation. No additional architectural changes appear necessary.

### Assessment

**Ready to merge?** Yes

**Reasoning:** The implementation matches both halves of the stated plan, preserves genuine agent errors, and has strong unit and mount-level regression coverage. Only minor outdated comments remain.

---

catalan-adobe:
Both stale comments from the review are fixed (channel-close listener header + `close()` doc block) — they now describe the actual contract: status update + cleanup + `onDisconnect`, error-level log, no transcript event. Also underscore-prefixed `executeLocalCDP`'s unused `localTargetId` param while in the file (kept positionally for wire-message symmetry). Full gate re-run green: lint 7/7, typecheck, sync suite 104/104.

PR #1773, opened 2026-07-30
feat(sliccstart): list iCloud tray sessions via CLI, with reveal consent

trieloff:
## Summary

Adds a way to see and use the iCloud-synced tray sessions from the command line. The macOS launcher gains a headless `Sliccstart --list-sessions [--reveal-urls]` subcommand, and the Go `slicc` CLI gains `list-sessions` plus a `<verb>-cloud` family (`follow`/`prompt`/`exec`/`watch`) that resolves a session's join URL from iCloud and runs the verb — so you never hand-copy a join URL.

## Why

Follow-up to the iCloud tray-session sync (v5.88.x). The sessions synced across a user's devices were only reachable through the launcher's SwiftUI list. This exposes them to the headless follower CLI. iCloud KVS is an Apple-only API readable only by the signed, iCloud-entitled `Sliccstart` binary, so putting it in the cross-platform Go CLI directly would need cgo + entitlement/profile signing and break the `CGO_ENABLED=0` cross-compile; instead the Go CLI shells out to the launcher.

**Swift (`packages/swift-launcher`)**
- `main.swift` parses `--list-sessions` **before** the SwiftUI app boots (`@main` moved off the struct); a normal launch still shows the GUI.
- `Models/TraySessionCLI.swift` (pure, unit-tested): arg parse, redacted-vs-full `SessionDTO` encoding (ISO-8601 dates for Go), consent outcome, dialog-button mapping, `consentKey`/`describeCaller`, `RevealConsentStore`.
- `Models/TraySessionCLIRunner.swift` (thin, untestable glue): reads the real store, the `NSAlert`, and best-effort caller identity (`getppid`/`proc_pidpath`/`SecCode`).

**Go (`packages/slicc-cli`)**
- `internal/cloud/cloud.go` (pure): `ParseSessions`, `ParseSelector` (`--index N` / `--session <id-prefix>`), `Select` (newest-first), `FormatTable`.
- `internal/cloud/resolve_darwin.go`: locates `Sliccstart.app` (`$SLICCSTART_APP` → `mdfind` → `/Applications` → `~/Applications`) and shells out. `resolve_other.go` returns `ErrUnsupported` off macOS.
- `cloud.go`: `cmdListSessions` + `cmdCloud`; a `cloudList` seam makes dispatch testable without a launcher. Verbs wired into `main.go`; telemetry allowlist updated.

## Security

Join URLs are **bearer secrets** (holder can attach as a follower and run commands on the leader), so:
- **Redaction by default** — `list-sessions` prints metadata only (opaque id, label, device, age); safe to pipe/log.
- **Consent gate on `--reveal-urls`** — a remembered "Always" (UserDefaults, keyed by caller code-signing id / path) wins; otherwise an `NSAlert` (Deny / Allow Once / Always Allow / Always Deny) naming the requesting process shows in a GUI session, and a **headless/SSH** caller is denied (exit 3) with guidance to grant once from the screen.
- Caller identity is spoofable by same-user code, so redaction-by-default is the real control and the dialog is a speed bump — called out in the docs, not oversold.

## Cross-cutting impact

- **Floats exercised**: Sliccstart (macOS) + slicc-cli (Go follower). Off macOS the cloud verbs report "macOS only".
- No change to the WebRTC follower path itself — the `-cloud` verbs resolve a URL then reuse the existing `cmdFollow`/`cmdPrompt`/`cmdExec`/`cmdWatch`.

## Test plan

- [x] Swift: `swift build`; full suite `swift test` (332 tests, 0 failures); coverage gate `swift-coverage-check.sh` (Lines 40.28 / Functions 44.20 / Regions 51.08, all above floor); SwiftLint (0 serious); `swift format lint --strict` clean.
- [x] Go: `go build ./...`; `make check` (gofmt, tidy, vet, golangci-lint clean, coverage 61.2% ≥ 58%).
- [x] Docs: doc-size gate (swift-launcher 19883/20000), Prettier, and doc-refs all pass.
- [x] Smoke: `Sliccstart --list-sessions` on the local build prints `[]` and exits 0 without launching the GUI.
- [ ] Not covered by CI: the real reveal dialog + a live cross-device session (needs a signed, iCloud-entitled build on a Mac with sessions).

## Documentation

- [x] `packages/swift-launcher/CLAUDE.md` (Headless CLI section; provisioning section refreshed) and `packages/slicc-cli/CLAUDE.md` (verbs, `internal/cloud`).

## Risk & rollback

- Blast radius: a new headless code path (guarded by an explicit flag) + new opt-in CLI verbs; the GUI launch path is unchanged. Revert the PR to remove both. Worst case for the reveal path is a denied request (no token disclosed).

## Checklist

- [x] No secrets, credentials, or tokens in the diff (test fixtures use `slicc.test` placeholders)
- [x] Cross-runtime contract checked (Swift emits the JSON `internal/cloud` parses; iOS follower reuse of `SyncedTraySession` unaffected)

### DIFF
--- a/packages/slicc-cli/CLAUDE.md
+++ b/packages/slicc-cli/CLAUDE.md
@@ -12,6 +12,8 @@ slicc <join-url> watch [scoop]                  Tail the leader's live agent out
 slicc <join-url> follow [--no-banner] [runner]  Stay connected; run leader-issued commands via <runner>
 slicc <join-url> follow --eval [repl]           Same, into ONE persistent REPL (state persists; see README)
 slicc update [--check]                          Self-update to the newest released CLI binary
+slicc list-sessions [--json]                    List iCloud-synced tray sessions (macOS only; no join URLs)
+slicc <verb>-cloud [--index N|--session <id>]   Resolve a session's join URL from iCloud, then run <verb>
 ```
 
 `watch` is a passive `tail -f` on the agent that mirrors the browser thread: it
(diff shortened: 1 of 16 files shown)


trieloff:
Thanks — all four addressed in 0977d5905:

- **P1 (unbounded wait on an older launcher):** `internal/cloud.List` now runs the launcher via `exec.CommandContext` with a bounded timeout (20s for metadata, 2m for the reveal path, which blocks on the consent dialog). A pre-`--list-sessions` Sliccstart that boots its GUI is killed at the deadline with a clear "update Sliccstart" error instead of hanging.
- **P1 (consent key vs. cloud verbs):** you're right that granting via `Sliccstart …` directly keys the approval to the shell, not the `slicc` CLI. The onboarding message no longer tells the user to launch Sliccstart directly; it now says to **re-run the same command** from the Mac's screen. Since the cloud verbs always spawn Sliccstart with `slicc` as the parent, granting through `slicc <verb>-cloud` on-screen keys the "Always Allow" to the same caller that the later SSH invocation uses, so it carries over.
- **P2 (bare selector value):** `ParseSelector` now errors on a trailing `--index`/`--session` with no value instead of leaking it into verb args.
- **P2 (Unicode truncation):** `truncate` operates on runes now, so multibyte labels/device names are cut on rune boundaries and stay valid UTF-8. (Full wide-char *column* alignment via wcwidth is left out of scope; the correctness bug — invalid UTF-8 — is fixed.)

PR #1784, opened 2026-07-30
feat(layouts): collapse to a single dock-tree layout + agent sizing/placement API

froesef:
## Summary

- Removes the fixed-grid (shell-grid) and IDE-panel-dock layout modes so `<slicc-dock-tree>` is the sole, always-on layout system. Chat and the four fixed tool panels become permanent, independent tree leaves with a hover-reveal move button for drag; `DockTreeSpec.locked` lets Cherry push a fixed, unmovable arrangement into an embedded follower (`mountSlicc({ layout })`).
- Adds an agent-facing `layout open/close/move/size` shell surface backed by new `SliccDockTree.setSurfaceSize()` (px/percent resize of any leaf), so the agent can rearrange and size panels/sprinkles/chat programmatically — not just via the five canned presets.
- Fixes a resize bug where dragging a divider kept resizing after the mouse button was released (pointer capture was set on the divider element, which a mid-drag re-render then removed from the DOM, silently releasing capture per spec — capture now lives on the stable host element). Aligns the manual-drag minimum to the same 2%-of-pair floor `setSurfaceSize` uses.
- Fixes a related gap: `layout open`/`layout close` now fire the same tool-panel activation lifecycle a dock-rail click does (previously only the click path loaded panel content).
- Shrinks the top nav bar's minimum height (44px → 36px).
- Deletes `<slicc-workbench-body>`/`<slicc-workbench-pane>`, obsolete once every panel is a permanent dock-tree leaf.
- **Security fix (independent of the layout work, split into its own commit):** `sudo/panel-responder.ts` now binds the real `confirm`/`prompt` at module init, before any page-realm code (a sprinkle, dip, or agent-authored panel) can reassign `globalThis.confirm` to auto-approve its own sudo prompts. No native available (worker/headless) now denies rather than defaulting open.
- **New panel system**, added on top of the dock-tree work: `SliccPanel`/`SliccLayout` web components plus a five-zone layout engine, panel registry, save/load of layout documents (`/workspace/layouts/`, or `/etc/slicc/layouts/` when gated), and a trusted layer (`ui/wc/trusted-layer.ts`) that mounts fixed chrome and approval overlays in a spoof-proof, un-occludable way.

### Feature-flag scope — what's gated and what isn't

Only the **panel-layout feature itself** is behind `panel-layouts` (off by default, uniform across every float, no `floatDefaults`/URL-param escape hatch): whether `panelizeShell` re-parents the shell into panels at boot, and whether a Cherry-pushed layout gets applied. With the flag off, the dock-tree from this PR remains the shipping layout — that path is unchanged by the panel work.

Not gated, and shipping unconditionally in every float regardless of the flag:
- The `SliccPanel`/`SliccLayout` custom elements themselves — `customElements.define(...)` runs as a module-load side effect in `@slicc/webcomponents`, so the elements are registered (just unused/dormant) with the flag off.
- The `trusted-layer.ts` hardening and the `confirm`/`prompt` capture fix — both are called out in their own commit messages as security fixes independent of the layout work, intentionally not tied to the flag.

## Test plan

- [x] `npm run typecheck` (all 9 targets)
- [x] `npm test` (full suite, 12,994 passed — one unrelated `node-server` sudo endpoint network flake confirmed passing in isolation)
- [x] `npm run verify` / `npm run lint` (biome, prettier, custom-lints, complexity, manifest, deadcode)
- [x] New `slicc-dock-tree` tests for `setSurfaceSize` (percent/px, locking, no-op cases) and the resize-hang regression guard
- [x] New `layout-command`/`apply-layout`/`wc-sprinkles` tests for `open`/`close`/`move`/`size` parsing, dispatch, and direct tool-panel activation
- [x] New panel-system tests: `layout-interaction`, `layout-schema`, `panel-registry`, `slicc-layout`, `slicc-panel`, `zone-ops`, `trusted-layer`, `trusted-layer-stacking`
- [x] Feature-flag gate tested both directions: a pushed Cherry layout applies with `panel-layouts` on and is ignored with it off; the Experimental features… dialog lists the toggle and persists it
- [x] Manually verified `layout open`/`layout size` live in a rebuilt dev harness (file panel populated, exact percent resize confirmed via `getTree()`)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


trieloff:
Please use #1817 when it lands

PR #1817, opened 2026-07-31
feat: centrally-managed feature flag system with experimental settings pane

trieloff:
## Summary

Adds a runtime feature flag system shared by the web UI and Cherry, with flag values authored centrally in the Cloudflare worker and served from `GET /api/flags`. The first flag, `experimental-settings`, controls a standalone **Experimental features…** avatar-menu item and dialog — on for the web UI, off for Cherry. It is currently `userToggleable: false`, so the dialog honestly shows an empty state rather than a toggle.

## Why

There was no unified flag mechanism: user-facing toggles were scattered across independent localStorage modules (`discovery-preference.ts`, `soundscape.ts`, `theme.ts`, `timestamp-preference.ts`), each with its own key and defaults, and per-float capability gating lived inline (e.g. `SIDE_PANEL_FEATURES`). That made it impossible to ship a feature to one float but not another without a code change, and impossible to change a default without a webapp deploy.

Planned and reviewed in a workspace spec; implemented in three waves with a verifier pass between each.

- **String values, boolean-coerced.** Flag values are strings on the wire and in storage. `isFeatureEnabled(id)` applies one documented coercion rule; `getFeatureValue(id)` returns the raw string. Chosen over booleans so variant/experiment values can be added later without a schema change.
- **Three-level precedence** (highest wins): local user override → worker-served value → bundled registry default. The bundled default is what keeps behavior correct offline, on first paint, and when the worker is unreachable — the remote fetch is never on the critical path. Flags such as `experimental-settings` with `userToggleable: false` deliberately ignore local user overrides.
- **Registry in `core/`.** `packages/webapp/src/core/feature-flags.ts` holds flag ids, metadata and bundled per-float defaults. The float is passed in as an argument rather than resolved internally, so `core/` never imports `ui/` (keeps `lint:ui-back-edges` clean).
- **Worker as central config.** `GET /api/flags?float=<float>` resolves a base map with a per-float overlay, authored in `wrangler.jsonc`. Edge-cached ~5 min via the Cache API, keyed including the `float` param. CORS mirrors the `oauth-exchange.ts` allowlist with `Vary: Origin`, since the Cherry embed is a cross-origin iframe on third-party hosts. Bad config falls back to the base map rather than 500ing.
- **Origin routing.** Flags come from the worker, *not* the local node-server bridge, so `resolveApiUrl()` is deliberately not used — the hosted worker base URL is resolved instead. Relevant to review pattern #9.
- **Cherry is data, not code.** Cherry receives the full payload and simply has `experimental-settings` off. No `if (cherry)` branch in the UI. Deliberately kept separate from `SIDE_PANEL_FEATURES` in `cherry-panel-protocol.ts`, which is a mount-time hard gate rather than a user-visible flag; documented as such.
- **Lazy hydration.** Feature-flag hydration is dynamically imported after the boot-critical path, reducing the main entry from 25.71 kB to 23.59 kB and leaving about 410 B below the 24 kB budget without changing synchronous defaults/cache behavior.
- **New localStorage keys**: a single JSON override map plus a per-float last-known-good cache. Guarded so a missing or throwing `localStorage` (worker context) never throws.

## Cross-cutting impact

- **Floats exercised**: boot wiring goes through a single `setupFeatureFlagsForPage` entry point, verified across standalone, extension, extension-detached, electron-overlay, hosted-leader, connect, follower and cherry. This centralisation was added in the final wave specifically because the Cherry regression test — written to exercise the real boot path rather than calling the registry directly — surfaced a float-consistency gap. The follower's avatar menu deliberately omits **Experimental features…**; that exclusion is documented in code and follows review pattern #8.
- **Cache isolation**: the last-known-good cache is per-float, so a Cherry entry cannot leak into a web-UI float in the same browser profile. Explicitly probed in review.
- **Async lifecycle**: the worker refresh is timeout-bounded and non-blocking; timeout, non-2xx, network throw and malformed JSON all fall back silently with no unhandled rejection and no boot delay.
- **Flags resolve at boot.** A centrally changed value takes effect on next load, not mid-session. Documented.
- **Routes-mirror rule**: `ROUTES_INDEX_BODY`, `tests/index.test.ts` and `tests/deployed.test.ts` all updated.

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] 224 focused tests across `packages/webapp/tests/` and `packages/cloudflare-worker/tests/`, mirroring the changed `src/` paths
- [x] `npm run test`
- [x] `npm run test:coverage`
- [x] `npm run build`
- [x] `npm run bundle-size` — main entry 23.59 kB / 24 kB
- [x] Touched-file debt gate (`check-touched-exemptions.mjs`)
- [x] New behavior covered by unit tests: per-float defaults, remote-over-default, override-over-remote, coercion table, corrupt storage, absent `localStorage`, non-user-toggleable override ignored, all four fetch failure modes, per-float cache isolation, standalone dialog empty state, avatar-menu visibility, deliberate follower omission, and Cherry parity through the real boot path
- [x] **UI change — [screencast attached](https://github.com/ai-ecoverse/slicc/pull/1817#issuecomment-5141286976).** It shows the standalone **Experimental features…** dialog and honest empty state, confirms Account settings has no Experimental section, and verifies the genuine Cherry follower omits the menu item.

**Post-rebase verification**: strict lint/format/custom-lint/complexity/manifest/dead-code gates, typecheck, the full test suite, coverage, full build, touched-file debt gate, and bundle-size gate all pass on the rebased head.

## Documentation

- [x] `packages/webapp/CLAUDE.md` — add-a-flag procedure, value/coercion rule, precedence, boot-time resolution caveat
- [x] `packages/cloudflare-worker/CLAUDE.md` — the `/api/flags` endpoint, where config is authored, routes-mirror reminder
- [x] `packages/cherry/CLAUDE.md` — pointer to the shared registry with Cherry's own central values
- [x] `README.md` — no change needed; no existing user-facing behavior changed

## Risk & rollback

- **Blast radius**: additive. With the worker unreachable every float falls back to bundled defaults, which reproduce today's behavior.
- **Kill switch**: central values can be neutralised by a worker-only deploy setting every flag to its bundled default — no webapp deploy required.
- **Rollback**: clean single-branch revert. No persisted data migration; clearing the override key restores resolved values.
- **Needs manual verification**: the Cherry path in a real cross-origin embed (CORS + `frame-ancestors` interaction) is not something CI covers.

## Checklist

- [x] No secrets or credentials in the diff
- [x] Public API and endpoint changes are intentional and reflected in docs
- [x] Contract used by other floats checked (worker → hydration → registry → UI traced end to end)
- [ ] Commits are focused and package-local where possible — **partially**: `7b39a10da` mixes the worker endpoint with two webapp registry files because two agents wrote concurrently. Content is correct; only the commit boundary is off.

---

### DIFF
--- a/packages/cherry/CLAUDE.md
+++ b/packages/cherry/CLAUDE.md
@@ -87,6 +87,11 @@ mountSlicc({
   at mount time and sent in `handshake.welcome`; there is no runtime toggle.
   Separate from `capabilities` (which gates agent _powers_ over the host page);
   features gate _UI surfaces_ shown to the user.
+- **Feature flags are separate:** the `?cherry=1` webapp boot uses the shared registry in
+  `packages/webapp/src/core/feature-flags.ts` with Cherry defaults (`experimental-settings`
+  is off). This is not `SIDE_PANEL_FEATURES` in
+  `packages/chrome-extension/src/cherry-panel-protocol.ts`: that constant is host-provided,
+  mount-time `CherryFeatures` panel visibility sent in `handshake.welcome`, not remote flag config.
 - `theme` accepts a `SliccTheme` object (`{ id, name, base, tokens, css?,
 disableShader?, components? }`) that the SDK serializes as JSON in the
   handshake welcome. The follower applies it on boot via `applyCherryTheme`,
--- a/packages/cloudflare-worker/CLAUDE.md
+++ b/packages/cloudflare-worker/CLAUDE.md
@@ -22,6 +22,7 @@ the built SLICC webapp as static assets to browser visitors.
 | `src/install-cli.ts`                                                                   | `/install-cli` POSIX installer script + `/download/slicc-cli/:target` release-asset resolver                        |
 | `src/llms-txt.ts`                                                                      | `/llms.txt` builder                                                                                                 |
 | `src/rel-docs.ts`                                                                      | `/rel/:name` — dereferenceable docs for SLICC custom rels                                                           |
+| `src/flags.ts`                                                                         | `/api/flags` string-map resolution, CORS, and per-float edge caching                                                |
 | `src/oauth-exchange.ts`, `src/oauth-registry.ts`                                       | OAuth callback relay (`/auth/callback`)                                                                             |
 | `src/auth/cloud-callback.ts`                                                           | `/auth/cloud-callback` IMS popup callback for the cloud dashboard                                                   |
 | `src/cloud/cloud-sessions-do.ts`                                                       | `CloudSessionsDurableObject` — per-user state for `/api/cloud/*`                                                    |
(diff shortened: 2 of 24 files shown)


trieloff:
## Experimental features screencast

Leader: the avatar menu shows **Experimental features…** immediately after **Export transcript**. It opens the standalone **Experimental** dialog with the honest empty state, “No experimental features are available right now.” After dismissing with **Done**, **Account settings** is opened and contains no Experimental section.

The final beat uses a genuine Cherry follower mounted through the Cherry SDK and connected to the leader; its avatar menu omits **Experimental features…** because the flag is off for that float.

**Video:** [Play or download the Safari-compatible H.264 MP4 screencast](https://ai-aligned-gh--ai-ecoverse.agentbin.net/13eac2c3175b64b422b1d12516405afaa376d2daba6b7f7eed7e30b128a28d39.mp4)

PR #1857, opened 2026-08-02
fix(tray): restrict follower filesystem access

### LINKED ISSUE #1833: fix(tray): follower fs.request reads the leader's VFS with no path restriction
Found while implementing #1794 (iOS `fs.*` client). Not iOS-specific — it applies to every follower, and the browser follower has been able to do this all along.

## The route

`follower-dispatch.ts:279-294` routes a follower's `fs.request` with `targetRuntimeId: 'leader'` to `FsRouter.executeLocalFs`, which calls `handleFsRequest(vfs, request)` with the leader's own VFS handle:

```ts
const vfs = this.context.options.vfs;
...
const responses = await handleFsRequest(vfs, request);
```

There is **no path restriction on this route**. No `RestrictedFS`, no allowlist, no approval. The handle is passed through as-is (`fs-router.ts:38-60`), and `handleFsRequest` is a pure dispatcher over eight VFS methods with no notion of who is asking.

For contrast, the agent's own handle *is* wrapped: `createSudoFs` at `scoop-context.ts:448`. That wrapper is what makes writes to `/etc/sudoers` and `/etc/sudoers.d/*` require approval — hardcoded self-protection, independent of policy (`shell/sudo/sudoers.ts:43-46`). The tray route does not go through it.

## What is actually reachable today, and why that is luck

I want to be precise here, because my first read of this was wrong and worth correcting explicitly.

The object the webapp hands the leader as `vfs` is **not** a `VirtualFS`. `createLeaderOptionsFactory` (`ui/wc/wc-tray.ts:229-238`) builds a lazy proxy with exactly two methods, cast to satisfy the type:

```ts
vfs: {
  async stat(path) { return (await deps.openFs()).stat(path); },
  async readFile(path, options) { return (await deps.openFs()).readFile(path, options); },
} as import('../../fs/virtual-fs.js').VirtualFS,
```

`handleFsRequest` calls `readFile`, `writeFile`, `stat`, `readDir`, `mkdir`, `rm`, `exists`, and `walk`. Six of those hit an undefined property, throw a `TypeError`, and come back as `{ ok: false, error: '... is not a function' }`.

So the situation is:

- **Arbitrary read is live.** Any connected follower can `readFile` and `stat` **any path** in the leader's VFS. That includes anything under a mounted directory, which bridges to real local disk via `FileSystemDirectoryHandle`.
- **Write, delete and enumerate are unreachable — by accident.** Nothing denies them. The proxy just does not implement them. The comment above it says it exists "for `preview.request` handling", so the narrowness is incidental to a different feature, not a security boundary.

That second point is the part worth acting on. The protective effect vanishes the moment anyone widens the proxy, and **widening it is on the roadmap**: the files tree and the freezer rail (#1795) both need `readDir`. Whoever adds `readDir: (p) => fs.readDir(p)` to that object is unlikely to realise they are simultaneously exposing `writeFile`, `rm` and `mkdir` to every connected follower — including writes to `/etc/sudoers`, which the agent cannot perform without approval.

Secrets themselves are not exposed this way: `docs/secrets.md:249` is explicit that PATs live in `~/.slicc/secrets.env` or extension options, not in the VFS.

## Suggested fix

Gate the route rather than relying on the proxy's shape:

1. Wrap the handle used by `executeLocalFs` in a `RestrictedFS` (or an equivalent path gate) scoped to what followers are actually meant to reach — `/sessions`, `/workspace`, `/tmp/upload`, rather than `/` and mount roots.
2. Decide write policy explicitly. Either refuse mutating ops on the follower route outright, or route them through the same approval path the agent uses. The video-attachment surface (#1797) wants `writeFile` to `/tmp/upload/`, which is a good argument for a narrow write allowlist rather than a blanket refusal.
3. Add a regression test that a follower-originated `fs.request` for a path outside the allowlist is refused, so the guarantee does not depend on which methods a UI-layer proxy happens to define.

A comment on the `wc-tray.ts` proxy noting that its narrowness is load-bearing until (1) lands would be cheap insurance in the meantime.

## Note

`packages/ios-app` ships an `fs.*` client in #1832. It is a requester only and adds no capability the browser follower did not already have; it documents the two-method limitation rather than working around it.

trieloff:
Closes #1833

## Solution
Authenticated followers can read and mutate the leader VFS except `/proc`; the page now forwards follower writes through the existing writable VFS RPC.

## Misc
The full suite has an unrelated ordering failure in `fake-llm/server.test.ts`; JS builds and bundle checks pass, while Swift is unavailable in this VM.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG&panel=chat)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

👋 I've got this PR.

trieloff:
This is too restrictive. List the paths that exist here in a comment table and include arguments for and against excluding each.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

These are the code-defined leader-VFS namespaces; a profile may also add mount points at arbitrary absolute paths.

| Path | What lives there | Argument for excluding | Argument against excluding |
|---|---|---|---|
| `/` | Namespace root | Raw listing reveals every private and mounted namespace | A filtered listing would let followers discover safe capabilities |
| `/workspace` | Project files, installed skills, packages | May contain source or user data not intended for another device | It is the primary working tree and the strongest cross-device use case |
| `/sessions` | Frozen transcripts, snapshots, cost index | Conversation history can contain sensitive prompts and outputs | Followers need it for cross-device history and session restoration |
| `/shared` | Shared instructions, sprinkles, sounds, workflow output, notifications | Shared artifacts can still contain user-generated sensitive data | It is explicitly cross-scoop/shared and followers may need assets and outputs |
| `/scoops` | Per-agent workspaces and local sudoers | Exposing it breaks isolation between agents and leaks intermediate work | Debugging or resuming a scoop from a follower may need selected output |
| `/home` | User/profile state such as onboarding data | Identity and configuration data should remain leader-local | Some follower UX may need non-secret preferences or profile metadata |
| `/tmp` | Screenshots, downloads, scratch files, uploads | Arbitrary tools place transient sensitive material here | Followers need more than uploads to preview screenshots and generated artifacts |
| `/mnt` | User-mounted local, S3, or DA content | Mounts can expose personal or enterprise data and remote credentials' reach | Cross-device access to deliberately mounted content is valuable |
| `/etc` | Global sudo policy and grants | Reveals security policy and persisted approvals | Read-only policy visibility can help explain denied operations |
| `/proc` | Live process metadata | Leaks active commands, ownership, cwd, and runtime state | Useful for remote diagnostics and process monitoring |
| custom mount path | A mount may be placed outside `/mnt` | No static allowlist can infer the user's sharing intent | Explicitly selected mounts may be the exact content the follower needs |

My default would be read-only `/workspace`, `/sessions`, `/shared`, and `/tmp`; a filtered `/`; keep `/scoops`, `/home`, `/etc`, and `/proc` excluded; and require explicit opt-in for each mount. `/dev/null` is synthetic in `RestrictedFS`, not a leader-VFS root.

trieloff:
The follower is most likely a human, who should be authorized to access, not a bot. So we should err on the side of openness. I'd make everything writable except proc.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

Implemented in `151f3587b`: authenticated followers can now read and mutate every normalized path except `/proc`, and write operations are wired through the existing writable VFS RPC.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

@trieloff all test checks passed, but [`gated-review`](https://github.com/ai-ecoverse/slicc/actions/runs/30740625134/job/91477279820) requires approval; please approve the workflow run.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

Addressed in `f97ae4bbb`: the iOS docs now describe the writable proxy accurately, and the architecture matrix explicitly records that authenticated follower writes to `/etc/sudoers*` intentionally bypass the agent approval wrapper.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ0QE9R8B3K9ESDJ5N9NQQNG)⚡</sup>

Both notes are addressed in `f97ae4bbb`: the stale iOS proxy documentation is corrected, and the intentional `/etc/sudoers*` approval bypass is recorded in the architecture matrix.

PR #1907, opened 2026-08-04
fix(ios): stop transcript horizontal scrolling from switching scoops

### LINKED ISSUE #1892: bug: iOS: swipe left/right conflicts with horizontal scrolling of code blocks
### Summary

If you have a code block in chat that exceeds the width, it scrolls horizontally. Good. This horizontal scroll is also interpreted as a horizontal swipe gesture causing a switch from cone to scoop. annoying.

### Float (runtime)

Not sure

### Surface (best guess is fine)

UI (chat panel, terminal, files, tabs)

### Area (apply matching `area/*` labels)

- [ ] `area/webapp` — browser app core (packages/webapp)
- [ ] `area/extension` — Chrome extension (packages/chrome-extension)
- [ ] `area/node-server` — Node CLI / Electron server (packages/node-server)
- [ ] `area/worker` — Cloudflare worker (packages/cloudflare-worker)
- [ ] `area/swift` — native macOS apps (swift-launcher / swift-server)
- [x] `area/ios` — iOS follower app (packages/ios-app)
- [ ] `area/docs` — README, CLAUDE.md, AGENTS.md, docs/, SKILL.md
- [ ] `area/build` — build, CI, release, monorepo tooling

### Reproduction

```markdown
.
```

### Suspected root cause / location (optional)

```markdown

```

### Logs / error output (optional)

```shell

```

### SLICC version / commit (optional)

_No response_

### OS / browser / Node (optional)

_No response_

### Pre-flight

trieloff:
Closes #1892

## Problem

`ChatView.liveConversation` attaches the scoop-switch drag as a `.simultaneousGesture` over the whole transcript. Because it is *simultaneous*, a horizontal drag that a nested `ScrollView(.horizontal)` consumes for scrolling **also** reaches the outer gesture and fires `swipeToNextScoop()` / `swipeToPreviousScoop()`. Scrolling a wide code block sideways silently switched scoops.

Affected nested scrollers: fenced code blocks and GFM tables (`MarkdownText`), expanded tool bodies (`MessageBubble`), and the attachment chip strip (`AttachmentChips`).

## Approach

A rubber-band handoff. The inner scroller owns the drag until it is pinned at the edge the drag pulls away from; only then does scoop navigation take over. Drag right hands off only at the leading edge, drag left only at the trailing edge. A non-overflowing scroller is at both edges, so it never blocks.

Edge state is sampled at drag **start**, before the inner scroller moves — sampling at drag end would read the post-scroll position and get the wrong answer.

- `Models/SwipeArbiter.swift` — the pure rule, unit-testable with no view hierarchy.
- `Views/HorizontalScrollGuard.swift` — the `horizontalScrollGuard()` modifier that publishes each scroller's frame and scroll metrics and resolves a drag's origin against them.

Deployment target is iOS 17, so the iOS 18 scroll APIs (`onScrollPhaseChange`, `onGeometryChange`) were unavailable; the modifier uses preference keys and a named coordinate space instead.

## Three non-obvious defects found while verifying

The pure-logic unit tests passed while the fix still failed on a live fixture. Instrumented repro on the simulator surfaced three separate issues, all now fixed and covered:

1. **Callback order is parent-first.** SwiftUI does not guarantee the order of `onChanged` for simultaneous gestures, and empirically the outer 40pt `DragGesture` fires *before* the inner zero-distance one (outer at 08:15:24.044, inner at 08:15:24.045). The outer gesture therefore captured `nil` context for the entire drag. Capture is now order-independent: the inner gesture late-fills the origin when the outer is already active.
2. **Unknown context failed open.** `nil` context and "ordinary content" were indistinguishable, and both allowed the swipe — which is why the plumbing gap became a user-visible bug rather than a no-op. Unknown origin inside a guarded region now fails **closed**.
3. **Edge math used the wrong width.** The modifier applies `-8pt` horizontal padding, so the physical viewport is 16pt wider than measured. A scroller at offset 294.33 with a 338pt effective viewport was at its trailing max but reported as "middle". Edge math now uses the effective viewport.

## Testing

- `SwipeArbiterTests` covers the rule plus both gesture orderings and the fail-closed case.
- `FixtureConversationUITests.testCodeBlockScrollUsesRubberBandScoopHandoff` covers the code-block path end to end on the leaderless fixture route.
- SwiftLint, strict `swift format lint`, and the full iOS build/test/coverage gate pass. Coverage rose to 67.99% lines / 59.65% functions / 60.59% regions.

## Known gaps (non-blocking)

No end-to-end UI coverage yet for the table, tool-body, or attachment guards, frozen-session dismissal, the rightward leading-edge handoff, or the fail-closed unknown-metrics state. The arbiter logic and the code-block path are covered.

---

### DIFF
--- a/packages/ios-app/CLAUDE.md
+++ b/packages/ios-app/CLAUDE.md
@@ -63,9 +63,10 @@ the drag pulls away from (right at leading, left at trailing). Freeze edge state
 at drag start. Capture must tolerate either inner/outer callback order, and an
 unknown context in a guarded region fails closed. Edge math uses the effective
 viewport, including both 8pt expansions from negative horizontal padding.
-The transcript observer is a UIKit pan recognizer that explicitly permits
-simultaneous recognition; iOS 26 no longer makes a descendant SwiftUI gesture
-simultaneous with an ancestor gesture. The inner gesture still snapshots edges.
+On iOS 18+, the transcript observer uses `UIGestureRecognizerRepresentable` to
+explicitly permit simultaneous recognition; iOS 26 no longer makes a descendant
+SwiftUI gesture simultaneous with an ancestor gesture. iOS 17 keeps the SwiftUI
+fallback, and the inner gesture still snapshots edges.
 Ordinary transcript navigation and vertical scrolling stay unchanged. Because
 the target is iOS 17, use preference/geometry APIs, not iOS 18 scroll APIs.
 
--- a/packages/ios-app/CLAUDE.md
+++ b/packages/ios-app/CLAUDE.md
@@ -63,10 +63,10 @@ the drag pulls away from (right at leading, left at trailing). Freeze edge state
 at drag start. Capture must tolerate either inner/outer callback order, and an
 unknown context in a guarded region fails closed. Edge math uses the effective
 viewport, including both 8pt expansions from negative horizontal padding.
-On iOS 18+, the transcript observer uses `UIGestureRecognizerRepresentable` to
-explicitly permit simultaneous recognition; iOS 26 no longer makes a descendant
-SwiftUI gesture simultaneous with an ancestor gesture. iOS 17 keeps the SwiftUI
-fallback, and the inner gesture still snapshots edges.
+On iOS 18+, each guarded scroller uses `UIGestureRecognizerRepresentable` to
+resolve its own handoff; iOS 26 no longer makes a descendant SwiftUI gesture
+simultaneous with an ancestor gesture. The parent handles ordinary content and
+iOS 17 keeps the SwiftUI arbitration path.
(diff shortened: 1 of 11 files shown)


trieloff:
Reviewed the automated review's three non-blocking observations. No follow-up code change is warranted in this PR: the negative horizontal padding is part of the effective viewport behavior and changing it to hit-shape-only risks the verified trailing-edge math; the Combine import and environment default are unrelated optional cleanup, and both environment injection sites are present. Keeping scope focused on #1892.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ6ET7C47C716T0APD36SDZX)⚡</sup>

👋 I've got this PR.

Picking reviewers is your call — I won't request anyone myself.

PR #1908, opened 2026-08-04
feat(ios-app): render scoop lifecycle state and context fullness

### LINKED ISSUE #1826: feat(ios-app): render scoop lifecycle state and context fullness from ScoopSummary
Part of the iOS follower catch-up tracking issue (#1785). No blocking dependency; the wire fields land with the segmented-agent-tabs work.

## Background

The leader toolbar was reworked from "agent pills" to segmented agent tabs. Each segment now carries two signals beyond identity:

- **lifecycle state** — `working` | `broken` | `initializing` | `idle`, driving the glyph and the eyes treatment
- **context fullness** — a 0...100 number, drawn as an arc around the glyph, plus a `near-limit` severity band

Both are computed on the leader in `packages/webapp/src/ui/wc/wc-live.ts` and rendered by `packages/webcomponents/src/switcher/slicc-agent-tabs.ts`.

Followers had neither. `ScoopSummary` carried only identity metadata, so a follower rendered every scoop as a confident `idle` at 0% fullness — not "unknown", but *wrong*, which is worse. A scoop that is mid-turn or nearly out of context looked calm on the phone.

## What lands on the TypeScript side

`ScoopSummary` (`packages/shared-ts/src/tray-sync-protocol.ts`) gains two fields:

- `state` — the lifecycle state above
- `fill` — context-window fullness on the same 0...100 scale the tabs use

The leader populates them in `packages/webapp/src/ui/wc/wc-tray-scoops.ts`, which joins registered scoop metadata with the descriptors the leader toolbar has already rendered, so the state mapping stays single-sourced in `wc-live.ts` rather than being recomputed for followers. `scoops.list` is broadcast on state transition, not only on the 5s reconciliation timer, so a follower tracks the leader rather than sampling it.

The Swift mirror in `packages/ios-app/SliccFollower/Models/SyncProtocol.swift` gains both fields as optional, so a new follower driven by an older leader decodes cleanly.

## What this issue covers

Decoding the fields is not the deliverable — that is a one-line mirror change and ships with the protocol work. **Rendering them on the phone is the deliverable.**

iOS has no scoop switcher surface today. `App/AppState.swift` tracks scoops and the selected JID; there is no per-scoop status affordance anywhere in the view layer. So this is additive UI, not a port of an existing iOS control.

- Surface the lifecycle state per scoop. The web treatment is a glyph plus an eyes state (`broken` shows dead eyes, `initializing` shows none), which is idiomatic to the web component set and should not be copied literally into SwiftUI. Pick a native treatment and say why.
- Surface context fullness. The web draws an arc that sweeps with fill and rotates while `working`. A determinate `Gauge` or a ring around the scoop avatar is the natural SwiftUI analogue. Do not animate it into meaninglessness — the arc's job is to be readable at a glance, not to spin.
- Respect `UIAccessibility.isReduceMotionEnabled`. The web version pauses the arc animation rather than removing the arc, so the fullness reading survives with motion off. Mirror that: the cue may not depend on motion to exist.
- Handle the absent case honestly. Both fields are optional; a leader that does not send them must not render as `idle` at 0%, which is the exact bug this closes. Absent means unknown and should look unknown.
- Accessibility labels should carry both signals. The web `aria-label` reads roughly `"researcher: working, 64% context fill"`.

## Note on where this sits

The scoop switcher is not on the iOS deliverable list in #1785 as its own surface, and the narrow-viewport dock/workbench IA sub-issue (#1802) is where a scoop-selection surface would naturally live. If that lands first, this slots into it. If this lands first, it should not invent a second navigation model that #1802 then has to unpick — coordinate rather than racing.

## Acceptance

- A scoop that is mid-turn on the leader reads as working on the phone.
- A scoop near its context limit is visually distinguishable from an empty one.
- A leader that sends neither field produces an unknown treatment, not a false idle.
- The fullness reading is present with reduced motion enabled.
- The corpus fixture populates both fields and the Swift decoder round-trips them.
- XCUITest coverage of the state and fullness treatments, driven by fixtures rather than a live leader.

trieloff:
## Summary

- model scoop lifecycle and context fullness without coercing absent fields
- render lifecycle badges and a determinate context ring in the existing iOS scoop switcher
- add leaderless XCUITest fixtures, reduced-motion coverage, and package documentation

## Why

The iOS follower previously treated missing scoop status as a confident `idle` at 0%, so a working or nearly full scoop looked calm on the phone. This closes that false-idle bug: absent values now receive an honest unknown treatment, while transmitted lifecycle and fullness values remain visible and accessible.

The switcher uses a native SwiftUI circular `Gauge` around the existing avatar plus SF Symbol lifecycle badges rather than copying the web glyph/eyes SVG treatment. This keeps both signals legible at navigation-bar scale and follows platform conventions. Working and initializing badges may pulse, but reduced motion pauses that animation without removing the fullness ring or its accessibility value.

## Verification

- `npm run lint`
- `npm run verify`
- `node packages/dev-tools/tools/check-touched-exemptions.mjs`
- `npm run typecheck`
- `npm run test`
- `npm run test:coverage`
- `npm run build`
- `npm run bundle-size`
- `npm run lint:swift`
- `npm run lint:swift:format`
- `swift format lint --strict --parallel --recursive SliccFollower Package.swift`
- `./packages/dev-tools/tools/swift-coverage-check.sh --xcodebuild SliccFollower packages/ios-app SliccFollower`

Closes #1826

---

### DIFF
--- a/packages/ios-app/CLAUDE.md
+++ b/packages/ios-app/CLAUDE.md
@@ -14,6 +14,7 @@ This file covers the iOS follower app in `packages/ios-app/`.
 | ---------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
 | `SliccFollower/App/SliccFollowerApp.swift`, `App/AppState.swift`                                     | App entry + central `@MainActor AppState`                                                                               |
 | `SliccFollower/Models/SyncProtocol.swift`                                                            | Partial `Codable` mirror of `packages/shared-ts/src/tray-sync-protocol.ts`                                              |
+| `SliccFollower/Models/ScoopStatus.swift`                                                             | View-independent scoop lifecycle and context-fullness presentation model                                                |
 | `SliccFollower/Models/ChatMessage.swift`, `Models/TrayTypes.swift`, `Models/TrayChunkFraming.swift`  | Chat/signaling types and transport chunk reassembly                                                                     |
 | `SliccFollower/Sync/Keepalive.swift`                                                                 | `DataChannelKeepalive` ping/pong actor (used by `AppState`)                                                             |
 | `SliccFollower/Sync/TerminalClient.swift`                                                            | Single-flight `exec.*` client for the leader shell, byte output, cancellation, and disconnect handling                  |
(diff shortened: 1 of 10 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZ6ER6Z55F9963CRQQ8DXCV1)⚡</sup>

👋 I've got this PR.

Picking reviewers is your call — I won't request anyone myself.

PR #1934, opened 2026-08-05
fix(ios-app): move avatar status into header

trieloff:
## Summary

- Moves the iOS agent avatar out of the session pill and removes the fullness ring, lifecycle badge, and connection banner row.
- Shows connection trouble as deterministic TV static in the avatar eyes and keeps the header avatar position stable across label lengths and handedness.
- Mirrors the static-eye connection treatment in the browser web components.

## Why

The avatar was visually trapped inside the session pill, duplicated status with a layout-shifting banner, and could drift as assistant labels changed length. Connection status now stays in the avatar eyes without moving the transcript.

- iOS 26 uses a `.principal` avatar with separate material capsules; the older toolbar-group path remains for iOS 18.
- A fixed identity-label width stabilizes the avatar while the label tail-truncates.
- Connection trouble overrides lifecycle pupils with deterministic 1pt TV static; browser and iOS constants remain in parity.

## Cross-cutting impact

- iOS follower and browser follower/tray surfaces now share the connection-state eye treatment.
- No persisted state, protocol, dependency, or security boundary changes.

## Test plan / Verification status

This PR is intentionally a draft pending on-device visual verification.

### Passed

- `swiftlint lint` and `swift format lint --strict` pass. SwiftLint reports pre-existing warnings only and no errors.
- The browser-side work passed the full repository gates: lint, typecheck, tests, coverage, builds, and bundle budgets.

### Not completed

- `xcodebuild build` was **not completed on the final tree**.
- The iOS UI suites were **not run**, including `ScoopStatusUITests.testAvatarPositionStaysStableAcrossLabelLengthsAndHandedness`.
- No simulator screenshots were captured. Visible `Scoo…` truncation is unconfirmed. XCUITest reports the full accessibility label when text is visually clipped, so this cannot be asserted in tests and needs a visual check.

## Documentation

- [x] Updated `packages/ios-app/CLAUDE.md`.

## Risk & rollback

The main risk is layout divergence between the iOS 18 and iOS 26 toolbar implementations. Revert the four commits in this PR to restore the previous pill, ring/badge, and connection banner behavior.

## Device verifier notes

- iOS 18 installs but exits before drawing.
- iOS 26.5 hangs in a CoreSimulator `locationd` migration.
- A freshly created iOS 26.4 simulator device is the working path.
- iOS 18 and iOS 26 use genuinely different header layouts: the iOS 18 toolbar group versus iOS 26 `.principal` plus a material capsule. Both layouts need checking.

## Checklist

- [x] Commits are focused and package-local where possible.
- [x] No hand-edited files under `dist/`.
- [x] No secrets, credentials, or tokens in the diff.
- [x] Cross-runtime follower paths were updated.
- [ ] UI screenshots or screencast attached — intentionally pending device verification.


trieloff:
Visual validation on the iPhone 17 Pro simulator (scoop-status fixture; `-uiTestConnectionState connected` / `stalled`, reduce-motion for the static shot):

| Connected, right-handed | Leader stalled (TV static) | Connected, left-handed |
| --- | --- | --- |
| connected | static | left-handed |

**What lands well:** the static-eye treatment reads instantly and stays deterministic under reduce-motion; the connection banner row is gone and the composer placeholder carries the state without moving transcript rows; no ring/badge clutter on the avatar.

**Blocking layout regression:** at compact width the session pill overflows the screen edge in **both** handedness modes — "Working Scoop" loses its W on the leading edge right-handed, and the chevron clips on the trailing edge left-handed. The avatar also floats mid-bar rather than sitting immediately beside the pill (the documented position). The stable-position mechanism appears to reserve max-label width and pushes the pill out of bounds instead of truncating it. An earlier state of this branch (before the stabilization commits) did not exhibit this — the pill+avatar cluster fit and sat adjacent.

Suggest: clamp the pill (`lineLimit(1)` + `truncationMode(.tail)` + a maxWidth that respects the bar) and let the avatar anchor the cluster instead of a reserved-width spacer.

trieloff:
Re-validated the in-progress fix (snapshot of uncommitted worktree state — apologies if this is mid-flight):

| Right-handed | Left-handed |
| --- | --- |
| right | left |

**Fixed**: the session pill no longer overflows the screen edge in either handedness mode — the blocking regression from the previous comment is gone.

**Remaining**: the avatar now renders *on top of* the pill's leading text in both modes ("W⬤rking Scoop") — the adjacency became an overlap. Looks like the avatar is positioned with a negative offset/ZStack over the pill rather than as an HStack sibling with spacing.

Happy to re-run this validation (fixture states, both handedness, static eyes) on any push — takes about two minutes.

trieloff:
I prefer the #1940 visual treatment

PR #1944, opened 2026-08-05
ci(ios-app): parallelize simulator tests as a device × iOS matrix

trieloff:
The `ios-app` job ran the iPhone coverage suite and the iPad regular-width regression **sequentially** on one runner, so iOS wall-clock was the sum of both legs. This splits the tests into a new `ios-app-tests` matrix job so the slowest leg, not the sum, bounds wall-clock.

## Matrix

| | iOS 26 (`macos-latest`, Xcode 26.6) | iOS 27 (`xcode-27` beta image) |
|---|---|---|
| **iPhone** | unit suite, coverage-gated (canonical; unchanged floors + `coverage-ios-app` artifact) | unit suite, no coverage gate |
| **iPad** | regular-width browser regression | regular-width browser regression |

- All four cells run concurrently; `fail-fast: false` so one cell's failure doesn't cancel the rest.
- **iOS 27 cells are informational** (`continue-on-error: true`): they ride the beta runner image, and a beta breakage must not hold the merge queue. Flip when 27 goes GA.
- The iPad leg always compiled into its own derived-data path, so splitting it out duplicates no build work vs the sequential version.
- iPhone/27 skips the coverage gate deliberately — line coverage shifts between SDKs and the floors track the stable runtime.
- `ios-app` keeps SwiftLint/swift-format, both builds (simulator + device), and the Periphery scan.
- Per-cell xcresults upload as `test-timings-ios-app-<ios>-<device>`; the aggregate `ci` job now needs `ios-app-tests`.

No runtime downloads: iOS 26 uses the default Xcode on `macos-latest`, iOS 27 uses GitHub's dedicated `xcode-27` beta runner label.

https://claude.ai/code/session_01FXn7dzJEwXceMx3ABAhfPg

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1361,15 +1361,108 @@ jobs:
             -scheme SliccFollower \
             -destination 'generic/platform=iOS' \
             CODE_SIGNING_ALLOWED=NO
+      - name: Install Periphery
+        run: brew install peripheryapp/periphery/periphery
+      - name: Dead code detection (Periphery)
+        run: |
+          # Informational scan; surfaces unused Swift declarations without
+          # blocking PRs while existing findings get cleaned up incrementally.
+          # Unlike the SPM packages, an xcodegen project has no Package.swift for
+          # Periphery to infer from, so the project/scheme/target are explicit.
+          cd packages/ios-app && periphery scan \
+            --project SliccFollower.xcodeproj \
+            --schemes SliccFollower \
+            --targets SliccFollower \
+            --disable-update-check || true
+
+  # Simulator test matrix: device × iOS runtime, all cells concurrent so the
+  # slowest leg (not the sum of legs) bounds iOS wall-clock. The iPhone/26
+  # cell is the canonical coverage-gated run; iPhone/27 repeats the unit suite
+  # against the beta SDK without the coverage gate (line coverage shifts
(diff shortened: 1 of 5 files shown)


trieloff:
Both Major findings addressed in 08e7b52f7:

**1. `xcode-27` runner label** — it is a documented GitHub-hosted label: the [actions/runner-images available-images table](https://github.com/actions/runner-images#available-images) lists `xcode-27` / `xcode-27-xlarge` (arm64) as the beta image carrying Xcode 27, alongside the GA `macos-26` images (whose default Xcode 26.6 serves the iOS 26 cells). This PR touches `packages/ios-app/`, so its own run exercises all four cells — if the 27 cells sit queued rather than starting, that's the empirical tell the label isn't enabled for this org, and the fallback is moving the 27 column to a scheduled nightly.

**2. Copy-paste divergence** — gone. `ios-sim-select.sh` is now the single home of the simctl JSON parse (regex-parameterized; `swift-coverage-check.sh` sources it behind its existing `select_iphone_for_sdk` name), and the new `ios-sim-test.sh` runs the non-coverage slices sharing the `SLICC_IOS_SIM_UDID` override, the pre-boot + `bootstatus` wait, and `run_with_runner_init_retry` with the coverage-gated leg. Both matrix test steps in ci.yml are now one-line script calls.

The `if:`/`continue-on-error` note and the artifact Minor are acknowledged as-is (intended behavior).

PR #1945, opened 2026-08-05
feat(ios-app): settings rework — auto-connect, session probes, honest buttons

trieloff:
Live design-review rework of the Settings sheet, stacked on the Kokoro branch (#1899) because that's where the Speech section lives.

## The four changes

1. **Connect first, Settings only when dead.** Launch tries the Keychain-stored last-good session (`AppState.attemptStoredConnection`, gated on the existing Auto-reconnect toggle). Settings presents only when there's nothing to try, or when retries exhaust into `.gaveUp`.
2. **Session liveness probes.** `SessionReachability` GETs each iCloud session's join URL (4 s timeout; a GET doesn't attach a participant, so the leader never sees the probe; the URL is never logged — verdicts key off the one-way session id). Non-responding sessions sink to the bottom, dimmed, with an `icloud.slash` badge and a "not responding" caption. Still tappable — the probe can be wrong about a flaky network.
3. **Honest Connection block.** The list-item/button hybrid and the premature "Status: Disconnected" row are gone. Connect is a real prominent capsule (spinner while connecting, "Retry" after a failure), and a status line exists only when it carries information: connected, reconnecting, or why the last attempt failed.
4. **One-click voice download.** The button copy carries the 83 MB / Wi-Fi disclosure, so the separate consent step (`awaitingConsent`) is deleted. An in-flight download shows progress, a wall-clock-extrapolated ETA (appears past 5% so early estimates can't swing wildly), and Cancel.

## Proof

Probed sessions (all fixture URLs are dead, so all three sink + dim — newest-first within the cohort), the real Connect button, one-click Speech:

Settings with probed sessions and capsule Connect button

Downloading state — single click from idle, progress + Cancel (ETA appears on real downloads via `downloadStartedAt`; the UI-test fixture pins state without a start time):

One-click download with progress and cancel

## Deferred until the design settles

- `KokoroModelInstallationTests` still reference the deleted consent states; UI tests and the CLAUDE.md consent wording need the same pass.
- `-uiTestKokoroState awaiting` fixture value removed.

https://claude.ai/code/session_01FXn7dzJEwXceMx3ABAhfPg


trieloff:
The connect button should look like the cancel and clear buttons, just blue.

trieloff:
Both notes addressed:

**Plain blue button** (2b89b19a7) — Connect/Retry is now a standard Form action row like Cancel Download and Clear Stored Data. Shown here after a failed attempt (typed URL re-enables it), with the failure note that only exists because there is something to say:

Plain blue Retry row with failure note

**Apple-terse copy** (08a1ae897) — one-sentence footers ("Sessions started with Sliccstart on your other devices appear automatically.", "Connect to a SLICC session with its Join URL."), "83 MB. Downloads over Wi-Fi only.", and the download line now reads "Downloading · 42% · About 40 seconds remaining":

Terse copy across all sections

trieloff:
Review findings addressed in 1c4a747c0:

- **Probe SPA-fallback bug (Critical)** — confirmed and fixed; the probe now appends `?json=true` so it reaches the real tray handler instead of the SPA shell (which I had, embarrassingly, already proven answers 200 for everything while fixing `/privacy` earlier today). Verdicts now reflect tray existence.
- **Launch race** — the stored-connection attempt no longer depends on the Join URL field being empty; leftover typing can't strand a launch with neither a connection nor Settings.
- **Stop Reconnecting** — `.reconnecting` gets its own destructive label; the action stays the full teardown, and the comment above it says exactly what's abandoned.
- **ETA ticker** — only invalidates the Speech section while `.downloading` is on screen.
- **Kokoro tests** — rewritten to the one-click contract (request → `.downloading` immediately, `downloadStartedAt` set/cleared); all 5 pass locally.

Kept as-is, deliberately: the launch auto-connect is gated on the **Auto-reconnect** toggle. That toggle is the user's one switch for "reconnect to things on your own" — honoring it at launch matches its label; decoupling would leave no way to opt out of launch auto-connect.

PR #1956, opened 2026-08-06
fix(ios-app): follow TRAY_SUPERSEDED redirects when attaching to a tray

trieloff:
Fixes the iOS follower's refusal to join a session whose leader has reconnected.

## Summary

Before: tapping an iCloud session in iOS Settings could fail with a raw error card —
`Tray follower attach returned invalid response (409): {"trayId": …, "role": "follower", …}` —
and the session stayed unreachable, even though the row rendered as live.
After: the follower recognises that 409 as the tray hub's supersede redirect and re-attaches
to the replacement tray, landing in the session the row promised.

## Why

That 409 is a valid reply, not a malformed one. When a leader's tab reconnects it mints a
fresh tray and `POST`s `/api/tray/:trayId/supersede` against the old one
(`tray-leader.ts:506`), after which every attach on the old join URL answers
(`session-tray.ts:546`):

```json
409 { "role": "follower", "result": { "action": "fail", "code": "TRAY_SUPERSEDED",
      "error": "This session moved to a new tray after the leader reconnected",
      "joinUrl": "<replacement>" } }
```

The browser follower and the Go CLI both follow that redirect, bounded at five hops. iOS
never listed the code in its `fail` allowlist, so `validateAttachResponse` threw
`invalidAttachResponse` and the raw body reached the UI. The Go package comment said so out
loud: *"the TRAY_SUPERSEDED redirect the TS follower handles and iOS omits."*

| Float | Follows `TRAY_SUPERSEDED` | Where |
| --- | --- | --- |
| Browser follower | yes, 5 hops | `tray-webrtc.ts:487` |
| slicc CLI (Go) | yes, 5 hops | `internal/tray/conn.go:139` |
| iOS liveness probe | yes, 5 hops | `SessionReachability.swift:89` |
| iOS join path | **no — hard error** | `TraySignaling.swift:240` |

**6.29.0 is what made it user-visible.** The new `SessionReachability` walks the replacement
chain and reports `reachable` on the terminal hop, so a superseded session renders live and
sorts to the top of Settings while `sessionRow` still dials the stale `session.joinUrl`. The
probe and the join path disagreed.

**Repro** (what I ran on the simulator):

1. `POST https://www.sliccy.ai/tray` → a throwaway tray A.
2. `POST /api/tray/<A>/supersede` with `Authorization: Bearer <A's controllerToken>` and
   `{"joinUrl": "<a live leader's join URL>"}`.
3. Launch the app with `-joinUrl <A's join URL>`.

Expected: the follower lands in the live leader's session.
Actual (before): the 409 error card above; the session never connects.

- **Layering.** The chase policy is its own type, `SliccTrayKit/Networking/SupersedeRedirect.swift`,
  because iOS has **two** attach loops (`AppState.attachWithRetry` for the app,
  `TrayFollowerConnector.connectOnce` for the File Provider appex) and fixing one would have
  left the other on the old behavior. Bound and delay are pinned to
  `MAX_SUPERSEDE_REDIRECTS` / `SUPERSEDE_REDIRECT_DELAY_MS` and asserted equal in a test.
- **A hop is not just a new URL.** It regenerates the controller id (the old one is a
  participant of the tray being left) and moves the tray the connection *owns* —
  `activeJoinUrl` / `currentJoinUrl` — so reconnect budgets and the Keychain credentials
  written on connect carry the live tray instead of resurrecting the superseded one.
- **Rejected:** treating `URL(string:)` success as sufficient. It accepts
  `join/relative`, which would dial an unusable request; a replacement tray is always
  absolute, so scheme + host are required and a blank/relative one is terminal.
- **`joinUrl` is scoped to its code.** A stray `joinUrl` on `TRAY_EXPIRED` is ignored, so a
  dead tray cannot masquerade as a redirect.
- **Deleted a mirror rather than fixing it.** `Models/TrayTypes.swift` carried a second copy
  of the attach envelopes that nothing decoded — which is exactly why it had silently missed
  this variant. The envelopes now live only beside their decoder in `TraySignaling.swift`,
  with a comment saying why.
- `TraySignalingClient` gains a `Transport` seam (the shape `SessionReachability` already
  uses) so the attach state machine is testable without a live hub.
- `AppState`'s signaling loop moved into a same-file extension to stay under SwiftLint's
  `type_body_length` ceiling. **Pure move, no logic change** beyond the attach loop itself.
- No new dependencies.

## Cross-cutting impact

- **Floats exercised:** iOS follower only. The browser follower, Go CLI and launcher already
  walked the chain and are untouched; the worker contract is unchanged (this PR only teaches
  a client to read a reply the hub has always sent).
- **Both iOS attach paths** are covered (review checklist #8), not just the one that showed
  the bug. The File Provider connector keeps `public let joinUrl` as dialed and tracks the
  live tray in a new private `currentJoinUrl`, so its reconnect loop stops re-dialing a tray
  it already knows is superseded.
- **Other callers of the changed contract:** `attachWithRetry` now returns
  `(plan, client)` because a redirect swaps the client and bootstrap polling has to speak to
  the tray that issued the bootstrap. `runSignalingLoop` is the only caller.
  `FollowerAttachPlan` gained an optional field; `FollowerAttachResult` /
  `FollowerAttachResponse` were removed and had no callers (`rg` clean).
- **Persisted state:** a hop rewrites `activeJoinUrl`, so `persistTrayCredentials` stores the
  replacement in the Keychain on connect and `attemptStoredConnection` resumes the live tray.
  Deliberate — that is the browser's behavior too. iOS still never republishes to iCloud KVS
  (read-only consumer); the Mac launcher advertises its own fresh tray.
- **Security:** the join URL carries the session secret, so no log line or user-facing
  failure message contains it — asserted in `testFailureMessagesNeverLeakTheJoinUrl`. The
  redirect log records only that a redirect happened.
- **Bundle size:** N/A (native).

## Test plan

- [x] `swiftlint lint` — 0 errors (`type_body_length` regression from the added code fixed by
      the extension move, verified against `main`'s clean baseline)
- [x] `swift format lint --strict --parallel --recursive SliccFollower SliccTrayKit` — clean
- [x] **New behavior is covered by unit tests** —
      `SliccFollower/Tests/SliccFollowerTests/TraySupersedeTests.swift`, 10 cases over the
      decoder and the redirect policy; all passing
- [x] iOS coverage gate — `swift-coverage-check.sh --xcodebuild SliccFollower packages/ios-app SliccFollower`:
      70.86% lines / 68.42% functions / 68.88% regions vs floors 66 / 58 / 59.
      `SupersedeRedirect.swift` at 100%
- [x] `xcodebuild build -destination 'generic/platform=iOS Simulator'` — BUILD SUCCEEDED
- [x] `npm run verify` — 7/7 checks
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — 8 changed files, 0 on any debt list
- [x] `cd packages/slicc-cli && make check` — 61.1% vs 58% floor (comment-only change there)
- [x] **Manual smoke (iOS simulator, real hub):** the repro above. The stale tray returned the
      exact 409; the app followed the redirect and landed in the live leader's session
      (streaming transcript, avatar reporting scoop state). Dialing the live join URL
      directly still connects, so the happy path is unchanged.
- [ ] N/A: `npm run typecheck` / `npm run test` / both `build`s / extension smoke — no TS or
      extension code in this diff (the only non-Swift change is a Go package comment)

**No screencast attached** despite the UI-facing fix: a connected follower renders the
leader's private transcript, and the only frame that proves the fix is the connected one.
The `TraySupersedeTests` cases plus the manual repro above stand in.

**Pre-existing failures:** none observed. The `TraySessionDescription.SDPType` Sendable
warning in `TrayTypes.swift` reproduces on `main`.

## Documentation

- [x] Relevant `CLAUDE.md` — `packages/ios-app` (the supersede obligation next to iCloud
      Sessions) and `packages/swift-traysession` (the probe discards the replacement by
      design, so a consumer's join path must walk the chain itself)
- [x] Go package comment in `internal/signaling` no longer claims iOS omits the redirect
- [ ] `README.md` / `docs/` / agent-facing — no user-facing or agent-facing surface changed

`packages/ios-app/CLAUDE.md` sits at the 20k size ceiling, so the docs commit reclaims room
first: the "Simulator QA path" pointer already existed verbatim under Related Guides, and the
Formatting section restated the root npm scripts it tells you to prefer (its raw command also
omitted `SliccTrayKit`). Now 19,981/20,000.

## Risk & rollback

- **Blast radius:** iOS follower only, and only the `fail` branch of attach — a code path that
  previously always errored. `wait` / `signal` are untouched.
- **No flags or env knobs.** Reverting this PR is clean: no persisted-state migration, and the
  Keychain simply holds whichever join URL last connected (already true).
- **Worth verifying by hand before approving** (CI cannot): the redirect needs a real hub, so
  the supersede chase is only covered end-to-end by the manual repro above. CI covers the
  decode + policy, not the live WebRTC hop.

## Checklist

- [x] Commits are focused and package-local where possible (fix / test / docs)
- [x] No hand-edited files under `dist/`. `SliccFollower.xcodeproj/project.pbxproj` is
      regenerated by `xcodegen generate` for the two new files, as in prior commits
- [x] No secrets, credentials, or tokens in the diff — and the new failure messages are
      asserted free of the join URL
- [x] Public API changes are intentional: `FollowerAttachPlan.supersededByJoinUrl` and the
      `TraySignalingClient(joinUrl:transport:)` init are additive; the removed
      `FollowerAttachResult` / `FollowerAttachResponse` had no callers
- [x] Contract used by other floats: checked the leader (`session-tray.ts`), the browser
      follower, the Go CLI, and the launcher. iOS was the only float missing the behavior; no
      peer needed a change


trieloff:
Took both review notes in 924c1845a:

- **Blank `joinUrl` now fails at the wire.** `validateAttachResponse` trims and requires content for `TRAY_SUPERSEDED` rather than leaning on `SupersedeRedirect` to map it to `.terminal` downstream. Covered by `testAttachRejectsSupersededPlanWithABlankJoinUrl` (empty, whitespace, newline).
- **The two bounds no longer share a budget.** A supersede hop used to spend one of the 30 attach attempts; it now spends only the supersede bound, since a redirect is a different tray rather than another try at this one. The chase stays capped by `SupersedeRedirect`, so the loop still cannot spin.

Re-verified: 11 `TraySupersedeTests` passing, coverage 70.87% lines vs 66% floor, swiftlint 0 errors, `swift format lint --strict` clean, `npm run verify` 7/7, touched-exemptions 0.

PR #1966, opened 2026-08-06
fix(webapp): restore one-panel-at-a-time layout and working minimize for sprinkles

froesef:
## Summary
- Sprinkles were landing in an unused `middle` dock-tree zone while tool panels (VFS/terminal/memory/monitor) shared `right`, so a newly opened sprinkle always claimed its own column instead of tabbing alongside them.
- The dock-tree-collapse refactor (`58040db62`) deleted the old show-one "workbench body" without replacing its behavior: every tool panel/sprinkle became a permanently-visible independent leaf, and minimize was left clearing a dock-rail attribute with nothing left to hide.
- Clicking an already-open sprinkle's rail icon fires the dock's `collapse` event, but the handler only handled tool-panel ids, so nothing happened.

This restores, for the classic (`panel-layouts` flag off) dock-tree only:
- Sprinkles default to the same zone as tool panels (`right`), not a separate `middle` zone.
- `placeSurface` detaches every other non-chat surface before placing the new one, and syncs the rail's active indicator — one panel visible at a time, matching the pre-panels model.
- `#minimize` actually parks the sprinkle's leaf (`dockTree.removeSurface`, which hides via `display:none` rather than destroying it) instead of no-op'ing.
- `SprinkleManager.activate()` re-runs `addSprinkle` on an already-open sprinkle's existing container so a minimized/parked one re-places on rail click.
- Attention-pulse and background session-restore adds skip placement entirely (must not steal the one visible slot), matching the pre-refactor `#activate` skip for those modes.
- Clicking an open sprinkle's rail icon now routes to `manager.minimize`, the same path its own in-panel minimize button uses.
- Paid down `sprinkle-manager.ts`'s `noMisusedPromises` debt-list entry (an inline suppression on the one flagged line — a null-check dedupe guard, not a missed `await`) since this PR touches the file.

Panelized shells (`panel-layouts` on, a `hostSprinkleSurface` hook) are untouched — they own their own show/hide model.

## Test plan
- [x] `npm run verify` (lint gate, 7/7 checks)
- [x] `npm run typecheck`
- [x] `npm run test` (13797 passed)
- [x] `npm run test:coverage` (thresholds met)
- [x] `npm run bundle-size` (within budget)
- [x] New/updated unit tests for show-one collapse, minimize parking, panelized-mode no-op, and rail-collapse-to-minimize routing
- [ ] Manual verification in a running dev instance (sprinkle opens alongside VFS in the right zone, minimize hides it, rail click reopens it, one panel at a time)

### DIFF
--- a/packages/webapp/src/ui/sprinkle-manager.ts
+++ b/packages/webapp/src/ui/sprinkle-manager.ts
@@ -985,16 +985,18 @@ export class SprinkleManager implements SprinkleManagerHandle {
    * Route a rail-icon click to the right action based on current
    * state. When the sprinkle was surfaced in attention mode, promote
    * it to user-opened (`markActivated`). When it has a registered
-   * rail icon but no content yet, open it. Already-open user-driven
-   * entries get no extra work — the rail-zone toggle has already
-   * shown the panel.
+   * rail icon but no content yet, open it. An already-open sprinkle
+   * re-runs `addSprinkle` on its existing (already-rendered) container —
+   * idempotent when it's already shown, and the layout's re-place if it
+   * was minimized/parked.
    */
   async activate(name: string, zone?: string): Promise<void> {
     if (this.attentionOnly.has(name)) {
       this.markActivated(name);
       return;
     }
-    if (!this.openSprinkles.has(name)) {
+    const entry = this.openSprinkles.get(name);
+    if (!entry) {
       try {
         await this.open(name, zone);
       } catch (err) {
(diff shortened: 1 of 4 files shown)


trieloff:
Verified and unblocked:

**Bug confirmed on a live leader** (read-only CDP inspection): the dock spec showed two sprinkles as a column split in the unused middle zone (`middle: split[sprinkle:eclipse-viewing, sprinkle:loose-ends]`) with the tool zone (right) separate — each new sprinkle claims its own slice instead of tabbing. The dead minimize and the tool-panel-only collapse gate are both visible in main source.

**Provenance**: commit 58040db62 maps to **PR #1784** (the dock-tree collapse refactor). #1935 only gated the drag handle — not involved.

**Rebased onto current main** (force-pushed): the standalone noMisusedPromises debt commit was dropped entirely — main already paid that debt via the `!== null` form (PR #1995), which also removes the biome.json conflict that was blocking CI (a conflicting PR gets no merge ref and therefore zero check runs). The three functional commits survive unchanged. Local gates: verify 7/7, debt gate, typecheck, full test suite 14,039 passed.

PR #1978, opened 2026-08-07
ci: narrow swift-launcher trigger and cache Swift .build

trieloff:
## Problem

`swift-launcher` (~10.5 min on `macos-latest`, on the `ci` gate's critical path) fired on ~40% of PRs over the last 45 days, and **two thirds of those firings came from `root-config` alone** — 169 of 251. Of 128 distinct such PRs only 3 mentioned node/npm/esbuild; the rest were renovate bumps of `wrangler`, `vite`, `playwright`, `prettier` and friends, none of which reach a Swift compile.

## Solution

- Split a `native-root-config` filter (`patches/**`, `ci.yml`) out of the TS-shaped `root-config` and give `swift-launcher` only that. `patches/**` stays because patch-package rewrites the installed node_modules the bundled overlay build reads; the lockfile does not, and a lockfile break surfaces on the fast ubuntu jobs first. Other `root-config` consumers are untouched.
- Cache `packages/swift-launcher/.build` and `packages/swift-server/.build`. The `Build swift-server (dependency)` step is ~6 of the job's ~10.5 min and runs from scratch even when `swift-server` itself is unchanged and its own job is skipped. The key includes a `swift --version` fingerprint so an `macos-latest` Xcode roll cannot serve mismatched artifacts, and `coverage/` is excluded so the upload steps can never see a stale profile.

## Misc

This narrows *when* the job runs, not *what* it verifies — a genuinely native-affecting PR still triggers it via the existing `swift-*`, `spoon`, `assets`, and `swift-config` filters.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01KZEBRN27K8DBAWWBSBN0Y3RV&panel=chat)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       slicc-cli: ${{ steps.filter.outputs.slicc-cli }}
       swift-config: ${{ steps.filter.outputs.swift-config }}
       root-config: ${{ steps.filter.outputs.root-config }}
+      native-root-config: ${{ steps.filter.outputs.native-root-config }}
       speech: ${{ steps.filter.outputs.speech }}
       is-queue-leader: ${{ steps.queue-position.outputs.leader }}
     steps:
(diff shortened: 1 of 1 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01KZEBRN27K8DBAWWBSBN0Y3RV)⚡</sup>

👋 I've got this PR.

PR #2029, opened 2026-08-09
feat(shell): add vpod command — WASM Linux sandbox via @capsule-run/vpod

trieloff:
Adds a `vpod` supplemental shell command that runs Linux commands in a real 64-bit guest compiled to WebAssembly ([capsule-run/vpod](https://vpod.sh) — a genuine kernel + userland, not a shell reimplementation). First-cut integration to evaluate the engine alongside `v86`.

## Pattern

Follows the v86 install-gated trio (`vpod-command.ts` / `vpod-pods.ts` / `vpod-loader.ts`): the whole SDK comes from an ipk-installed `@capsule-run/vpod@0.6.0` in the VFS `node_modules` — nothing bundled, no CDN fallback, version pinned as a source literal, `ipx`/`npx` redirected via the builtin shadow map.

## Load mechanism (the interesting part)

vpod cannot use the blob-URL glue pattern (pyodide/ffmpeg/v86): it is a deeply URL-relative multi-file ESM distribution — chunked entry, `new Worker(url, {type:'module'})` derived from `import.meta.url`, and a jco-transpiled component that fetches its 27 MB core wasm relative to itself. Blob URLs are not hierarchical, so instead the loader imports `dist/index.js` **through the preview service worker** (`toPreviewUrl`), which serves VFS bytes at real same-origin hierarchical URLs with correct MIME types. Relative chunk imports, worker spawning, and the wasm fetch then resolve natively. Precedent: pyodide's wheel loader already fetches through this exact surface from inside a realm worker. The browser graph has no bare specifiers (verified against the tarball), so no import maps are needed.

## Surface

- `vpod start [--name] [--snapshot] [--net]` — named background pod, ProcessManager-tracked (`ps` sees it, `kill` closes it)
- `vpod run [--name] [--timeout <secs>] <cmd...>` — auto-boots the default pod; flags are only recognized before the first command word so guest args pass through intact; guest stdout/stderr/exit code map 1:1 onto the shell result
- `vpod ls` / `vpod net [--port]` / `vpod stop` / `vpod --version`
- One command at a time per pod (busy guard); no TTY, no streaming — matches the SDK contract

## Networking honesty

Guest networking rides on SharedArrayBuffer → needs COOP/COEP, which SLICC's origins deliberately do not serve (the sync-FS bridge was built without cross-origin isolation). Pods are therefore **offline** in this first cut; `vpod net` reports capabilities and per-port reachability with the upstream explanation. Adding COOP/COEP would be a platform-wide decision (node-server, CF worker, extension hosted-tab model) and is explicitly out of scope here.

## Tests & docs

- `vpod-command.test.ts`: 26 tests — arg parsing, lifecycle with injected fake SDK, auto-boot semantics, busy guard, PM registration/abort teardown, net reporting; shadow-map bootstrap assertion added to `builtin-shadow-map.test.ts`
- Agent skill `packages/vfs-root/workspace/skills/vpod/SKILL.md` (auto-bundled), `docs/shell-reference.md` rows (WASM list, command table, shadow-map table), README capability bullet

## Verification

`npm run verify` ✓ · debt gate ✓ · `typecheck` ✓ · `test` (14129 ✓) · `test:coverage` ✓ (one unrelated known-flaky `playwright-command.test.ts > iframe integration` Chrome-launch failure; passes in isolation 282/282) · both builds ✓ · `bundle-size` ✓ (runtime install → zero bundle impact)

Not yet done (follow-ups): live smoke test against the hosted origin's preview SW (`ipk add` + boot in a real leader tab), SW-side caching for `node_modules` reads (the preview channel does SW → BroadcastChannel → page → kernel RPC → OPFS per file), and a decision on guest networking.

https://claude.ai/code/session_01G8Pt2nX6koz4K868SWBMN9

---

## Live smoke test (Tier-1 harness, wrangler :8787 + bridge :5715)

Full cold path verified on a fresh ephemeral profile:

| Step | Result |
|---|---|
| `ipk add @capsule-run/vpod@0.6.0` | 7.5s (29.5 MB unpacked into VFS `node_modules`) |
| First `vpod run uname -a` | **`Linux vpod 6.18.0-3-lts … riscv64`** — 20.5s cold (SDK import through preview SW + 27 MB component + wasm compile + registry snapshot pull + boot); 17.2s on a second fresh profile |
| Warm `vpod run` | **16–17 ms** |
| `vpod run python3 -c "print(6*7)"` | `42` |
| `vpod ls` / `ps` integration | pod listed with ProcessManager pid, snapshot `vsnap-base-256mb` |
| `vpod net` / `vpod net --port 443` | backend `none` + COOP/COEP note + upstream `explainUnreachable` text |
| `vpod stop` | closes pod, unregisters, `pm.exit` |
| Not-installed gate | canonical `ipk add` guidance verbatim |

Two bugs found live and fixed in follow-up commits:

1. **`vpod run` argv quoting** — the host shell eats the user's quotes, so `python3 -c "print(6*7)"` reached the guest as bare `print(6*7)` → guest sh syntax error. `run` now re-quotes each token (`sh -c "..."` payloads survive as one token).
2. **Preview-SW concurrent-read collapse (platform bug, #2034)** — parallel module-graph fetches through `/preview/*` returned the *same file's bytes for every chunk URL* (hash-verified), so the vpod import linked against the wrong module and died with a missing-export error, and the worker's module map cached the poisoned chunk until reload. Root cause is below the RPC correlation layers (ZenFS WebAccess-on-OPFS same-context concurrent reads); this PR ships the responder-level mitigation — the preview-vfs responder now serializes its read pipeline (acks stay synchronous) with a regression test. FS-layer fix tracked in #2034.


trieloff:
Restructured per review direction: the preview-vfs responder serialization fix moved to its own PR — **#2037** — and this branch was rebased onto current main without it. **This PR now depends on #2037 landing first**: without it, the cold parallel module-graph import that boots the vpod SDK hits the #2034 read-collapse. Also updated the guest-networking wording to feature-detect `crossOriginIsolated` (naming Document-Isolation-Policy, #2036) instead of asserting isolation is never available.

trieloff:
Rebased onto main @ 6.46.0 — both dependencies have now landed (#2037 responder serialization, #2039 Document-Isolation-Policy), plus the DIP unlocks (#2044 ffmpeg-mt, #2046 meminfo). The branch is three clean commits (command, quoting fix, isolation-aware wording); full local gate + tests green. On this main, `ipk add @capsule-run/vpod@0.6.0` → `vpod run` cold-boots correctly (serialized preview reads) and `vpod start --net` gets a live fetch backend (DIP). Staying in draft for the visual/UX pass.

trieloff:
**Custom image story** (the natural follow-up once this merges): upstream PR [capsulerun/vpod#79](https://github.com/capsulerun/vpod/pull/79) adds Dockerfile-built snapshots to vpod itself — the Dockerfile is built for `linux/riscv64` by Apple's `container` CLI (BuildKit executes `RUN` under riscv64 emulation), the flattened rootfs replaces the Alpine minirootfs in Mohamed's existing pipeline, and `vpod-native` captures the machine state as a normal `.snap`.

Verified end to end: `FROM alpine:3.23` + `apk add python3 nodejs git` + `COPY` → 84 MB snapshot that resumes and answers all probes (python through vpod's warm shim, node v24.18.1, git 2.52.0, riscv64). ~3.5 min per build with a warm image cache. Notable gotcha found on the way: BuildKit's tar exporter rewrites absolute symlink targets with a platform prefix (`/bin/sh -> /linux_riscv64/bin/busybox`) — unfixed, the guest panics on init.

AOT support for custom images is in flight on the same branch: `aot-snapshot.sh --workload custom --trace-cmd '<hot command>'` so a node/git image traces its *own* hot code instead of the stock python/apk workload (which such images may not even contain).

Implications for this PR's surface, all deferred until upstream lands:
- `vpod start --snapshot <name>` already accepts registry names — custom snapshots become reachable the moment they're catalogued (self-hosted registry is just static JSON + blob; our CF worker + R2 could serve one org-wide).
- A `vpod build -f Dockerfile` shell command handing a VFS Dockerfile to this pipeline is the follow-up feature, host-side (needs Docker-ish tooling, not browser-runnable).
- Custom-AOT wasm components are a drop-in file replace in the ipk-installed package (we load everything from VFS), pending Mohamed's take on runtime-loadable AOT modules.

No changes to this PR — it stays draft for the UX pass.

PR #1938, opened 2026-08-05
Constrain iPad chat transcript width

trieloff:
## Solution
Caps the centered transcript at 680pt so iPad conversations remain readable without changing compact layouts.

## Testing
Adds iPad conversation-fixture capture to the PR screenshot check; the resulting proof is posted inline on this PR.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01KZ8YGAYDGQSQZ32KKDBSYZCT&panel=chat)

PR #2048, opened 2026-08-09
fix(ios): center the iPad transcript column

trieloff:
## The bug

#1938 capped the chat transcript at `MessageListLayout.maximumReadableWidth` (680pt) on regular-width iPads. The column got narrower but never moved: it rendered flush against the **leading** edge, stranding the trailing third of the window.

## Root cause

The classic double-frame idiom was already there and, read on its own, it is correct:

```swift
.scrollTargetLayout()
.frame(maxWidth: MessageListLayout.maximumReadableWidth)  // 680
.frame(maxWidth: .infinity)                               // fill + center
```

The problem is *where* it sat. Those frames wrapped the `LazyVStack` that carries **`scrollTargetLayout()`**, and a scroll view anchors its content on the target layout. So the centering happened and was then undone: the outer frame did resolve to the full viewport and did place the 680pt child at the center, and the scroll view then applied an equal and opposite horizontal content offset to bring the target layout's leading edge back to the viewport's leading edge. Net movement: zero.

I tinted both frames and measured the pixels on an iPad Pro rather than guessing. On a 984pt conversation column:

| band | measured extent |
| --- | --- |
| inner 680pt cap | `0 → 680pt` |
| outer `.infinity` frame | `0 → 832pt` |
| unclaimed | `832 → 984pt` |

Everything is shifted left by exactly `(984 − 680) / 2 = 152pt` — precisely the centering offset the frame had applied. Temporarily deleting `scrollTargetLayout()` (control experiment) made the same code center perfectly: outer frame `0 → 984`, column `152 → 831`. That pins the scroll-target anchor as the cause, not the frames.

## The fix

`scrollTargetLayout()` has to stay directly on the `LazyVStack` — `scrollTo(id:)` and `viewID` restoration (`ChatPresentationState`) need every message to be its own scroll target, so it cannot simply be hoisted outside the frames.

Instead the cap moves off the stack and onto the **rows** (`readableTranscriptColumn()`). The target layout then keeps the full viewport width, so there is no offset left to scroll away, and the stack's own `.center` alignment does the centering. The transcript's invisible bottom anchor is what holds the stack at full width, so it is now explicitly greedy and commented as load-bearing.

This also keeps the whole viewport draggable — capping the `ScrollView` itself would have worked too, but would have made the outer gutters dead to scrolling.

## Proof

Local iPad Pro capture of the `conversation-fixture` screen, gutters measured from the PNG:

| | left gutter | right gutter |
| --- | --- | --- |
| before | 16pt | 320pt |
| after | **168pt** | **168pt** |

The `ios-screenshots` workflow attaches its own iPad `conversation-fixture` capture to this PR, so the centering is visible directly in the sticky screenshot comment.

## Tests

New `SliccFollowerUITests/TranscriptColumnUITests`, wired into the existing iPad CI leg. It compares the center of a transcript timestamp header against the center of the navigation title — both are centered in the same container, so they coincide exactly when the column is centered and diverge by half the leftover width when it is not, which keeps the assertion independent of the dock rail's width. It also asserts the column still keeps a real gutter, so a silently-uncapped column fails too.

Verified the test actually catches the regression by running it against the unfixed layout: it fails with column center `340.0` vs container center `392.75` — the exact flush-left signature (`340 = 680 / 2`, i.e. the column pinned at `x = 0`).

Verification: 516 iOS unit tests pass with coverage above every floor (lines 72.75/72, functions 71.24/70, regions 68.93/68); `npm run verify` 7/7; `npm run typecheck` clean; swiftlint 0 serious; `swift format lint --strict` clean; doc-size and touched-file debt gates clean.


trieloff:
## Visual proof — iPad `conversation-fixture`

Both captured locally on the same iPad Pro simulator (984pt conversation column beside the 48pt dock rail), via `ios-screenshots.mjs --device=ipad --screen=conversation-fixture`. The "before" is a build of the merge-base `MessageListView.swift`; the "after" is this branch.

| Before — `main` | After — this PR |
| :---: | :---: |
|  |  |
| capped but flush left | capped **and centered** |
| left `16pt` · right `320pt` | left `168pt` · right `168pt` |

The column width is identical in both — 680pt, exactly as #1938 intended. The only thing that changes is where it sits: the leftover `304pt` stops piling up on the trailing side and splits evenly into `152pt` gutters (the extra `16pt` in the measurements above is the rows' own horizontal padding).

Gutters were measured off the PNGs by scanning for the first and last non-canvas pixel on several transcript rows, not estimated by eye.

Issue #2155, opened 2026-08-18
Discovery licks: dismiss gesture backed by /etc/llmstxtignore


trieloff:
## Problem

Discovery licks of kind `llms-txt` are informational only: the card says «there is no card action», so the same hosts keep pinging forever and every ping costs context in the receiving agent.

I audited every `llms.txt` mention across 67 archived sessions (about 126 MB). After removing help-text boilerplate and the raw-JSON copy each session export embeds in its own header, there were **55 discovery events**:

| Count | Host |
| --- | --- |
| 18 | `github.com` |
| 9 | `app.slack.com` (plus 2 `adobe.enterprise.slack.com`, 1 `adobe-dx-support.slack.com`) |
| 8 | `www.sliccy.ai` / `www.sliccy.com` |
| 4 | `www.paypal.com` |
| 3 | `ampcode.com` |
| 2 | `time.com` |
| 5 | one each: `etailment.de`, `xing.com`, `replicate.com`, `exp-tech.de`, `api-dashboard.search.brave.com` |
| 1 each | `valkanik.com`, `absoluteeclipse.eu`, `metaculus.com` |

Seven were fetched, three produced value (a Shopify store publishing UCP endpoints and product JSON, a shop catalogue with SKUs, and a developer-docs index). That is a %5 payoff on pings. The three highest-volume hosts (`github.com`, `app.slack.com`, `www.sliccy.ai`, 35 of 55 events) produced value zero times: what they publish is marketing copy, in Slack's case an ASCII banner.

Two secondary findings from the same audit:

- Six events were delivered into `agent-memory-curator` sub-agent sessions, which have no browser and cannot act on a URL at all.
- Three of those six fired because the curator was *reading an archived transcript that itself contained old discovery events*, so historical text re-triggered live licks.

## Ask

Give the discovery card the same binary gesture the **upgrade lick** and the **sudo lick** already have, where dismiss means «never ask me about this host again», and back it with a file the discovery engine consults before emitting.

## Proposed mechanism

1. **`/etc/llmstxtignore`**, following the existing `/etc/sudoers` precedent:
   - one entry per line, hostname or glob, for example `github.com`, `*.slack.com`
   - `#` starts a comment, blank lines ignored
   - matching is on the advertising host, case-insensitive
2. **Discovery engine reads it before emitting the lick.** A match means the event is skipped silently: no card, no lick, no context cost. This must happen upstream of lick creation, not in the renderer.
[...]

Issue #2062, opened 2026-08-10
Sudo over tray


trieloff:
When driving from the iOS app, there is no way to approve sudo requests. IOS is actually a great surface for this with push notifications (for turn complete), priority notifications (for sudo) and Face ID (for extra security)

trieloff:
## Plan: sudo over tray, Face ID approval, turn-end + priority push (one PR)

### Where we are

Two sudo paths exist and only one is broken for remote driving:

1. **Scoop → cone** (`webapp/src/scoops/scoop-approval-router.ts`) — the cone *agent* decides; the `sudo-request` lick card is already rendered on iOS, informational only. Unchanged by this work.
2. **Cone → human** (`webapp/src/sudo/index.ts` → `createSudoBroker()`) — a native gesture: node-server OS dialog (`http-broker.ts`), extension-panel `confirm` (`panel-responder.ts`), or panel-RPC to the page realm. **None of these can reach a tray follower.** Mac leader: the dialog pops on the Mac while you hold the phone. Hosted leader (headless Chromium): nobody to ask, fails closed.

Transcript export already solved this once, bespoke: `transcript.export.approve.request/response` (v4) in `shared-ts/src/tray-sync-protocol.ts`, decided by `headlessLeader` + `RequesterTracker` ("last user message came from a follower") in `tray-leader-sync.ts`. That gate is a sudo action wearing a costume; it gets folded in.

### 1. Wire protocol (`shared-ts`, additive, v5)

```ts
// leader → follower
| { type: 'sudo.approve.request'; requestId; kind: SudoKind; detail: string;
    suggestedPattern?: string; scoopName?: string; expiresAt: number }
| { type: 'sudo.approve.cancel'; requestId }          // answered elsewhere / timed out
// follower → leader
| { type: 'sudo.approve.response'; requestId; decision: 'allow'|'deny'|'always';
    pattern?: string; attestation?: 'biometric'|'passcode'|'none' }
| { type: 'push.register'; platform: 'ios'; token: string; environment: 'sandbox'|'production' }
```

* `SudoKind` gains `'export'` — `detail` is the export selector rendered as text (`"transcript export: <selector>"`), `suggestedPattern` is the selector glob so `always` can persist a `NOPASSWD Export …` sudoers rule like any other kind.
[...]

Issue #2043, opened 2026-08-09
DIP unlock: Atomics/SAB fast path for the sync-FS and sync-exec bridges (lift snapshot caps on isolated leaders)


trieloff:
The sync-FS design (PR #1519, commits 079944d08 → b69546f75) explicitly demoted the Atomics/SharedArrayBuffer bridge to "a documented future perf lever" because isolation via COOP/COEP was untenable. With Document-Isolation-Policy on the leader (#2039), the lever is now cheap to pull — per-document, SW control intact, embedded floats unaffected.

Scope (the large item in the DIP unlock series):

- Realm `readFileSync`/`writeFileSync`/`execSync`/`spawnSync` block on `Atomics.wait` over a SAB ring filled by the kernel worker (realm workers share the document's agent cluster — inheritance proven live by vpod's nested workers under #2039).
- Retires **on isolated leaders**: the 500-file / 1 MB / 10 MB snapshot bounds (`ENOSYNC` ceiling), per-call sync-XHR + SW round-trip latency, the `syncFsBridgeEnabled` SW-confirmation boot gating, and the standing exposure to synchronous-XHR deprecation.
- The SW sync-XHR path **stays as the universal baseline** — Cherry/spoon/Electron embedded leaders can never isolate (`docs/kernel/process-model.md` already reframed by #2039).
- Follow the #1519 spec history for the original Approach-A design notes (`docs/superpowers/specs/2026-07-15-sync-fs-browser-realm-design.md` at 079944d08, deleted from main, lives in git history).

Part of the DIP unlock series (#2036, PR #2039).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01G8Pt2nX6koz4K868SWBMN9

Issue #2034, opened 2026-08-09
Concurrent VFS reads through preview-vfs responder return wrong file contents (ZenFS WebAccess-on-OPFS)


trieloff:
Found while live smoke-testing the vpod loader (PR #2029).

**Symptom:** when several `/preview/*` URLs are fetched concurrently (e.g. the browser fetching a module graph's chunks in parallel), every response carries the same file's bytes. Hash-verified live: 8 parallel fetches of distinct `dist/*.js` files of an ipk-installed package all returned the entry file's content (identical SHA-256, equal to `index.js`). Sequential fetches of the same URLs are always byte-correct against the npm tarball. Module-graph consequence: the graph links but fails with `does not provide an export named …`, and the worker's module map then caches the poisoned chunk for the life of the realm.

**Repro (leader page console):**
```js
const ch = new BroadcastChannel('preview-vfs');
const read = (p, i) => new Promise(res => { const id = 'probe-' + i; ch.addEventListener('message', function h(ev){ const d = ev.data; if (d?.id !== id || d.type !== 'preview-vfs-response') return; ch.removeEventListener('message', h); res(p + ' => ' + String(d.content).slice(0, 30)); }); ch.postMessage({ type: 'preview-vfs-read', id, path: p, asText: true }); });
await Promise.all(['/dist/index.js', '/dist/chunk-a.js'].map((p, i) => read('/workspace/node_modules/@capsule-run/vpod' + p, i)));
```

**Localization:** correlation ids are matched correctly at every layer by inspection (`preview-sw-handler.readViaMainPage`, `preview-vfs-responder`, `RemoteVfsClient.pending` map, `VfsRpcHost` per-request closures) — the collapse is below the RPC, in the kernel's ZenFS `WebAccess`-on-OPFS read path under same-context concurrent `readFile` calls (shared sync-access-handle state is the suspect).

**Mitigation shipped in PR #2029:** `preview-vfs-responder` now serializes its stat+read pipeline (receipt acks stay synchronous), with a regression test. That shields the preview surface only — any other caller issuing concurrent reads against the kernel VirtualFS may still hit the underlying bug; a FS-layer fix (or per-file read locking in the VirtualFS adapter) is the real close-out.

[...]

PR #2163, opened 2026-08-18
feat(ci): migrate the five Cosmos scheduled agents to GitHub Actions + Claude Code

trieloff:
## Summary

Migrates the five scheduled agents we ran on Augment Code's Cosmos onto this repo's existing `claude-code-action` pattern, so they live next to `rum-error-triage.yml` / `agentic-debt-triage.yml` instead of on a retired external platform. Each becomes a deterministic Node selector step plus one or two Claude steps.

| Workflow | Cadence | Selector | Claude step |
| --- | --- | --- | --- |
| `boy-scout-debt-dispatcher.yml` | daily 02:17 UTC | `boy-scout-debt/select-debt-file.mjs` | cleans one debt-listed file, opens the PR |
| `pr-fix-dispatcher.yml` | every 2h | `pr-fix-dispatcher/scan-failing-prs.mjs` | matrix job, one fixer per queued PR |
| `backlog-dispatcher.yml` | daily 09:35 UTC | `backlog-dispatcher/select-backlog-issues.mjs` | triage phase labels ready issues, author phase opens a PR each |
| `claude-md-compactor.yml` | Sat 22:40 UTC | `claude-md-compactor/measure-claude-guides.mjs` | compacts oversized guides, one PR |
| `flaky-ci-hunter.yml` | Mon 06:50 UTC | `flaky-ci-hunter/scan-flakes.mjs` | one determinism fix PR |

## Why

Cosmos is being retired, and these five agents are the automation that pays down debt, unsticks failing bot PRs, turns ready issues into pull requests, keeps the instruction files inside budget, and chases nondeterministic CI. Rebuilding them here means one platform, one auth path, secrets we already have, and selector logic that is unit-tested in the repo instead of living in a prompt on someone else's service.

Three Cosmos concepts have no GitHub Actions equivalent, so each needed a deliberate substitution:

- **Dispatcher + `PR Author` worker → selector step + Claude step.** Cosmos dispatchers launched a separate worker agent. Here every rule-expressible decision lives in a pure, unit-tested `lib.mjs`, and only genuine judgement is left to the model. The PR fix dispatcher keeps its 3-per-run cap through a `max-parallel: 1` matrix over the scanner's queue, so each PR gets its own checkout and its own turn budget.
- **VFS YAML registries → GitHub-native state.** No state file, state branch, or Actions cache anywhere in the five. `rerun_shas` is derived from `run_attempt > 1` (re-running is what bumps it, so the runs endpoint *is* the ledger); skips and dispatches are `<!-- pr-fix-skip:<sha> -->` / `<!-- pr-fix-dispatch:<sha> -->` marker comments, following the `<!-- rum-fp:… -->` precedent in `rum-error-triage.yml`; flake cooldown reads the `automation/flaky-fix/<slug>` PR dates; the compactor dedups on an already-open PR. The `ci-fix-*` labels are human-visible markers and deliberately **not** the dedup key, so relabelling a PR by hand cannot cause a double dispatch or a double re-run.
- **VFS digest → step summary + `flaky-ci-digest` artifact**, uploaded `if: always()` because the quiet weeks are exactly when the digest matters.

The flake hunter's Cosmos "source 2" (reading the PR fix dispatcher's registry) collapses into source 1: a dispatcher re-run is what creates attempt 2, so the same evidence is already there.

Two details that are easy to get wrong, and are therefore pinned by tests:

- **The compactor's policy is stricter and wider than the repo gate.** `PACKAGE_CLAUDE_MAX_CHARS` is 20,000 for `packages/*/CLAUDE.md`; the compactor's policy is 10,000 → 9,500 across *every* tracked `CLAUDE.md`. The committed gate is untouched, the relationship is spelled out in three places so the two cannot be confused, and the 3,000-byte agent-facing runtime guide is excluded by construction. A deterministic `--check` step re-measures after Claude runs and fails the job rather than trusting the model to have finished the work.
- **The flake hunter lists runs one day at a time.** `GET /actions/runs` returns at most 1000 items however you page it, and this repo produces ~2,400 runs a week, so a window-wide query would silently truncate to ~2.5 days and manufacture a quiet week. Retrieved-vs-`total_count` is reconciled per day, and truncation is reported in the digest rather than as a clean scan.

Selector logic reuses what already exists: the boy-scout selector imports `size-exemption-lib.mjs`, so the multi-rule test-file override stays excluded by construction and the debt-list parser is not cloned (the duplication gate would catch it).

No new dependencies. Node builtins, global `fetch`, `gh`, and `git` only.

### The backlog dispatcher needed its rubric reconstructed

Its prompt body was a `kb://expert-templates/prompts/backlog-dispatcher.md` include that did not survive the platform's retirement, so the readiness rubric had to come from somewhere. It came from the agent's own history: **~50 of its decisions are still recorded on this repository** as `cosmos-dispatched` / `cosmos-skipped` labels plus its verbatim skip comments. The hard-override catalog is its own words:

| Issue | Cosmos' skip reason |
| --- | --- |
| #2062 Sudo over tray | "touches the sudo/approval authorization surface, and the ask spans tray and host surfaces rather than one localised change" |
| #2072 iOS transcript jump | "upstream iOS 26 SwiftUI bug (FB20979569) needs on-device experimentation and a design call, not a localised fix" |
| #2043 Atomics/SAB fast path | "large cross-cutting redesign … architectural scope, not a small contained PR" |
| #2034 concurrent VFS reads | "root cause is unconfirmed … the real fix is an FS-layer locking design — needs a human to pick the approach" |
| #2155 | "A PR already exists for this ticket (#2156), so no additional dispatch is needed." |

The classes it *did* dispatch: an agentic-debt item naming both the sin and its file, a small localised bug with a concrete symptom, a missing shell command with a clear spec, one named flaky test, doc drift naming the stale file, a narrow test addition. **Reviewers: this catalog in `backlog-dispatcher/lib.mjs` is the single highest-value thing to review here**, since it is reconstructed rather than handed over.

That history is also load-bearing: the legacy `cosmos-*` labels are treated as **equivalent** to the new `backlog-*` ones when screening, because otherwise the first run would re-propose everything Cosmos already refused, starting with sudo-over-tray.

Structurally it copies `rum-error-triage.yml` rather than inventing a shape — a triage phase that labels `backlog-ready`, then an author phase that implements and opens the PR and swaps the label. Splitting them means an issue is never labelled ready by a run that then exhausts its turns before a PR exists.

**Three deliberate departures from the original**, each flagged before implementation:

- **Scope is this repository only.** The original scanned 21 `ai-ecoverse` repos. `BOT_PAT` is a fine-grained PAT scoped to slicc, and no workflow here touches another repo's API, so a fleet version would mean putting write access to 20 other repos into this repo's secret store where every workflow could reach it. Nothing hardcodes the repo name (it comes from `GITHUB_REPOSITORY`), so promoting this to `ai-ecoverse/.github` later is a packaging change — adding `workflow_call` and a repo input — not a rewrite.
- **The stale sweep never closes a pull request.** The original closed dispatcher PRs after 7 idle days, discarding work whose only sin was waiting for review. This one labels `backlog-stale` and comments once.
- **Skip decisions are made once.** The original re-posted its skip comment on every tick (#2072 has two, five days apart). The label is now the dedup key, and removing it is the documented way to re-queue an issue.

## Cross-cutting impact

- **Floats exercised**: none. This is CI-only tooling — no runtime, UI, bundle, or float code is touched, so there is nothing to check across CLI / extension / Electron / Sliccstart / worker.
- **Persisted state side-effects**: none in the app. The five deliberately persist nothing; all cross-run memory is read back out of GitHub (see above).
- **Repo-level writes these workflows can perform once merged**: new branches (`automation/boy-scout/*`, `automation/backlog/issue-*`, `automation/weekend-claude-compaction-*`, `automation/flaky-fix/*`), PRs, PR and issue comments, the `ci-fix-*` / `backlog-*` / `boy-scout-debt` / `flaky-fix` labels, pushes to the head branch of an automation PR, and `rerun-failed-jobs`. Nothing merges, closes, or pushes to `main`; the flake scanner has no write verb at all; the backlog dispatcher never closes or assigns an issue, and never edits an issue title or body.
- **Interaction between two of them**: backlog PRs use an `automation/backlog/` branch prefix, which `isAutomationPr()` matches, so the PR fix dispatcher will pick up CI failures on them. That coupling is deliberate and pinned by a test importing both libs, so renaming the prefix fails a test instead of silently severing it.
- **Security**: `BOT_PAT` is required rather than optional — GitHub's anti-recursion guard suppresses workflow runs for `GITHUB_TOKEN`-authored pushes, so CI would never run on the PRs these open (same constraint already documented in `coverage-ratchet.yml` and `renovate-patch-reconcile.yml`). Job permissions are least-privilege per job; the PR fix dispatcher's `scan` job holds `contents: read` and cannot push. Values interpolated into a prompt from PR data (`title`, failing-check summaries) are flattened to one line with `${{` neutralised so an Actions expression cannot be re-expanded. The `pr_number` dispatch input waives only the two time-based waits and cannot be used to aim a fixer at a human's PR.
- **Bundle size**: unaffected.

## Review rounds

Eight findings from Codex and the house reviewer, all real, all fixed. Recorded here because two of them were failure modes no test would have caught locally.

**Round 1** (`406b1cd9e`): `gh label create` needs `issues: write`, not `pull-requests: write`, on two workflows — every non-dry run would have died at the bootstrap step. And `rerun-failed-jobs` returns **201 with an empty body**, so `res.json()` threw *after* the re-run had already fired; the REST layer is now split into `request` / `gh` / `ghWrite`, where writes never parse a body. That refactor also caught a latent bug the review had not flagged: the old success check counted a tolerated 403/409 as a successful re-run.

**Round 2** (`3dd3b3bb4`):

- **P1, and I got this wrong the first time.** I dismissed the same finding on round 1 with "boy-scout is unaffected because it uses `BOT_PAT`", having assumed that PAT was broadly scoped. Its own header says contents + pull-requests only. So the nightly fixer would have selected a file, completed the refactor, run the whole verification pass, and *then* died at `gh label create` — the most expensive possible place to fail. The label is now bootstrapped by a workflow step under `GITHUB_TOKEN`, and a test asserts the brief contains no `gh label create` invocation.
- **Forked automation PRs were queued for a fixer.** The fix job checks out the bare `head.ref` in *this* repository, so a fork PR either fails on a missing branch or — if a branch of the same name exists here — edits and pushes the wrong one, with the label and SHA marker already written so nothing retries. Now refused at the screen, deleted forks included.
- **The compactor verified the wrong threshold.** `--check` only received `MAX_CHARS`, so a guide rewritten to just under the oversized band passed even though the brief promised the target, and would be re-selected the following week having achieved nothing. The measure step now carries its worklist through.
- **The flake window was short by ~17 hours.** Seven UTC calendar dates at the Monday 06:50 cron covered ~6d 7h, hiding the previous Monday morning — enough to drop a job under the two-commit threshold and make a real flake look quiet.
- **Merge-queue runs were invisible to source 3.** The CLI required `head_branch === 'main'`, but `merge_group` runs carry `gh-readonly-queue/main/…`, and `ci.yml` fires on `merge_group` here. The useful part was *why*: the CLI filter was stricter than the pure logic beneath it, which already listed `merge_group` in `MAIN_EVENTS`. Fixing only the CLI would have left the same bug one layer down, so both layers now share one helper.

## Test plan

- [x] `npm run verify` — **7/7 checks pass** (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod)
- [x] `npm run lint:docs` + `check-doc-refs` + `node packages/dev-tools/tools/check-touched-exemptions.mjs` — clean
- [x] **New behavior is covered by unit tests**: 273 new cases across the five `lib.test.mjs` suites; the whole `dev-tools` vitest project is **944 passing / 41 files** (`node node_modules/vitest/vitest.mjs run --project dev-tools`)
- [x] Offline selector runs against this repo: the boy-scout selector finds 236 tractable candidates and picks `packages/cloudflare-worker/src/cloud/error-envelope.ts`; the compactor finds exactly one oversized guide (`packages/swift-server/CLAUDE.md`, 10,012 chars); the flake scanner's day-by-day pagination and flip detection are driven end to end over a fixture; the PR fix scanner is driven end to end with a stubbed `fetch` (dispatch / rerun / skip all reached, zero writes under `DRY_RUN`)
- [x] The backlog selector was driven offline against a fixture: an issue carrying the **legacy** `cosmos-skipped` label was suppressed (the re-propose regression), a fresh issue was held back by the 1h settling age, an issue with `Closes #N` in an open PR was dropped as in-flight, and the `debt:paranoia` issue was selected — plus one case per remaining reason code and a stale-sweep pass that picked only the idle `automation/backlog/` PR and ignored a `renovate/` one
- [x] All five workflow files parse, and their expressions were reviewed by hand. `actionlint` is not a dependency here. The `fromJSON` matrix is guarded by `has_dispatch`, so an empty queue cannot reach it
- [ ] N/A — `npm run typecheck`, `npm run test` (full), `npm run build`, `npm run build -w @slicc/chrome-extension`: this PR adds only `.mjs` tooling, workflow YAML, and Markdown. No TypeScript, no bundled code, and nothing in the build graph changes. CI will run all of them anyway; I ran the gates that can actually fail on this diff rather than claiming a full pass I did not do.
- [ ] N/A — manual float smoke / screencast: no UI or runtime surface changes.

**Pre-existing failures**: none observed in anything I ran.

## Documentation

- [x] Relevant `CLAUDE.md` — one grouped entry in `packages/dev-tools/CLAUDE.md` covering all five. Paid for it by dropping text already duplicated in its own deep-ref pages, so the guide sits at 9,880 chars and does not become the compactor's first customer
- [x] `docs/` — new `docs/dev-tools-details.md#scheduled-agentic-workflows`: the shared shape, the state substitutions, the two `CLAUDE.md` budgets, the runs-endpoint cap, the backlog dispatcher's recovered rubric and never-close policy, and the on-demand dispatch commands
- [x] A `README.md` per new `packages/dev-tools/<dir>/`
- [ ] `README.md` (root) — not needed; no user-facing behavior change

## Testing it on demand (no waiting for a cron)

All five take `workflow_dispatch` with `dry_run` (stops after the selector, fully read-only), plus an input that forces work to exist:

```bash
gh workflow run boy-scout-debt-dispatcher.yml -f file=packages/webapp/src/fs/sudo-fs.ts
gh workflow run pr-fix-dispatcher.yml         -f pr_number=1234
gh workflow run backlog-dispatcher.yml        -f issue=2071
gh workflow run claude-md-compactor.yml       -f max_chars=9000 -f target_chars=8500
gh workflow run flaky-ci-hunter.yml           -f window_days=14 -f job=node-server
```

`pr_number` is the one that needed building: the dispatcher otherwise only acts on a failing automation PR that is at least 20 minutes stale with no recent human activity, which is impossible to arrange on demand. Naming a PR waives those two waits and nothing else. `issue=<n>` does the same for the backlog dispatcher, waiving only the settling age.

## Risk & rollback

- **Blast radius**: CI/automation only. Nothing ships to a float, and no product behavior changes. The realistic failure mode is a noisy or wrong PR opened by one of these agents, not a regression in SLICC.
- **Kill switches**: each workflow is independently revertable, and every one can be disabled from the Actions UI without touching code. `dry_run` makes each one read-only on demand. Caps are conservative by design: one file per day, one flake per week, three fixers per tick with at most five in flight, two attempts per PR, one re-run per SHA, five backlog dispatches per tick with a standing ceiling of ten open backlog PRs.
- **Rollback**: revert cleanly, in whole or per workflow. No migrations, no persisted state to unwind — that is the main practical benefit of keeping state in GitHub rather than in a committed ledger. The only durable side effects are labels and comments, which a human can remove.
- **What CI cannot catch, and what I would watch on the first live runs**: the Bedrock auth path and per-workflow model vars are only exercised at runtime; `allowed_bots: '*'` is required for scheduled runs and unverifiable locally; the `fromJSON` matrix has not run against a real non-empty queue; and the first boy-scout and backlog runs are the first time Claude has to do real work and pass `check-touched-exemptions` on its own PR. Suggested order after merge: `dry_run=true` on all five, then a targeted `pr_number` run and an `issue=<n>` run on one well-understood issue, then let the crons take over.

## Checklist

- [x] Commits are focused and package-local where possible (one per workflow, plus review-fix and docs commits)
- [x] Public API / tool / shell-command / lick-channel changes: none
- [x] No contract used by other floats is changed, so there are no follower/leader/offscreen paths to check
- [x] Labels are net-new (`ci-fix-*`, `backlog-*`, `boy-scout-debt`, `flaky-fix`); the retired `cosmos-fix-*` labels on existing PRs are ignored, while the retired `cosmos-dispatched` / `cosmos-skipped` labels on ~50 issues ARE honoured as skip/dispatch history, because ignoring them would re-propose work already refused


trieloff:
Thanks — all three P1 findings were real and are fixed in 406b1cd9e.

**1 + 2. `issues: write` for label bootstrapping.** Correct, and it would have broken every non-dry run before it did any work. `gh label create` is `POST /repos/{repo}/labels`, which is gated on `issues`; `pull-requests: write` only covers applying an existing label to a PR, and `--force` is still a write. Both steps run under `GITHUB_TOKEN`, so job permissions govern them. Added `issues: write` to the PR fix dispatcher's `scan` job and to the flaky hunter, with a comment explaining the distinction — this is exactly why `rum-error-triage.yml` and `agentic-debt-triage.yml` both carry it. (The boy-scout dispatcher creates its label through `BOT_PAT`, whose scope is independent of job permissions, so it was unaffected.)

**3. Empty `201` from `rerun-failed-jobs`.** Also correct, and the failure ordering was the bad part: the re-run is already triggered by the time `res.json()` throws `Unexpected end of JSON input`, so the tick would have re-run the jobs and *then* died before emitting its queue and summary — losing the dispatch queue and leaving no record of what it did.

Rather than special-case one more status code, writes now go through a `ghWrite` helper that reports whether GitHub accepted the call and never touches the body (right for `rerun-failed-jobs`, label add/remove, and comments alike), and reads treat an empty body as `null`. That also fixed a latent bug in the same code path: the old `if (res !== null)` success check counted a tolerated `403`/`409` the same as a real acceptance, so the "re-ran N/M" figure could over-report.

Verified against a stubbed `201`-with-no-body response driving the real scanner end to end: the POST is issued, the scan survives it, takes the re-run path, and still writes its summary (exit 0). The pre-fix behaviour reproduces as `Unexpected end of JSON input`.

trieloff:
@codex review — please re-review at 406b1cd9e, which addresses your three P1 findings (issues: write on both label-bootstrapping jobs, and the empty-201 handling on rerun-failed-jobs).

trieloff:
Second round addressed in `3dd3b3bb4` — all five findings were real, including one I wrongly dismissed the first time round. Thanks both.

### P1 — the boy-scout fixer could not label its own PR

Correct, and my reasoning last round was wrong. I waved this away with "boy-scout is unaffected because it uses BOT_PAT", having assumed that PAT was broadly scoped. Its own header says otherwise: **contents + pull-requests write**. Label creation is `POST /repos/{repo}/labels`, gated on Issues, and `--force` is still a write — so the nightly fixer would have completed its refactor, run the whole verification pass, and then died at `gh label create` before opening the PR. The most expensive possible place to fail.

The label is now bootstrapped by a workflow step under `GITHUB_TOKEN` with `issues: write`, and the brief tells Claude the label already exists and that it must not try to create one. A test asserts the prompt contains no `gh label create` invocation, so this cannot regress quietly.

### P2 — forked automation PRs were queued for a fixer

Real, and the failure mode is worse than a wasted run: the fix job checks out the bare `head.ref` in this repository, so a fork PR either fails on a missing branch or — if a branch of the same name happens to exist here — **edits and pushes the wrong branch**. And as you note, the dispatch label and SHA marker are written before the checkout, so nothing ever retries it.

`screenPr` now refuses any head that is not a branch in this repository, including the deleted-fork case (`head.repo: null`). I kept one distinction: a PR object that never *states* a head repo is treated as unknown rather than as a fork, so fixtures and flattened shapes keep their meaning instead of silently becoming skips. Five tests, including the nested `head.repo.full_name` shape and the flattened `headRepo` one. I took the "require same-repo" option rather than threading the head repository through the matrix: pushing to a contributor's fork from a scheduled job is a privilege I would rather this workflow not have.

### P2 — the compactor verified the wrong threshold

Right: `--check` only ever received `MAX_CHARS`, so with `max_chars=9000 target_chars=8500` a guide left at 8,900 passed verification even though the brief promised 8,500 — and it would be re-selected the following week, having achieved nothing.

The measuring run now emits its `worklist`, and the check enforces both invariants: selected guides against the target, every tracked guide against the max. The worklist has to be carried in precisely because a successfully rewritten guide no longer looks oversized to a fresh measurement. New pure helpers `selectAboveTarget` / `parseWorklist` with seven tests, including the exactly-at-target boundary and the "no worklist, so a bare `--check` keeps its old meaning" case.

### P2 — the flake window was short by about 17 hours

Confirmed, and the arithmetic is as you describe. At the Monday 06:50 UTC cron, seven calendar dates spanned Tuesday 00:00 → Monday 06:50, roughly 6d 7h, so the previous Monday morning through Tuesday midnight was invisible — which is exactly enough to drop a job under the two-commit threshold and make a real flake look quiet.

`dayWindows` now returns every UTC day the trailing interval touches (`windowDays + 1` dates unless `now` is exactly midnight), and the extra boundary day is queried whole and then trimmed to the exact `now - windowDays` cutoff by a new `withinWindow`. A run with an unparseable `created_at` is deliberately kept: losing evidence to a bad timestamp is a worse failure than scanning slightly too much. The regression is pinned by a test at the real cron slot asserting the previous Monday is in the listing.

### Minor — merge-queue runs were invisible to source 3

Good catch, and the diagnosis of *why* it drifted is the useful part: the CLI filter was stricter than the pure logic beneath it, which already had `merge_group` in `MAIN_EVENTS`. Since `ci.yml` fires on `merge_group` in this repo, that filter was discarding most post-merge CI before it could corroborate anything.

Both layers now share one `isDefaultBranchRef` helper that accepts `main` or a `gh-readonly-queue/main/…` ref, so they cannot drift apart again. Fixing only the CLI would have left the same bug one layer down, since the pure logic compared `o.branch` to the default branch by equality too. Tested end to end with `event: 'merge_group'` and a real queue ref, which is the case the suite was missing.

---

Verification: `npm run verify` 7/7, `dev-tools` 868 tests (850 + 18), `check-touched-exemptions` clean. Extracting `describeForeignHead` was needed to keep `screenPr` under the cognitive-complexity gate, which this repo enforces. The three READMEs whose documented behaviour changed were updated in the same commit.

The fifth and last Cosmos expert (Backlog Dispatcher) is stacked on this branch in #2165 rather than added here, so this PR stays at the four workflows you have already reviewed.

Issue #2166, opened 2026-08-18
sprinkle: send cannot target a runtime, and unknown flags like --runtime are silently ignored with exit 0


trieloff:
## Summary

Two related gaps in `sprinkle` made a routine sprinkle problem undiagnosable for about an hour today, and one of them actively produced false confidence.

1. `sprinkle send` cannot target a runtime, so on a leader-plus-follower setup there is no way to push data to the panel instance that asked for it.
2. Unknown flags such as `--runtime=<id>` are silently swallowed and the command exits 0, so probes that appear to prove delivery prove nothing.

## Context

The `loose-ends` sprinkle hydrates itself by reading its store from the VFS, and emits a `request-load` lick when it cannot. On a `slicc-standalone` follower, `slicc.exec('cat …')` and `slicc.readFile()` both time out inside the sprinkle sandbox, because the follower's bridges are not backed by the leader's storage. So the follower's panel instance legitimately asks its owner to push the list.

The owner answers with `sprinkle send loose-ends '{"action":"load-items","tasks":[…]}'`. Whether that push ever reaches the follower's instance is currently unknowable.

## Repro: `--runtime` is accepted and ignored

```
$ host
status: leader
followers:
  - follower-8a474d67… (slicc-standalone) connected 5m ago [playwright]

$ sprinkle list --runtime=TOTALLY-FAKE-ID
  loose-ends [open]  Loose ends  (/shared/sprinkles/loose-ends/loose-ends.shtml)
  …
EXIT: 0

$ sprinkle list --runtime=follower-8a474d67-…
  loose-ends [open]  Loose ends  (/shared/sprinkles/loose-ends/loose-ends.shtml)
  …
EXIT: 0
```

Identical output and exit 0 for a fabricated runtime id and for the real follower. Same with the skill-level wrapper:

```
$ loose-ends reseed --runtime=BOGUS-NONEXISTENT-RUNTIME-ID
loose-ends: reseeded 'loose-ends' with 14 task(s) from /shared/loose-ends.json
EXIT: 0
```

## Repro: `send` has no targeting at all

```
$ sprinkle send loose-ends --runtime=$FOL '{"action":"noop-probe"}'
sprinkle send: invalid JSON                        EXIT: 1

$ sprinkle send loose-ends '{"action":"noop-probe"}' --runtime=$FOL
sprinkle send: invalid JSON                        EXIT: 1

[...]

Issue #2164, opened 2026-08-18
Memory curator has no writable scratch directory, so each pass gambles on a sudo prompt (/workspace/.s.txt, python3 -)


trieloff:
## What happened

During a single memory-curator pass today (session 65, 2026-08-18 15:52–15:58 UTC) the curator raised two cone-mediated sudo requests, both outside its declared grants:

| time | kind | detail | outcome |
| --- | --- | --- | --- |
| 15:52 | write | `/workspace/.s.txt` | denied |
| 15:58 | command | `python3 -` | denied |

The shipped curator prompt (`packages/vfs-root/shared/MEMORY.md`) grants `writablePaths: [/workspace/CLAUDE.md]` and an `allowedCommands` list of text tools. It also tells the curator, in the rules section:

> Write the memory file only. Do not create backups or scratch copies next to it — the file is versioned, and you have no write access anywhere else.

So the agent is instructed not to use scratch files, has nowhere to put one, and asks for one anyway. `python3 -` is worse than useless here: `python3` resolves on this runtime but dies with «pyodide is not installed in node_modules», so approving it would have burned a turn and failed.

## Why it happens

The pass is a measure-then-rewrite job under a hard character budget. The natural shape is: draft the new file, measure it, commit it. With exactly one writable file and no scratch directory, the only way to measure a draft is to write it into the live memory file first, which is precisely what the prompt warns against because a kill mid-write corrupts the store. An agent that takes the warning seriously has no legal move left, so it improvises: a hidden file in `/workspace`, or an interpreter to hold the draft in memory and do the arithmetic.

Cost of each improvisation: one sudo prompt to the cone, one wasted turn inside a pass whose whole design goal is «finish in well under 10 minutes», and, if the human is not watching, an approval that widens the sandbox for a reason nobody reviewed.

## Proposed change

Give the curator a scratch directory and tell it that interpreters do not exist here. Diff against `main`:

```diff
--- a/packages/vfs-root/shared/MEMORY.md
+++ b/packages/vfs-root/shared/MEMORY.md
@@ -1,6 +1,7 @@
 ---
 writablePaths:
   - /workspace/CLAUDE.md
+  - /shared/memory-curator/ # scratch space for draft-then-measure; deliberately outside /workspace so a stray write cannot litter the cone's tree or install skills
 visiblePaths:
   - /sessions/
   - /shared/
@@ -96,7 +97,7 @@
 upskill list                      # what is already installed — never suggest a duplicate
 ```
[...]

PR #2167, opened 2026-08-18
fix(dev-tools,ci): three live-run fixes — backlog dispatcher denied every issue, flake hunter chased a gate job, automation PRs got no CI

trieloff:
Follow-up to #2163: three bugs found by **running** the five migrated workflows on `main`. None was catchable by a unit test or a reviewer — each one needed a live run to see.

| # | Workflow | Symptom on the first live run | Fix |
|---|---|---|---|
| 1 | Backlog Dispatcher | Selected **0 candidates from 17 open items** and looked healthy doing it | Drop `skill issue` from the denylist; log per-issue rejection reasons |
| 2 | Flaky CI Hunter | **Dispatched a fixer at `CI / ci`**, a pure aggregator job, on a signature of Actions plumbing | Refuse to localize a signature containing no non-plumbing line |
| 3 | Boy Scout, Compactor, Backlog, Flaky | **Every PR they opened had ZERO checks** and waited on a human click | Claude pushes; a deterministic step opens the PR with BOT_PAT |

---

# 1. The Backlog Dispatcher denied every issue in the repo

The Backlog Dispatcher's first live dry run selected **0 candidates from 17 open items**. That looked like a clean backlog. It wasn't: `skill issue` was in `DENYLIST_LABELS`, and [`issue-skill.yml`](.github/workflows/issue-skill.yml) applies that label to **every** issue on `opened`, as a joke. In this repository the label means "an issue exists", not "user error", so the dispatcher was disabled by construction — 14 of 17 open items carried it, and every future issue is born with it.

Dropping it surfaces 2 genuine candidates and still rejects everything that should be rejected.

## Why

Two failures inside this one workflow, and the second is why the first survived a review round.

**1. The denylist entry was wrong for this repo.** I added `skill issue` reasoning from the label's name, which reads exactly like "user error, not a bug". Its actual semantics here are an unconditional joke label on `issues: [opened]`.

**2. The selector never said why it rejected anything.** `0 candidates` is this workflow's most common *healthy* outcome — a quiet backlog is normal and a no-op tick is a success. Without per-issue reasons, a correct quiet tick and a completely disabled selector produce identical logs, which is how this got through a review cycle and 76 unit tests. Both siblings (`pr-fix-dispatcher`, `boy-scout-debt`) log a per-item verdict; this one only logged a count. The reasons were already computed and returned by `selectCandidates` — nothing printed them.

- Removed `'skill issue'` from `DENYLIST_LABELS`, with a comment naming the workflow that applies it so nobody helpfully adds it back.
- Added `reportRejections()`: groups screened-out issues by reason code, prints the codes with issue numbers and one sample reason. Grouped rather than one line per issue so a repo with hundreds of open issues stays readable.

Verified live on this branch (`gh workflow run backlog-dispatcher.yml --ref bb/backlog-dispatcher-smoke-fixes -f dry_run=true`, run [32164236968](https://github.com/ai-ecoverse/slicc/actions/runs/32164236968)):

```
backlog-dispatcher: 2 candidate issue(s) after screening
  - #2157 (bug, score 40) Zero-byte files cannot be deleted — rm -f and rm -rf exit 0 while deleting nothing
  - #2164 (other, score -60) Memory curator has no writable scratch directory…
backlog-dispatcher: screened out 16 issue(s):
  - already-decided (12): #2072 #2062 #2043 #2042 #2034 #2022 #1979 #1957 #1666 #1619 #1490 #211
      Already carries "cosmos-skipped" — this dispatcher decides once. Remove the label to re-queue.
  - pull-request (3): #2086 #2029 #1782
  - too-young (1): #2166
      Opened 0.1h ago (< 1h settling window).
```

Every remaining rejection is correct: 3 are pull requests, 12 are the issues Cosmos itself already refused (the legacy-label equivalence doing its job), and #2166 was opened minutes earlier and is held by the 1h settling window.

## Cross-cutting impact

- **Floats exercised**: none. CI-only tooling.
- **Behaviour change**: the dispatcher goes from "never proposes anything" to "proposes up to 5 per day, capped at 10 open PRs". That is the intended behaviour from #2163, now actually reachable. The triage phase still has to agree an issue is ready, and the hard-override catalog is unchanged.
- **No other consumer** of `DENYLIST_LABELS` exists; `skill issue` keeps its meaning everywhere else (it is still a joke).

## Risk & rollback

- **Blast radius**: the dispatcher can now actually dispatch. Bounded by the existing caps (5 per tick, 10 open PRs) and the triage phase's judgement; every result is a reviewable PR.
- **Rollback**: revert this commit and it returns to doing nothing.
- **Worth watching on the first live run**: whether triage agrees with the two candidates the selector surfaced. #2164 scores −60 with `security-surface, architectural` smells, which is exactly the shape the hard-override catalog should refuse — a good first test of whether the reconstructed rubric actually bites.

---

# 2. The Flaky CI Hunter dispatched a fixer at a gate job

The first live run scored `CI / ci` at 3 (three attempt flips on distinct commits) and dispatched a fixer at it. `CI / ci` is `if: always()` over `needs: [everything]` — a pure aggregator that fails whenever any child job fails, and therefore *flips whenever any child flips*. Its "common failure signature" was:

```
echo "::error::One or more jobs failed or were cancelled"
##[error]One or more jobs failed or were cancelled
##[error]Process completed with exit code 1.
```

That names no failure mode. An aggregator has no nondeterminism of its own — it mirrors everyone else's — so a fixer sent at it chases a phantom, and the one PR per week this workflow is allowed to open would have been spent on nothing.

**Fix.** Rather than pattern-match job names (plenty of repos have a real job called `ci`), `localizeFlake` now requires the shared signature to contain at least one line that is not GitHub Actions plumbing. This generalizes past gate jobs: whenever the scanner cannot see a real error line it declines to dispatch and the job lands in the digest as "not localized", which is the honest description of what is known. The signature is still reported so a human can see what was observed.

Verified against the verbatim signature from the live digest (rejected) and a real flake with an `AssertionError` riding along with the same plumbing line (still localized). Both are pinned as tests.

`CI / e2e` — score 2, "no common failure signature across two flips" — was already handled correctly and is unchanged.

## Test plan (both fixes)

- [x] `npm run verify` — 7/7 pass
- [x] `node node_modules/vitest/vitest.mjs run --project dev-tools` — **947 passing / 41 files** (+4 tests for these two fixes)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs origin/main` — `OK (6 changed file(s), 0 still on any debt list)`
- [x] Fix 1 verified by a **live dry run on this branch** (log quoted above); a future re-add of `skill issue` now fails a test
- [x] Fix 2 verified against the **live digest artifact's exact signature**
- [ ] N/A — `typecheck` / full `test` / `build`: `.mjs` tooling only, nothing in the build graph
- [ ] UI change — N/A

## Documentation

- [x] `flaky-ci-hunter/README.md` — the filter list documents the plumbing-only rule and why it is signature-based rather than name-based
- [x] `backlog-dispatcher/README.md` — no change needed; it points at `DENYLIST_LABELS` rather than duplicating the list

## Checklist

- [x] No secrets in the diff

---

# 3. Every automation PR landed with zero checks

The compactor's PR (#2168) and the boy-scout PRs (#2169, #2170) all arrived with **no checks at all**. Not failing — never started. CI, Claude PR Review, AI Comment Detection and both Renovate workflows were queued as `action_required`, waiting for a human to click "Approve and run". I had to approve them by hand to find out whether the work was even correct.

**Cause.** `claude-code-action` forces its Bash `gh` onto the action's own `github_token:` input, overriding the step-level `GH_TOKEN: ${{ secrets.BOT_PAT }}` that these workflows set for precisely this reason. (The comment next to it in #2163 even said "BOT_PAT (not GITHUB_TOKEN) so the created PR triggers CI" — the intent was right, the mechanism was not.) So `gh pr create` ran as GITHUB_TOKEN, the PR came out authored by `github-actions[bot]`, and GitHub gates workflow runs on such a PR behind manual approval. A nightly agent whose output needs a human click before it can even be evaluated is not automated.

**Fix — the shape this repo already uses.** `coverage-ratchet.yml` has been creating its PR in a plain shell step with `GH_TOKEN: ${{ secrets.BOT_PAT }}` all along, and its PRs pick up the full suite unattended. So: Claude pushes the branch and writes the PR title/body to files, and a deterministic step opens the PR with BOT_PAT. Labelling is a separate step on GITHUB_TOKEN, because a label is an Issues API write and BOT_PAT carries contents + pull-requests only — `gh pr create --label` under BOT_PAT would 403 and lose the PR.

**Proven live**, dispatched from this branch:

| PR | Opened by | Author | Checks |
| --- | --- | --- | --- |
| #2170 | Claude, old flow | `github-actions[bot]` | **0** — still stuck |
| #2171 | new deterministic step | `trieloff` | **34, ran unattended** |

(#2171 was a test artifact and is closed: because the run was dispatched with `--ref` on this branch, Claude branched off it and swept this PR's own 20 files in beside the debt payoff.)

Two things improve beyond the checks. PR creation is no longer the model's job, so it cannot be half-done — and a run where Claude deliberately pushed nothing (the "prohibited fix" escape hatch) is now a logged no-op rather than an ambiguity. In the backlog dispatcher the `backlog-ready` → `backlog-dispatched` swap also moves into the deterministic step, so an issue can no longer be marked dispatched with no PR behind it.

One assumption in this design was worth checking rather than trusting: that a step-level `env:` var reaches Claude's Bash tool at all, since no existing workflow in this repo relies on that. The live run settles it — the PR body file existed and the PR opened, where a missing var would have produced the step's `::warning::` and no PR.

## Test plan (fix 3)

- [x] `node node_modules/vitest/vitest.mjs run --project dev-tools` — **950 passing / 41 files**
- [x] `npm run verify` — 7/7 · `check-touched-exemptions.mjs origin/main` — `OK (19 changed file(s), 0 still on any debt list)` · `npm run lint:docs` — clean
- [x] Prompt tests assert `gh pr create` never appears as an imperative in any of the four briefs
- [x] End-to-end live run, table above
- [x] The four new shell steps were also exercised offline against a scratch repo with a stub `gh` across every path: no branch pushed, branch + body, PR already open, branch but no body

### DIFF
--- a/.github/workflows/backlog-dispatcher.yml
+++ b/.github/workflows/backlog-dispatcher.yml
@@ -15,8 +15,12 @@ name: Backlog Dispatcher
 #      rejected ones `backlog-skipped` (with ONE comment saying why). It writes
 #      no code and opens no PR.
 #   2. AUTHOR — a second claude-code-action run takes the open `backlog-ready`
-#      issues, implements the minimal change, opens the PR with `Closes #<n>`,
-#      and swaps the label to `backlog-dispatched`.
+#      issues, implements the minimal change, pushes one branch per issue, and
+#      writes each PR's title + body to `$RUNNER_TEMP/backlog-pr-<n>.{title,md}`.
+#      It opens no PR and touches no `backlog-dispatched` label.
+#   2b. a deterministic shell step opens ONE PR per pushed branch with BOT_PAT,
+#      labels it, and swaps the issue's `backlog-ready` label for
+#      `backlog-dispatched` (see that step's note on why Claude cannot do it).
 #   3. a deterministic stale sweep (sweep-stale-prs.mjs) labels + comments once
 #      on dispatcher-owned PRs idle for a week. It NEVER closes a PR.
 #
@@ -61,11 +65,16 @@ name: Backlog Dispatcher
 #                             bearer token) for claude-code-action, shared with
 #                             boy-scout-debt-dispatcher.yml / rum-error-triage.yml.
 #   BOT_PAT                   Fine-grained PAT scoped to this repo (contents +
-#                             issues + pull-requests write). The branch is pushed
-#                             and the PR opened with BOT_PAT — NOT GITHUB_TOKEN —
-#                             because GitHub's anti-recursion guard suppresses
-#                             workflow runs for GITHUB_TOKEN-authored pushes, and
-#                             these PRs MUST get the full check suite.
+#                             pull-requests write; NO issues). The branches are
+#                             pushed and the PRs opened with BOT_PAT — NOT
+#                             GITHUB_TOKEN — because GitHub's anti-recursion
+#                             guard suppresses workflow runs for
+#                             GITHUB_TOKEN-authored pushes and queues the checks
+#                             of a GITHUB_TOKEN-authored PR as
+#                             `action_required`, while these PRs MUST get the
+#                             full check suite unattended. Every label and issue
+#                             edit is an Issues API write BOT_PAT cannot make, so
+#                             those calls use GITHUB_TOKEN.
 # Optional repo variables: BACKLOG_BEDROCK_MODEL / RUM_BEDROCK_MODEL,
 # RUM_AWS_REGION.
 #
@@ -234,16 +243,23 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
-      - name: Author the PRs for backlog-ready issues
+      - name: Author the branches for backlog-ready issues
         if: steps.ready.outputs.has_ready == 'true' && github.event.inputs.dry_run != 'true'
         uses: anthropics/claude-code-action@be7b93b1907a4abad570368f3c74b6fe3807510b # v1
         env:
           # Same Bedrock auth as phase 1 (see the notes on that step).
           AWS_BEARER_TOKEN_BEDROCK: ${{ secrets.AWS_BEARER_TOKEN_BEDROCK }}
           AWS_REGION: ${{ vars.RUM_AWS_REGION || 'us-east-1' }}
-          # BOT_PAT (not github.token) so the opened PR gets the full check
-          # suite — see the header note on the anti-recursion guard.
-          GH_TOKEN: ${{ secrets.BOT_PAT }}
+          # Claude's Bash `gh` always ends up with the action's own
+          # `github_token:` input, NOT whatever GH_TOKEN is set here — the action
+          # overrides it. That is why the PRs are opened by the step below
+          # instead; this phase's own `gh` work (the skip-path label swaps) needs
+          # only the job's GITHUB_TOKEN, and branch pushes use the BOT_PAT
+          # credential persisted by actions/checkout.
+          GH_TOKEN: ${{ github.token }}
+          # One body + title file per issue; the step below opens the PRs from
+          # them. Passed as env so the brief and that step cannot drift apart.
+          PR_BODY_FILE_TEMPLATE: ${{ runner.temp }}/backlog-pr-<issue-number>.md
         with:
           use_bedrock: 'true'
           # Job permissions above scope this token; Claude's own `gh`/`git` work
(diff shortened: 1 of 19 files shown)


trieloff:
@codex review — a second fix landed after your pass (`b7d3692ff`): the Flaky CI Hunter was dispatching a fixer at `CI / ci`, an `if: always()` aggregator over `needs: [everything]`, on a "common signature" of pure Actions plumbing. `localizeFlake` now requires at least one non-plumbing line before it will localize. Please review that commit too.

trieloff:
Both P2s fixed in `35b480bf5`, plus a third thing your second comment made me realise.

**P2 — `skill issue` still documented as denied.** Correct, and worse than you flagged: my own PR body claimed "no change needed; the README points at `DENYLIST_LABELS` rather than duplicating the list". It duplicated the list. The README now names the authoritative source and explains why that label must never go back on the denylist, so the next person who reads the name and reaches for it hits the reason first.

**P2 — grouped output misattributed reasons.** Also correct, and it defeats the point of the diagnostic: `denylisted` covers every label, so two issues denied for `question` and `invalid` would both be reported under whichever came first — hiding exactly the label the log exists to reveal. `too-young` has the same problem with ages.

Reasons that differ are now attributed per issue; a reason shared by the whole group still collapses to one line, so the common case stays compact:

```
  - already-decided (12): #2072 #2062 #2043 … #211
      Already carries "cosmos-skipped" — this dispatcher decides once. Remove the label to re-queue.
  - denylisted (2):
      #1 — Carries "question".
      #2 — Carries "invalid".
```

You asked for a test, and the formatting lived in a private function in the thin I/O CLI where nothing could reach it — so it moved to `lib.mjs` as the pure `formatRejections`, which is where the repo's rule/judgement split says it belongs. Covered by a mixed-label test asserting per-issue attribution *and* that a uniform group does not fragment.

**Third fix, prompted by yours.** This branch moved PR creation out of Claude into a deterministic BOT_PAT step (a PR Claude opens is authored by `github-actions[bot]`, and GitHub then queues every check on it as `action_required` — observed on #2168 and #2169, both landed with zero checks). The prompts now instruct Claude not to open the PR, but your point about a diagnostic that can quietly lie applies here too: a model that ignored the instruction would recreate the `action_required` PR, and my new step would find it already open and just label it — a regression indistinguishable from success. So it is now enforced, not requested: `--disallowedTools "Bash(gh pr create:*),Bash(gh pr edit:*),Bash(gh pr merge:*)"` on boy-scout / compactor / flake hunter, and the backlog author's allowlist narrowed from `Bash(gh pr:*)` to `list`/`view`. I checked `--disallowedTools` against the installed Claude CLI's `--help` rather than assuming the flag exists.

Verification: `npm run verify` 7/7, **952 tests** in the `dev-tools` project, `check-touched-exemptions` OK on 19 changed files, `lint:docs` clean.

PR #2177, opened 2026-08-18
fix(ci): review our own agents' pull requests instead of skipping them

trieloff:
## Summary

The PR reviewer skipped every bot-authored pull request, which meant the PRs our five scheduled agents open got **no automated review at all**. #2168, #2169 and #2170 each arrived with zero reviews; I only got any by dispatching this workflow by hand.

## Why

The `if:` rule was written for Renovate dependency bumps, and for those it is right — a version bump does not need an AI review. Applied to an agent's PR it removes review from the code that needs it most: a model wrote it unattended, and no human put their name on it as author.

There is no fallback, either. Codex and Copilot both ignore bot-authored PRs — Codex stayed silent on #2168 through two explicit `@codex review` requests — so this workflow is the only review such a PR can get.

And the manual reviews I had to trigger found real things: on #2169 it caught that the PR description still described an abandoned first attempt and contradicted the shipped diff. That is exactly the class of defect nobody else was going to catch on an unattended PR.

## Approach

Two changes, and either one alone does nothing:

1. **The job-level `if:` now admits `automation/*` heads regardless of author.** Since #2167 those PRs are opened with `BOT_PAT`, so their author is a User and the existing type check already lets them through — but that makes review contingent on which token happened to open the PR, a property we could regress without noticing. Keying on the head ref states the intent directly: our automation's PRs get reviewed.

2. **`allowed_bots: '*'` on the review step.** A PR opened by `github-actions[bot]` makes the bot the triggering actor, and `claude-code-action`'s human-actor gate would reject the run outright ("Workflow initiated by non-human actor") — the exact noisy failure the blanket exclusion was avoiding. Without this the job starts and then refuses, which is worse than skipping.

Renovate stays excluded, because its heads are `renovate/*`. The fork guard is untouched: in-repo heads only, since `pull_request` from a fork cannot see secrets.

## Cross-cutting impact

- **Floats exercised**: none. CI-only.
- **Blast radius**: one more review per automation PR (roughly one a day from the boy-scout dispatcher, plus occasional compaction/flake/backlog PRs). The reviewer has `contents: read`, so a review can never push.
- **Security**: `allowed_bots: '*'` permits bot-*initiated* runs; it does not widen what the reviewer may do, and the job-level `if:` still restricts which PRs reach it.

## Test plan

- [x] `npm run verify` — 7/7
- [x] `npm run lint:docs` — clean
- [x] `node node_modules/vitest/vitest.mjs run --project dev-tools packages/dev-tools/pr-review-gate/` — 8 passing
- [x] `check-touched-exemptions.mjs origin/main` — `OK (2 changed file(s), 0 still on any debt list)`
- [x] `npx prettier --check` on the workflow; the `if:` expression and both `with:` keys verified by parsing the YAML rather than by eye
- [ ] **Verified live after merge**: the next boy-scout PR should draw a review with no human dispatch. That is the only real proof, and I will confirm it on the next nightly (or a manual dispatch) rather than assume.
- [ ] N/A — `typecheck` / `test` / `build`: workflow YAML and a README

## Documentation

- [x] `packages/dev-tools/pr-review-gate/README.md` — documents who is eligible before the gate runs, the `automation/*` carve-out, and why (the gate's own skip rules were documented; eligibility was not)

## Risk & rollback

- **Rollback**: revert; automation PRs go back to being unreviewed.
- **Watch for**: a review loop. The reviewer only comments and the gate skips any PR that already has inline review comments, so a review cannot trigger another review — but the first automation PR after this merges is worth a look to confirm exactly one review lands.

## Checklist

- [x] No secrets in the diff

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -64,14 +64,26 @@ jobs:
     # Skip fork PRs: secrets are unavailable on `pull_request` from forks, and
     # reviewing them safely would require `pull_request_target` + a security
     # review (out of scope). Only run when the PR head lives in this repo.
-    # Also skip bot-authored PRs (e.g. Renovate dependency bumps): they don't
-    # need an AI review, and claude-code-action rejects bot-initiated runs by
-    # default ("Workflow initiated by non-human actor"), which only produced
-    # noisy failed runs. Manual dispatch always runs (operator intent).
+    #
+    # Bot-authored PRs are skipped — EXCEPT the `automation/*` heads our own
+    # scheduled agents push. That exclusion was written for Renovate dependency
+    # bumps, which genuinely do not need an AI review. Applied to an agent's PR
+    # it removes review from the code that needs it most: a model wrote it
+    # unattended, and nobody assigned themselves as its author. Observed on
+    # #2168/#2169/#2170, which arrived with no automated review of any kind
+    # (Codex and Copilot ignore bot-authored PRs too, so this workflow is the
+    # only reviewer they can get).
+    #
+    # Those PRs are normally authored by the BOT_PAT owner — a User, not a Bot —
+    # so the type check alone usually lets them through. The head-ref carve-out
+    # is what makes that independent of which token happened to open the PR.
+    # Manual dispatch always runs (operator intent).
     if: >-
       github.event_name == 'workflow_dispatch' || (
-        github.event.pull_request.head.repo.full_name == github.repository &&
-        github.event.pull_request.user.type != 'Bot'
+        github.event.pull_request.head.repo.full_name == github.repository && (
+          github.event.pull_request.user.type != 'Bot' ||
+          startsWith(github.event.pull_request.head.ref, 'automation/')
+        )
       )
     steps:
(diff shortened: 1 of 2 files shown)


trieloff:
Fixed in 13dbb4bbc — good catch, and the reasoning behind it is worth recording.

You are right that this is not a defect: the trust boundary was never the author, it is the head repo. Pushing an `automation/*` branch to this repository already requires write access, so the code the reviewer executes was authored by someone trusted either way. But the header asserted a property the `if:` no longer guarantees, and a security comment that misdescribes its own gate is worse than no comment — the next person to touch that `if:` would reason from a false premise.

Rather than just deleting the stale clause, the block now states the boundary explicitly, so it stays true if the author check changes again:

```
# Note the trust boundary is the HEAD REPO, not the author: bot-authored
# `automation/*` PRs are reviewed (see the job `if`), and that is safe for the
# same reason a human's PR is — pushing an `automation/*` branch to this repo
# already requires write access, so the code was authored by someone trusted.
```

Also worth noting for the record: your verification of the `automation/` prefix across all five agents plus the coverage ratchet is the check I should have run myself and did not — I reasoned from the four workflows I had just edited.

PR #2184, opened 2026-08-19
refactor(cloud-core): extract shared poll-join-file loop in polling.ts

### LINKED ISSUE #2183: Complicatification: cloud-core polling.ts forks one poll-loop into two ~90%-identical 40-line functions differing only in a freshness source, an error code, and one label string
## Where

`packages/cloud-core/src/polling.ts` — two exported functions:

- `pollCloudStatus` — `packages/cloud-core/src/polling.ts:19-52`
- `pollForRefreshedStatus` — `packages/cloud-core/src/polling.ts:54-98`

## What

The two functions are copy-paste forks of the same ~40-line poll-until-ready loop. Both:

- run the identical `while (Date.now() - start < opts.timeoutMs)` loop;
- `await handle.readFile('/tmp/slicc-join.json')`, `JSON.parse`, and check `parsed.joinUrl`;
- compare `parsed.updatedAt` against a freshness floor with the same `>` string-compare and the same "no floor ⇒ accept any well-formed read" branch;
- capture `lastStalePayload` / `lastError`;
- `await new Promise((r) => setTimeout(r, opts.intervalMs))`;
- build a nearly word-for-word three-branch `errSuffix` (stale / last-error / never-appeared) and `throw new CloudError(...)`.

They differ in only **three** details:

1. where the freshness threshold comes from — an optional `opts.minUpdatedAt` field (`:16`, `:28-29`) vs a positional `baselineUpdatedAt` argument (`:56`, `:70-71`);
2. the `CloudError` code — `SANDBOX_NOT_READY` (`:49`) vs `LEADER_NOT_READY` (`:95`);
3. one clause in the stale-suffix string — `minUpdatedAt=` (`:40`) vs `baseline.updatedAt=` (`:86`), plus a slightly different timeout message.

## Why this is Complicatification

An entire retry/timeout/stale-vs-missing-error-reporting loop is duplicated when the actual variation is two scalars and one label. This is the same failure mode the cdp layer's shared `PendingRequestTable` was extracted to eliminate (closed #1926): any future change to the loop — honoring an `AbortSignal`, jittering the interval, fixing the `>` vs `>=` freshness semantics, or adjusting the error contract — must now be made in two places and can silently drift. The freshness-comparison semantics in particular are load-bearing (see `packages/cloud-core/CLAUDE.md` "Stable Contracts" → `ConeEntry.trayId and lastJoinUpdatedAt`), so drift here is a correctness risk, not just cosmetic.

## Scope / occurrence

- 2 functions in 1 file, ~40 lines each, ~90% identical.
- Each has exactly **one** caller: `pollCloudStatus` ← `packages/cloud-core/src/operations/start.ts:256`; `pollForRefreshedStatus` ← `packages/cloud-core/src/operations/resume.ts:155`.

## Suggested remediation

Extract one private helper that takes the freshness floor and error code as parameters, e.g.:

```ts
async function pollJoinFile(
  handle: SandboxHandle,
  opts: PollOpts & { floor?: string; errorCode: CloudErrorCode; timeoutMessage: (ms: number) => string }
): Promise<CloudStatus> { /* the single shared loop */ }
```

Then the two exports become thin wrappers:

- `pollCloudStatus(handle, opts)` → `pollJoinFile(handle, { ...opts, floor: opts.minUpdatedAt, errorCode: 'SANDBOX_NOT_READY', ... })`
- `pollForRefreshedStatus(handle, baselineUpdatedAt, opts)` → `pollJoinFile(handle, { ...opts, floor: baselineUpdatedAt, errorCode: 'LEADER_NOT_READY', ... })`

This removes ~40 duplicated lines and gives the freshness/timeout/error logic a single source of truth. The existing freshness-poll tests in `packages/cloud-core/tests/` pin the behavior on both paths, so the refactor is verifiable.

trieloff:
Closes #2183

## What

`packages/cloud-core/src/polling.ts` had two exported functions —
`pollCloudStatus` and `pollForRefreshedStatus` — that were ~90%-identical
copies of the same poll-until-ready loop over `/tmp/slicc-join.json`. They
differed only in three details: where the freshness floor comes from
(`opts.minUpdatedAt` vs the positional `baselineUpdatedAt`), the `CloudError`
code (`SANDBOX_NOT_READY` vs `LEADER_NOT_READY`), and two label/message
strings.

This change folds the shared loop into a single private `pollJoinFile` helper
parameterized by `floor`, `floorLabel`, `errorCode`, and a `timeoutMessage`
builder. The two exports become thin wrappers over it. Public signatures are
unchanged, and so is all observable behavior:

- strict-newer (`>`) freshness comparison, with the "no floor ⇒ accept any
  well-formed read" branch;
- the three-branch timeout suffix (stale / last-error / never-appeared), with
  the same `minUpdatedAt=` vs `baseline.updatedAt=` labels;
- the same two error codes and timeout messages.

The freshness/timeout/error logic now has a single source of truth, so future
changes (AbortSignal support, interval jitter, `>` vs `>=` semantics, error
contract tweaks) can no longer silently drift between the two paths.

## Why

Flagged as Complicatification debt (see the issue). The freshness-comparison
semantics are load-bearing per `packages/cloud-core/CLAUDE.md` → "Stable
Contracts" → `ConeEntry.trayId and lastJoinUpdatedAt`, so duplicated drift here
is a correctness risk, not just cosmetic.

## How verified

- New focused test `packages/cloud-core/tests/polling.test.ts` (8 tests) pins
  both wrappers: no-floor acceptance, strict-newer rejection-then-accept, and
  each timeout error (stale suffix + label, last-error suffix, and the distinct
  `SANDBOX_NOT_READY` / `LEADER_NOT_READY` codes and messages).
- Existing `start.test.ts` (11) and `resume.test.ts` (11) still pass — the two
  callers are unchanged.
- `npm run verify` (lint gate, 7/7), `npm run typecheck`,
  `npm run test:coverage:cloud-core` (116 tests; `polling.ts` at 100%
  statements/functions/lines), and the touched-file exemptions gate all pass.

### DIFF
--- a/packages/cloud-core/src/polling.ts
+++ b/packages/cloud-core/src/polling.ts
@@ -1,3 +1,4 @@
+import type { CloudErrorCode } from './errors.js';
 import { CloudError } from './errors.js';
 import type { SandboxHandle } from './substrate.js';
 import type { CloudStatus } from './types.js';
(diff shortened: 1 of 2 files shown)


trieloff:
Addressed the Codex P2 on the 40 ms wall-clock deadline in `packages/cloud-core/tests/polling.test.ts` (ab8ae6263).

Rather than fake timers, the fixture options are split so no test depends on scheduler timing:

- `SUCCESS_OPTS` (`timeoutMs: 2_000`) for the tests that expect the poll to resolve. The deadline is never actually reached — the fixture serves an acceptable read on the first or second pass — so the larger value costs no wall-clock time while removing the race between an event-loop stall and the fresh second read.
- `TIMEOUT_OPTS` (`timeoutMs: 40`) stays for the three timeout-error tests, where reaching the deadline *is* the expected outcome; a slow worker can only make it arrive sooner in loop-iteration terms, never change the result.

Re-verified: `npm run verify` (7/7), `check-touched-exemptions` (OK), `npm run typecheck`, `npm run test` (14,769 passed), `npm run test:coverage:cloud-core` (`polling.ts` still 100% statements/functions/lines). Builds untouched — the change is test-only.

Issue #2146, opened 2026-08-17
VFS: phantom directory entries return ENOTDIR on every unlink path but escape via rename — regression of #2006 on 6.54.8→6.54.12, wedges the whole .jsh runtime


trieloff:
Regression of #2006 (closed) observed live on a production leader across the `6.54.8` → `6.54.12` upgrade, plus two findings that were not in the original report: an **in-session recovery that actually works**, and a **blast radius** that takes down the whole in-process `.jsh` runtime.

## Timeline (2026-08-17, production leader, live)

The session was mid-task with a `one.newrelic.com` tab under active `playwright-cli` automation. `6.54.12` was released at `12:30:04Z` while the session was running on `6.54.8`.

At ~13:43Z, `/.playwright/session.md` — a file that `playwright-cli` had been writing to successfully for hours — became a **directory** in the VFS index:

```
$ ls -ld /.playwright/session.md
drwxr-xr-x 1 user user 15014 Aug 17 15:43 /.playwright/session.md/

$ ls -la /.playwright/session.md/
ls: /.playwright/session.md/: No such file or directory
```

Note the size, `15014`, is identical to the sibling `snapshots/` directory. `lstat` claims directory; enumerating it says it does not exist.

Every access path disagrees:

| Operation | Result |
| --- | --- |
| `fs.lstatSync` | reports **directory**, size 196 |
| `fs.rmSync(recursive, force)` | `ENOTDIR` |
| `fs.rmdirSync` | `ENOTDIR` |
| `fs.unlinkSync` | `ENOTDIR` |
| `fs.writeFileSync` | `EINVAL` |
| `fs.readFileSync` | `EINVAL` |
| `rm -rf` | silent no-op, entry persists |
| **`mv`** | **succeeds** |

## Finding 1: rename is a working in-session recovery

#2006 concluded there is no in-session recovery path. There is one: **`rename` escapes the poisoned entry when every unlink/rmdir path returns `ENOTDIR`.**

```
mv /.playwright/session.md /.playwright/session.md.broken   # works
printf '' > /.playwright/session.md                          # recreates as a real file
```

After that, `playwright-cli` wrote to `session.md` normally again. Whatever validation rename takes clearly bypasses the type check that unlink and rmdir trip over. That asymmetry looks like the actionable clue: the index entry's type is wrong, delete paths dispatch on the wrong type and refuse, rename does not care.

## Finding 2: it spreads, and it wedges the whole in-process runtime

[...]

PR #2187, opened 2026-08-19
fix(fs): materialize zero-byte files on OPFS so they can be deleted

### LINKED ISSUE #2157: Zero-byte files cannot be deleted — rm -f and rm -rf exit 0 while deleting nothing
## Summary

Zero-byte files under `/workspace` **cannot be deleted**. The delete is accepted in-memory, then the sync-fs flush to the backing store fails with `ENOENT` on `unlink`, and the VFS entry **reverts to existing**.

The exit codes are wrong in every variant, and two of them are silent:

| Command on a 0-byte file | exit code | stderr | file afterwards |
| --- | --- | --- | --- |
| `rm f` | **1** | `cannot remove 'f': No such file or directory` | **survives** |
| `rm -f f` | **0** | *(nothing)* | **survives** |
| `rm -rf dir/` (contains one) | **0** | *(nothing)* | **dir + 0-byte file survive** |
| `mv f g` | 1 | `cannot stat 'f': No such file or directory` | not renamed |

`rm -f` and `rm -rf` are the dangerous ones: they report complete success while deleting nothing, so any cleanup step believes it ran. This is the same silent-wrong-answer class as #1733 (`cmp`) and #1728 (`merge-base`) — the caller cannot detect the failure.

A **1-byte** file in the same directory deletes correctly every time, which is what isolates the trigger to size 0.

## Root cause (evidence below)

A zero-byte write is never materialised on the backing store — the flush log reports **`0 write(s)`** for a file that was just created. The host therefore has no such path, so the subsequent `unlink` raises `ENOENT` and the optimistic in-memory delete is rolled back:

```
[sync-fs] ERROR: flush failed — 0 write(s) + 1 delete(s) were NOT persisted:
sync-fs flush failed for 1 path(s): /workspace/zbtest/flushprobe.txt:
ENOENT: no such file or directory, unlink '/workspace/zbtest/flushprobe.txt'
```

So the defect is on the **write** side (an empty write is dropped as a no-op), and it only becomes visible on delete.

## Environment

SLICC 6.55.4 (upgraded from 6.15.1 on 2026-08-18). Reproduced in the cone shell with unrestricted filesystem access — this is not a sudoers/permission issue.

Scope: the file **survives** only under `/workspace` (host-synced). Under `/shared` and `/scoops/<folder>` the file *is* removed, but `rm` still exits **1** with the same spurious `No such file or directory`, so the wrong exit code is universal while the failed deletion is `/workspace`-specific.

## Reproduction

```bash
rm -rf /workspace/zbtest; mkdir -p /workspace/zbtest; cd /workspace/zbtest

: > redir.txt          # 0 bytes
touch touched.txt      # 0 bytes
printf 'x' > onebyte.txt   # 1 byte — control

# 1. plain rm: exits 1, file survives
rm redir.txt; echo "rc=$?"
node -e 'console.log("survives:", require("fs").existsSync("/workspace/zbtest/redir.txt"))'

# 2. rm -f: exits 0, prints nothing, file STILL survives
rm -f redir.txt; echo "rc=$?"
node -e 'console.log("survives:", require("fs").existsSync("/workspace/zbtest/redir.txt"))'

# 3. control: 1-byte file deletes fine
rm onebyte.txt; echo "rc=$?"
node -e 'console.log("survives:", require("fs").existsSync("/workspace/zbtest/onebyte.txt"))'

# 4. rm -rf silently no-ops on a directory holding a 0-byte file
mkdir -p sub && : > sub/empty.txt
rm -rf sub; echo "rc=$?"
node -e 'const fs=require("fs");console.log("dir:",fs.existsSync("/workspace/zbtest/sub"),"file:",fs.existsSync("/workspace/zbtest/sub/empty.txt"))'
```

### Actual

```
rc=1
survives: true
rc=0                   <- rm -f claims success
survives: true         <- but it is still there
rc=0
survives: false        <- 1-byte control deletes correctly
rc=0                   <- rm -rf claims success
dir: true file: true   <- nothing was deleted
```

### Expected

`rm f` exits 0 and removes the file; `rm -f f` likewise; `rm -rf dir` removes the directory entirely. A 0-byte file should be no less deletable than a 1-byte one.

### Partial-deletion variant

A directory holding both an empty and a non-empty file is left half-cleaned, with `rc=0`:

```bash
mkdir -p mixed && : > mixed/empty.txt && printf 'data' > mixed/full.txt
rm -rf mixed; echo "rc=$?"     # -> rc=0
# mixed/ -> still exists
# mixed/empty.txt -> still exists
# mixed/full.txt  -> deleted
```

### The read-back is also briefly inconsistent

`fs.rmSync` returns without throwing and an immediate `existsSync` reports `false`; after the flush fails, the same check reports `true`. Code that verifies its own cleanup can therefore observe the delete succeeding and only later find the file back:

```bash
: > flushprobe.txt
node -e 'const fs=require("fs");fs.rmSync("/workspace/zbtest/flushprobe.txt");console.log("immediately after:",fs.existsSync("/workspace/zbtest/flushprobe.txt"))'
node -e 'console.log("moments later:",require("fs").existsSync("/workspace/zbtest/flushprobe.txt"))'
# immediately after: false
# moments later: true
```

All four ways of creating an empty file behave identically (`: >`, `touch`, `printf '' >`, `fs.writeFileSync(p,"")`), so the trigger is the resulting size, not the creating command. `fs.unlinkSync` fails the same way as `fs.rmSync`.

## Workaround

Give the file a byte first, then delete it:

```bash
printf 'x' > f && rm f     # verified: rc=0, actually gone
```

## Impact on agents

Cleaning up scratch files is a reflex at the end of almost every task, and empty files arise constantly and unremarkably — a log nothing was written to, a marker/lock file, a result file for a query that returned no rows (this was found via a `vfs-*.txt` file that legitimately held zero matches). The agent runs `rm -f`, gets exit 0 and silence, and reports the workspace clean; the file is still there. Because `rm -rf` on a directory also returns 0 while leaving that directory in place, "remove my temp dir" can silently do nothing at all.

The `mv` failure removes the natural fallback — the file cannot even be renamed out of the way — so without knowing the `printf 'x'` trick an agent has no way to clear it.

I also suspect this is the mechanism behind the long-standing "phantom deletions" seen in git clones inside `/workspace` (files that `git status` shows as deleted but that will not go away, and empty files a checkout cannot remove). Worth checking whether that report resolves to this same root cause.

Suggested fix: materialise zero-length files on write (do not treat an empty write as a no-op), and make a flush failure surface as a non-zero exit from `rm`/`rm -f` rather than being swallowed.

trieloff:
Fixes #2157

## Summary

A zero-byte file under `/workspace` could not be deleted. Before: `rm f` exited 1 with a spurious "No such file or directory" and the file survived, while `rm -f f` and `rm -rf dir/` exited 0 with no diagnostic at all and deleted nothing. After: a 0-byte file is materialised on the backing store when it is created, so all three delete it and exit 0 — a 1-byte file in the same directory always worked, which is what isolated the trigger to size 0.

## Why

The reporter's hypothesis was correct — **a zero-byte file is never materialised on the backing store**, so the later `unlink` cannot find it. The exact mechanism, verified rather than assumed:

1. ZenFS' VFS layer flushes a file by walking its **dirty byte ranges** (`vnode.sync()` in `@zenfs/core`). An empty file has none, so the loop body never runs and the backend's `write` is **never called**.
2. `IndexFS.createFile` only records an inode in the in-memory index. `WebAccessFS` (`@zenfs/dom`, the OPFS backend behind `/workspace`) therefore never creates the `FileSystemFileHandle`.
3. `stat` / `exists` answer from the index, so the file looks real (size 0). `IndexFS.unlink` drops the index entry and then calls `WebAccessFS.remove` → `removeEntry` → `NotFoundError` → **`ENOENT: no such file or directory, unlink '<path>'`**, byte-for-byte the message in the issue.
4. ZenFS clears the vnode cache only **after** a successful `unlink` (`cacheOf(fs).remove(resolved)` sits below the throwing call), so the failed removal leaves the path cached and the entry **reverts to existing**.

That settles the reporter's open question 4: the later `existsSync` is served from metadata (the ZenFS index / vnode cache), not from the host — the broken half is the **create**, as suspected.

It also explains every variant with no `rm`-specific behaviour anywhere. `rm` is a native shell command against `ctx.fs` (`symlink-aware-file-commands.ts`), so `rm f` surfaced the ENOENT as exit 1 and `rm -f` / `rm -rf` swallowed it per POSIX and exited 0. The `/shared` + `/scoops` split (removed, but still exit 1) is the same ENOENT: the index entry is already gone by the time the backend throws.

### Repro

Against the real ZenFS OPFS backend (`VirtualFS` + the faithful FSA mock in `tests/fs/fsa-test-helpers.ts`), which is what the issue's `node -e` probe was reaching:

```
before: exists zero: true
        zero rm err: ENOENT: no such file or directory, unlink '/workspace/zbtest/redir.txt'
        one  rm err: undefined              <- 1-byte control deletes fine

after:  exists zero: true
        zero rm err: undefined
        one  rm err: undefined
```

Shell-level, through `AlmostBashShell` over an OPFS-backed VFS:

| Command on a 0-byte file | Before | After |
| --- | --- | --- |
| `rm f` | exit 1, `cannot remove 'f'`, survives | exit 0, gone |
| `rm -f f` | exit 0, no stderr, **survives** | exit 0, gone |
| `rm -rf dir/` (contains one) | exit 0, dir + file survive | exit 0, dir gone |

Both halves land in `patches/@zenfs+dom+*.patch`, the repo's established mechanism for this dependency (it already carries the `keepExistingData` and #2146 ino fixes). Nothing special-cases `rm`.

1. **`WebAccessFS.createFile` materialises the OPFS handle.** Every file the index knows about also exists on the backing store, whether or not bytes ever arrive. Net-neutral on cost: the first `write` currently pays a *failing* `get` followed by a create; now `get` hits the warmed `_handles` cache.
2. **`WebAccessFS.remove` tolerates an already-absent OPFS entry.** `IndexFS` has already validated existence against the authoritative index (and dropped the entry) before the backend sees the path, so an absent OPFS entry means the requested end state is already reached. A genuinely unknown path still raises ENOENT from `IndexFS` — asserted.

Part 2 is what makes the fix retroactive: **0-byte files already on disk in the field** predate part 1 and would otherwise stay undeletable forever.

Alternative rejected: forcing the empty write from `VirtualFS.writeFile` (e.g. write one byte, then truncate). `IndexFS.touch` only narrows the in-memory inode and never truncates the handle, so that leaves a real stray byte on disk while the inode claims 0 — trading one index/store disagreement for another.

### Phantom deletions in `git` clones

The root cause matches the reporter's suspicion. Any path the ZenFS index knows but OPFS does not was permanently undeletable: every `unlink` threw, and the failed removal kept the vnode cached, so the file came back. That covers both shapes described — empty files a checkout cannot remove, and files `git` reports as deleted that will not go away (a sidecar entry whose OPFS counterpart is gone). Part 2 closes that class. Worth re-testing those clones against this build; if they clear, both issues can be closed together.

### Deliberate scope decisions

- **`rm -f` keeps suppressing diagnostics.** That is POSIX, and it was not the defect — the spurious error underneath it was. With the root cause gone, `rm -f` deletes and exits 0.
- **`flushSyncFsCache`'s exit code stays out of scope, stated rather than left silent.** None of the reported failures went through it: `rm` never enters a realm. Reflecting a post-run durability failure in the exit code remains a real behaviour decision (the existing comment's "weighed but deferred"), and folding an exit-semantics change into a data-loss fix would make both harder to review. The loud `[sync-fs] ERROR` breadcrumb still fires on a genuine flush failure, and the new realm tests assert it does **not** fire for a 0-byte create+delete.
- **`mv` is not asserted, and this is why.** The issue lists `mv f g` failing on a 0-byte file. `rename` over this backend fails for a **1-byte** file too: ZenFS' `Async` mixin mirrors each op onto an InMemory sync cache that is only populated once `ready()` has run, and `renameSync` against the empty cache raises `no such file or directory (Out of sync!)`. Size-independent, therefore a separate defect rather than part of the zero-byte asymmetry — reported here instead of silently folded in, and worth its own issue.

## Cross-cutting impact

- **Floats**: the patch is in a shared dependency, so every float on the OPFS backend benefits (CLI, extension hosted leader, Electron, cloud). Node/Vitest falls back to ZenFS `InMemory`, which was never affected.
- **Other callers of the changed contract**: `WebAccessFS.remove` is reached from `IndexFS.unlink` / `rmdir` / `rename`. All three gate on the index first, so swallowing an OPFS `ENOENT` cannot mask a real miss — a test asserts a genuinely unknown path still throws. Non-`ENOENT` errors (`ENOTEMPTY`, `EACCES`) propagate unchanged. `createFile` gains one FSA round-trip per created file, offset by the `get` the first write no longer fails on.
- **Persisted state**: no schema or migration change. The `/.metadata.json` sidecar shape is untouched; the fix makes the store agree with what the sidecar already claims.
- **Security / bundle size**: no change on either axis.

## Test plan

- [x] `npm run verify` — 7/7 (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod)
- [x] `npm run typecheck`
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — 5 changed files, 0 on any debt list
- [x] `npm run test` — 14,772 passing, no new failures
- [x] `npm run test:coverage` / `test:coverage:webapp`
- [x] `npm run bundle-size`
- [x] **New behaviour is covered by unit tests**:
  - `packages/webapp/tests/fs/zero-byte-file-delete.test.ts` (new) — asserts against the **real backing store**, not the VFS view, since an index entry with nothing behind it is the entire bug. Covers materialisation, `rm`, recursive `rm` of a mixed directory (the partial-deletion variant), the phantom-entry case, the still-`ENOENT` guard, and shell `rm` / `rm -f` / `rm -rf`. Every case carries the **1-byte control**, so the suite would have caught the size-0 asymmetry.
  - `packages/webapp/tests/kernel/realm/sync-fs-bridge.test.ts` — the two cases the issue asked for: 0-byte create+delete inside one realm run, and create in run 1 / delete in run 2 (the realm boundary it called out). Both assert no flush error is emitted; both carry the 1-byte control.
  - `cjs-realm-harness.ts` — `makeTreeFs.writeFile` now registers ancestor directories. Without it, a file written after construction left `/workspace` unknown to `stat`, so run 2's snapshot walk bailed at the root and came up empty; the harness could not model the two-run scenario at all.
- [ ] UI change — N/A, no UI surface touched.

**Pre-existing failure** (environmental in this sandbox, unrelated to the diff): `packages/webapp/tests/shell/supplemental-commands/playwright-command.test.ts > iframe integration` — `Chrome exited with code null before reporting CDP port` (no launchable Chrome here).

## Documentation

- [x] `docs/` — `docs/pitfalls.md`, in the existing "OPFS Writes" section: a file the index knows is not necessarily a file OPFS has, plus the two rules that follow (materialise on create; treat the index as authoritative on removal).
- [x] `patches/patches.json` — `reason` / `removeWhen` / `verify` updated for the new behaviour, per `patches/README.md`.

## Risk & rollback

- **Blast radius**: every float on the OPFS backend. The change is additive at the backend boundary — one extra handle creation, and one errno newly treated as success on a path the index has already cleared.
- **Rollback**: revert the PR. `patch-package` reapplies the previous patch on `postinstall`; no persisted-state migration to undo. Files created while this was live are ordinary empty files on disk and stay valid either way.
- **Not covered by CI**: real Chromium OPFS behaviour. The FSA mock is faithful to the shapes `WebAccessFS` uses, but a reviewer should confirm the issue's verbatim reproduction against a real build, and re-check the `/workspace` `git`-clone phantom deletions.

## Checklist

- [x] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs (no surface changed; behaviour now matches the documented contract)
- [x] If this changes a contract used by other floats, I checked the followers/leader/offscreen paths — the patched backend is shared, and Node/Vitest uses `InMemory` and was never affected


trieloff:
Addressed the P2 in `3034dcf65`.

You were right, and it is the same divergence reached from the other side: `super.createFile` inserts the inode before the backing handle exists, so a `getFileHandle(..., { create: true })` rejection left the index claiming a file OPFS never got — `stat` reporting a phantom, the next create hitting `EEXIST`, and `writeFileInner` having already marked the path dirty so a flush could persist it into the sidecar.

Fixed by deleting the inode (and any cached handle) on the error path before rethrowing, rather than reordering: the handle creation needs the inode's parent-directory resolution to have happened, so create-then-roll-back is the ordering that stays correct.

The regression test drives a `QuotaExceededError` through the FSA mock's **prototype** — ZenFS caches its own handle instances, so a wrapper returned from `getDirectoryHandle` would never be the object it calls. It asserts no entry survives, the backing store is untouched, and the path is still writable afterwards (no `EEXIST` tombstone). Confirmed it fails without the rollback and passes with it.

`verify` 7/7 · typecheck · debt gate (6 files, 0 debt) · `npm run test` 14,773 passing.

PR #2193, opened 2026-08-19
feat(upskill): add `update` with recorded provenance, never touch skill dotfiles

### LINKED ISSUE #2186: upskill: add 'update|upgrade [--dry-run] [<skill>]' with recorded install provenance, and never touch dotfiles in a skill directory
## Summary

`upskill` can install a skill but cannot **update** one. The only way to refresh an installed skill is to re-run the original install with `--force`, which means the caller must remember where the skill came from, and which deletes everything in the skill directory that upstream does not ship — including the dotfile where the skill keeps its credentials.

Requested:

```bash
upskill update [<skill>] [--dry-run]     # or `upgrade`
```

- with a skill name, update that one; with none, update every installed skill
- `--dry-run` reports what would change without writing
- provenance is recorded at install time (e.g. a `.upskill` file in the skill directory) so the update needs no arguments
- **new rule: `upskill` never creates, modifies or deletes dotfiles in a skill directory**

That last rule is not a separate nicety. A `.upskill` provenance file is destroyed by the very command that would consume it (measured below), so update-without-arguments cannot work until dotfiles are preserved.

## Current behaviour

`upskill --help` lists `search`, `list`, `tabs`, `info`, `read`, and the install forms. There is no `update`/`upgrade`, and no `--dry-run` anywhere. The only refresh path is `--force`, documented as "Overwrite existing skills".

Without `--force`, an existing skill is refused outright:

```
upskill: skill "mixtape" already exists (use --force to overwrite)
```

So there is no middle ground: either refuse, or replace the directory wholesale.

`upskill info <name>` already reports a `Source:` field, but it is the **root kind**, not the origin:

```
Skill: mixtape
Source: native
Source root: /workspace/skills
```

Nothing records the repo, branch or commit a skill was installed from. To refresh, a caller has to already know it was `ai-ecoverse/skills` on `main` — and if they misremember the branch, `--force` silently installs a different skill's worth of content over the top.

## Reproduction — `--force` deletes dotfiles and local files

Measured on SLICC 6.57.4, on a credential-free skill so nothing real was lost:

```bash
printf 'PROBE_SECRET=keep-me\n' > /workspace/skills/mixtape/.config
printf 'source: ai-ecoverse/skills\n' > /workspace/skills/mixtape/.upskill
printf 'user notes\n'              > /workspace/skills/mixtape/NOTES-local.md

upskill ai-ecoverse/skills --skill mixtape --force
```

### Actual

```
.config          *** DELETED ***
.upskill         *** DELETED ***
NOTES-local.md   *** DELETED ***
```

Only the three files upstream ships survive. The install replaces the directory contents rather than reconciling them.

### Expected

`.config` and `.upskill` survive. (Whether a non-dot local file like `NOTES-local.md` should survive is a separate judgement call — the dotfile rule alone is enough to fix the credential and provenance problems.)

## Why this bites: it destroys live credentials

Skill config is stored **as a dotfile inside the skill directory**. Three installed skills currently keep credentials there:

```
skills/bb/scripts/.config
skills/gmail/scripts/.config
skills/fastly/scripts/.config
```

So `upskill <repo> --skill gmail --force` silently logs the user out of Gmail. This is not hypothetical — it happened here today. After refreshing the `bb` skill, the next command failed with:

```
bb: no bb server configured — run 'bb pair --code <code> --server https://<handle>.getbb.app'
```

The credential was gone, and re-pairing that particular skill normally requires a shell on the machine that owns the bb server, so recovery can be genuinely expensive. Nothing warned that a refresh would revoke it.

## Why `--dry-run` matters

Twice today I refreshed a skill and had no way to know beforehand what would change, or afterwards whether it worked. Both times I ended up hand-rolling verification: fetch `GET /repos/<owner>/<repo>/git/trees/main?recursive=1`, recompute `sha1("blob " + len + "\0" + content)` for every local file, and compare. That is a lot of ceremony for "is my skill current?", and it is the only way to catch the interesting cases:

- One refresh was a **no-op** (already current).
- The other took a skill from 5 files to 12 and `gh.jsh` from 3409 to 3730 lines, restoring three script modules and four test files that a bad PR had reverted months earlier. The local copy had been silently stale that whole time, missing whole features.

Neither outcome was predictable from the command, and `upskill` reported the identical `Installed 1 skill(s)` for both. A `--dry-run` that listed added/changed/removed paths — and an update that reported "already current" — would have answered in one step.

## Proposed shape

```bash
upskill update                    # every installed skill, from recorded provenance
upskill update <skill>            # just that one
upskill update --dry-run          # report only, write nothing
upskill update <skill> --branch x # override the recorded ref
```

- **Provenance**: write `.upskill` in the skill directory at install time — source kind (github / tessl / browse.sh), `owner/repo`, ref, resolved commit sha, install timestamp. The resolved sha is what makes "already current" cheap and exact.
- **Classify each path** like the `upgrade` command already does for bundled workspace files: `unchanged`, `updated`, `added`, `removed`, `kept-local`. Reusing that vocabulary would make the two commands feel like one system.
- **Dotfiles are never touched** — not created, not modified, not deleted. This preserves `.config` (credentials), `.upskill` (provenance), and any `.gitignore` a skill ships. Note some skills legitimately ship dotfiles upstream (`skills/xlsx/scripts/.gitignore`, `skills/tiktok/.gitignore`, `skills/elevenlabs/scripts/.gitignore`, `skills/oryx/.config.example`), so "never touch" needs to mean "never touch after first install" rather than "never write" — otherwise those files could never be installed at all. Worth deciding explicitly.
- Consider `upskill outdated` (or `update --dry-run` with no skill) to list every installed skill whose recorded sha is behind its source.

## Impact on agents

Refreshing a skill is currently a destructive operation dressed as a routine one. An agent that runs `--force` to pick up a fix silently revokes the user's credentials for that skill, and cannot tell the user what changed. The absence of provenance also means an agent must carry the install source in its own memory — I have exactly that note in mine, and it went stale and became actively wrong, which is how a skill sat silently 321 lines behind upstream. Making `upskill update` self-describing removes a whole class of quiet drift.

Filed at the request of the user (Lars), who additionally proposed the `.upskill` file and the dotfile rule; the measurements above are mine.

trieloff:
Closes #2186.

## Problem

`upskill` could install a skill but not update one. The only refresh path was re-running the original install with `--force`, which meant the caller had to remember where the skill came from — and which deleted everything upstream does not ship, including the dotfile where the skill keeps its credentials. Per the issue, refreshing the `bb` skill silently revoked its pairing; `gmail` and `fastly` store credentials the same way. There was also no way to know beforehand what a refresh would change, or afterwards whether it did anything.

## What changed

**Dotfiles are never touched.** `upskill` no longer creates, modifies, or deletes a dotfile that already exists in a skill directory — on install, `--force`, or `update`. A reinstall clears the skill's non-dot files and keeps the directory skeleton that reaches nested dotfiles (`scripts/.config`). Upstream dotfiles (`.gitignore`, `.config.example`) still land on **first** install, since "never write" would mean they could never be installed at all.

**Provenance (`<skill>/.upskill`).** Every install records source kind, `owner/repo` (or browse.sh slug), ref, resolved commit sha, timestamp, and the file list it wrote. `upskill info` now reports `Installed from:` instead of only the discovery root kind. The sha lookup is gated on a configured `github.token`: the install path deliberately prefers codeload over the API, so an anonymous install must not spend its one rate-limited request on bookkeeping (a `tessl` test pins that invariant, and a new test pins it for provenance).

**`upskill update` / `upgrade`.**

```
upskill update                    # every skill with provenance
upskill update <skill>            # just that one
upskill update --dry-run          # report only, write nothing
upskill update <skill> --branch x # override the recorded ref
upskill update --json             # { ok, dryRun, results[] }
```

Paths are classified with the vocabulary `upgrade apply` already uses for bundled workspace files — `unchanged`, `updated`, `added`, `removed`, `kept-local` — so the two commands read as one system. Only files the recorded file list attributes to a previous install are removable, so user-added files (the `NOTES-local.md` case from the issue) are reported `kept-local` and survive. Skills installed before provenance tracking report how to recover.

Sample dry run:

```
Skill update (dry run — nothing written):
  alpha: would update from github:octo/skills
      updated    SKILL.md
      added      scripts/new.sh
      removed    scripts/gone.sh
      (1 kept-local — dotfiles and local additions are never touched)
```

## Docs

- `docs/shell-reference.md` — new `upskill update` section with the classification table, plus a "Dotfiles are never touched" section.
- `packages/vfs-root/workspace/skills/upgrade/SKILL.md` — the runtime upgrade skill now recommends `upskill update --dry-run` as a separate pre-decision step alongside the changelog review. The upgrade card only covers **bundled** files, so installed skills drift silently; a release is the natural moment to check them. The dry run is the safe default when the agent volunteers the check.
- `packages/vfs-root/workspace/skills/skill-authoring/SKILL.md` — notes that a skill should keep credentials in a dotfile, since that is what survives updates.

## Verification

`lint` · `typecheck` (all 9 projects) · `test` (14,802 passed) · `test:coverage:webapp` (80.06 / 71.03 / 80.33 / 82.14, all above floors) · both builds · `check-touched-exemptions` OK.

The one local failure is `playwright-command.test.ts > iframe integration`, which cannot launch Chrome in this worktree (`Chrome exited with code null before reporting CDP port`) — environmental, unrelated to this change; the coverage numbers above are from a run with that file excluded.

Boy-scout gate required clearing pre-existing debt on touched files: `upskill-command.ts` no longer imports `cdp/` (the `tabs` browser type is derived from `handleTabs`), `install-pipeline.ts` names its two hook globals instead of `Record<string, unknown>`, and `browse-sh.ts` compares its in-flight promise cache with `!== undefined`. Baselines ratcheted down accordingly.

15 new tests cover dotfile survival at depth, the never-overwrite rule, provenance round-trip, the anonymous no-API-call invariant, dry-run classification, apply semantics, `--json`, `--branch` override + injection rejection, the missing-provenance path, and upstream-removal failure.


trieloff:
## Review from a parallel implementation of #2186

I implemented the same issue independently on `bb/implement-2186-upskill-update-upgrade-with-prove-thr_u7zj9zpvzw`, so I read this one closely and ran probes against `8552222d8` to check my code-reading. Everything below is reproduced empirically, not inferred. A lot of this PR is better than mine — the `files` list in the provenance record, `--json`, the `prepareBrowseShSkill` extraction that makes browse.sh skills actually updatable, and the runtime-skill doc updates are all things I either punted on or got wrong. But there is one blocker and two correctness bugs.

---

### 1. Blocking: the update write path has no path-traversal guard

`upstreamFiles()` in `update.ts` filters ZIP entries on prefix and trailing slash only:

```ts
for (const [path, content] of Object.entries(files)) {
  if (!path.startsWith(prefix) || path.endsWith('/')) continue;
  const relative = path.slice(prefix.length);
  if (relative) upstream.set(relative, content);
}
```

`planUpdate` then composes `${destDir}/${relative}` and `applyPlan` does `mkdir -p` + `writeFile` on it. Nothing rejects `..` segments, so an archive entry that climbs out of the skill directory is written outside it.

Probe against `8552222d8` — a skill with recorded provenance, `upskill update mixtape`, where the upstream ZIP contains `skills/mixtape/../../../etc/pwned.txt`:

```
Skill update:
  mixtape: updated from github:acme/skills
      added      ../../../etc/pwned.txt
      updated    SKILL.md

/etc/pwned.txt exists = true
content = "ESCAPED\n"
```

It writes outside the skill directory *and* prints the traversal in the classification report as a normal `added` line.

This is not a pre-existing hole being inherited — `installSkillFromZip` in this same PR still carries the guard it has always had:

```ts
if (normalizedPath.includes('/../') || normalizedPath.includes('/..') ||
    !normalizedPath.startsWith(destDir + '/')) continue; // skip malicious entry
```

The new extraction path in `update.ts` just doesn't reuse it. Simplest fix is to drop entries with a `..` segment (or an absolute `relative`) in `upstreamFiles`, which also keeps them out of the `files` list that later decides what `update` may delete.

Worth noting while you're in there: `writeZipFilesToDir` (the ZIP fast path in `installFromGitHub`) has never had the guard either. Pre-existing, but this PR now also feeds its return value into provenance, so it's the natural moment to fix it.

### 2. The resolved commit sha is recorded but never used to avoid work

The issue's requirement 1 is explicit about why the sha matters: *"The resolved sha is what makes 'already current' cheap and exact — prefer comparing shas over refetching and diffing every file."*

In `updateGitHubSkill` the order is `fetchRepoZip` → `planUpdate` → `resolveCommitSha`. The ZIP download happens unconditionally at the top, so the recorded sha can never short-circuit it; it only ever populates the `from`/`to` display fields. On top of that, `resolveCommitSha` early-returns unless `github.hasToken`, so for a token-less user the sha is never recorded at all.

Probe on `8552222d8` — content already identical to upstream, recorded sha equal to HEAD, no token configured:

```
Skill update:
  mixtape: already current (github:acme/skills)

requests = ["https://codeload.github.com/acme/skills/zip/refs/heads/main"]
```

One full repository archive downloaded, no commits call, and the "already current" line carries no commit. The same probe on my branch:

```
upskill update:
  mixtape: already current (aaaaaaa)

requests = ["https://api.github.com/repos/acme/skills/commits/main"]
```

I agree with the reasoning behind the gate for *installs* — keeping the codeload path API-free is a real invariant and `tessl.test.ts` pins it. But an explicit, user-initiated `upskill update` is exactly where spending one API call is correct: a ~200-byte commits response against a whole repo archive, and it is what makes "already current" *exact* rather than "the bytes I happened to compare matched". Suggest keeping the `hasToken` gate on the install path, dropping it for `update`, and comparing `headSha === provenance.sha` before touching codeload. As written, the sha field is decorative for the majority of users, and `upskill outdated` (issue item 6) can't be built cheaply on top of it.

### 3. `writeProvenance` spreads the caller's record last, defeating its own version and timestamp stamping

```ts
const full: UpskillProvenance = {
  version: PROVENANCE_VERSION,
  installed: record.installed ?? new Date().toISOString(),
  ...record,   // ← last
};
```

Every update caller passes `{ ...provenance, ref, sha, files }`, i.e. an object still carrying the on-disk `version` and `installed`, and the trailing spread wins. Probe: plant `version: 0, installed: "1999-01-01T00:00:00.000Z"`, run `upskill update mixtape`, read the record back:

```json
{"version":0,"installed":"1999-01-01T00:00:00.000Z","kind":"github","source":"acme/skills", ...}
```

Two consequences. `installed` never advances, so `upskill info` can't tell you when a skill was last refreshed and the `?? new Date().toISOString()` fallback is dead code on the update path. And `PROVENANCE_VERSION` can never be applied to an existing record — which is the one thing a version field is for; a future v2 writer would silently keep emitting `version: 1`. Moving `...record` to the front fixes both. (`version` reaches the function despite `Omit<…, 'version'>` because excess-property checking doesn't apply to spread-derived properties, so the type doesn't catch it.)

### 4. `--force` and `update` disagree about user-added files

The PR body says *"user-added files (the `NOTES-local.md` case from the issue) are reported `kept-local` and survive"*. That holds for `update`, but `--force` still routes through `clearSkillDirPreservingDotfiles`, which removes every non-dot entry regardless of the `files` list. Probe on `8552222d8`:

```
upskill acme/skills --skill mixtape --force
  .config kept        = true    ← the credential fix works
  NOTES-local.md kept = false
```

So "refresh this skill" gives two different answers depending on which command the user reaches for, and the documented guarantee only covers one of them. Either extend the `files`-list logic to the `--force` path or scope the claim in `docs/shell-reference.md`. (I delete it on both paths and say so, which is a worse policy than yours but at least consistent — the point is that the two paths should agree, not which one wins.)

### 5. Requirement 7's second half is unmet — no way to record provenance for an existing skill

The recovery message is:

```
upskill: no install provenance for "bb" — reinstall it once (upskill <source> --skill bb --force) to record where it came from
```

That requires the user to already know `<source>`, which is the exact knowledge the issue says goes stale: *"I have exactly that note in mine, and it went stale and became actively wrong, which is how a skill sat silently 321 lines behind upstream."* The issue asks for "a clear message rather than a crash, **and ideally record provenance when the user supplies it**" — the message is there, the recording isn't. A `--from <owner>/<repo>` flag on `update` is a few lines and turns a blind full reinstall into a recorded, classified update with a dry-run available first.

### 6. Smaller things

- `upskill update <typo>` reports `no install provenance for "<typo>"` even when no such skill is installed, which points the user at the wrong problem. Worth distinguishing "not installed" from "installed but unrecorded".
- `writeZipFilesToDir` is now shared with `plugin-command.ts`, so agent-plugin installs silently inherit "never overwrite an existing dotfile" via `canWriteSkillFile`. Probably desirable, but it's an unannounced behavior change in a different subsystem and the helper name no longer describes all its callers.
- browse.sh updates always rewrite `SKILL.md` when the upstream `updated` date moves, because the date is spliced into the preamble — so those skills will report `updated` on a date bump with an identical body. Cosmetic, but it makes `--dry-run` noisier than it needs to be for that registry.

---

### What I'd take from this PR over mine

For whoever lands one of these: the `files` list is the better answer to the issue's explicitly-open question about `NOTES-local.md`, and it's what gives `kept-local` real meaning instead of being a dotfile-only bucket. `--json` matches `upgrade apply`'s machine-readable contract better than my text-only report. `prepareBrowseShSkill` is the right refactor — I punt on browse.sh with a "use `--force`" message, which is weaker. And updating `vfs-root/workspace/skills/upgrade/SKILL.md` so the runtime upgrade flow suggests `upskill update --dry-run` is a connection I missed entirely.

My branch has the traversal guard, the sha short-circuit, `--from`, and `upskill outdated`, if any of that is useful to lift: `bb/implement-2186-upskill-update-upgrade-with-prove-thr_u7zj9zpvzw` (`958cd046e`).

Happy to share the four probe tests as a patch if that's easier than re-deriving them.

trieloff:
Addressed all three review findings in b7e5cf7, each with a regression test.

**P1 — dot directories on the Contents-API install path.** Confirmed and fixed. `downloadGitHubDir` passed only `item.name` to the dot check, so once recursion descended into `scripts/.config/`, a nested credential (`.config/token`) looked like a plain `token` and `--force` overwrote it. It now tests the accumulated skill-relative path. The ZIP path was never affected — it sees full relative paths — so this was the fallback path only. The new test drives a codeload-500 install through the Contents API and asserts a user-written `scripts/.config/token` survives `--force`; I verified it fails against the old check.

**P2 — authenticated API fallback on update.** Fixed. `update` bailed out as soon as the unauthenticated codeload request failed, which made private repos and repos whose default branch is neither `main` nor `master` permanently non-updatable even though the install path handles them. New `fetchGitHubDirFiles` reads the directory through the same authenticated request context the install path uses; when neither source works, both failures are reported (`<codeload error>; Contents API fallback failed: <api error>`). Two tests: a successful API-fallback update, and the both-sources-fail error.

**P2 — failed updates reported as current.** Fixed. `applied` alone gated the closing line, so a batch where everything failed still printed "all skills are current" while stderr carried the errors and the exit was 1. That line is now also gated on the batch being clean (no failures, no missing provenance). Test covers the mixed batch: one current skill plus one without provenance still exits 1 and does not claim everything is current.

**Bundle size (CI).** The `bundle-size` job failed on the webapp kernel worker. Measured on this machine: `main` is already at **1850.3 kB** against the 1850 kB budget on its own — the job only runs on PRs that touch webapp, so drift arriving through other paths stays invisible until the next such PR — and this PR's modules add **7.5 kB** (1857.85 kB total). Per the budget's stated policy ("raising one needs a reason in the PR body"), I raised the kernel-worker ceiling to **1865 kB**, which restores ~7 kB of headroom rather than pinning it to the current size. The other five budgets (webapp main entry / total payload, extension service worker / side panel / total) are untouched and comfortably under.

trieloff:
Thanks — this is a genuinely useful review, and the probes made it fast to confirm. All six items are fixed in b576c6e; each was reproduced first and is pinned by a test.

**1. Path traversal on the update write path — confirmed and fixed.** Your read was exact: `upstreamFiles` filtered on prefix and trailing slash only. I reproduced it as a test (upstream archive containing `alpha/../../../etc/pwned.txt` → `/etc/pwned.txt` written, traversal printed as a normal `added` line), and verified it fails against the old code before keeping it.

Rather than patch the one site, I pulled the guard into `skill-paths.ts` as `isSafeSkillRelativePath` and routed every write path through it: it replaces the inline check in `installSkillFromZip` (so there's one implementation, not two that can drift), and covers `writeZipFilesToDir` and both Contents-API readers, which — as you noted — never had it. It rejects `..`/`.` segments, absolute paths, empty segments, backslash separators, drive prefixes, and NUL. Rejected entries are also excluded from the recorded `files` list, so a later `update` can't be talked into deleting something outside the skill directory either.

**2. The sha never short-circuiting — fixed, and you're right about where the trade sits.** `update` now resolves the ref's head *first*; if it equals the recorded sha and every recorded file is still on disk, it reports `already current` and never touches codeload. `resolveCommitSha` grew an `allowAnonymous` parameter that only `update` passes, so the install-side invariant your review acknowledged (codeload stays API-free, pinned by `tessl.test.ts`) is untouched while an explicit update spends one ~200-byte request instead of a repo archive.

The one refinement: I gate the short-circuit on the recorded files still being present. A matching sha proves *upstream* hasn't moved, not that the local copy is intact — a deleted file falls through to the full compare. Three tests: no codeload request on the sha path, the commits call happening without a token, and a locally-deleted file defeating the short-circuit.

**3. `writeProvenance` spread order — confirmed and fixed.** Exactly as described, including why the type didn't catch it. `...record` now goes first and both `version` and `installed` are stamped unconditionally, so the timestamp advances on every update and a future v2 writer can actually migrate a v1 record. Test plants `version: 0, installed: "1999-01-01"` and asserts both move. I also documented `installed` as the last install/update rather than the first install, since that's what it now means.

**4. `--force` vs `update` disagreeing — fixed by making `--force` follow `update`.** `clearSkillDirPreservingDotfiles` now takes the recorded file list and deletes only paths a previous install actually wrote, so `NOTES-local.md` survives both commands. A skill with no record keeps the old clear-everything behavior, because nothing is attributable to an install — which is another reason to want item 5. Test asserts `NOTES-local.md` survives `--force` while `SKILL.md` is still replaced.

**5. `--from` — added.** `upskill update <skill> --from <owner>/<repo> [--path <dir>]` records a source for a skill installed by hand or before provenance tracking. Since a bare `--from` has no path, it locates `<skill>/SKILL.md` in the archive (shallowest match wins, so a top-level copy beats a vendored one) and records the resolved path, so the next update needs no arguments. That first update can add and overwrite but never delete, because there's no file list to attribute yet — tested.

**6. Smaller items — all three.** `update` now distinguishes "no skill named X is installed — check `upskill list`" from "installed but unrecorded", and the latter points at `--from` instead of a blind reinstall. `writeZipFilesToDir` takes an explicit `keepExistingDotfiles`, and `plugin-command.ts` passes `false` with a comment — it wipes the destination first, so nothing changes in practice, but the plugin subsystem no longer silently inherits a skill-specific rule and the helper's contract is visible at both call sites. browse.sh comparison now strips the SLICC preamble before comparing, so a date bump with an identical body reports `unchanged`; test installs at one `updated` date and updates at another.

**Bundle size:** kernel worker is 1861.10 kB after this round; I set the budget to 1870 kB for headroom (it was 1850, and `main` alone was already at 1850.3).

Verification: lint, typecheck (9 projects), 14,951 tests, webapp coverage 80.16 / 71.17 / 80.51 / 82.24 (above floors), both builds, `bundle-size`, and the touched-file debt gate.

Two things I did not take from your branch: `upskill outdated` (item 6 of the issue is explicitly a "consider", and `update --dry-run` with no skill name already answers the same question — happy to add the alias if you'd rather have the verb), and your delete-on-both-paths policy, since the `files` list makes the keep-user-files rule enforceable on both commands instead. If you'd still like to send the four probe tests as a patch I'll fold them in — mine cover the same ground but yours are the ones that found this.

PR #2179, opened 2026-08-18
refactor: pay down boy-scout debt in packages/cloud-core/src/errors.ts

trieloff:
## Boy-scout debt cleanup — `packages/cloud-core/src/errors.ts`

Pays down the boy-scout debt on this one file so it is clean from every debt rule, in a single focused PR.

### Debt list cleared

**`record-string-unknown` (untyped string-keyed bags)**
- Replaced the untyped `Record<string, unknown>` on `CloudError.details` with a named `CloudErrorDetails` interface describing the shape cloud-core actually attaches:
  - `running?`, `paused?`, `cap?` — set at the two `CAP_EXCEEDED` throw sites in `operations/start.ts`.
  - `sandboxId?` — set at the `SANDBOX_NOT_READY` throw site in `operations/start.ts`.
- `CloudErrorDetails` is a structural subset of the worker's existing local `CloudErrorDetails` (`cloudflare-worker/src/cloud/error-envelope.ts`), so `err.details` remains assignable everywhere the worker forwards it via `errorResponse`.
- Ratcheted the baseline with the supported command `node packages/dev-tools/tools/check-record-string-unknown.mjs --update`; the only change is the removal of the `packages/cloud-core/src/errors.ts` entry (`-1`). No other baseline entry was touched.

### Behaviour

Behaviour-preserving: the new type is a compile-time refinement only. `CloudError`'s runtime shape (code, message, `details` payload) is unchanged, and the `details` objects constructed at every throw site already match the named fields.

### Tests

Added `packages/cloud-core/tests/errors.test.ts` (mirrors `src/errors.ts`) pinning:
- `CloudError` code/message/name and default-`undefined` details,
- the typed running-cap and paused-cap `details` shapes,
- `isCloudError` narrowing and its rejection of plain errors and non-error values.

### Verification

- `npx biome check --write` on the touched files — clean.
- `npm run typecheck` — passes.
- `npx vitest run packages/cloud-core/tests/errors.test.ts packages/cloud-core/tests/start.test.ts` — 16 passed.
- `npm run test:coverage:cloud-core` — passes, coverage floor held.
- `GITHUB_EVENT_NAME=pull_request GITHUB_BASE_REF=main node packages/dev-tools/tools/check-touched-exemptions.mjs origin/main` — `OK (2 changed file(s), 0 still on any debt list)`.
- `npm run verify` — all 7 lint-gate checks passed.

### Confirmation

[...]

PR #2198, opened 2026-08-19
feat(ci): answer reviewer feedback on our agents' PRs automatically

trieloff:
## Summary

When a reviewer leaves findings on an `automation/*` PR, nothing picks them up. The only way to get them addressed today is to open a local Claude session and say "take over PR X and address the review comments". This adds the workflow that does it: it reads the feedback, fixes what is worth fixing, replies in every thread — including to say "this finding is wrong, and here is why" — and pushes to the PR's existing head branch.

## Why not persist and resume the Claude session

That was the original idea (encrypt the transcript into a workflow artifact, resume it in a second workflow). It is technically possible — the CLI takes `--session-id` and `--resume` — and it would be worse:

- **This repo is public**, so workflow artifacts are world-downloadable. Encryption protects the transcript only as long as key handling is perfect, and the failure is a permanent, silent public leak of everything the model read.
- The transcript format is internal and unversioned; a CLI bump inside the action can break resume months later.
- **A resumed session defends the plan it already committed to.** #2169 and #2170 are the proof: the correct response to Codex was to *abandon* the original approach, and the session that produced it was the one least likely to do that. A fresh session reading the shipped diff and the reviewer's argument is less anchored, not more.

GitHub is already the durable state: the diff is on the branch, the argument is in the threads, the verdict is the check suite.

## Shape

The house two-part pattern. A deterministic Node selector (`packages/dev-tools/review-responder/`) answers "is there unseen feedback from someone other than us, on a PR in scope?" in pure unit-tested functions; one Claude step does the part that needs judgement.

The gate deliberately does **not** classify a comment as actionable-vs-noise. Whether a reviewer is *right* is judgement, so the prompt carries the real precedent from this repo: on #2170 the house reviewer LGTM'd a change while Codex correctly identified that it defeated the spelling-based debt gate. Codex was right. The model is told to reason about the code and say which reviewer it agrees with, not to count votes — and that changing code to satisfy a finding it believes is wrong is the worst available outcome.

## Loop safety

This workflow posts comments, and comments are its trigger, so this is the part worth reviewing hardest.

**The trigger gate tests for our marker, not our author.** Excluding `github-actions[bot]` as a triggering author is the obvious way to write it and it is wrong here: `claude-pr-review.yml` posts the house review under that same login, and on an `automation/*` PR that is often the **only** review there is — Codex ignores bot-authored PRs, and Copilot degrades to "unable to review this pull request" once its quota is spent (observed on #2179 today). An author-based gate would leave this workflow unreachable in exactly the case it exists for, and it would look healthy while doing it: no failed runs, just silence. `<!-- review-response` is a prefix of both markers we write, so one clause covers the summary comment and the crash notice.

**Claude's inline replies are caught structurally instead**, by `isSelfOutput` (our login + an inline comment that is a reply). Requiring the model to mark its own replies would put loop safety at the mercy of the model remembering. Those replies do start a run, which the selector immediately skips — one idle run per burst, bounded by the concurrency group, and a bounded waste is the right trade for a guard that cannot be forgotten.

**Admitting our own login means the other dispatchers' bookkeeping comments reach the selector**, so `dropReason` filters `<!-- pr-fix-skip:` / `<!-- backlog-skip:` by marker, plus two unmarked machine notices by a short body-pattern list. `partitionFeedback` returns **everything it dropped with the reason why**, and the CLI prints it — because a filter that removes everything is indistinguishable from "no feedback arrived", which is precisely how a `DENYLIST_LABELS` entry silently disabled the backlog dispatcher last week.

**A three-reviewer burst is one run.** `concurrency` keyed per PR with `cancel-in-progress: true`, plus a leading 5-minute debounce: the earlier runs die mid-sleep having done nothing, and the survivor reads a feedback set that is by then complete. Two responders must never race pushes on one branch.

**The push uses `BOT_PAT`, not `GITHUB_TOKEN`**, or GitHub's anti-recursion guard suppresses `synchronize` and the PR sits on stale green checks that never saw the fix — the `checks: 0` bug from #2168/#2169.

**Merging, opening a PR and closing anything are removed from the tool surface** via `--disallowedTools`, not discouraged in prose. A model that ignores a "do not merge" instruction regresses silently in a way that looks exactly like success.

## All three feedback endpoints

`/pulls/{n}/reviews` alone sees **nothing** from the house reviewer, which posts a top-level issue comment. I made that exact mistake twice this week while shepherding #2167–#2170 — reported "the reviewer found nothing" when it had posted substantive findings both times. The scanner reads reviews, inline comments and issue comments, and there is a regression test for the case where the house reviewer's top-level comment is the only feedback.

## Cross-cutting impact

- **Floats exercised**: none. CI-only.
- **Blast radius**: one run per reviewer burst on `automation/*` PRs (v1 does not touch human PRs). Job holds `contents: read`; the push is a separate explicit credential.
- **Security**: fork heads refused in both the `if:` and the selector. `allowed_bots` names one bot, never `'*'` (this repo is public and the action's docs warn that `'*'` lets external Apps invoke it with prompts they control — acute here, since the prompt contains reviewer-authored text and the step can push).

## Test plan

- [x] `npm run verify` — 7/7
- [x] `node node_modules/vitest/vitest.mjs run --project dev-tools` — 1001 passing (49 new)
- [x] `npm run lint:docs` — clean
- [x] `npx prettier --check` + `biome check` on every touched file — clean
- [x] `check-touched-exemptions.mjs origin/main` — OK
- [x] Workflow YAML parsed and asserted: marker-based `if:` with no author test, concurrency group present, pinned SHAs, all four `--disallowedTools` entries, `allowed_bots` not `'*'`
- [x] **Live read-only dry run against #2179, #2181, #2184.** On #2181 it drops Copilot's quota notice and the semantic-release notice with reasons and reduces to the single house-reviewer finding — the case an author-based gate would have missed entirely. All three correctly skip on `state != open`.
- [ ] **Not verified: a real run.** The Claude step has never executed — Bedrock auth, the actor gate, `--disallowedTools` parsing and `SUMMARY_FILE` reaching the Bash env are all unproven. `SUMMARY_FILE` follows the `PR_BODY_FILE` pattern now working in production in boy-scout and backlog, which is the strongest evidence available without merging.
- [ ] **Not verified: the debounce/cancel interaction.** I have not observed GitHub cancelling a queued run mid-`sleep`.
- [ ] N/A — `typecheck` / `build`: no TypeScript in this change.

## Documentation

- [x] `packages/dev-tools/review-responder/README.md` — includes the full loop-safety argument, why neither guard can be author-based, and why sessions are not resumed
- [x] `docs/dev-tools-details.md` — new `## review-responder` section
- [x] `packages/dev-tools/CLAUDE.md` — registered (10,161/20,000 chars, gate passes)

## Risk & rollback

- **Rollback**: delete the workflow. Nothing else depends on it.
- **Watch for on the first live run**: (1) exactly one run per burst, not three; (2) the response commit carrying fresh checks; (3) no second response at the same head SHA; (4) the model replying rather than silently agreeing with a wrong finding. I will verify these on the next `automation/*` PR that draws review rather than claiming them now.

## Checklist

- [x] No secrets in the diff


trieloff:
All three confirmed against the YAML and fixed in f2ad790a1. Taking them in order.

**P1, the self-cancelling concurrency group — correct, and the most valuable finding of the three.** I had read the loop-safety argument in this PR and accepted it, because it reasons carefully about the job `if:` and I did not think about ordering. Concurrency is evaluated *before* the `if:`, so the exclusion logic never gets the chance to matter: the first inline reply cancels the responder that wrote it, mid-run, before the SHA marker exists, and the replacement then sees the original feedback still unanswered and starts over. Ping-pong, presenting as cancelled runs rather than as a failure, billed the whole way.

Now `cancel-in-progress: false`. The later runs of a burst queue, wake, find the marker at an unmoved head SHA with nothing newer behind it, and skip in seconds. That puts idempotence in the pure tested selector instead of a YAML expression, which is where this repo prefers it anyway. The debounce comment claimed cancellation as load-bearing; it no longer does, and the README and `docs/dev-tools-details.md` are corrected too, including why the obvious version is a trap.

**P1, the scanner running from the PR's own checkout — correct, and the comment above it asserted the opposite.** There was one checkout with no `ref:`, carrying `token: BOT_PAT`, while the comment claimed "run BEFORE the head branch is checked out … so it must not execute code from the branch it is deciding about. Same trust ordering as claude-pr-review.yml's gate." The intent was right and the implementation contradicted it, which is the worst combination, because the comment tells the next reader not to look.

Split in two: a default-branch checkout (`persist-credentials: false`, no token, shallow) for the scan, then the head with BOT_PAT only after the gate approves it. Asserted by parsing the YAML rather than by eye — trusted checkout → scan → head checkout → `npm ci`, with no token on the first and no `npm ci` before the second.

On the part you flagged that is *not* fully gone: `npm ci` and the model's own verification still execute branch code with BOT_PAT in `.git/config`. I documented that in place rather than implying the fix removed it. It is the same exposure `renovate-patch-reconcile.yml` already accepts, bounded the same way — an `automation/*` head can only exist if something with write access pushed it. What changed is that the decision to admit the branch is no longer made by the branch.

**P2, `allowed_bots` making the workflow unrunnable — correct, and it was a self-inflicted wound from #2196.** I narrowed these lists repo-wide an hour earlier and then wrote a new workflow gated on the narrow value without checking that its own triggering actors could pass. Exactly the failure mode #2177 was about: not a break, a quiet non-functioning.

All three reviewer bots are named now (`github-actions[bot]`, `chatgpt-codex-connector[bot]`, `copilot-pull-request-reviewer[bot]`), never `'*'`. The reasoning I put in the file, since it looks like it contradicts #2196: admitting a third-party App as an *actor* grants it no content control it lacks anyway, because its findings are this workflow's input by design whichever event starts the run. Excluding them, by contrast, breaks the feature silently — you ignore bot-authored PRs, and the house reviewer stands down once inline comments exist, so Copilot's review is sometimes the only event an `automation/*` PR ever gets.

I considered your alternative of dispatching the Claude step from an actor-independent event, which is cleaner in principle. I went with named actors because the dispatch hop's failure mode is a run that silently never fires, and this repo has already been bitten by that class of bug this week (automation PRs landing with zero checks because a push used the wrong token). Happy to revisit if you think the trust argument does not hold.

trieloff:
@codex review

trieloff:
Fixed in 6f434fb6a, and your P2 was right in a way that also explains a failed check on this PR.

**The contradiction you found is real.** The prompt hard-required a commit and push while options (b) and (c) tell the model to leave the code alone when a finding is wrong or out of scope. Every way out of that is bad: fail on an empty `git commit`, push an empty commit and spend a full CI run saying nothing, or quietly disregard a constraint labelled HARD. Committing is now conditional on having changed something, and the prompt states that a reply-only response is a complete and expected outcome rather than a degenerate one — which matters, because the case where a reviewer is wrong is the case this workflow most needs to handle gracefully.

**Separately, your review event made the responder run on this PR, and it failed — usefully.** `MODULE_NOT_FOUND` on `scan-review-feedback.mjs`. That is the trusted checkout from your earlier P1 doing its job: it pins the gate to the default branch, which does not have the scanner until this PR merges, and it must not fall back to the PR's copy or the branch would be judging itself. Correct behaviour, terrible presentation — a red X that reads like a broken workflow, and it would recur on any later PR that moves this package. It now emits a `::warning::`, sets `should_respond=false`, and exits 0.

That run also taught me something I had wrong when I answered your P1, and it narrows the fix's scope: for `pull_request`-family events Actions takes the **workflow file** from the head branch, which is how the branch's copy ran at all. So pinning the ref moves the gate's *logic* out of the branch's reach, not the *workflow's*. I have said so in the file rather than leaving my earlier answer to imply more than it delivers. `claude-pr-review.yml` has the same shape and the same limit; what bounds both is that pushing an in-repo branch requires write access.

Four findings from you on this PR so far, all valid, and two of them things I had read past in my own review of it.

trieloff:
@codex review

trieloff:
Both fixed in a590d76d4, and the first one is the best finding on this PR. Answering them separately.

**Untrusted commenters reaching the prompt — correct, and I had reasoned my way past it.** When I narrowed `allowed_bots` I argued that a reviewer's text is input by design, so trusting the content was fine. That holds for *reviewers*. It does not hold for arbitrary third parties, and this repo is public: any account can comment on any PR here. The path you traced is exactly right — comment body → feedback JSON → prompt → a step with unrestricted `Bash` and `BOT_PAT` persisted in `.git/config`. `allowed_bots` gates the triggering actor and does nothing about feedback a run collects, which I should have seen when I wrote the sentence claiming otherwise.

The filter is now deterministic and in the pure module: authors pass if they are one of the three named reviewer bots (by login) or their `author_association` is OWNER, MEMBER, or COLLABORATOR — the associations that imply write access. Everything else is dropped, **including a missing or unrecognised association**, so it fails closed rather than trusting a payload shape we didn't anticipate. Every drop is named in the run log with the association that caused it (`untrusted-author (association=NONE)`); a filter that silently discarded a real reviewer's comment would be the next bug.

One subtlety worth stating because it looks like a hole and isn't: a stranger's comment can still *start* a run. Gating the trigger in YAML would mean trusting `github.event` to make a security decision in a context where the workflow file comes from the head branch; instead the run starts, finds nothing trusted, and skips. Cheap, and the decision stays in the tested module.

Verified live against this PR (read-only): all nine Codex items and all four owner comments kept, the only drop being Copilot's quota notice.

trieloff:
**Mid-run feedback being lost — also correct, and worse than a missed reply.** The bound was "when we posted", so anything created between the scan snapshot and the marker comment was permanently invisible: not deferred, not retried, gone, with no trace that a finding had been dropped. Silent, and unbounded in the sense you describe — nothing short of further feedback would ever surface it.

The marker now records what was actually processed: `<!-- review-response:<sha>:<iso8601> -->`, where the timestamp is the newest feedback item in the answered snapshot. That bound is by construction no later than any item the run did not see, so a comment posted while the model is working is still unanswered on the next pass. Legacy single-field markers still parse and fall back to the comment's `created_at`, so PRs already carrying an old marker don't suddenly re-answer their whole history.

Two things I tightened beyond the report:

- The `decideResponse` input is renamed `respondedWatermark` rather than keeping `lastResponseAt` with new meaning. A parameter whose name says "when we posted" while holding a watermark is precisely the confusion that produced this bug.
- The comparison stays **strict** `>`, and I've written down why, because it leaves one hole: GitHub timestamps are second-granular, so a comment in the same second as the watermark but absent from the snapshot is still lost. `>=` closes it by re-answering every item sharing the watermark on every subsequent run — and since each response records that same watermark again, that is a self-sustaining duplicate-reply loop. A vanishingly rare miss beats a permanent one; if it ever bites, the fix is to record handled item IDs, not to loosen the comparison.

Regression test covers your exact timeline (10:00 in-snapshot, 10:02 mid-run, 10:05 marker) and asserts both that the old bound would have swallowed 10:02 and that the new one doesn't. 1028 dev-tools tests green, 7/7 verify. Also fixed a latent bug the shared marker regex was hiding: it is `/g`, so `matchAll` resumed from `lastIndex` and could miss a marker depending on call order.

Six findings from you on this PR, all six real, and two of these were things my own review of it missed.

trieloff:
@codex review

PR #2203, opened 2026-08-19
feat(ci): weekly canary for the Bedrock model IDs our workflows resolve to

trieloff:
## Summary

Adds a weekly canary over every Bedrock model ID our workflows can actually resolve to. It sends one 1-token `InvokeModel` per distinct ID, and files (or updates) a single issue when an ID is genuinely unusable. No Claude in the loop — this is a deterministic monitor, and it holds `issues: write` and nothing else.

## Why

This week a repo variable held `us.anthropic.claude-opus-4-9`, which does not exist. Five scheduled agents failed on every run for a full day and **nothing noticed** — the crons fired, the jobs went red, and it was found only by reading run logs by hand. Bedrock model IDs are mutable infrastructure: they get retired, renamed, and superseded with no warning, and the failure is silent because a dead ID looks like an ordinary job failure in a workflow nobody watches.

A canary that invokes each ID once a week would have caught it on the first Monday.

## What it does

- **Derives the ID list from the workflows**, rather than a hand-kept copy of it — 10 `vars.*_BEDROCK_MODEL` names plus the `global.anthropic.claude-sonnet-4-6` literal default. A scout with a hardcoded list goes stale exactly like the thing it monitors.
- **Fails loudly if a workflow references a variable the scout was not given** (exit 3). Actions tokens cannot read repository variables — that is a hard 403 — so values must be interpolated in as env, which makes that env block the one hand-maintained thing here. This guard is what stops a future workflow from silently escaping the canary.
- **Reports actionably**: which ID failed, the evidence, which variable holds it, which workflows consume it, and a verified replacement from the same family if one probed OK. It states plainly that a human must set the variable, because Actions cannot write repo variables either.

## The design constraint: `invalid` vs `inconclusive`

One false "this model is dead, go change that variable" issue and nobody reads the next one. So 403 quota, 429, 5xx, timeouts, transport errors, and **anything unrecognised** classify as `inconclusive`, are retried with backoff, and are never reported as a dead model. Only three positive signals produce `invalid`. A week where every probe was inconclusive files nothing and instead warns that the canary was **blind** — a canary that cannot tell you it saw nothing is worse than none.

My review of the worker's implementation found one door this left open, fixed in ec8363295: Bedrock rejects a retired ID and an account that lost its entitlement *identically*, so one revoked token fails every probe at once — and that would have been reported as eleven simultaneously dead models, asking someone to re-value every variable in the repo. Model IDs die one family at a time; credentials fail all at once. When every probe fails it is still filed (a total outage must be loud) but diagnosed as a credential problem, with the token, region, and model access to check first.

## Verification

- `npm run verify` 7/7 · `dev-tools` 1061 tests (33 in this package) · `lint:docs` clean · debt gate clean.
- YAML-parsed the workflow and asserted the cron, the permissions block, and that **every** variable the extractor finds is present in the env block (10 found, 0 missing, 0 unused).
- Ran the extractor against the real `.github/workflows/` — output is the variable→workflow mapping in `docs/dev-tools-details.md`.
- End-to-end CLI runs against recorded Bedrock payloads with a stubbed `fetch`: the dead-ID case, the unwatched-variable case (exit 3), and the all-throttled case (files nothing, warns).

**Not yet proven live, and I want to be exact about it:** `AWS_BEARER_TOKEN_BEDROCK` is a repo secret, so the actual HTTP path — URL, bearer header, request body — has never executed. The classifier is tested against recorded payloads whose status codes and error-type fields I am confident about; the `message` strings are representative, and no test asserts one verbatim. If a real message differs in wording, the failure mode is `inconclusive` (a quiet miss), not a false alarm. I will `workflow_dispatch` it once this is on main, which is the first real proof, and report back.


trieloff:
Both fixed in cafa05340. They are the same mistake in two places: the scout drawing a confident conclusion it had not earned.

**P1 — the auto-close condition.** Correct, and the reason it slipped through is that `has_invalid == 'false' && all_inconclusive == 'false'` *reads* strict. Your case is narrow and entirely plausible: the very ID the open issue is about comes back throttled while the other IDs answer fine. Neither condition trips, so the workflow closes a live outage report having never retested the thing it reports — and then files it again next week. A monitor that closes real issues on a technicality is how people learn to ignore it. Closing now requires `inconclusive_count == 0`: every ID definitely `ok`, or the issue stays open. Closing is the one hard-to-reverse action here, so it gets the strictest gate rather than the most convenient one.

**P2 — commented-out references.** Also right, and worse than cosmetic in the first direction you name. A commented-out `vars.NEW_BEDROCK_MODEL` would trip the unwatched-variable guard, which exits 3, so **every scheduled run would abort** — the canary silently dead in exactly the way it exists to detect. The second direction is a weekly false alarm about a model ID nothing calls. YAML comments are stripped before scanning now, tracking quote state so a `#` inside a string survives (`echo 'issue #42'` keeps a reference on the same line). Tests cover the disabled variable, the commented-out retired ID, a trailing comment after a live reference, and the quoted `#`.

I re-ran the extractor against the real `.github/workflows/` after the change: the same 10 variables and the same `global.anthropic.claude-sonnet-4-6` literal, so nothing reachable was lost to the stripper.

7/7 verify, 1064 dev-tools tests (36 in this package). Still worth stating plainly: the live Bedrock HTTP path has never executed, because the credential is a repo secret. I will `workflow_dispatch` this once it is on main and report what the real probe returns.

trieloff:
@codex review

PR #2204, opened 2026-08-19
feat(ci): let the model scout prove its alerting path on demand

trieloff:
## Summary

Adds a `probe_extra_ids` input to the model scout's manual dispatch, so the classifier's `invalid` path can be exercised against real Bedrock on demand. The extra IDs are probed and logged but never enter the results, so they cannot move a count, change the report, or open or close an issue.

## Why

The scout's [first live run](https://github.com/ai-ecoverse/slicc/actions/runs/32302711794) worked: both reachable IDs returned 200, the extractor found all 10 variables and the literal default, and it correctly filed nothing.

That verifies the request shape, the auth, the region, and the `ok` path. It verifies none of the behaviour anyone depends on. The `invalid` path is the only one that ever alerts a human, it is unreachable while every configured ID is healthy, and **its failure mode is silence** — if real Bedrock words its rejection differently than `VALIDATION_MODEL_PHRASES` expects, a genuinely dead model classifies as `inconclusive`, nothing is filed, and the canary is decorative. That is precisely the outage it was built to catch, so leaving that branch unproven defeats the point of building it.

```
gh workflow run model-scout.yml -f dry_run=true \
  -f probe_extra_ids=us.anthropic.claude-opus-4-9
```

`us.anthropic.claude-opus-4-9` is the ID whose absence took out five scheduled agents for a day. The run now shows what Bedrock actually returns for it and how it classifies.

## Why this can't become a hazard

A diagnostic that can file or close an issue is worse than no diagnostic:

- The extra probes run **after** the real targets and never enter `results`, so `summarizeResults`, `buildReport`, and every issue decision are untouched by them.
- Entries that are not Bedrock Anthropic model IDs are **dropped, not probed**, so a typo becomes a logged rejection rather than a request to an arbitrary URL path.
- The input only exists on `workflow_dispatch`, which needs write access, and is empty on every scheduled run.

## Verification

`npm run verify` 7/7 · `dev-tools` 1067 tests (39 in this package: parse/trim/dedupe, the non-model-ID rejection, and empty/undefined/whitespace input) · `lint:docs` clean · debt gate clean.

I will dispatch the self-test once this is on main and post what real Bedrock returns for the dead ID, including whether the classifier gets it right. If it does not, the phrase list needs an entry and I would rather find that out now than during the next outage.

### DIFF
--- a/.github/workflows/model-scout.yml
+++ b/.github/workflows/model-scout.yml
@@ -47,6 +47,16 @@ on:
         type: string
         required: false
         default: ''
+      # Turns the self-test from a log line into an assertion. Printing a verdict
+      # nobody reads would leave the regression that matters — Bedrock rewording a
+      # rejection until a dead ID reads as `inconclusive` — exactly as quiet as the
+      # outage it causes. Set this and a mismatch fails the run.
+      probe_extra_expect:
+        description: 'Fail the run unless every probe_extra_ids entry classifies as this'
+        type: choice
+        required: false
+        default: any
+        options: [any, ok, invalid, inconclusive]
 
 concurrency:
   group: model-scout
@@ -94,6 +104,7 @@ jobs:
           SINS_BEDROCK_MODEL: ${{ vars.SINS_BEDROCK_MODEL }}
           # Empty on every scheduled run; only a manual dispatch can set it.
           PROBE_EXTRA_IDS: ${{ inputs.probe_extra_ids }}
+          PROBE_EXTRA_EXPECT: ${{ inputs.probe_extra_expect }}
         run: node packages/dev-tools/model-scout/scan-models.mjs
 
       - name: Warn when the canary learned nothing
--- a/.github/workflows/model-scout.yml
+++ b/.github/workflows/model-scout.yml
@@ -34,6 +34,19 @@ on:
         type: boolean
         required: false
         default: false
+      # Self-test hook. The `ok` path proves itself every week, but the `invalid`
+      # path — the only one that ever alerts anybody — cannot be exercised while
+      # every configured ID is healthy, and it is the branch whose failure mode is
+      # silence. Pass a known-dead ID here (`us.anthropic.claude-opus-4-9` is the
+      # one that caused the outage this canary exists to catch) to see what real
+      # Bedrock returns and how it classifies. These IDs are probed and logged but
+      # deliberately excluded from the report, the summary, and every issue
+      # decision, so a self-test can never file or close anything.
+      probe_extra_ids:
+        description: 'Extra model IDs to probe and classify, comma-separated (logged only, never reported)'
+        type: string
+        required: false
+        default: ''
 
 concurrency:
   group: model-scout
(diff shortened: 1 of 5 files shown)


trieloff:
Fixed in 36075bdfe — you are right, and the reason it matters is the same reason the self-test exists at all.

A printed verdict is not a test. The regression I built this hook to catch is Bedrock rewording its rejection until a dead model ID starts classifying as `inconclusive`, and the entire problem with that regression is that it is **quiet**: the scout keeps running weekly, keeps reporting nothing, and looks healthy while its only alerting path is dead. A log line nobody reads leaves it exactly as quiet as the outage it would cause. So I shipped a diagnostic with the same failure mode as the bug.

`probe_extra_expect` (`any|ok|invalid|inconclusive`) makes it an assertion: any self-test probe classifying differently exits 4 and fails the run, and the error says what a failure means — Bedrock changed its rejection, the phrase lists in `lib.mjs` need an entry, and until then the scout cannot report a dead model. Setting an expectation with no IDs to probe also fails rather than passing on a vacuous truth.

```
gh workflow run model-scout.yml -f dry_run=true \
  -f probe_extra_ids=us.anthropic.claude-opus-4-9 \
  -f probe_extra_expect=invalid
```

Three unit tests cover it: no expectation checks nothing, a dead ID reading `inconclusive` is reported as a mismatch, and matching verdicts pass case-insensitively.

Unrelated but worth admitting since it happened on this PR: I had been piping `npm run verify` through `tail -1`, which discards the exit code, and it hid a real biome failure (import ordering) for one commit. Caught it before pushing, fixed, and I am checking the status now instead of the last line. 7/7 verify, 1070 dev-tools tests.

trieloff:
@codex review

trieloff:
Live proof, now that this is on main.

**The alerting path works against real Bedrock.** `us.anthropic.claude-opus-4-9` — the ID whose absence took out five scheduled agents — returns `400 ValidationException: The provided model identifier is invalid.` and classifies as `invalid`. So the phrase list does match Bedrock's real wording, which was the thing I could not verify from recorded payloads and the thing everything else depended on. [Run 32309853764](https://github.com/ai-ecoverse/slicc/actions/runs/32309853764).

**And the assertion can actually go red**, which matters because an assertion never observed failing is not known to work. Asserting `ok` for that same dead ID fails the run with `❌ self-test mismatch: us.anthropic.claude-opus-4-9 expected \`ok\`, got \`invalid\`` and the note about the phrase lists. [Run 32309981597](https://github.com/ai-ecoverse/slicc/actions/runs/32309981597).

**And the diagnostic stayed inert:** both runs reported `2 ok, 0 unusable, 0 inconclusive` and `Every reachable model ID answered. Nothing to file.` The dead ID never entered the summary, and zero issues exist with the `bedrock-model-scout` label. The failing run failed at the probe step, so it also never reached the issue steps.

The scout is now verified end to end: extraction, request shape, auth, region, the `ok` path, the `invalid` path, and the guard that would tell us if the `invalid` path ever broke.

PR #2210, opened 2026-08-20
feat(tools): bound and kill long-running bash runs

trieloff:
## Summary

A `bash` command that never returns no longer holds the turn open. Two knobs bound the wait — `background_after` (default 600s) detaches a slow command and reports its outcome later as a **`bash` lick**, and `timeout` (no default) kills it — and every invocation is now a real kernel process, so `ps` lists a background job, `kill <pid>` reaches it, and the timeout's SIGKILL fans out to realm-backed descendants (`node`, `python3`, `.jsh`) as an uncatchable `worker.terminate()`.

Before: a wedged command forced the user to cancel the whole turn in the cone, and in a scoop nothing could stop it — no user is there, and a blocking `agent` caller waits with it. Now: the agent gets a job id (and a pid) at the budget, keeps working, and receives the exit code, a 2 KB preview, and the full output path when the command finishes.

## Why

Asked to check how pi handles long-running bash tools first: pi's bash tool takes an optional model-set `timeout` in seconds, has no default, and has no backgrounding, so there was nothing to inherit. The failure mode this fixes is specific to unattended execution — the cone at least has a human with a cancel button; a scoop has neither.

Defaults are configurable per context because that is what the unattended case needs:

| Surface | How |
| --- | --- |
| Per call | `bash({ command, background_after, timeout })` |
| Per scoop | `scoop_scoop({ background_after: 60 })` → `ScoopConfig.backgroundAfterSeconds` |
| Per agent call | `agent … --background-after 60` → `AgentSpawnOptions.backgroundAfterSeconds` |

- **Layering.** `tools/bash-tool.ts` owns the budget logic and takes a `BashJobHost` handle (`pid` / `signal` / `kill()` / `exit()`) defined in `tools/types.ts`; `scoops/scoop-context.ts` implements it over the `ProcessManager` (`spawnBashJob`). `tools/` sits below `kernel/` in the layer stack, so the manager is never imported there.
- **Why a pid, not just an abort.** `runInRealm` takes no `AbortSignal` — `pm.signal(pid, …)` is its only stop path. So a cooperative abort can never interrupt a realm-backed command; only a signal delivered to a pid the realm parents under can. The job pid is therefore passed as the third argument of `executeCommand`, which makes realm children descendants of *that job* rather than of the whole turn.
- **Detach lifecycle.** The job record stays `running` after detaching (that is what makes it `ps`-visible and killable) and is reaped with the command's exit code on completion. Parented to the turn pid, so a normal turn end (`pm.exit`, no cascade) lets a detached job survive to deliver its lick, while an explicit cancel / `stop()` / `drop_scoop` — all of which *signal* the turn pid — reaps the job and its realm descendants (#1166 semantics preserved).
- **New runtime invariant: per-run realm parentage.** Detaching makes concurrent runs on one `AlmostBashShell` real, and `activeShellPid` is a single mutable field — a detached run spawning its realm child late would attach it to whichever run started most recently, so `kill` would hit the wrong tree. just-bash hands each command context the `signal` its exec started with, so parentage now lives in a `WeakMap<AbortSignal, number>` (`jobPidByRunSignal`), with `activeShellPid` (panel terminal) and `getCurrentShellPid` (turn pid) as fallbacks. `ctx.signal` is threaded through the `node` and `python` commands to reach it. The regression test fails without this (it mis-parents to the other job's pid).
- **Alternative rejected: a realm per bash invocation.** That would preempt everything, including CPU-bound builtins, but hosting just-bash in a realm means an RPC for each of the ~108 supplemental command modules that close over CDP, the `ProcessManager`, sudo brokers, secrets, and the orchestrator. Out of scope here; the residual gap is documented instead (below).
- **Known limit, stated not implied.** In-worker just-bash builtins (`grep`, `sed`, `jq`) have no worker to terminate: SIGKILL aborts cooperatively and just-bash observes it only at its next statement boundary, and a CPU-bound builtin blocks the event loop the detach/timeout timers run on. Written up in `docs/kernel/process-model.md § Bash jobs`.
- **New lick channel** `bash` wired end to end: `LickEvent` type + fields (`bashJobId`, `bashCommand`, `bashExitCode`, `bashJobPid`), `EXTERNAL_LICK_CHANNELS`, `LICK_LABELS`, `formatBashLick`, host name/id resolvers, `LICK_CHANNELS`, and a `square-terminal` lick-card icon.
- **No new dependencies.** No persisted-state or IndexedDB shape changes.

## Cross-cutting impact

- **Floats exercised**: the `bash` tool runs in the kernel worker on every float (standalone CLI, Electron/Sliccstart, cloud hosted leader, and the thin extension via the hosted leader tab's worker), so all of them get the budgets. The kill path is float-correct because the realm runner terminates a `DedicatedWorker` in standalone and removes the sandbox iframe in the extension — both uncatchable.
- **Contexts without a `ProcessManager`** (in-process test factory, any float built without a kernel host) get `jobHost.spawn() → null` and run exactly as before: no pid, no `ps` entry, cooperative abort only. Covered by a test.
- **Shell contexts**: the panel-terminal `AlmostBashShell` path is untouched — it still passes its own shell pid and now also gets the per-run `WeakMap` entry as a side benefit. Offscreen / side-panel shells are unaffected.
- **Other callers of the changed contracts**: `buildProcessConfig` gained an optional parameter; its only callers are the `node` and `python3`/`python` commands, both updated. `executeCommand`'s third argument already existed (the panel terminal host passes it) — the bash tool is simply the second caller to use it. `ScoopConfig.structuredOutputSchema` was retyped from `Record<string, unknown>` to a named `JsonSchemaObject` as part of the debt payoff; `agent-bridge`, `agent-command`, and `structured-output-tool` were checked and typecheck clean.
- **Follower wiring / transcript export**: the `bash` lick is deliberately local-only (`LOCAL_ONLY` in the lick-manager exhaustiveness map) — a background command belongs to the runtime that started it and is not forwardable to a follower. It reaches the transcript through the normal channel-message path, so no export change is needed.
- **Bundle size**: `webapp: kernel worker` raised to 1886 kB. Measured against `origin/main` at `dc706cb5b` (which had itself raised the budget to 1875 kB): base 1,871,663 B, this branch 1,882,230 B, so this change is ~10.6 kB — nearly all user-facing string payload (tool description, schema descriptions, the detach/timeout guidance the model acts on, and the delegation `SKILL.md` section, which `vfs-root` bundles into the worker). I trimmed ~1.5 kB of redundant prose first. The new limit leaves 3.8 kB of headroom, matching the 3.3 kB `main` currently holds over its own base, so the ratchet moves by this change's size and no further.
- **Security**: no new secret surfaces. Full command output goes to `<tempDir>/bash-<jobId>.txt` via the ungated fs, the same per-context temp dir (`/tmp` for the cone, `/scoops/<folder>` for a scoop) the existing truncation path already writes to, and the lick preview is bounded at 2 KB and passes through the normal tool-result scrubber path.

## Test plan

- [x] `npm run verify` — all 7 checks pass (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode ×2)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — OK, 32 changed files, 0 on any debt list
- [x] `npm run typecheck`
- [x] `npm run test` — 15,464 passing, no new failures
- [x] `npm run test:coverage` — floors met
- [x] `npm run build`
- [x] `npm run bundle-size` — passes with the raised kernel-worker limit
- [x] **New / changed behavior is covered by unit tests**:
  - `npx vitest run packages/webapp/tests/tools/bash-tool.test.ts` — 45 tests; 20 new (schema advertisement, detach, lick contents + routing, failed/thrown jobs, job registration, pid threaded into the shell, exit-code reaping, SIGKILL at timeout, external `kill`, turn-cancel exit code, no-job-host fallback)
  - `npx vitest run packages/webapp/tests/tools/bash-job-process.test.ts` — new; drives a **real** `ProcessManager`: `ps` visibility, SIGKILL fan-out to a realm-kind descendant, `kill <pid>` on a detached job, survives-turn-end vs dies-on-cancel
  - `npx vitest run packages/webapp/tests/kernel/bash-lick-routing.test.ts` — new; cone vs targeted-scoop routing, dropped-target discard
  - `npx vitest run packages/webapp/tests/scoops/scoop-realm-parenting.test.ts` — the concurrency regression (verified it fails without the `WeakMap` fix: parents to the other job's pid)
  - plus `scoop_scoop`, `agent --background-after`, `agent-bridge`, and `lick-formatting` coverage
- [x] N/A — no UI change, so no screencast. The only UI-adjacent line is a lucide icon name for the existing lick-card component.

**Pre-existing failures**: none. The `packages/webapp/tests/shell/ipk/*-e2e` suites install real packages from the npm registry against a 5 s per-test budget, so they time out in bursts under load (I saw three fail together, pass on a rerun, and pass on `main` under the same conditions). Unrelated to this change.

## Review follow-up (Codex, 3 findings — all fixed in `fix(tools): close three review gaps…`)

All three were real. Each fix has a test that fails without it.

1. **P1 — detached output bypassed the secret scrubber.** A foreground bash result is scrubbed at the `adaptTools` boundary, but a detached job reaches the agent as a lick preview plus a file on disk, and neither crosses it. The tool now applies the same `getToolResultScrubber()` pass itself, **before** the file write (a preview-only scrub would just move the leak to disk), and withholds the output entirely if the scrubber throws — the lick still reports the exit code. My original PR text claimed this path was already covered; it was not.
2. **P1 — a detached job outlived its scoop.** By dispose time the turn record it was parented to is gone, so `dispose()`'s turn-pid SIGTERM could not reach it: `drop_scoop` (or one-shot `agent` teardown) deleted the scoop directory and left the command running against it. `ScoopContext` now tracks live job pids (`liveBashJobPids`, entries removed as each job exits or is killed) and SIGKILLs whatever remains.
3. **P2 — both `.jsh` dispatch paths dropped the run signal.** Confirmed and fixed: they called `buildJshProcessConfig()` bare, so a discovered `.jsh` fell back to the mutable `activeShellPid`. The new test reproduces the mis-parenting exactly (realm child attached to job B's pid, 1026 instead of 1025) and passes with `ctx.signal` / the exec signal threaded through. `executeScriptFile` still calls it bare on purpose — it is a by-path API with no run context and no production caller.

## Review follow-up, round 2 (Codex, 2 findings — fixed in `fix(scoops): deliver bash licks to gated scoops…`)

1. **P1 — the completion lick never reached a gated scoop.** `ScoopMessageRouter.passesTriggerGate()` keeps its own hardcoded list of machine-addressed channels and `bash` was not on it, so for a `scoop_scoop` scoop (`requiresTrigger: true`) the lick was filtered and the watermark advanced: the scoop never learned its own job had finished. `bash` is now in that gate. I deliberately did **not** swap the gate for `isExternalLickChannel()` — that predicate also covers `session-reload` / `navigate` / `upgrade` / `cherry` / `preview` / `discovery` / `workflow`, and letting those bypass a scoop's trigger is a routing change outside this PR. The narrow list now carries a comment saying why it is narrower.
2. **P1 — a nested `agent` outlived the run that started it.** `buildSpawnOptions` never set `signal`, so nothing propagated a stop into the spawned scoop: not the new `timeout` kill, and not a cancelled turn either. `AgentBridge.spawn` already honoured `options.signal`. Threaded `ctx.signal` through, and added it to the `AgentSpawnOptions` mirror in `agent-command.ts` (`shell/` is below `scoops/`, so the shape is restated rather than imported).

## Documentation

- [x] `docs/` — `tools-reference.md` (both knobs, the pid/`ps`/`kill` affordance, the non-preemptible builtin case), `kernel/process-model.md` (new **Bash jobs** section, new spawn-site row, per-run parentage, residual limit; also corrected a stale claim that the realm runner only terminates on SIGKILL — it escalates every terminating signal), `architecture.md`
- [x] `packages/vfs-root/workspace/skills/delegation/SKILL.md` — `--background-after` on `agent`, a `background_after` subsection for `scoop_scoop`, and the `kill <pid>` affordance
- [ ] `README.md` — not user-facing surface
- [ ] `CLAUDE.md` — no convention change

## Risk & rollback

- **Blast radius**: the `bash` tool on all floats. The riskiest new behaviour is the *default* 600s detach: a command that previously blocked forever now returns a job id at ten minutes, so an agent that ignores the "do not re-run it" guidance could double-run a slow command. The wording in the tool result and in the lick is deliberately explicit about that.
- **No flags or env knobs.** Reverting this PR restores the previous behaviour cleanly; nothing is persisted and no migration is involved. Only side effect left behind by a revert would be already-written `bash-<jobId>.txt` files in temp dirs.
- **CI cannot catch**: the actual `worker.terminate()` of a real realm (tests assert the signal delivery and `terminatedBy`, and the existing in-process realm test covers the runner). Worth one manual check by a reviewer: run `bash({ command: "node -e 'while(true){}'", timeout: 5 })` in a dev instance and confirm the worker dies (CPU drops, `ps` shows the `jsh` pid gone) rather than lingering.

## Checklist

- [x] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs (new `bash` lick channel, two new `bash` tool params, `agent --background-after`, `scoop_scoop({ background_after })`)
- [x] If this changes a contract used by other floats, I checked the followers/leader/offscreen paths (see Cross-cutting impact — the new lick channel is intentionally local-only)


trieloff:
Thanks — all three Codex findings were real and are fixed in `d1da78ea6`, each with a test that fails without the fix. Also rebased onto `96dc58782`.

**P1 scrub the detached output** — correct, and my PR text was wrong to claim this path was covered. A detached job reaches the agent as a lick preview plus a file on disk, neither of which crosses the `adaptTools` boundary. The tool now applies the same `getToolResultScrubber()` pass itself, and scrubs **before** the write rather than only the preview, since a preview-only scrub would just relocate the leak to disk. A throwing scrubber withholds the output instead of passing it through; the lick still carries the exit code. Tests: scrubbed in both the preview and the file, and the withhold-on-throw path.

**P1 reclaim detached jobs when the scoop is removed** — correct. `ScoopContext` now tracks live job pids in `liveBashJobPids` (added at spawn, removed as each job exits or is killed) and SIGKILLs whatever remains in `dispose()`. Kill rather than await: for `drop_scoop` and one-shot `agent` teardown the directory the output would be written to is being deleted, so waiting would only delay the teardown. Tests cover both that a live job is signalled and that an already-exited pid is not.

**P2 propagate the run signal into discovered `.jsh`** — confirmed and fixed on both paths (the discovered-command handler gets `ctx.signal`, `tryJshFallback` gets the originating exec signal). The new test in `scoop-realm-parenting.test.ts` reproduces the mis-parenting precisely: without the fix the realm child attaches to job B's pid (1026 instead of 1025), so `kill` on job A misses it. One deliberate exception: `executeScriptFile` still calls `buildJshProcessConfig()` bare — it is a by-path API with no run context and no production caller, so `activeShellPid` is the right fallback there.

One unrelated change in the same push: the `webapp: kernel worker` budget is now 1888 kB rather than 1880 kB. The base grew 7.7 kB over the 16 commits this branch rebased onto, which left the earlier pick with 601 B of headroom. Measured against `96dc58782`: base 1,868,842 B, branch 1,879,399 B (~10.6 kB, nearly all user-facing strings), and the new limit holds 8.6 kB of headroom — the same margin the old 1870 kB limit had over its own base.

trieloff:
@codex review — the three findings from the first pass are fixed in the second commit (secret scrub on detached output before both the file write and the preview, `liveBashJobPids` reaped with SIGKILL at `ScoopContext.dispose()`, and the run signal threaded through both `.jsh` dispatch paths). Please re-check those specifically.

trieloff:
Both second-pass findings were real too, and are fixed in `25c64728c` — again each with a test that fails without the fix.

**P1 bypass the scoop trigger gate for bash licks** — this one would have quietly broken the feature exactly where it matters most. `ScoopMessageRouter.passesTriggerGate()` keeps its own hardcoded list of machine-addressed channels (`webhook` / `cron` / `fswatch` / `sprinkle`) and `bash` was not on it, so for a `scoop_scoop` scoop (`requiresTrigger: true`) the completion was filtered and the watermark advanced: the scoop never learned that the job it started had finished. `bash` is now in that gate, and the existing `it.each` parameterised test covers it.

I deliberately did **not** switch the gate to `isExternalLickChannel()`, even though the duplication is what caused the miss: that predicate also covers `session-reload`, `navigate`, `upgrade`, `cherry`, `preview`, `discovery` and `workflow`, and letting all of those bypass a scoop's trigger is a routing change well outside this PR with no tests behind it. The narrow list now carries a comment saying why it is narrower, so the next reader does not "fix" the duplication blindly.

**P1 propagate timeout cancellation into nested agent commands** — correct, and worse than the timeout case alone: `buildSpawnOptions` never set `signal`, so nothing propagated a stop into a spawned scoop — not the new `timeout` kill, not a cancelled turn. `AgentBridge.spawn` already honoured `options.signal`; it just never received one. `ctx.signal` is now threaded through, with tests for both the present and absent cases.

Verification after both fixes: `npm run verify` 7/7 · `check-touched-exemptions` OK (35 files, 0 on any debt list) · typecheck clean · `npm run test` 15,464 passing · coverage floors met · both builds · `bundle-size` passing.

trieloff:
Third pass on the two second-round fixes plus one CI catch: the `signal` field was missing from the `AgentSpawnOptions` **mirror** in `agent-command.ts` (`shell/` sits below `scoops/`, so the shape is restated rather than imported). My local typecheck grep swallowed it — `tsc` colorizes output and the ANSI codes sit between `error` and `TS2339`, so a plain `grep 'error TS'` matched nothing while tsc had exited 1. CI was right; fixed in `c61f0b701`, and all 18 checks are green.

@codex review — please re-check the trigger-gate and nested-agent-signal fixes.

PR #2211, opened 2026-08-20
feat(ui): link file mentions in chat and enrich the file preview

trieloff:
## What

Agents name files constantly — "I rewrote `bb.jsh`", "see `packages/webapp/src/main.ts:42`" — and those names were inert text. They are now links.

A mention is checked against the VFS in the background and **only becomes clickable once the file is confirmed to exist**, so the transcript never shows a link that leads nowhere. Clicking opens a preview that identifies files by their **content** rather than a hardcoded extension list, so `.jsh` previews as source instead of hitting "Preview not available".

.jsh preview with shell syntax highlighting

## Why the preview was broken

`getMimeType()` answers `application/octet-stream` for any extension it has not been told about — correct for *serving* bytes, fatal for *reading* them. `slicc-quick-look` keyed off that and gave up. There were also **two** drifted extension maps (`core/mime-types.ts` and a copy in `ui/wc/file-actions.ts`).

`sniffFileType()` replaces them for the reading path: **magic bytes → extension → UTF-8 decodability**. The precedence matters — a `.txt` full of PNG bytes still reports as an image; only files nothing else identifies get the permissive text fallback. The serving path is deliberately left alone (sniffing a type you then put in a `Content-Type` header is how MIME-confusion bugs happen).

## How the mention pipeline is split

Guessing and verifying have opposite failure costs, so they are separate modules:

| Module | Job |
|---|---|
| `core/file-mentions.ts` | The heuristic. Pure; finds candidates, knows nothing about the VFS. |
| `core/file-mention-resolver.ts` | Verification, via a lazily built, bounded basename index. |
| `ui/file-mention-linker.ts` | Walks a rendered message; links only what resolved. |
| `ui/wc/wire-file-mentions.ts` | Lifecycle; waits out streaming before resolving. |

The heuristic can afford to be permissive **because** verification gates it. It rejects versions (`1.2.3`), domains (`diffs.com`), and sentence-enders (`done.`); `.sh` is pointedly kept despite being a TLD.

Streaming bubbles are skipped entirely — a half-arrived path (`packages/webapp/src/ma`) resolves to nothing and would burn a lookup per token.

## Pierre libraries

- **`slicc-quick-look`** renders text through **`@pierre/diffs`** (diffs.com). A file with uncommitted changes opens on its **diff**.
- **`slicc-file-tree`** renders through **`@pierre/trees`** (trees.software), gaining search, virtualization, inline rename and git status lanes.

file tree with search, type icons, sizes and git status lanes
Quick Look showing a unified diff for a modified file

Quick Look renders text **twice on purpose**: a synchronous `<pre>` first, then the `@pierre/diffs` view once that lazily-imported chunk arrives. The overlay is never empty and never blocked, and if the import fails the `<pre>` stays — a degraded preview, not a broken one.

### Two behaviour changes worth reviewing

1. **Row actions moved into the context menu.** `@pierre/trees` renders row decorations as text or an icon only — there is no slot for interactive controls — so the four hover buttons (Preview / Reference / Download / Overflow) are now context-menu items. The events are unchanged; only the click target moved.
2. **`file-preview` also fires on Enter / double-click**, which the old tree had no equivalent for.

### Upstream bug worked around

`@pierre/trees` 1.0.0-beta.6 renders a **single blank row and no files** for leading-slash absolute paths (`['/a.ts','/b.ts']`) — the `/` becomes an empty first segment. The adapter strips it going in and restores it in events, so absolute VFS paths stay absolute at the component boundary.

## Bundle size — please read

**I raised the total-JS-payload budget from 31 MB to 34 MB.** The honest numbers:

| | Size |
|---|---|
| `main` baseline | **30.36 MB** (98% of the 31 MB budget) |
| This PR | **33.02 MB** |

The previewer needs a syntax highlighter and there was 0.64 MB of headroom. Pulling `@pierre/diffs` into the Vite graph costs ~6.4 MB of lazily-fetched chunks; **3.7 MB of that is given back** by shrinking the *pre-existing* `slicc-diff.js` sprinkle bundle from 9.4 MB → 5.7 MB (it was inlining all 332 Shiki grammars into one file served to an iframe).

What did **not** move is what users pay on every load:

- main entry chunk: **22.84 kB**, unchanged (limit 24 kB)
- kernel worker: **1.86 MB**, unchanged (limit 1.87 MB)

The added bytes are per-language chunks fetched only when a file of that type is previewed.

Both builds now resolve `shiki` to `src/shims/shiki-bundle.ts` — Shiki's `bundle/web` subset (78 languages) plus Go, Rust, Swift, TOML and diff, which `bundle/web` omits and this repo is written in. Trimming further is not worth it: **78 → 37 languages moved the payload by under 200 KB**, because the weight is Shiki's engine and theme machinery, not the grammars.

**Known follow-up:** `@pierre/diffs` is now bundled twice — once by Vite for the main page, once by esbuild for sprinkle iframes. Teaching Quick Look to reuse the already-built `slicc-diff.js` would put the payload *below* today's baseline, but it means inverting the component's dependency on the library, which does not belong in this PR.

## Verification

- **15,138** node tests + **2,036** browser tests (real Chromium, real Pierre libraries)
- `npm run verify` (all 7), full `typecheck`, both coverage gates, boy-scout debt gate
- Coverage on the rewritten tree: 83% → **94%** statements, **100%** functions

The file-tree suite was rewritten to assert the **contract** (accessible row names, events) rather than markup, since the DOM now belongs to the library. Synthetic events need `composed: true` to cross its shadow boundary — worth knowing before writing more.

Not verified end-to-end in a live instance: the webapp loads from the hosted origin, so seeing mentions light up in a real transcript needs a deploy.


trieloff:
please provide screenshots of the file listing side rail in a narrow and wide side rail, so that we can evaluate the responsive design

trieloff:
## Responsive rail — narrow vs wide

@trieloff here are the two ends of the range. I also found (and fixed) a regression while producing them: the rewrite had dropped the host chrome entirely. The old tree declared `width: 190px; flex: 0 0 auto` plus a right divider and padding; the replacement declared nothing, so the rail lost its divider and had no defined sizing behaviour in a flex row.

The chrome is back, with one deliberate change: **the rail now fills its container instead of pinning to 190px.** The surrounding layout already knows the right width — the `files` panel declares `minWidth: 220` and the dock zone is user-resizable — and a virtualized tree is happy at any of them. `min-width: 0` is what lets it shrink below its content, so a narrow rail elides instead of overflowing.

Both are added as stories (`NarrowRail`, `WideRail`) so this stays reviewable.

**Narrow — 200px**

file rail at 200px: names middle-truncate, git badges hold the right edge

**Wide — 420px**

file rail at 420px: full names with size and git status

### What to judge

The tradeoff is in the right-hand decoration slot, and it is the library's behaviour rather than something I chose:

- At **420px** everything fits: name, size (`2K`, `3K`) and git status (`M`, `D`).
- At **200px** the size column yields to the git badge, and names **middle-truncate** — `hero.css` becomes `h...css`, `components` becomes `co...nents`. Indentation and the type icons still read, but the truncation is aggressive at that width.

If `h...css` is too lossy, the options are to raise the `files` panel's `minWidth` above 220, or to drop the size decoration below some width so names keep the space. Happy to do either — it's a one-line change in `renderRowDecoration`.

## Review findings

Thanks — four of the six were real and are fixed:

| Finding | Status |
|---|---|
| Resolver never invalidated (both reviewers) | **Fixed** — 30s TTL on the index, checked on lookup so an idle transcript costs nothing. Negative answers expire with it. |
| `~/` and `../` mentions could never resolve | **Fixed** — both prefixes are stripped to the meaningful tail, which the existing segment-boundary suffix rule then matches. |
| Context menu didn't emit `file-preview` / `file-reference` / `file-download` | **Fixed** — this was the worst of the four. The menu called the implementations directly, so it worked while every other consumer of the documented events, including my own RowActions story, saw nothing. It re-dispatches now. |
| Menu offered directory ops that always fail (`EISDIR`, non-recursive `rm`) | **Fixed** — Rename/Duplicate/Delete are file-only, matching the old tree, which drew no row actions on directories at all. |
| `path:42` ignored in diff mode | **Fixed** — `setSelectedLines` now runs on the diff branch too, using the `additions` side. |
| Copilot review | Blocked on quota, not a finding. |

Each fix has a test: `file-actions-menu.test.ts` (new, 8 cases) pins the event contract and the directory gating specifically, since that is the pair most likely to rot.

## CI

Green except the `e2e` job, which failed on `speech-roundtrip.test.ts` — the real-model kokoro spec — timing out in its own `bootLeader` waiting for the terminal.

That is not this PR: **`git-clone-live.test.ts` passed in the same run using the identical `dock.selectItem('term')` + `waitForFunction(__slicc_terminal_view)` sequence.** The speech spec is #17 of 18, runs `retries: 0` by deliberate choice, and its own comments document repeated timeout tuning ("observed CI runs land between 9 and 15 minutes").

I did harden the mention wiring regardless: it sat ahead of the block that lazy-mounts tool panels — whose own comment notes that skipping it leaves "the terminal open with no session" — so a throw there really could have taken the terminal down. It now catches, logs and returns a no-op teardown.

PR #2214, opened 2026-08-20
perf(webcomponents): frame-budget the slicc-shader background (idle 44% GPU → ~3%)

karlpauls:
## What

Gives the `<slicc-shader>` full-viewport WebGL background a **frame budget** so it stops burning GPU/CPU while the leader tab sits idle-but-visible.

- **15 fps ambient cadence** (was display-rate 60/120 fps) via a new pure, DOM-free gate `packages/webcomponents/src/freezer/frame-budget.ts`.
- **800 ms full-rate burst** on interactive stimuli (`pulse()`, attribute/`scroll`/`tint`/`mode` changes, theme flips, genuine resizes).
- **True static stop**: `cone` with `speed=0` or `prefers-reduced-motion` renders once per stimulus and **stops the loop** — `speed=0` is now literally paused, not just a zeroed uniform.
- **DPR-1 default backing-store cap** (was 2) with a new `dpr` attribute (0.5–2) escape hatch for hero/showcase uses.
- **Context-loss safety**: the scheduler parks entirely while the WebGL context is lost (`#contextLost` guard) so a stimulus can't spin an empty rAF loop.

## Why (measured, root cause confirmed live)

On a thin-extension install (Chrome 151, Apple Silicon, leader tab visible in a background window, panel closed, agent idle):

| Process | With shader | After removing `<slicc-shader>` |
| --- | --- | --- |
| Chrome GPU process | **~44%** | ~2% |
| Leader-tab renderer | ~18% | 0.0% |

rAF only pauses for *hidden* pages; a visible-but-background window renders at full rate, and the tray's open WebRTC connection exempts the tab from timer throttling — so nothing reined the ungated loop in. Combined 60→15 fps (÷4) and DPR 2→1 (÷4 pixels) targets **~44% → ~3% GPU process**.

## How it was built

Executed the approved spec + plan task-by-task (TDD):

1. `frame-budget.ts` pure gate + unit tests (8).
2. Scheduler restructure in `slicc-shader.ts`: one `#wake()` entry point + a self-terminating `#tick` (+6 behavioral tests).
3. `dpr` resolution cap (+2 tests, proven under a stubbed `devicePixelRatio=2`).
4. Docs: webcomponents `CLAUDE.md` convention, `docs/webcomponents-details.md` pattern, `docs/pitfalls.md` incident.

**Plan defect found & fixed during implementation:** plan §3h wired `new ResizeObserver(() => this.#wake({ burst: true }))`, but ResizeObserver fires a *mandatory initial notification* on `observe()`, which opened an 800 ms burst on every connect and made the ambient-budget acceptance test read ~120 fps. Fix skips exactly that first (mount-artifact) notification while keeping the burst for genuine resizes; the spec + plan (§3h / Step 4) were reconciled in the same commit so the audit trail stays truthful. The acceptance test was left exactly as written.

## Review

The spec + plan survived an adversarial **Codex (gpt-5.5, xhigh) pre-implementation review**: REWORK with 9 findings, all folded in before implementation (exact-indentation snippets, `#contextLost` guard, no-burst mode-switch/context-restore, per-wake reduced-motion pulse contract, DPR-under-stubbed-ratio tests, tighter ambient margins, full CI gate list).

## Planning artifacts

The spec and plan ship in this branch by repo convention (`docs/superpowers/specs/2026-08-19-shader-energy-budget.md`, `docs/superpowers/plans/2026-08-19-shader-energy-budget.md`) — the release process purges them from `main`; they are intentionally **not** stripped here.

## Verification

All plan-scope gates green: `verify` (7/7), `lint:ci`, `deadcode`/`deadcode:production-files`, `typecheck`, webcomponents suite (2063/2063, incl. shader 37/37 + frame-budget 8/8), `test:coverage:webcomponents` (floor met), `npm run build` + `build -w @slicc/chrome-extension`, `bundle-size` (all under limit), `lint:docs`.

Note: local `npm run test` shows 4 failures in `packages/dev-tools/tools/check-skill-router-sync.test.mjs` — a pre-existing macOS + Node-24 `rmSync`-on-symlink quirk in a subsystem untouched by this PR; it runs green on CI (Linux).

## Screenshots

Storybook PR screenshots will re-capture the `freezer` area stories — expected visually near-identical (CI captures at `devicePixelRatio` 1, where the DPR cap is a no-op).


karlpauls:
## Addressed review feedback (commits `e001cb5b4`, `f8c194a64`)

Ran a multi-agent review pass (correctness, tests, comments, silent-failures, type-design) and folded the findings in:

**Correctness**
- `pulse()` now rejects non-finite input. A `NaN` energy never cleared the `#energy < 0.001` rest floor, so the `#energy === 0` static-stop could never fire — a static field would spin forever uploading a `NaN` `u_energy`. Regression test added.
- `webglcontextrestored`: on `#setupGlResources()` failure it now degrades to the CSS fallback (+ a `console.error`) exactly like the initial `#initGl` path, instead of leaving the field parked/blank with `#contextLost` stuck `true` and no signal.

**Docs**
- Corrected the new pitfall's WebRTC claim: WebRTC exempts the leader tab from timer throttling only — **not** freezing (that's disabled via launch flags). It had contradicted the authoritative Memory-Saver pitfall in the same file.
- Comment precision: ResizeObserver initial notification fires for a rendered non-zero-size element; noted the seconds-vs-ms clock split; "~0.43 Hz **at max speed**".

**Tests (webcomponents shader suite now 40 in-file, all green)**
- `cone speed=0` renders **exactly one** frame (pins "renders once", complementing the existing "then stops").
- A theme change repaints and **re-stops** a static field (the previously-untested theme→`#wake` burst path).
- `pulse(NaN)` does not spin a static field (guards the fix above).

The automated Claude review returned **LGTM**; its three observations were confirmed non-actionable (the `speed=0` focal-orbit freeze is the intended "literally paused" semantics; the reduced-motion `#energy` latch is harmless since reduced-motion always self-terminates after one frame; `this.dpr` per-render is a cheap no-reflow read).

Local gates green (`verify` 7/7, typecheck, webcomponents suite, `lint:docs`). `package-lock.json` remains intentionally uncommitted.

karlpauls:
## Review round 2 applied (commits `3453b13ea`, `869d82515`; rebased onto latest main)

Addressed the four-agent post-merge review.

**MUST-FIX — ResizeObserver skip swallowed the first real notification**
The one-shot skip suppressed the *wake*, not just the burst. An element mounted inside `display:none` gets no initial notification, so the skip then ate the genuine first "became visible" resize, leaving a static field a silent 1×1 canvas. Now the initial notification **wakes without a burst** (`this.#wake({ burst: !wasInitial })`).

**MUST-ADD tests** (webcomponents shader file: 46 in-file, all green; stable across repeated runs)
- Hidden→visible reveal renders at real size (guards the fix above).
- Burst expiry on an animated field returns to the ambient budget (uses a **pure `scroll` burst** — see note below).
- Context restore resumes rendering (animated) + a static variant that re-renders and re-stops.
- Failed restore degrades to the CSS fallback **and emits a diagnostic** (asserted).
- Reduced-motion renders one frame per wake and stops.
- Mode-switch performs exactly one synchronous draw (pins `#lastFrameTs = performance.now()`).

**Hardenings** — clear `#contextLost` before the failed-restore degrade (so the stale flag can't outlive the disposed `#gl`); clamp `#energy` non-negative in `pulse()` (self-documents the `#energy === 0` termination invariant; finite guard kept).

**Doc accuracy** — WebRTC exempts **timer throttling only** (freezing is a separate launch-flag / `autoDiscardable` opt-out — fixed both `pitfalls.md` and `webcomponents-details.md`, plus the dangling "Memory-Saver" cross-ref → "Local Network Access: Launched Chrome Must Disable the Check"); `FRAME_EPSILON_MS` fixes **pacing, not the average**; DPR budget is **per-mode**; docblock notes mode-switch is burst-free + cites `AMBIENT_FPS`/`BURST_MS`. Spec's ResizeObserver bullet + plan's "Review round 2" log reconciled.

**Two intentional deviations from the literal instructions, flagged for transparency:**
1. **Burst-expiry test uses a `scroll` change, not `pulse()`.** `pulse(1)`'s glow energy decays over *rendered frames* (~135 frames from 1.0), which keeps an animated field at display rate well past `BURST_MS` in a frame-rate-dependent way — with `pulse()` the measurement read 34 draws (> the approved ≤22). A `scroll` change opens a **pure** burst (no lingering energy), which is exactly the finding's named "continuous scroll bursts" scenario and makes the ≤22-over-1000ms assertion valid and CI-robust.
2. **The round-1 "cone speed=0 renders exactly one frame" and reduced-motion assertions were relaxed to a 1–2 range.** Fix #1 makes the RO initial notification a real (ambient) wake, so a static/reduced field may legitimately render one *extra* ambient frame on connect before stopping — "exactly 1" is no longer correct; "≥1 and ≤2, then stops" is.

**Out of scope (pre-existing) — filed as follow-ups:** #2232 (mode-switch link failure renders stale program), #2233 (context-lost-with-no-restore unlogged), #2234 (disconnect→reconnect CSS-degrade via reused force-lost context).

**Gates:** `verify` 7/7, `lint:ci`, `deadcode`(+prod), `typecheck`, root `test` (15683 pass), webcomponents suite (2062) + coverage (94.22%), both builds, `bundle-size`, `lint:docs` — all green. `package-lock.json` left uncommitted. Known non-blocker: 4 local `check-skill-router-sync` failures are a macOS/Node-24 quirk, green on CI Linux.

PR #2253, opened 2026-08-21
perf(webapp): cut the JS payload 26% by de-duplicating build graphs

trieloff:
## Summary

Cuts the webapp's total emitted JS from **33,022,925 to 24,304,182 bytes (−26.4%)** with no feature removed. The savings are pure de-duplication: the page graph, the kernel-worker graph, and each standalone esbuild bundle are separate builds that share nothing, so a handful of import edges were shipping the same megabytes two and three times.

`size-limit`'s total-JS-payload budget is ratcheted **34 MB → 25 MB** to lock it in.

## Why

Lazy `import()` bounds what a user *downloads* at boot; it does not bound what a build *ships*, and the payload budget counts every emitted file. Three edges were doing the damage:

| Waste | Cause |
| --- | --- |
| **5.8 MB** `dist/ui/slicc-diff.js` | an esbuild IIFE — esbuild inlines dynamic imports where Rollup code-splits them, so the curated Shiki grammar set (5.1 MB) landed eagerly in one file, a second copy of chunks the app already ships |
| **~1.8 MB** in the kernel worker | `kokoro-js` + `@huggingface/transformers`, unreachable in a worker (speech needs Web Audio) |
| **0.5 MB** `dist/ui/slicc-editor.js` | same IIFE problem, CM6 this time |
| **0.35 MB** in the kernel worker | xterm, reached through a shell class whose view layer the worker never mounts |

The `~1.8 MB` had a particularly cheap origin: `ensure-speech-assets.ts` (worker-side) imported `KOKORO_MODEL_ID` — a *string* — from `kokoro-engine.ts`.

- **`<slicc-diff>` / `<slicc-editor>` become Rollup entries** (`build.rollupOptions.input`). `dist/ui/slicc-diff.js` and `slicc-editor.js` are now stable-name **loader shims**: a classic script (so the `<script src>` consumers in `sprinkle-renderer.ts` are untouched) that dynamic-imports the hashed ES entry, resolved from the shim's own `src` so it works inside a srcdoc iframe. Generated in `closeBundle` from `.vite/manifest.json`; dev serves the identical shape so load ordering matches. The entries emit ~2 KB each.
- **`stubPageRealmSpeechPlugin`** replaces `speech/{speak,hear}.ts` in the **worker build only** (`worker.plugins`, which is not derived from `plugins`). `say`/`hear` keep their page and panel-RPC branches in one module; the page branch is gated on `speechSynthesis`, i.e. dead in a DedicatedWorker. The stub throws rather than no-ops, so a realm-check regression is loud. Stub chunk: **501 bytes**.
- **`speech/model-ids.ts`** holds the two model ids and stays dependency-free.
- **`scoop-context` constructs `AlmostBashShellHeadless`** — the follow-up the `AlmostBashShell` class comment already proposed. That orphaned the view layer entirely (knip's production-files gate caught it, cf. #1018), so the second commit deletes `shell/almost-bash-shell.ts` + `shell/index.ts` and retargets the tests; `kernel/remote-terminal-view.ts` has been the real terminal view for a while.

**Behavior change:** the element bundles load asynchronously now. Both elements adopt pre-upgrade properties (`ui/upgrade-own-properties.ts`), so the documented `el.patch = …` / `el.value = …` sprinkle patterns still work — that race already existed on the inline-sprinkle path, which appended the script tag asynchronously. A sprinkle that calls a **method** synchronously right after the script tag must now await `window.__SLICC_SPRINKLE_ASSETS__['slicc-editor.js']`; the style guide documents it. `lucide-icons.js` is deliberately left as an IIFE because `LucideIcons.createElement()` is a synchronous API sprinkles call inline.

## Cross-cutting impact

- **Floats exercised**: CLI/standalone harness (live), hosted leader tab, kernel worker. The extension build is unaffected in shape — it loads UI from the hosted origin — and `npm run build -w @slicc/chrome-extension` passes.
- **Speech parity**: the worker stub is only reachable if a realm check regresses; `packages/webapp/tests/speech/worker-realm-boundary.test.ts` guards both edges (no engine import from `ensure-speech-assets`, plugin wired to `worker.plugins` only, stub covers every runtime export of the real modules).
- **Sprinkles**: `sprinkle-renderer.ts` is unchanged — same `<script src="/slicc-diff.js">` contract, same stable URLs.
- **Bundle size**: quantified above; the size-limit budget is lowered, not raised.
- **Persisted state**: none touched.

## Test plan

- [x] `npx prettier --check` on the diff — clean
- [x] `npm run verify` — all 7 gates (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod)
- [x] `npx vitest run --project webapp` — **12,538 passing**, 42 skipped
- [x] `npm run test:coverage:webapp` — floors met
- [x] `npm run build` (JS workspaces) and `npm run build -w @slicc/chrome-extension`
- [x] `npm run size -w @slicc/webapp` — main entry 22.92 kB / 24 kB, kernel worker 1.89 MB / 1.91 MB, total **24.3 MB / 25 MB**
- [x] New behavior covered: `tests/ui/upgrade-own-properties.test.ts` (pre-upgrade property adoption) and `tests/speech/worker-realm-boundary.test.ts` (realm edges)
- [x] **Manual smoke (real Chrome, standalone harness on an isolated bridge port)** — sprinkle bundles: `<slicc-diff>` and `<slicc-editor>` define through the shim, pre-upgrade `.patch` / `.value` are adopted, diff renders in shadow DOM, `__SLICC_DIFFS__` / `__SLICC_CM6__` present; verified against both `dist/ui` and the dev server.
- [x] **Manual smoke (speech, same harness, real weights)**:
  - `say --list` → Web Speech voices, then kokoro voices lead once warm
  - `say -l en-US "…"` → audible, exit 0
  - `say --warmup` → `ensure-speech-assets` staged onnxruntime-web + espeak-ng and **both** HF repos (`whisper-tiny`, `Kokoro-82M-v1.0-ONNX`)
  - **TTS** `say -l en-US -o /tmp/tts.wav "the quick brown fox jumps over the lazy dog"` → valid 158,444-byte RIFF WAV
  - **STT round trip** `hear -i /tmp/tts.wav -l en` → "The quick brown fox jumps over the lazy dog."
  - **STT** `hear -i` on the whisper.cpp `jfk.wav` sample → exact transcript
  - `say -l es-ES "…"` → exit 0 (non-English kokoro via espeak); `say -l es-ES -o …` still correctly refused as English-only
  - console watcher: **0 errors** for the whole session
- [ ] N/A — no visual UI change, so no screencast

**Pre-existing failures** (unrelated, environment-only in this sandbox): `playwright-command.test.ts > iframe integration` ("Chrome exited with code 21 before reporting CDP port") and `check-skill-router-sync.test.mjs` (`mktemp` denied). Both fail identically without this change.

## Documentation

- [x] `packages/webapp/CLAUDE.md` — two new never-rules (sprinkle bundles ride the app graph; the worker stubs the page-realm speech modules)
- [x] `docs/pitfalls.md` — new section "Bundle Graphs: The Page and the Kernel Worker Are Built Separately"
- [x] `docs/architecture.md` — shell rows point at the headless module and `remote-terminal-view.ts`
- [x] `packages/vfs-root/workspace/skills/sprinkles/style-guide.md` — async element bundles + `__SLICC_SPRINKLE_ASSETS__`

## Risk & rollback

- **Blast radius**: sprinkle-rendered custom elements (both floats) and the worker's `say`/`hear` bridging. Both smoke-tested live.
- **Watch for**: a sprinkle in the wild calling `slicc-editor` / `slicc-diff` *methods* synchronously at parse time — properties are handled, methods need the new promise.
- **Rollback**: revert cleanly; no persisted-state or data-shape migration.
- **CI can't catch**: the real-browser speech round trip above (the repo's own `RUN_REAL_SPEECH_E2E` spec is opt-in) — done by hand, results in the test plan.


trieloff:
### CI triage + review response

Both failures on the first run were investigated rather than re-run blindly; neither came from this change, and both are now addressed.

**1. `node-matrix-tests (24)` — flake, confirmed.** `playwright-command.test.ts > iframe integration` died at `Timed out waiting for Chrome CDP port (60000ms)`. That suite launches its own Chrome through `chrome-launch.ts` and imports nothing this PR touches (it does not reference the deleted `shell/index.ts` barrel); 15,793 tests passed alongside it. **Green on re-run, same commit.**

**2. `e2e` — a latent bug in the speech spec's own recovery path, now fixed here.** The 92 MB `model_quantized.onnx` write tripped the transient this spec's header documents (`@zenfs/dom` + kerium / Chromium mojo pipe): 3 in-place retries on **both** attempts, then the spec's reboot recovery, then a 20s timeout on

```
Expected substring: "Welcome to SLICC"
Received string:    "New messages"
```

That assertion cannot hold on a re-boot, by product design. `detectWelcomeFirstRun` (`scoops/welcome-detection.ts`) returns `isFirstRun: false` once the cone's history already holds a welcome lick — the first boot in that same test posted one — and returns `false` outright when the marker read throws, which is precisely what a wedged OPFS session does. The Playwright page snapshot confirms it: full shell, composer and rails rendered, chat thread containing nothing but the "New messages" follow chip. So the recovery added in 9e1e5cd4 has never been able to get past its own re-boot; it simply had not run before, because this spec only executes when the `speech` path filter matches.

`bootLeader({ firstRun })` now gates the welcome assertion to the cold boot. The `__slicc_terminal_view` wait already in that function is the readiness signal that holds for both, since `mountWorkbenchTerminal` publishes it only once the kernel session is live. **Green on re-run before this commit; the fix keeps the documented recovery working the next time the transient fires.**

### Review response

Thanks for the deterministic checks — the layer-baseline and deletion-fallout verification in particular.

On the nit (25 MB total-JS budget leaves ~0.7 MB headroom): keeping it tight, deliberately. Same reasoning as the kernel-worker budget — a ratchet that trips loudly on the next eager-bytes regression is the point, and raising it later is a one-line, reviewable decision. If it starts tripping on unrelated dependency bumps we can revisit.

Codex and Copilot both returned quota-limit messages, so neither produced review content.

PR #2268, opened 2026-08-21
fix(memory): raise curator run ceiling and make a killed pass fail safe

### LINKED ISSUE #2263: Memory curator: default timeoutSeconds is too low, and a killed pass leaves the memory over budget
## Context

This installation has run the memory curator on every session since 2026-08-09 ⸺ 21 passes, 18 of them in the last five days. Along the way `/shared/MEMORY.md` drifted from the shipped `packages/vfs-root/shared/MEMORY.md`. The drift is small and every hunk of it was paid for by a failed pass, so it is worth folding back.

Two of the hunks are already filed as #2164 (scratch directory, «shell tools only», the `awk` per-section recipe, and the stale «no write access outside {{MEMORY_PATH}}» sentence). That issue has a human decision on it («widen the grant, ideally in /tmp/») and is still open. This issue covers **the rest of the drift plus one change the drift does not yet contain**, and adds five more days of evidence for #2164 at the bottom.

Full local-vs-shipped diff, for reference:

<details>
<summary><code>diff -u packages/vfs-root/shared/MEMORY.md@main /shared/MEMORY.md</code> (this installation)</summary>

```diff
@@ -1,6 +1,8 @@
 ---
 writablePaths:
   - /workspace/CLAUDE.md
+  - /shared/loose-ends.json
+  - /shared/memory-curator/ # scratch space for draft-then-measure; ...
 visiblePaths:
@@ -37,7 +39,7 @@
-timeoutSeconds: 600
+timeoutSeconds: 1200
 thinkingLevel: medium
@@ -45,7 +47,7 @@
-**Work fast: a full pass should finish in well under 5 minutes.** The run is hard-stopped at 10 minutes, and a kill mid-write can corrupt the memory file — so mine the three signals, rewrite the file, and stop, rather than exploring the archive exhaustively.
+**Work fast: a full pass should finish in well under 10 minutes.** The run is hard-stopped at 20 minutes, and a kill mid-write can corrupt the memory file or leave it over budget — so mine the three signals, rewrite the file, compact it in the same pass, and stop, rather than exploring the archive exhaustively.
```

The remaining hunks (`/shared/memory-curator/` grant, the `awk` recipe, the drafting paragraph, the two rewritten rules bullets, the customization comments) are exactly the diff proposed in #2164. `/shared/loose-ends.json` is a local grant for a local store and is **not** proposed for upstream.

</details>

## Change 1: the default timeout is too short, and 1200 s is not enough either

`timeoutSeconds: 600`. Raised here to `1200` on 2026-08-18 after passes started dying mid-work. Result across the 18 passes since 2026-08-17:

| date (UTC) | exit | turns | how the log ends |
| --- | --- | --- | --- |
| 08-17 16:54 | 1 | 34 | mid `node -e`, inner command `exit code: 124` |
| 08-18 07:29 | 1 | 19 | mid command, inner `exit code: 124` |
| 08-18 14:02 | 1 | 35 | mid work (the pass documented in #2164) |
| 08-18 15:20 | 1 | 21 | mid work |
| 08-18 17:31 | 1 | 21 | mid work |
| 08-19 18:14 | 1 | 47 | mid `sed`, output `sed: invalid command: -` |
| 08-20 17:47 | 1 | 6 | prose «Still ~2.7KB over — compacting further.» then nothing |
| 08-21 08:35 | 1 | 30 | mid `write_file`, no result recorded |

Twelve passes exited 0, eight exited 1 with a truncated log. Every truncation lands in the compaction phase, because compaction is last.

**What a truncated pass costs.** After the killed 08-21 08:35 pass the memory file was left at **24.404 bytes against a budget of 16.608** ⸺ 47% over. The next pass had to spend its own budget cleaning that up before it could mine anything (first command of the 11:24 pass: `wc -c /workspace/CLAUDE.md` → `24404`). So the failure is not «one pass produced nothing», it is «one pass produced a file that violates the invariant the prompt calls hard, and the following pass pays for it».

Raising the ceiling helped (the 08-18 cluster of three consecutive kills has not recurred) but did not fix it: 08-21 08:35 was killed under the 1200 s setting.

Proposed, and running here since 2026-08-18:

```diff
-timeoutSeconds: 600
+timeoutSeconds: 1200
```

```diff
-**Work fast: a full pass should finish in well under 5 minutes.** The run is hard-stopped at 10 minutes, and a kill mid-write can corrupt the memory file — so mine the three signals, rewrite the file, and stop, rather than exploring the archive exhaustively.
+**Work fast: a full pass should finish in well under 10 minutes.** The run is hard-stopped at 20 minutes, and a kill mid-write can corrupt the memory file or leave it over budget — so mine the three signals, rewrite the file, compact it in the same pass, and stop, rather than exploring the archive exhaustively.
```

The prose has to move with the number, otherwise the prompt tells the agent it has 10 minutes while the runtime gives it 20 ⸺ and an agent that believes it is nearly out of time takes shortcuts in the wrong place.

## Change 2 (new): make a killed pass fail safe

Change 1 buys headroom, it does not remove the failure mode. A wall-clock kill is always possible, so the prompt should order the work such that a kill at any point leaves a **valid** file rather than an over-budget one. Today it does the opposite: the agent mines, drafts, writes, and only then compacts, so the most interruptible phase is the one that enforces the budget.

Suggested addition to «Working within the budget», replacing steps 3 and 4:

```diff
-3. Write the complete file once.
-4. `wc -c {{MEMORY_PATH}}` to confirm. If you are still over, cut a whole section rather than shaving a few characters at a time.
+3. Write the complete file once, already inside the budget. Never write a version you know to be over budget «to fix in the next step»: the pass can be killed at any moment, and the file you leave behind is the one the next session inherits.
+4. `wc -c {{MEMORY_PATH}}` to confirm. If you are still over, cut a whole section rather than shaving a few characters at a time.
+
+If the pass is running long, drop the oldest-dated section wholesale and write. An under-budget file missing one stale section is a good outcome; an over-budget file is a broken one.
```

Cheap, and it converts «pass killed» from data loss into an ordinary lossy compaction.

## Evidence for #2164, five days on

The scratch grant proposed in #2164 has been live here since 2026-08-18 16:03. It is used on every pass, and no curator sudo request has been raised since. Two follow-ups for whoever implements it:

1. **`/tmp` works as a grant, but the prompt must name a fixed path.** The decision on #2164 was «ideally in /tmp/ as this will be cleaned out automatically». Checked just now with a throwaway scoop granted `writablePaths: ["/tmp/memory-curator/"]` and nothing else: `mkdir`, write, read and `rm` all succeed, and the file lands in the shared `/tmp` the cone sees, not in the scoop's private `/scoops/<name>/tmp`. So the location is viable. Two caveats: `mktemp` does not exist in this shell (`bash: mktemp: command not found`, both in my probe and in the 08-18 14:02 curator log, where the curator reached for it before falling back to `/workspace/.s.txt` and getting denied), so the prompt has to name the directory literally; and I could not confirm that `/tmp` is in fact cleared automatically, which is the property the decision rests on ⸺ worth verifying before the customization comment promises it.
2. **Scratch is never pruned, and drafts multiply.** `/shared/memory-curator/` here holds 15 files, among them `d2.md` through `d7.md` written minutes apart in the 08-19 pass ⸺ six successive full rewrites, which is precisely the trial-and-error convergence the budget section forbids. Worth one sentence in the prompt: reuse a single draft filename, and do not leave more than one behind.

## Separately: the legacy `Auto-extracted` block appeared again

The 24.404-byte file described above contained a `## Auto-extracted (consolidated)` section, the same artefact reported in the first comment on #2164 and supposedly guarded by #2002. I have not isolated whether the killed curator wrote it or the legacy extractor appended it after the kill, so this is a pointer rather than a bug report ⸺ but if the guard keys off «the curator completed», a killed pass is exactly the hole in it.

## Environment

SLICC on the `adobe` model realm, curator running `thinkingLevel: medium`, memory file `/workspace/CLAUDE.md`, budget 16.608–16.644 chars over the passes cited. Session logs `/sessions/agent-memory-curator-2026-08-*.md`.

trieloff:
Fixes #2263

The memory curator's prompt told the agent it had 10 minutes, the runtime hard-stopped the run at 600 s, and compaction ran *last* — so the most interruptible phase was the one enforcing the budget. A killed pass left the memory file over budget (24404 bytes against a 16608 budget after the 2026-08-21 08:35 kill) and the next pass paid to clean it up.

## Changes — `packages/vfs-root/shared/MEMORY.md`

1. **Raise the run ceiling.** `timeoutSeconds: 600` → `1200`, and the prose moves with the number (10 minutes soft, 20 minutes hard, "or leave it over budget", "compact it in the same pass"). Text verbatim from the issue. An agent that believes it is nearly out of time takes shortcuts in the wrong place.
2. **Make a killed pass fail safe.** «Working within the budget» step 3 now says to write a file that is *already* inside the budget and never one known to be over "to fix in the next step", plus a closing paragraph: when the pass runs long, drop the oldest-dated section wholesale and write. An under-budget file missing a stale section beats an over-budget one.
3. **`rg` on the `allowedCommands` list.** 32 invocations across the 21 analysed passes, and it was absent — while the list is read as documentation of what to reach for.
4. **Trim the never-used entries.** `date`, `diff`, `du`, `mount`, `nl`, `readlink`, `stat`, `touch`, `xxd` — zero uses in 417 curator bash calls. `od` stays for corrupt-store forensics. Removing an entry cannot remove a base command.

Out of scope, deliberately: no change to `writablePaths`/`visiblePaths` and no scratch grant — that is #2164 and it moves an authorization surface. `thinkingLevel` untouched.

## One extra change, outside `MEMORY.md`, flagged for a decision

`agentic-memory.ts` clamps `timeoutSeconds` with `MAX_MEMORY_TIMEOUT_SECONDS = 600`. Without touching it, change 1 is **inert**: `1200` in the frontmatter is pinned straight back to 600, and the prompt would promise 20 minutes while the runtime kills at 10 — the exact mismatch change 1 exists to remove. Verified against the existing test, which asserted `maxWallClockMs: 600_000` from the shipped prompt and still passed after the frontmatter bump.

So this PR also raises `MAX_MEMORY_TIMEOUT_SECONDS` to `1200`. `DEFAULT_MEMORY_TIMEOUT_SECONDS` (the value used when frontmatter omits the key) stays at 600. Worth noting for #2263's evidence: an installation running `timeoutSeconds: 1200` locally was still being killed at 600 s, which is consistent with "raising it helped but did not fix it".

## Tests

`packages/webapp/tests/scoops/agentic-memory.test.ts` — three assertions pinned the old numbers and were updated: the shipped-prompt `maxWallClockMs` (`600_000` → `1_200_000`), the soft-budget prose ("well under 5 minutes" → "10 minutes"), and the clamp test (600 → 1200 second maximum). `agentic-memory-default.test.ts` needed no change. The trim needed no test change; the `rg` addition needed none either, since frontmatter commands are unioned onto the base set.

```
$ npx vitest run packages/webapp/tests/scoops packages/webapp/tests/ui/session-freezer.test.ts
 Test Files  96 passed (96)
      Tests  1767 passed (1767)

$ npm run lint            # clean
$ npm run typecheck       # clean
$ node packages/dev-tools/tools/check-touched-exemptions.mjs
check-touched-exemptions: OK (3 changed file(s), 0 still on any debt list)
```

Draft on purpose: the `MAX_MEMORY_TIMEOUT_SECONDS` bump is a runtime-ceiling change and is yours to accept or split out.

### DIFF
--- a/packages/vfs-root/shared/MEMORY.md
+++ b/packages/vfs-root/shared/MEMORY.md
@@ -10,9 +10,6 @@ allowedCommands:
   - cat
   - cp
   - cut
-  - date
-  - diff
-  - du
   - echo
   - file
   - find
@@ -21,23 +18,18 @@ allowedCommands:
   - jq # structured reads of JSON stores such as /shared/loose-ends.json
   - ls
   - mkdir
-  - mount # listing mount state; mutations blocked by the write grant, not this list
   - mv
-  - nl
   - od # read-only byte inspection of corrupted stores
   - printf
-  - readlink
+  - rg # recursive, .gitignore-aware search; what the curator actually uses over find|xargs grep
   - sed
   - sort
-  - stat
   - tail
-  - touch
   - tr
   - uniq
   - upskill
   - wc
-  - xxd
-timeoutSeconds: 600
+timeoutSeconds: 1200
 thinkingLevel: medium
 ---
 
@@ -45,7 +37,7 @@ thinkingLevel: medium
 
 Curate the durable memory for session {{SESSION_COUNT}}. Today's date is {{TODAY}}.
 
-**Work fast: a full pass should finish in well under 5 minutes.** The run is hard-stopped at 10 minutes, and a kill mid-write can corrupt the memory file — so mine the three signals, rewrite the file, and stop, rather than exploring the archive exhaustively.
+**Work fast: a full pass should finish in well under 10 minutes.** The run is hard-stopped at 20 minutes, and a kill mid-write can corrupt the memory file or leave it over budget — so mine the three signals, rewrite the file, compact it in the same pass, and stop, rather than exploring the archive exhaustively.
 
 Read the entire current memory at {{MEMORY_PATH}} if it exists. Then mine the archived session at {{SESSION_ARCHIVE_PATH}} using the reading recipes below. Rewrite the entire {{MEMORY_PATH}} file with the durable information worth carrying into future sessions. Every part of the file is editable and counts toward the budget.
 
@@ -104,8 +96,10 @@ Measure, decide, then write once. Do not converge on the budget by trial and err
 
 1. `wc -c {{MEMORY_PATH}}` to get the current size, and subtract {{BUDGET_CHARS}} to get the exact surplus.
 2. Decide up front which sections absorb that surplus, oldest-dated first, and roughly what each one costs. Budget the whole cut before editing anything.
-3. Write the complete file once.
+3. Write the complete file once, already inside the budget. Never write a version you know to be over budget «to fix in the next step»: the pass can be killed at any moment, and the file you leave behind is the one the next session inherits.
 4. `wc -c {{MEMORY_PATH}}` to confirm. If you are still over, cut a whole section rather than shaving a few characters at a time.
+
+If the pass is running long, drop the oldest-dated section wholesale and write. An under-budget file missing one stale section is a good outcome; an over-budget file is a broken one.
 
 Rules:
 
--- a/packages/webapp/src/scoops/agentic-memory.ts
+++ b/packages/webapp/src/scoops/agentic-memory.ts
@@ -16,7 +16,7 @@ const log = createLogger('agentic-memory');
 
 export const MEMORY_INSTRUCTIONS_PATH = '/shared/MEMORY.md';
 export const DEFAULT_MEMORY_TIMEOUT_SECONDS = 600;
-export const MAX_MEMORY_TIMEOUT_SECONDS = 600;
+export const MAX_MEMORY_TIMEOUT_SECONDS = 1200;
 
 /**
  * Exactly the memory file, not `/workspace/`. The curator is given `upskill` so
--- a/packages/webapp/src/scoops/agentic-memory.ts
+++ b/packages/webapp/src/scoops/agentic-memory.ts
@@ -190,7 +190,7 @@ export async function runAgenticMemoryPass(
     }
     if (outcome.result.exitCode !== 0) {
       // A name-in-use rejection means a PRIOR curator is still running and holds
-      // the fixed `memory-curator` name (its window is now up to 10 min). Unlike
+      // the fixed `memory-curator` name (its window is now up to 20 min). Unlike
       // a curator that spawned and failed, no run has released — the legacy
       // append is NOT safe: the running namesake read the memory file before
       // this session's append and its whole-file rewrite would clobber it. Defer
--- a/packages/webapp/tests/scoops/agentic-memory.test.ts
+++ b/packages/webapp/tests/scoops/agentic-memory.test.ts
@@ -124,9 +124,10 @@ Memory={{MEMORY_PATH}} archive={{SESSION_ARCHIVE_PATH}} count={{SESSION_COUNT}} 
     expect(spawn.mock.calls[0][0]).toMatchObject({
       persistSession: true,
       name: 'memory-curator',
-      // Shipped MEMORY.md sets timeoutSeconds: 600 → a 10-minute wall-clock
-      // ceiling, generous enough that a slow pass is not killed mid-write.
-      maxWallClockMs: 600_000,
+      // Shipped MEMORY.md sets timeoutSeconds: 1200 → a 20-minute wall-clock
+      // ceiling, generous enough that a slow pass is not killed mid-write
+      // and left over budget (#2263).
+      maxWallClockMs: 1_200_000,
     });
   });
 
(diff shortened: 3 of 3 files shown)


trieloff:
Addressed the review's one finding in `ef81ed3`: `agentic-memory.ts:193` said "its window is now up to 10 min" — the last place still naming the old 600 s window — now 20 min. Swept the rest of the package, `docs/webapp-details.md` and the two `CLAUDE.md` files for the same string; that comment was the only other occurrence.

Agreed on keeping `MAX_MEMORY_TIMEOUT_SECONDS` in this PR rather than splitting it — the frontmatter bump is inert without it.

`agentic-memory.test.ts` (26 tests) and `npm run typecheck` green after the change.

Issue #2255, opened 2026-08-21
Shell commands silently ignore unknown flags and exit 0 (sweep beyond `sprinkle`)


trieloff:
Split out from #2166 (fixed there for `sprinkle` only, PR #2254).

## The pattern

A verb-dispatching shell command that scans for its own flags with `args.indexOf('--name')` / `args.includes('--json')` and treats everything else as positional will **silently ignore** any flag it does not recognise, and exit 0. That is worse than not supporting the flag: a probe that exits 0 is indistinguishable from one that honoured the flag, so it produces false confidence rather than an error. In #2166 that behaviour cost about an hour of misdiagnosis, because `sprinkle list --runtime=TOTALLY-FAKE-ID` and `sprinkle list --runtime=<real follower>` printed identical output and both exited 0.

`ssh-command.ts` shows the shape of the fix (`ssh: unknown flag: --x`, exit 1); `sprinkle-command.ts` now has a per-subcommand version of the same walk that also accepts flags in any position and honours a `--` terminator.

## Scope

A grep for `unknown flag` across `packages/webapp/src/shell/supplemental-commands/*-command.ts` shows roughly 20 of 70 commands reject unknown flags today. The rest split into two groups, and only the first is in scope:

**In scope — commands with a fixed, small flag vocabulary of their own.** Confirmed by inspection to use the silent-swallow scanning pattern:

- `webhook-command.ts`, `crontask-command.ts` — `args.indexOf('--name' | '--filter' | '--scoop' | '--cron')`
- `models-command.ts`, `plugin-command.ts`, `theme-command.ts` — `args.includes('--json' | '--all' | '--help')`

Worth auditing in the same sweep (same class, not yet individually confirmed): `open`, `mcp`, `secret`, `discover`, `fswatch`, `upgrade`, `oauth-token`, `oauth-domain`, `imgcat`, `screencapture`, `cherry-emit`, `slicc-fs-cleanup`, `di`, `cost`, `uname`, `which`, `kill`, `ps`, `nuke`, `playwright-cli`.

[...]

PR #2269, opened 2026-08-21
feat(shell): add mktemp, ported from just-bash#377

### LINKED ISSUE #2267: Shell: no `mktemp`, so agents improvise scratch paths they do not own
## What is missing

`mktemp` does not exist in the shell. Neither a `just-bash` builtin nor a supplemental command:

```
$ command -v mktemp
$ mktemp -p /tmp/memory-curator
bash: mktemp: command not found
```

`xxd`, `unlink`, `cmp`, `df` and `which` all made it into `packages/webapp/src/shell/supplemental-commands/`; the one command whose whole job is «give me a path I am allowed to write» did not.

## Why it is worth a supplemental command

Agents reach for `mktemp` reflexively, and its absence does not fail cleanly ⸺ it cascades. From the memory curator's 2026-08-18 14:02 pass, three consecutive turns:

```
S=$(mktemp)                       →  bash: mktemp: command not found
sed '…' $F > $S                   →  bash: : Is a directory     # $S empty, redirect hits the cwd
sed '…' $F > /workspace/.s.txt    →  EACCES: sudo: approval denied '/workspace/.s.txt'
```

Command substitution swallows the first failure, so the agent does not learn that `mktemp` is missing ⸺ it learns that «`>` into a variable is a directory», concludes its scratch strategy is wrong, and improvises a path it does not own. The third line is a cone-mediated sudo prompt, which is the subject of #2164: an agent with no temp path of its own escalates into somebody else's tree. That pass was time-boxed at ten minutes and spent three turns on this.

The same shape shows up whenever a scoop needs an intermediate file. `/tmp` is not a safe guess: verified with throwaway scoops today, a `/tmp` write is denied both with `writablePaths: []` and with an unrelated single-file grant (`EACCES: sudo: approval denied '/tmp/filegrant-probe.js'`), while an explicit `["/tmp/memory-curator/"]` grant works.

## Requested behaviour

Small surface, GNU-compatible where it costs nothing:

| form | result |
| --- | --- |
| `mktemp` | create a unique empty file, print its absolute path |
| `mktemp -d` | same for a directory |
| `mktemp -p DIR` / `--tmpdir=DIR` | place it under `DIR` |
| `mktemp TEMPLATE` (`…XXXXXX`) | honour the template, ≥3 trailing `X` |
| `mktemp -u` | print a name without creating it |

The part that actually matters for this runtime is **which directory is the default**, and the answer cannot be a hardcoded `/tmp`:

- For the cone, `/tmp` is writable and fine.
- For a scoop, `/tmp` is out of grant unless someone granted it, so a `mktemp` that returns `/tmp/tmp.abc123` hands back a path the caller cannot write. That is worse than the current error, because the failure moves one step later, into whatever wrote to the path.

Proposal: resolve the default as `$TMPDIR` when set, otherwise the caller's own temp root ⸺ `/tmp` for the cone, `/scoops/<folder>/tmp` for a scoop, which already exists in the scoop tree and falls inside the default `writablePaths` of `/scoops/<folder>/`. And set `TMPDIR` in the scoop's environment so the convention is discoverable: `echo $TMPDIR` prints empty today, in cone and scoop alike.

Two smaller asks:

- Exit non-zero on an unrecognized flag, per #2255. A `mktemp` that ignores `--suffix` and prints a path anyway would be a fresh source of silent breakage.
- Mention it in `man`/`commands` output so an agent can find it without guessing.

## Implementation pointers

- `packages/webapp/src/shell/supplemental-commands/mktemp-command.ts`, modelled on `unlink-command.ts` / `cmp-command.ts` (both are small, FS-only, `defineCommand` from `just-bash`).
- Register in `packages/webapp/src/shell/supplemental-commands/index.ts`, next to `createUnlinkCommand()`.
- Test at `packages/webapp/tests/shell/supplemental-commands/mktemp-command.test.ts`: unique names across calls, `-d`, template expansion, `-p`, default resolution for cone vs scoop, unknown flag exits non-zero.

## Related

- #2164 ⸺ memory curator has no writable scratch directory; the `mktemp` failure above is its first domino.
- #2263 ⸺ curator prompt changes, including the finding that `allowedCommands` can neither deny a base command nor supply a missing one. Listing `mktemp` there would have done nothing; the command has to exist first.
- #2255 ⸺ shell commands silently ignoring unknown flags.

trieloff:
Fixes #2267. Takes over the original draft, which had gone stale.

## What changed since the draft

The draft did two things. **The `$TMPDIR` half has since shipped on main** via the per-cone TMPDIR work (#2566 / #2574 / #2596), in a better form than this branch had — `tmpDirFor()` resolves against the live roster, where this branch derived the path from `$HOME`. The rebase drops all of it; `scoop-context.ts`, `almost-bash-shell-headless.ts` and `scoop-shell-env.test.ts` are untouched by this PR now.

What is left is the half that never landed: **there is still no `mktemp` in any runtime.** `docs/shell-reference.md` on main already tells agents "`mktemp` resolves against `$TMPDIR`" — a promise nothing could keep. The failure does not surface where it happens:

1. `S=$(mktemp)` reports `command not found`, but command substitution swallows the diagnostic
2. `$S` expands to the empty string
3. `sed '…' $F > $S` redirects to the empty path and reports `bash: : Is a directory`

## Pulling in the just-bash PR instead of waiting

The generic POSIX command belongs upstream, and it is written: **[vercel-labs/just-bash#377](https://github.com/vercel-labs/just-bash/pull/377)**, CI green, still unmerged. Rather than keep waiting, this PR **ports that implementation** onto the supplemental-command surface — not a second design. Same flags, same GNU parse order, same diagnostics, same `--version` string, so deleting the overlay when the builtin lands changes nothing a caller sees.

| | this overlay | upstream #377 |
| --- | --- | --- |
| `-d`, `-u`, `-q`, `-t`, `-p DIR`, `--tmpdir[=DIR]`, `--suffix=SUFF`, `TEMPLATE`, `-h`/`--help`, `--version` | yes | yes |
| unknown option | exit 1, never ignored (#2255) | exit 1 |
| X-run anywhere (`fooXXXXbar`) | yes | yes |
| random source | `crypto.getRandomValues`, rejection-sampled over `[0-9A-Za-z]` | same |
| missing destination dir | rejected up front | rejected up front |
| `-u` probe | `lstat` (a dangling symlink counts as taken) | same |
| collision detection | errno, not message body | same |
| modes | 0600 file / 0700 dir | same |
| default directory | `scratchDir()` — the per-unit `$TMPDIR` pin | `$TMPDIR`, else `/tmp` |
| creation | probe-then-create | atomic `IFileSystem.createExclusive` |

The last two rows are the only deliberate differences, and both are in the file header:

- **`scratchDir()`** is this repo's `$TMPDIR` rule with its blank-value handling, so a unit gets a path it may actually write instead of a `/tmp` it discovers is unwritable later. Upstream reads `$TMPDIR` too, so the two agree.
- **Creation is weaker.** Upstream adds `IFileSystem.createExclusive` in the same PR; the pinned `just-bash` (3.4.2) does not expose it, so the sequence here is `lstat` → create → `chmod`. Between the probe and the write another writer can take the name, and the entry is briefly readable at the backend default mode. The window is narrow in this runtime — one kernel worker, no other OS processes on the VFS — but it is real, and it is the strongest single reason to delete this overlay. A failing `chmod` takes the entry back rather than returning a path the caller would treat as private.

## The tripwire

`packages/webapp/tests/shell/supplemental-commands/mktemp-builtin-tripwire.test.ts`.

A supplemental command **shadows** a builtin of the same name. So when `just-bash` ships `mktemp`, nothing breaks — the overlay keeps winning, every test in this PR keeps passing, and the weaker probe-then-create keeps running in place of upstream's atomic one. The failure mode is silence, which is what the tripwire is for.

It boots a bare `Bash` with no supplemental commands and asserts `mktemp` is *absent*: `exec('mktemp')` exits 127, and `command -v mktemp` exits 1 with empty stdout. A third case asserts the overlay is what answers today, so a green tripwire can never be misread as "mktemp works now".

The file header carries the removal steps — `git rm` the overlay, its test, the registration in `index.ts`, the `docs/shell-reference.md` row, and the tripwire itself — and says explicitly to delete rather than adjust the test.

**Verified non-vacuous rather than assumed:** registering a stand-in `mktemp` as a custom command flips both assertions (`127 → 0`, and `command -v` `1 → 0` printing `/usr/bin/mktemp`). A green tripwire I had not seen go red would not be worth committing.

## Registration is a stub

`mktemp/run.ts` behind a `defineCommand` stub in `mktemp-command.ts`, lazily imported on first use. `supplemental-commands/index.ts` is in the kernel worker's boot-critical graph, and the first-load gate caught the static import:

```
FAIL: worker graph: this change adds 6.0 kB to the eager first-load closure
      (4292 kB -> 4298 kB), over the 5 kB per-change allowance.
```

After the split: `worker 4292 kB — +0.2 kB vs base`.

## Verification

On this branch, rebased onto `origin/main` (`c30d9efe0`):

```
npm run lint                → exit 0
npm run lint:ci             → exit 0
npm run lint:autofix-drift  → npm run lint is a no-op on this tree
npm run typecheck           → exit 0 (all 9 tsconfig projects)
npm run test                → 1166 files passed | 5 skipped
                              21350 tests passed | 61 skipped
npm run test:coverage:webapp
                            → exit 0; supplemental-commands/mktemp 98.64% stmts,
                              96.06% branches, 98.93% lines
node .../check-first-load-size.mjs
                            → First-load OK (page +0.0 kB, worker +0.2 kB)
node .../check-touched-exemptions.mjs
                            → no debt lists found — nothing to gate
npm run build               → exit 0
npm run build -w @slicc/chrome-extension
                            → exit 0
```

**55 tests** across the two files: default directory and `$TMPDIR` precedence, per-call uniqueness, modes 0600/0700, the chmod-failure take-back, all four `-p`/`--tmpdir` spellings plus a `-up DIR` cluster, `--suffix` in both spellings and its three rejections, `-t` including its `$TMPDIR`-beats-`-p` inversion, GNU meta-flag ordering (`--bad --help` vs `--help --bad` vs `-- --help` vs `-p --help` vs `--help=1`), `chartXXXX` not being read as a short cluster, missing and non-directory destinations, `-q`, collision retry, and an unrelated error whose *message* contains `EEXIST` not being retried as a collision.

Docs: a `mktemp` row in `docs/shell-reference.md`. The `$TMPDIR` section there needs no change — it already described this behaviour, and now the command exists to match it.


trieloff:
## Upstream cross-link

The `just-bash` side of this is now real, in its own bb thread (`thr_t6h48sqv4i`, project `just-bash`):

**https://github.com/trieloff/just-bash/pull/1** — *feat(mktemp): add GNU-compatible mktemp command* (draft, in the `trieloff/just-bash` fork; branch `mktemp-command`, 29 tests). A PR against `vercel-labs/just-bash` is being raised from that same branch — as of this comment there is no open `mktemp` PR on `vercel-labs/just-bash` yet, so treat the fork PR as the canonical link until that one exists.

### How the two halves fit

The upstream command resolves its default directory as **`$TMPDIR` when set and non-empty, else `/tmp`** — deliberately, so that "sandboxed embedders set `TMPDIR` to a path the caller may actually write". That embedder is us: the `TMPDIR` pinning in this PR (`/tmp` for the cone, `/scoops/<folder>/tmp` for a scoop) is exactly the input the upstream command expects, and it **stays after the shim is deleted**. Nothing about the sandbox-aware half is duplicated upstream.

### What changes when `just-bash` ships it

Delete `packages/webapp/src/shell/supplemental-commands/mktemp-command.ts`, its registration in that directory's `index.ts`, and `packages/webapp/tests/shell/supplemental-commands/mktemp-command.test.ts`. The swap only widens the surface — no caller that works against the shim breaks against the builtin:

| | shim (this PR) | upstream PR |
| --- | --- | --- |
| `-d`, `-u`, `-q`, `-p DIR`, `--tmpdir[=DIR]`, `TEMPLATE`, `-h`/`--help` | yes | yes |
| `-t`, `--suffix=SUFF`, `--version` | **exit 1**, unrecognized option | implemented |
| unknown flags | exit 1 (#2255) | exit 1 |
| default dir | `$TMPDIR` → `/scoops/<folder>/tmp` derived from pinned `$HOME` → `/tmp` | `$TMPDIR` → `/tmp` |
| missing destination dir | surfaced by the VFS error | rejected up front |
| modes | VFS default | 0600 file / 0700 dir |

The one shim-only behaviour is the `$HOME`-derived scoop fallback, which exists purely to cover a shell constructed outside `buildScoopShellEnv`. Once `TMPDIR` is pinned it is belt-and-braces, and it is correct for it to disappear with the shim.

trieloff:
## Correction to the swap table above

The [earlier cross-link comment](https://github.com/ai-ecoverse/slicc/pull/2269#issuecomment-5370107628) listed `-t`, `--suffix=SUFF` and `--version` as *shim: exit 1, unrecognized option* against *upstream: implemented*. That table is stale as of this takeover — the whole point of the rewrite is that the row no longer exists. All three are implemented here, ported from [just-bash#377](https://github.com/vercel-labs/just-bash/pull/377), along with X-run-anywhere templates, rejection-sampled `crypto` names, up-front missing-directory rejection, `lstat`-based `-u` probing, errno-based collision detection, and 0600/0700 modes.

The one shim-only behaviour that comment described — the `$HOME`-derived scoop fallback — is gone too, and for the reason it predicted: `$TMPDIR` is now pinned per unit on main (#2566 / #2574 / #2596), so the belt-and-braces branch has nothing left to cover. `scratchDir()` is the single resolver.

Upstream #377 is still open and still green. Nothing in this PR is a reason not to merge it; when it lands, `mktemp-builtin-tripwire.test.ts` goes red and carries the removal steps.

trieloff:
## Both findings fixed in d5e39a309

**P1 — mode 0600/0700 is fiction here.** Confirmed: `VfsAdapter.chmod()` is an explicit no-op (`packages/webapp/src/shell/vfs-adapter.ts:717`, "Our VFS doesn't track permissions"), and `/tmp` is granted to every scoop by design (`builtinScoopGrants()` + `ALWAYS_WRITABLE_PREFIXES`). So the entry is not private, and no amount of care in this command makes it so.

I did not try to synthesise an access boundary that the VFS has no model for. Instead the command stops claiming one. The help text now reads:

> What you get back is a UNIQUE name, not a private one. This VFS tracks no permission bits, so the usual 0600/0700 modes cannot be applied and the /tmp tree is readable by every unit. Do not put secrets in a mktemp file.

`docs/shell-reference.md` says the same. The `chmod` call stays, for two reasons spelled out at `createExclusive`: it is correct on any backend that does implement it, and a *throwing* chmod must still take the entry back rather than leave a path behind at a mode the caller would misread as private. This is now the one place the overlay's output deliberately differs from upstream's — when the builtin lands it will print GNU's 0600/0700 line, which will be just as untrue on this VFS, and that is upstream's text to fix.

New test covers the production path directly: a filesystem whose `chmod` is a no-op, asserting the command still returns a usable path. The previous suite only exercised a *throwing* chmod, which is the rarer case.

**P2 — `--directory=no` created a directory.** Correct, and it is exactly the #2255 failure: the stripped name granted the opposite of what the caller asked for. `applyLongFlag` now rejects a non-`undefined` attached value for `--directory` / `--dry-run` / `--quiet` / `--help` / `--version`, and — for the ordering half of your point — `metaFromLong` returns `'stop'` on such a token so the parser emits the diagnostic instead of help short-circuiting past it. `--directory=no --help` reports the option; `--help --directory=no` still prints help.

Five new cases, and I checked they are not vacuous: reverting both hunks turns 4 of them red (`--version=1` already passed, since the help/version branch handled it before).

Full pass re-run green: lint, typecheck, 21358 tests, coverage (`supplemental-commands/mktemp` 98.67% stmts / 96.23% branches), first-load (`worker +0.2 kB`), both builds.

PR #2282, opened 2026-08-21
feat(webapp): bash progress / ETA overlay

trieloff:
## Summary

Adds live progress / ETA feedback for `bash` tool calls, entirely on top of just-bash's public API (no patch — tracks [vercel-labs/just-bash#319](https://github.com/vercel-labs/just-bash/issues/319)). Before, a long `bash` call (a `sleep`, a `for` loop, a `curl` download) showed only a static "…" until it finished; now it shows one progress unit per tool call, folding every sub-step into a single "how long will this take" bar and ETA. Design doc: `docs/exploration/bash-progress-overlay.md`.

## Why

Agent workloads repeat long shell commands constantly and the UI gave no signal of how far along a call was. This is the "ship it today" overlay from the design doc; if upstream lands a first-class protocol later, `sleep`/`timeout`/wrapper pieces collapse onto it and `types.ts` is already a superset.

- **One unit per tool call** (`shell/progress/script-progress.ts`): before `bash.exec` the script AST (`bash.transform(script).ast` — the browser bundle exports no `parse`) is planned into registry-dispatch **steps** — `echo; echo; echo` = 3, static `for` loops expand, `timeout`/`env`/`nohup`/`nice` count their inner command, interpreter builtins count 0, `&&`/`||` count in full as an upper bound. Data-dependent shapes (`while`, globs, `$var` lists, `xargs`, `if`) make the total unknown → indeterminate. `fraction = (completedSteps + currentChildFraction) / totalSteps`.
- **Sub-steps fold in** via `ProgressEmitter.setAggregator`: `sleep`/`timeout` tickers, per-command start/end, and `curl` download bytes are routed into the active `ScriptRun` instead of the sink — users care about the tool call, not one sub-step. Steps tick at the dispatch wrapper **on completion** (just-bash's `trace` only fires for `find`, so it can't be the seam).
- **Determinate sources**: `sleep` via `BashOptions.sleep` (ticks inside just-bash's own timer allowance — defense-in-depth, no outer `setInterval`); `timeout DURATION cmd` against its limit; `curl -o`/`wget` byte progress streamed in `proxied-fetch.ts`, keyed to the node-server `X-Proxy-Content-Length` hint (identity encoding only).
- **Quiet row treatment** (reuses existing chrome, no new bar): the lucide icon chip fills bottom-up, the "…" badge becomes three dots (one per third, active dot blinking), an open terminal body gets a 3px top-border bar. % / ETA / `k of N` live in the row `title` + `aria-valuenow`.
- **Composer ring**: `WcChatController.onToolProgressChange` → the send button's existing `progress` attribute, so its tool-phase ring becomes a determinate arc (built for exactly this).
- **Cluster propagation**: when 3+ calls collapse into `<slicc-tool-cluster>` the rows hide, so the same treatment mirrors onto the visible cluster head (`aggregateClusterProgress`: mean fraction, indeterminate if any child is, longest child ETA), keeping its "N steps" count.
- New `tool_progress` agent event (`ToolProgressEvent` in `@slicc/shared-ts`) threaded scoop-context → lifecycle manager → orchestrator/facade → kernel message → agent-event-stream / offscreen-client → `WcChatController`. Named `ToolProgressEvent`, **not** `ProgressEvent`, to avoid the DOM global silently resolving in files that miss the import.

## Cross-cutting impact

- **Floats**: CLI + extension both exercised (proxied-fetch keeps all 4 realm branches; the emitter is a no-op when there's no `ToolExecutionContext`, i.e. the human terminal emits nothing).
- **Secrets**: progress **labels** are built from argv, so they run through the same `getToolResultScrubber()` path as tool results before leaving the emitter (a `curl -H "Authorization: …"` never reaches the card).
- **Tray/iOS**: `tool_progress` added to the tray-sync corpus with `ios: 'unknown'` (no Swift case yet — follower shows no bar), and the generated `tray-sync-corpus.json` regenerated.
- **Download memory**: the streaming reader preallocates a single buffer when the size is known, so it adds no extra full-body copy over the previous `arrayBuffer()` path.

## Test plan

- [x] `npx prettier --check .`
- [x] `npm run typecheck` (webapp + worker + node-server)
- [x] `npm run test` — 10135 passing, no new failures; new suites under `packages/webapp/tests/shell/progress/` (emitter, script-progress, sleep, timeout, wrap-command, fetch-progress, shell-wiring), `packages/webapp/tests/shell/proxied-fetch-progress.test.ts`, `packages/node-server/tests/fetch-proxy-content-length.test.ts`, and `packages/webapp/tests/ui/wc/wc-chat-controller-progress.test.ts`
- [x] `npm run test:coverage:webapp` — 81.3%, above floor
- [x] `npm run build -w @slicc/webapp` / `-w @slicc/node-server` / `-w @slicc/chrome-extension`
- [x] boy-scout debt gate (`check-touched-exemptions.mjs`) clean
- [x] **Manual smoke (CLI harness + Bedrock provider)** — screenshots below; console clean, prod ports untouched

**Screenshots** (live on the standalone dev harness):

Expanded tool row — the dark terminal body wears a 3px top-border progress bar; the three-dot badge and the composer's tool ring track the same fraction:

progress-terminal-bar

Collapsed row — icon chip fill + three dots + composer ring:

progress-row-dots

Tool cluster (3+ calls) — the aggregate treatment propagates onto the collapsed cluster head, keeping its "N steps" count:

progress-cluster

## Documentation

- [x] `docs/exploration/bash-progress-overlay.md` (design + implementation notes)
- [x] `packages/webapp/CLAUDE.md` (developer architecture line)

## Risk & rollback

- Blast radius: chat UI + shell progress only; the emitter no-ops without a tool context, so the human terminal and non-UI callers are unaffected. Revert is a clean single-branch revert (no persisted-state migration).
- Known limitation, out of scope here: very large `curl -o` (≥250 MB) still OOMs the renderer — a **pre-existing** browser-VFS memory ceiling (whole-body copies in the latin1 string / binary-cache / IndexedDB write), not introduced or worsened by this PR. Tracked in a separate thread.
- CI can't catch the real-Chrome rendering; verified by hand on the harness (screenshots above).

## Checklist

- [x] No secrets/tokens in the diff
- [x] Cross-float contract (`tool_progress`) checked through offscreen/facade/kernel/follower paths


trieloff:
Both review findings addressed in `1a9a639`, plus the branch relinearized.

**🟡 Major — truncation before scrub: fixed, good catch.** Confirmed the leak: the scrubbers replace *whole* known secret strings, so an 80-char cut through a token left the prefix unrecognised and it reached the card title. Truncation now happens in `ProgressEmitter.deliver()` **after** the scrub resolves (`capLabel`); `progressLabel`, `scriptLabel` and `fetchLabel` all return untruncated text. Regression test in `emitter.test.ts` builds a label whose secret deliberately starts before and ends after the cap, then asserts the scrubber was called with the *full* label and that no `sk-` fragment reaches the sink (it also pins that the label is still capped for display).

**🔵 Minor — swift-server parity: noted in code.** Kept as a deliberate degradation rather than a speculative Swift change: a swift-server-backed leader falls back to the indeterminate download bar. Added a comment at the hint site in `packages/node-server/src/routes/fetch-proxy.ts` pointing at `Sources/Server/APIRoutes.swift` so whoever wants determinate download progress there knows exactly what to mirror.

**Also:** #2284 (the `curl -o` OOM fix) was branched off this branch and merged back into it, which put a merge commit on the PR and broke the linear-history gate. Rebased so those two commits sit inline — the resulting tree is byte-identical to the merge commit (`git diff` empty), so nothing from that fix was lost.

PR #2293, opened 2026-08-21
fix(providers): bump pi-ai to 0.84.2 so Grok 4.6 reaches the model dropdown

trieloff:
## Why Grok 4.6 was missing

`packages/webapp/providers/xai-grok.ts` doesn't maintain its own model list — `config.getModelIds()` is a pass-through over pi-ai's bundled xAI catalog:

```ts
getModelIds: () => getNativeXaiModels().map(toModelMetadata),
// getNativeXaiModels() === getModels('xai')
```

So the dropdown can only ever show what the pinned `@earendil-works/pi-ai` ships in `dist/providers/data/xai.json`.

**Root cause:** `package.json` declared `^0.84.0`, but `package-lock.json` was pinned to exactly **0.84.0**, whose xAI catalog is:

| api | models |
| --- | --- |
| `openai-completions` | `grok-4.3`, `grok-build-0.1` |
| `openai-responses` | `grok-4.5` |

No `grok-4.6`. Confirmed against the **live running build** over CDP (port 9222) by fetching every loaded chunk and grepping the catalog data — the shipped bundle carried exactly those three ids.

pi-ai **0.84.2** adds it:

```json
{ "id": "grok-4.6", "name": "Grok 4.6", "api": "openai-completions",
  "provider": "xai", "reasoning": true, "input": ["text","image"],
  "contextWindow": 500000, "maxTokens": 500000 }
```

## The fix

- Bump `@earendil-works/pi-ai` to `^0.84.2` in `packages/webapp` + `packages/chrome-extension`, and refresh the lockfile.
- No provider code change needed — `grok-4.6` is `openai-completions`, which `resolveNativeXaiModel` / `streamXai` already dispatch (same path as `grok-4.3` and `grok-build-0.1`).
- Lockfile churn beyond the version bump is pi-ai's own dependency changes: it dropped the `@mistralai/mistralai` SDK (the Mistral *catalog* is still present, and slicc never used that SDK) and moved `openai` 6.26.0 → 6.40.0.

## Regression guard

`packages/webapp/tests/providers/xai-grok.test.ts` mocks `getModels`, so a stale pin is invisible to it. Added a test that asserts against the **real** pi-ai catalog via `vi.importActual`, so a lockfile that predates a Grok release fails in CI rather than silently in the UI.

## Verification

- `npm run lint` ✅ · `npm run typecheck` ✅ · `npm run test` (16 085 passed) ✅ · `npm run build` + extension build ✅ · `npm run bundle-size` ✅ (kernel worker still under budget) · `check-touched-exemptions` ✅
- Built `dist/` and re-grepped the catalog chunk: `grok-4.3 grok-4.5 grok-4.6 grok-build-0.1` — Grok 4.6 is now in the shipped catalog.
- `test:coverage:webapp` hit the known `playwright-command` iframe Chrome-launch flake locally; unrelated to this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1458,21 +1458,20 @@
       }
     },
     "node_modules/@earendil-works/pi-ai": {
-      "version": "0.84.0",
-      "resolved": "https://registry.npmjs.org/@earendil-works/pi-ai/-/pi-ai-0.84.0.tgz",
-      "integrity": "sha512-N9RDk8q0eglGiy+NqTZ3Ev2j+6oFNXSAJa8b0CYhvWB9HGiKZjsoCESXkUvMDLybrn0wXp75sdsoBzEtHxk9kA==",
+      "version": "0.84.2",
+      "resolved": "https://registry.npmjs.org/@earendil-works/pi-ai/-/pi-ai-0.84.2.tgz",
+      "integrity": "sha512-6MzsrYIYNVlE7SfpbL2yYb67Qo58p/7Q+xWG1RZvoX1P80aRCHSod2/13aFpxkow1lPO2LEh3c495J0Gwmyjig==",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/sdk": "0.91.1",
         "@aws-sdk/client-bedrock-runtime": "3.1048.0",
-        "@earendil-works/pi-telemetry": "^0.84.0",
+        "@earendil-works/pi-telemetry": "^0.84.2",
         "@google/genai": "1.52.0",
-        "@mistralai/mistralai": "2.2.6",
         "@opentelemetry/api": "1.9.0",
         "@smithy/node-http-handler": "4.7.3",
         "http-proxy-agent": "7.0.2",
         "https-proxy-agent": "7.0.6",
-        "openai": "6.26.0",
+        "openai": "6.40.0",
         "partial-json": "0.1.7",
         "typebox": "1.3.7"
       },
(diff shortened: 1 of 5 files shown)


trieloff:
What's up with renovate?

trieloff:
let's pin exact versions, so that we get minor version bumps from renovate, too and drop all the carets from package.jsons

trieloff:
### Rebased and fixed — the merge-queue failure was real, thanks for the bisect

Reproduced after rebasing onto current `main` (which now has #2283):

```
AssertionError: expected { …(35) } to deeply equal { …(36) }
packages/webapp/tests/shell/provider-env-seed.test.ts:60
```

pi-ai 0.84.2 adds exactly one entry to the `envMap` in `dist/env-api-keys.js`:

```
only upstream:    [ 'qwen-token-plan-individual => QWEN_TOKEN_PLAN_API_KEY' ]
only mirror:      []
value mismatches: []
```

`qwen-token-plan-individual` is a new provider id that reads the same `QWEN_TOKEN_PLAN_API_KEY` as `qwen-token-plan`. Since `PROVIDER_API_KEY_ENV` is meant to mirror pi-ai exactly (pi-ai does not export the table), the fix belongs in the source rather than the expectation — otherwise a user on that provider silently gets no key seeded into realm scripts. Added in 756b616.

Re-verified on the rebase: `npm run test` 16,198 passed / 0 failed, plus `lint`, `typecheck`, `lint:patches`, `check-touched-exemptions`, both builds and `bundle-size` clean. `grok-4.6` is present in the built catalog chunk.

Sorry for the batch poisoning on #2287 — that is a nasty failure mode, since none of it is visible from this PR's own checks.

PR #2306, opened 2026-08-22
fix(webapp): pair tool results by call id; fill the cluster head k-of-N

trieloff:
## Summary

A message's tool calls execute **concurrently**, but `tool_result` carried no tool-call id — so both the chat UI and the buffered transcript paired results by *"last unfinished call with this **name**"*. The moment two same-named calls overlapped, output landed on the wrong row. Before: three parallel `bash` calls rendered `echo BBB → CCC` and `echo CCC → BBB`. After: each row shows its own output.

With ids in hand, the tool cluster also becomes determinate — the whole batch is known before any call finishes — so the cluster head now fills as calls complete.

## Why

Found while investigating whether 3+ tool calls arrive in one turn (they do). Measured on the standalone harness with a real provider:

1. Prompt: *"Issue three PARALLEL bash tool calls in one turn: `sleep 5; echo AAA` / `BBB` / `CCC`"*
2. Observe the three rendered rows.

```
Expected            Actual (before)
echo AAA -> AAA     echo AAA -> AAA
echo BBB -> BBB     echo BBB -> CCC   *** crossed ***
echo CCC -> CCC     echo CCC -> BBB   *** crossed ***
```

**The agent's own view was always correct** — asked directly, it reported `ONE -> ONE`, `TWO -> TWO`, `THREE -> THREE`, because pi pairs by id internally. This was a display + persisted-transcript defect, not a data-integrity one. That is also why it went unnoticed: nothing downstream of the model was wrong.

Why calls overlap at all: `agent-loop.js` picks `executeToolCallsSequential` only when some tool declares `executionMode: 'sequential'`. SLICC declares it **nowhere**, so it takes `executeToolCallsParallel` → `await Promise.all(...)`. Measured: 3 × `sleep 8` completed in ~8 s, not 24 s.

- **Threads `toolCallId`** through `scoop-context` → `scoop-lifecycle-manager` → `orchestrator`/`facade` → kernel `AgentEventMsg` → `agent-event-stream` / `offscreen-client` → `WcChatController`, on `tool_use_start`, `tool_result` and `tool_progress`. The id is **optional on the wire** purely for back-compat with followers built before it existed.
- **Pairing** moves to `#findToolCall(message, toolName, toolCallId)`: match by id when present, else the old name scan. The fallback is kept deliberately — for a pre-id sender the ambiguity is unavoidable, and silently dropping those events would be worse than an occasional mispair.
- **`bufferToolEnd` matches by id too.** It had the identical name-based lookup, so *persisted transcripts* mis-attributed results as well — that defect outlived the session, and replaying a session reproduced it.
- **Rows adopt the provider id** instead of a local `uid()`, so progress and results address the same identity.
- **Cluster is now `(finished + partial) / N`** and never indeterminate: the batch size is known before any call finishes. A still-running call contributes its own fraction so the fill advances smoothly between completions instead of jumping a whole step. Returns to "no treatment" once every call settles.

### UI: the cluster gets no progress bar — by design

The row keeps its 3px top-border bar. A cluster is a *summary* of many calls, and a bar there reads as noise, so it advances only through its lucide **icon fill** and the **three dots**, keeping the existing "N steps" count. The CSS selector for the bar is now explicitly row-only, and a test pins that (asserts the cluster body has no `::before` rule while the row does) so it can't drift back.

## Cross-cutting impact

- **Floats**: the id rides the existing kernel/offscreen envelopes; extension and CLI both exercised. No new message types.
- **Follower parity**: `toolCallId` is inventoried in the tray-sync corpus as `dropped` for `tool_use_start` / `tool_result` / `tool_progress` — `SyncProtocol.swift` has no property for it yet, **so followers still name-pair and can still cross outputs on a parallel batch**. Adding the Swift property is the natural follow-up; the corpus keeps the gap visible rather than silent. Generated JSON regenerated.
- **Other callers**: every relay of `onToolStart` / `onToolEnd` / `onToolProgress` was updated; the id is appended as an optional trailing parameter so no positional call site changes meaning.
- **Side effect (good)**: previously all concurrent calls' progress collapsed onto one row (`prog:1` observed on the harness). Per-call progress under parallel batches now works.

## Test plan

- [x] `npm run typecheck` (webapp + node-server + worker)
- [x] `npm run test` — **14,013 passing**, no new failures
- [x] `npx prettier --check .` / `biome check .`
- [x] boy-scout debt gate, layer-back-edges, record-string-unknown, `lint:patches`
- [x] `npm run build -w @slicc/webapp` + `-w @slicc/chrome-extension`; bundle size 1.92 / 1.93 MB
- [x] **New tests**: a regression test that resolves the **middle** of three concurrent calls and asserts the *id* — not arrival order — decides which row receives it; and a cluster test covering k-of-N fill, the "no bar on cluster" CSS invariant, and clearing once all calls settle
- [x] **Manual smoke (CLI harness + real provider)**: parallel batch now renders `AAA→AAA`, `BBB→BBB`, `CCC→CCC`; cluster measured mid-flight at `title: "0/3 · 0 of 3 done — 30%"`, dots `● ○ ○`, computed cluster `::before` width `auto` (no bar), `3 steps` count intact

**Cluster mid-flight** — head at 30%, dots `● ○ ○`, lucide icon filling, and no bar; the per-call rows keep their own dots underneath:

cluster k-of-n fill

**Pre-existing failures**: none. (Three `zenfs-alloc-overflow` patch tests failed locally until `@zenfs/core` was reinstalled — a half-patched `node_modules`, not a code issue; CI installs clean.)

## Documentation

- [x] No documentation changes needed — no user-facing surface or agent-facing contract changed; the wire addition is documented inline on the type.

## Risk & rollback

- Blast radius: chat rendering + transcript attribution. Clean single-commit revert; no persisted-state migration (the id is additive and optional).
- Sessions persisted *before* this change keep their old (possibly mispaired) tool results — the fix is not retroactive.
- CI can't catch the real-provider parallel batch; verified by hand on the harness as described above.

## Checklist

- [x] No secrets or tokens in the diff
- [x] Cross-float contract change (`toolCallId`) checked through offscreen / facade / kernel / follower paths and inventoried in the corpus

### DIFF
--- a/packages/ios-app/SliccFollower/Tests/SliccFollowerTests/Fixtures/tray-sync-corpus.json
+++ b/packages/ios-app/SliccFollower/Tests/SliccFollowerTests/Fixtures/tray-sync-corpus.json
@@ -1024,12 +1024,14 @@
       "dropped": [
         "messageId",
         "progress",
+        "toolCallId",
         "toolName"
       ],
       "event": {
         "type": "tool_progress",
         "messageId": "m2",
         "toolName": "bash",
+        "toolCallId": "call-1",
         "progress": {
           "id": "sleep-1",
           "label": "sleep 30",
@@ -1052,13 +1054,16 @@
         "toolName",
         "type"
       ],
-      "dropped": [],
+      "dropped": [
+        "toolCallId"
+      ],
       "event": {
         "type": "tool_result",
         "messageId": "m2",
         "toolName": "read_file",
         "result": "file contents",
-        "isError": false
+        "isError": false,
+        "toolCallId": "call-1"
       }
     },
     {
@@ -1104,14 +1109,17 @@
         "toolName",
         "type"
       ],
-      "dropped": [],
+      "dropped": [
+        "toolCallId"
+      ],
       "event": {
         "type": "tool_use_start",
         "messageId": "m2",
         "toolName": "read_file",
         "toolInput": {
           "path": "/workspace/notes.md"
-        }
+        },
+        "toolCallId": "call-1"
       }
     },
     {
--- a/packages/shared-ts/src/agent-wire-types.ts
+++ b/packages/shared-ts/src/agent-wire-types.ts
@@ -88,11 +88,38 @@ export type AgentEvent =
       model?: string;
       usage?: ChatMessageUsage;
     }
-  | { type: 'tool_use_start'; messageId: string; toolName: string; toolInput: unknown }
-  | { type: 'tool_result'; messageId: string; toolName: string; result: string; isError?: boolean }
+  | {
+      type: 'tool_use_start';
+      messageId: string;
+      toolName: string;
+      toolInput: unknown;
+      /**
+       * Provider tool-call id. Optional only for back-compat with followers
+       * built before it existed; without it the UI must fall back to matching
+       * results by tool NAME, which mispairs concurrent same-named calls
+       * (the agent loop runs a message's tool calls with `Promise.all`).
+       */
+      toolCallId?: string;
+    }
+  | {
+      type: 'tool_result';
+      messageId: string;
+      toolName: string;
+      result: string;
+      isError?: boolean;
+      /** Provider tool-call id; see `tool_use_start`. */
+      toolCallId?: string;
+    }
   | { type: 'tool_ui'; messageId: string; toolName: string; requestId: string; html: string }
   | { type: 'tool_ui_done'; messageId: string; requestId: string }
-  | { type: 'tool_progress'; messageId: string; toolName: string; progress: ToolProgressEvent }
+  | {
+      type: 'tool_progress';
+      messageId: string;
+      toolName: string;
+      progress: ToolProgressEvent;
+      /** Provider tool-call id; see `tool_use_start`. */
+      toolCallId?: string;
+    }
   | { type: 'turn_end'; messageId: string }
   | { type: 'error'; error: string }
   | { type: 'screenshot'; base64: string; url?: string }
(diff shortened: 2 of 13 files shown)


trieloff:
Both P2 findings addressed in `3750a47` — thanks, the first one was a regression I introduced (the previous `uid()` key was collision-proof by construction).

**P2 — provider ids scoped per message.** Fixed as suggested: row keys are now `<messageId>:<toolCallId>`, so a reused provider id can no longer produce duplicate `data-tool-id`s or let a thread-wide `querySelector` paint a historical row. Identity-based pairing is unchanged. One addition beyond the suggestion: `#findToolCall` also accepts the *bare* id, because a transcript persisted by an older build stored it unscoped and would otherwise stop pairing on restore. Test: same id reused across `m1`/`m2`, asserting distinct row ids and that progress paints only the newer row.

**P2 — liveness from in-flight state, not the badge.** Fixed: `#sessionToolCalls` tracks the calls this session actually started (via `tool_use_start`); restored history never enters it, so an aborted turn's result-less rows can no longer pin a determinate 0% on a historical cluster. Entries deliberately outlive completion so a finished call still counts toward the batch total, and the set clears wherever the thread resets (`dispose`, `loadMessages`). Test: `loadMessages` with three result-less calls asserts the rebuilt cluster gets no treatment at all.

Full suite green at 14,016 (two provider-catalog tests failed locally beforehand purely because my `node_modules` had pi-ai 0.84.0 against the lockfile's 0.84.2 — environment drift, not the diff).

Issue #2278, opened 2026-08-21
feat(work-unit): generic supervisor APIs — createMany/join/detach/promote, nested delegation


trieloff:
Part of #1666 — RFC "Phase 8: expose generic parallel work-unit APIs".

## Goal
Orchestration built on `WorkUnitManager` rather than on cone-specific tools, with cone/scoop command names kept as product aliases.

## Scope
- `createMany` / `join` (wait) / `cancel` / `detach` + `promote` (child → independent root) / per-unit resource budgets (turns, tokens, processes, wall-clock — `ScoopConfig.maxTurns` / `maxWallClockMs` already exist).
- Nested delegation behind an explicit grant: `policy.canCreateChildren` for a child (today always `false`); a grandchild's capabilities stay ⊆ its parent's (`isPolicySubset`).
- `scoop_wait` constrained to the caller's own subtree (today `resolveScoopNames` matches globally; completion delivery is already grouped per parent).
- Parent closure handles descendants according to policy (cascade today; add `detach-on-close`).

## Exit criteria (from the RFC)
- [ ] Multiple root units and multiple children run independently and in parallel.
- [ ] A child can create grandchildren only when granted; ownership and inheritance are explicit and tested.
- [ ] Detach/promote is representable and tested.

PR #2262, opened 2026-08-21
feat(webapp): WorkUnit runtime for cones and scoops + multiple cones UI (#1666 phases 1–4)

trieloff:
Part of #1666 — phases 1–4 of the WorkUnit migration as one linear PR (8 commits, one per phase/follow-up; no squash). Supersedes #2252, #2256 and #2257, which are closed in favour of this one.

## Summary

Cone and scoop become roles over one runtime primitive, `WorkUnit`, and the UI can (behind a flag) run several cones.

**Phase 1 — seam** (`0a9903266`, `d672dd0e0`)
`RegisteredScoop.parentJid` is required (`null` = root; backfilled and persisted on restore). New `packages/webapp/src/work-unit/`: descriptor/policy/completion types, root/child presets with the child ⊆ parent invariant, `ScoopContextWorkUnit` adapter, `WorkUnitManager` (`orchestrator.getWorkUnits()`), conformance suite. ADR: `docs/work-unit.md`.

**Phase 2 — lifecycle ownership** (`b18891702`, `7f6fe4a0f`)
`ScoopLifecycleManager` hosts one `LiveWorkUnit` per scoop (its context, tab, observers); explicit legal-transition table; `teardown()` is the single runtime teardown, `close()` unregisters through the host (active-licks guard + record deletion).

**Phase 3 — hierarchy and policy instead of `isCone`** (`cacc2b61e`, `630c57d0d`)
`ScoopContext` reads a `WorkUnitDescriptor`; the lifecycle manager gates capabilities on the policy; completion, `scoop_wait`, idle notices and sudo requests route to the child's **parent**; unaddressed events resolve the default root; `bootstrapCone` only seeds a root when none exists. New `lint:iscone-ratchet` (4 grandfathered reads). `multi-root.test.ts` proves two independent roots.

**Phase 4 — add / remove cones in the UI** (`90544884b`, `f50fb5a99`)
Behind the new **`multiple-cones`** feature flag (Settings → Experimental, off by default): a *Cones* section in the freezer rail — click switches, ✕ removes after an inline confirm, *New cone* asks for a name and activates the cone once registered. Extra cones get `cone-<slug>` folders and per-folder chat sessions (primary keeps `session-cone`); dropping a cone cascades to its scoops; **the last cone can never be removed** (rail, panel client, bridge and kernel all refuse). Switching to a history-less unit now clears the thread.

Plus `ce1808636`: boy-scout paydown of `feature-flags.ts`' `Record<string, unknown>` entries.

[...]

PR #2361, opened 2026-08-24
fix(work-unit): scope scoop_wait and name-based child resolution to the caller's subtree

### LINKED ISSUE #2360: fix(work-unit): scoop_wait and name-based child resolution must be scoped to the caller's subtree
Part of #1666; extracted from #2278 (decided 2026-08-24) because with multiple cones on `main` it is a live cross-cone leak, not a future supervisor feature.

## Bug
`scoop_wait` resolves scoop names **globally**: `resolveScoopNames` matches against the whole roster, so cone A's `scoop_wait helper` can match and wait on cone B's scoop of the same name. Completion *delivery* is already grouped per parent (#2262 phase 3); the *matching* is not.

## Fix
- Constrain `scoop_wait` (and any other name-based child resolution reachable from a unit's tools — audit `scoop_scoop` name-dedup, `feed_scoop`, `drop_scoop` for the same hole) to the **caller's own subtree**: the calling unit plus what it transitively owns (`parentJid` walk / `subtreeOf`), falling back to nothing — an unmatched name is an error naming the caller's subtree, never a global match.
- A cone waiting on another cone's scoop is not a supported operation until #2278's supervisor APIs.

## Tests
- Two cones, same-named scoops: each cone's `scoop_wait` resolves only its own; unmatched name errors rather than crossing.
- A scoop's own tool calls (where permitted) resolve within its cone's subtree.
- Registry-order independence (both orders give the same answer — same pattern as `matchLickTargetAlias`).

## Non-goals
`createMany`/`join`/`detach`/`promote`, grandchildren, budgets — stay in #2278.

trieloff:
Closes #2360.

## Bug

`resolveScoopNames` matched scoop names against the **whole roster**, so with several cones on `main` cone A's `scoop_wait helper` could match and wait on cone B's identically named scoop. Completion *delivery* was already grouped per parent (#2262 phase 3); the *matching* was not.

## Fix

Every name-based child lookup reachable from a unit's tools now resolves against `subtreeOf(roster, caller.jid)` — the calling unit plus what it transitively owns (`parentJid` walk). An unmatched name is an error naming the caller's subtree; it never widens to a global match.

- `scoop_wait` / `scoop_mute` / `scoop_unmute` — `resolveScoopNames(names, config)`
- `feed_scoop` / `drop_scoop` — self-feed / self-drop are still reported as such; a sibling cone and its scoops are simply *not found* ("not found in your scoops (`<caller folder>`). Available: …")
- `list_scoops` — lists the caller's subtree, so what it advertises is exactly what the other tools can resolve
- `scoop_scoop` (audit result) — there was **no** name dedup at all, and folder derivation was collision-blind: two cones each spawning "helper" both got `/scoops/helper-scoop/`, i.e. one shared sandbox. Now a duplicate name inside the caller's own subtree is rejected, and the derived folder is suffixed (`helper-scoop-2`) when taken *anywhere* on the roster — folder uniqueness stays global on purpose because `/scoops/<folder>/` is one real VFS path.

A cone waiting on / feeding / dropping another cone's scoop stays unsupported until #2278's supervisor APIs.

`subtreeOf` moves from `transcript/collect.ts` to `work-unit/policy.ts` beside the other ownership-tree helpers (`childrenOf`, `rootOwnerOf`, `rootsOf`), generic over the record shape; `transcript/collect.ts` re-exports it so existing callers are untouched. That avoids a `scoops/ → transcript/ → scoops/` import back-edge.

## Tests

`packages/webapp/tests/scoops/scoop-management-tools.test.ts` — new `#2360` block with two cones, each owning a same-named `helper`, plus a grandchild under cone A:

- each cone's `scoop_wait helper` resolves only its own scoop
- registry-order independence (forward and reversed roster give the same answer, same pattern as `matchLickTargetAlias`)
- a cross-cone folder name errors instead of crossing (`scoop_wait`, `scoop_mute`)
- the transitive walk reaches a grandchild for its owner and nobody else
- `feed_scoop` / `drop_scoop` refuse a sibling cone and its scoops, still succeed on their own, and still say "Cannot feed/drop yourself"
- `list_scoops` shows only the caller's subtree
- `scoop_scoop` rejects a duplicate name in its own subtree; the same name under another cone succeeds with a uniquified folder

`packages/webapp/tests/work-unit/policy.test.ts` — `subtreeOf` at its new home: transitive ownership, order independence, empty (not global) for an unregistered root.

## Docs

`docs/work-unit.md` (Phase 3 detail), `docs/tools-reference.md`, `packages/vfs-root/workspace/skills/delegation/SKILL.md`, and the agent-facing tool descriptions.

## Verification

`verify` · `check-touched-exemptions` (0 debt) · `typecheck` · `test` · `test:coverage:webapp` (stmts 81.59 / branches 72.81 / funcs 81.67 / lines 83.60 — all above floors) · both builds · `bundle-size` (page 730/740, worker 4129/4160, kernel-worker 1595 kB). No UI change, so no screenshots.

### DIFF
--- a/docs/tools-reference.md
+++ b/docs/tools-reference.md
@@ -363,7 +363,10 @@ Scoop-only (not registered for the cone — it has no parent to message; its ass
 
 ### list_scoops
 
-Cone-only. List all registered scoops.
+Cone-only. List the scoops you own — your own subtree, not another cone's
+(#2360). Every name-based scoop tool (`feed_scoop`, `drop_scoop`, `scoop_mute`,
+`scoop_unmute`, `scoop_wait`) resolves names against exactly this list, so a
+name that is not here is an error, never a cross-cone match.
 
 | Property   | Value                                            |
 | ---------- | ------------------------------------------------ |
@@ -385,7 +388,12 @@ Cone-only. Create a new scoop.
 
 **Behavior**:
 
-- Folder is auto-derived from name (lowercase, slugified)
+- Folder is auto-derived from name (lowercase, slugified), and suffixed
+  (`helper-scoop-2`) when that folder is already taken anywhere on the roster —
+  `/scoops/<folder>/` is one shared VFS path, so two cones' scoops must not
+  collide there
+- A name already used inside your own subtree is rejected; the same name under a
+  different cone is fine
 - Scoop is registered but not activated
 - Use `feed_scoop` to give it a task
 
@@ -403,6 +411,8 @@ Cone-only. Delegate a task to a scoop.
 
 **Requirements**:
 
+- The scoop must be one you own (your subtree); another cone's scoop resolves as
+  "not found"
 - `prompt` must be complete and self-contained
 - Scoop has NO access to cone's conversation history
 - Include all context: file paths, URLs, instructions, expected output format
@@ -418,6 +428,8 @@ Cone-only. Remove a scoop.
 | **Name**   | `drop_scoop`                   |
 | **Input**  | `{ scoop_name: string }`       |
 | **Output** | `{ content: "Scoop removed" }` |
+
+Scoped to your own subtree — a cone cannot drop a sibling cone's scoop.
 
 ---
 
--- a/docs/work-unit.md
+++ b/docs/work-unit.md
@@ -326,6 +326,7 @@ the kernel host, for every float.
 - Unaddressed events (licks, sprinkles, workflow completions, follower snapshots) resolve the default root through `rootsOf(...)[0]` / `WorkUnitManager.resolveDefaultRoot()`; `bootstrapCone` only seeds a root when none exists.
 - `normalizeScoopRecord` sanitizes a root's trigger fields on register and restore; `ScoopPresentation` projects the wire's `isCone` from `isRootUnit`. Since #2279 the record has no role field at all, so nothing — `ui/` included — can branch on one.
 - `WorkUnitManager.close(id)` cascades to the unit's children first; closing root A leaves root B's subtree untouched.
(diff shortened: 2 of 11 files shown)


trieloff:
Fixed in `ef834cde` (rebased to `4f9b46b5`) — good catch, that TOCTOU is real: `register` awaited `db.saveScoop` before inserting into the live map, so two cones creating the same name concurrently could both pass the roster-only check in `scoop_scoop`.

`ScoopLifecycleManager.register` now claims the folder **and** the registry slot synchronously, before its first await (the roster read and the `scoops.set` happen in one synchronous run), renaming to the first free variant and re-deriving `trigger` / `assistantLabel`; if `saveScoop` then throws, the slot is released so a record that could not be persisted does not linger holding a folder. `uniqueFolder` moved to `work-unit/record.ts` so the tool-level pick and the authoritative reservation share one implementation.

Covered by two tests in `tests/scoops/scoop-lifecycle-manager.test.ts` — a concurrent registration with a deferred `saveScoop` (verified to fail without the fix) and the persist-failure rollback.

trieloff:
Merge-queue entry was dequeued by a **harness death, not a check regression**: the `merge_group` `e2e` run reported 18 failures, but only the first one is real — `multiple-cones-licks` timed out on a `New cone` click and every subsequent test failed with `net::ERR_CONNECTION_REFUSED at http://localhost:8787/`, i.e. the wrangler leader origin was gone for the rest of the run.

Evidence this is not the change:
- the `e2e` job passed on this exact head SHA (`4f9b46b5b`) in the PR run;
- the merge-group base differs from that run only by the `chore(release): 6.91.0` bump;
- all three multiple-cones specs pass locally against this branch on isolated ports (`SLICC_E2E_WRANGLER_PORT`/`BRIDGE_PORT`/`CDP_PORT`/`FAKE_LLM_PORT`): `licks land in the cone that owns them` ✓, `create, chat, spawn a scoop, switch and drop cones` ✓, `rail row actions` ✓ — 3 passed.

Re-enqueuing once (flake re-enqueue 1 of 2).

Issue #2274, opened 2026-08-21
feat(ui): WorkUnitClient — one client protocol for local and remote presentation


trieloff:
Part of #1666 — RFC "Phase 4: unify local and remote UI through `WorkUnitClient`" (now phase 5 after the UI landed first in #2262).

## Goal
One client protocol the WC shell mounts against, regardless of transport:

```ts
interface WorkUnitClient {
  list(): Promise<WorkUnitSummary[]>;
  snapshot(id): Promise<WorkUnitSnapshot>;
  send(id, input): Promise<void>;
  subscribe(id, listener): Unsubscribe;
  signal(id, processId, signal): Promise<void>;
}
```
with `LocalWorkUnitClient` (MessagePort to the kernel — today's `OffscreenClient`) and `RemoteWorkUnitClient` (tray/WebRTC — today's `FollowerSyncManager` + `wc-follower.ts` state machine).

## Scope
- Define `WorkUnitSummary` / `WorkUnitSnapshot` / `WorkUnitEvent` as the wire shapes (`work-unit/types.ts` already has the in-process versions).
- Adapt `OffscreenClient` and `FollowerSyncManager` to the interface behind adapters first; then move `mountWcUiLive` / `mountWcUiFollower` / `mountWcUiExtension` to one mount path consuming the client.
- Fixture parity test: leader and follower render the same fixtures from the same protocol.

## Exit criteria (from the RFC)
- [ ] Local leader and browser follower render the same fixtures from the same protocol.
- [ ] A follower feature no longer needs a parallel UI state machine (review-patterns category 8).
- [ ] Legacy extension-direct / offscreen UI paths removed or explicitly fenced.

Issue #2006, opened 2026-08-08
VFS: in-session in-memory index corruption has no recovery path — on-disk sidecar repair is futile mid-session


trieloff:
## Incident (2026-08-08, production leader — live)

The cone got stuck in a repair loop: files it depends on read as **directories** (`EISDIR`) even though they are genuine files on disk. Confirmed by read-only CDP inspection of the running leader:

| Path | Live VFS (agent sees) | On-disk sidecar | OPFS reality |
| --- | --- | --- | --- |
| `/workspace/CLAUDE.md` | **EISDIR (dir)** | file, 16249 B | file |
| `/workspace/skills/x-search/SKILL.md` | **EISDIR (dir)** | file, 2151 B | file |
| `/workspace/skills/x-search/x_search.jsh` | **EISDIR (dir)** | file, 3639 B | file |

The cone's own memory file reading as a directory is enough to derail it. A full scan of the on-disk sidecar (`slicc-fs/.metadata.json`, 11,914 entries) found **zero** files-typed-as-directories; OPFS reality is clean too. The on-disk `.metadata.json` had in fact been **rewritten ~12 minutes earlier** by a scoop that detected the corruption, obtained a sudo grant, and repaired it.

## Root cause / the gap

The corruption lives only in the **in-memory VFS index**, which is authoritative at runtime and never re-reads the on-disk sidecar. So:

- A scoop (or the cone) can repair the on-disk `.metadata.json` all it wants — the running agent keeps hitting `EISDIR` because the in-memory index still holds the path as a directory. The repair is **invisible until reload**, which is exactly the loop observed.
- The existing on-disk protections don't cover this: #1993 (merge-on-write, cross-context lock) and #1995 (boot-time sidecar self-heal + boot-error propagation) heal the sidecar **on disk at boot**. There is no **in-session** detection or heal for an already-diverged in-memory index.
- Worse, while the corrupt in-memory index is live, a subsequent VFS mutation risks flushing it back over the (correct) on-disk sidecar — the merge-on-write from #1993 mitigates but does not fully eliminate the re-corruption window.

Same corruption family as the boot-brick `EISDIR`/`ENOTDIR` incidents and #1979 (intermittent OPFS `Failed to write data to data pipe` on large writes).

## Impact

[...]

Issue #1666, opened 2026-07-25
feat: introduce a first-class WorkUnit runtime for cones and scoops


catalan-adobe:
> **Tracking overview — revised 2026-08-24** (previous revisions 2026-08-22/23). Goal: multiple concurrent cones that are useful day-to-day, then graduate the flag (#2280). The original RFC follows the overview.
>
> **Done / merged:** #2262 (WorkUnit runtime, phases 1–4, flag) · #2270/#2281 (parentId on the wire) · #2272 → #2285+#2317 (rail action row, dialogs, one Freezer, cones-first strip, `SLICC_LICK_TARGET`) · #2271 → #2286+#2322 (per-cone workspace + memory, curator per cone, owner-relative agent roots) · #2310 → #2318 (per-cone model on the record + wire) · #2311 → #2325 (crontask default, three-pass lick matcher, cone addressing, producer stamping) · #2313 → #2328 (two-instance e2e; three cone specs gate CI)
>
> **In flight:** #2312 → PR #2327 (read-only scoop view; awaiting user sign-off) · #2329 → PR #2330 (follower latches an empty model catalog; queued after CI)
>
> **Order from here**
> 1. #2327 lands (+ the pre-verified one-line follow-up flipping the read-only e2e spec live)
> 2. #2330 lands (+ delete the CI skip on the follower e2e spec)
> 3. #2279 — delete `isCone`/`type`, Phase-1 adapter, legacy paths; ratchet to zero (deletion before refactor)
> 4. #2334 — carve `scoop-context.ts` (pulled ahead 2026-08-24: most-churned file; #2275/#2274 land inside it). Scope: **mechanical extraction only** — module per responsibility cluster, no API redesign, no CapabilityBroker (#2276), coverage must not drop, one PR
> 5. #2275 — one conversation record per work unit (lands into the extracted session-persistence module)
> 6. #2274 — WorkUnitClient (local + remote presentation; the read-only-scoop and model surfaces added two more parity points it should fold in)
> 7. **Dogfood → #2280** flag graduation (judgment call on daily usefulness)
> 8. #2278 — supervisor APIs (incl. `scoop_wait` subtree scoping) → #2277 workspace modes → #2276 CapabilityBroker
>
[...]

Issue #2007, opened 2026-08-08
Kernel-ready 30s timeout kills a slow-but-progressing boot (no progress awareness)


trieloff:
The page-side kernel-ready watchdog (spawn.ts) rejects with `Kernel worker did not signal ready within 30000ms` if the worker does not post `kernel-worker-ready` within a single 30s window — regardless of whether boot is actively making progress.

## Incident (2026-08-08, production leader)

After a browser restart, a large restored session (588-message cone + 4 scoops, ~12k VFS entries, grok-4.5 models) mounted the VFS and began loading each scoop context (observed `[model] Cone using model: grok-4.5` etc. in the worker console) but did not reach ready within 30s. The watchdog fired, the recovery screen appeared, the user reloaded, and the same slow boot was killed again — a loop that never lets a genuinely-healthy boot finish. The time sink is `orchestrator.init()`’s per-scoop `createScoopTab` loop.

## Fix (this PR)

Make the timeout a **progress watchdog** rather than a hard cap: the worker emits a `kernel-worker-boot-progress` heartbeat at each boot milestone (providers registered, CDP connected, each restored scoop, lick manager, cone bootstrap), and the page re-arms the 30s timer on each. A slow-but-advancing boot survives; a genuine stall (no heartbeat for the window) still surfaces the timeout. Boot-error propagation (#1995) still fast-paths real throws.

Related: #1982, #1995, #2006.

PR #2364, opened 2026-08-24
refactor(persistence): one canonical conversation record per work unit (#2275)

### LINKED ISSUE #2275: refactor(persistence): one canonical conversation record per work unit
Part of #1666 — RFC "Phase 5: consolidate persistence".

## Goal
One canonical, append-only conversation record per work unit, from which Pi history, the UI projection, tray snapshots, frozen archives and child-completion summaries are derived. Today there are three stores with repair logic between them: `agent-sessions` (Pi), work-unit-relevant records in `slicc-groups` (routed messages/licks), and `browser-coding-agent` (UI projection, keyed `session-<folder>`).

## Scope
- Define `ConversationEntry` (user / assistant / tool-call / tool-result / external-event / child-result) and a `WorkUnitStore` keyed by `WorkUnitId` + `workspaceId`.
- Versioned, resumable migration with a read-old/write-new window; fixture profiles from historical schemas (including pre-`parentJid` records and the #2006 sidecar cases).
- Remove the buffer-repair paths in `Bridge.handleRequestScoopMessages` once the UI projection is deterministic.

## Exit criteria (from the RFC)
- [ ] One durable source represents settled conversation state.
- [ ] UI and Pi histories are deterministic projections.
- [ ] Reload needs no buffer-repair logic to prevent history truncation.
- [ ] Migration and rollback tested with real historical schemas; recovery screen behaviour defined.

trieloff:
Closes #2275. Part of #1666 (RFC Phase 5).

**Stacked on #2359** (`bb/2334-carve-scoop-context`, itself stacked on the merged #2356). Review the top commit only; I retarget to `main` once the stack lands.

## Summary

One work unit's conversation now has one canonical, append-only durable record — `ConversationEntry[]` in a new `slicc-work-units` IndexedDB store, keyed by workspace root × jid. Pi history, the chat projection, transcript text and the child-completion summary are **derived** from it. The legacy stores are still written and still the read fallback: this opens a read-old/write-new window, it does not cut over.

## Why

One conversation lived in three durable stores, none authoritative, each repaired against the others:

| Store | Key | Held |
| --- | --- | --- |
| `agent-sessions` | `<jid>` | Pi's `AgentMessage[]` |
| `browser-coding-agent` | `session-<folder>` | the chat panel's projection |
| `slicc-groups` | `chatJid` | routed messages / licks |

That repair logic is where "only the last few messages after a reboot" lived (`seedBuffersFromAgentState`, the three-step chain in `handleRequestScoopMessages`). The RFC's exit criterion is a single durable source with deterministic projections.

`packages/webapp/src/work-unit/conversation/`:

| Module | Owns |
| --- | --- |
| `types.ts` | `ConversationEntry` (`user` / `assistant` / `tool-call` / `tool-result` / `external-event` / `child-result`), the record, its version |
| `key.ts` | `<workspaceId>::<workUnitId>` identity, workspace root from `workspaceFor` |
| `entries.ts` | ingest from Pi messages (live path) and from a chat transcript (migration only) |
| `derive.ts` | `toAgentMessages` (Pi) · `toChatMessages` (UI) · `toTranscriptText` (tray / archives) · `toChildResultSummary` |
| `store.ts` | the IndexedDB store + the migration cursor |
| `migration.ts` | the versioned, resumable pass |

Non-obvious calls, and what I rejected:

- **The UI derivation reuses `agentMessagesToChatMessages`** rather than reimplementing the projection. A rebuild from the record and a rebuild from live agent state are therefore the same transcript, and the translator's existing tests stay the contract for both. A second translator would have been a fourth opinion about the same conversation.
- **A `tool-call` entry carries no arguments.** Pi keeps tool calls inside the assistant message that issued them, so the entry is an addressable handle (`toolCallId`, `name`, `assistantEntryId`) and the Pi derivation skips it. Copying a `write_file` payload onto it would double the record for nothing.
- **A `ui-projection` record derives to no Pi history, deliberately.** You cannot reconstruct a faithful Pi conversation from a rendered transcript. Replaying a reconstruction to the model is worse than restoring from the legacy store — and an empty derivation is exactly what makes that fallback happen.
- **Append-only with one exception.** Compaction (and `clear-chat`) replaces Pi's history wholesale; that shows up as a prefix divergence, is applied as a replace, and is counted on `rewrites`. Interleaving would splice a pre-compaction conversation into a post-compaction one. The prefix test is deliberately shallow (kind, seq, text) so a provider annotating a message this build already stored does not read as a rewrite.
- **`ui/lick-channels.ts` → `base/lick-channels.ts`** (pure constants, no UI imports). The layer guard requires it for `work-unit/` to read the channel type; it also clears the pre-existing `scoops → ui` back-edge, so the baseline **shrinks by one entry and gains none**.

### Migration / data-shape change

New DB `slicc-work-units` (v1): `conversations` keyed `key`, `migrations` keyed `id`. **No legacy store is read-modified or deleted.**

- **Versioned** — `CONVERSATION_RECORD_VERSION` is part of the cursor; a schema bump re-runs the pass over every unit.
- **Resumable** — the cursor is persisted after every unit, so a boot that dies mid-pass continues where it stopped.
- **A bad unit cannot break the boot** — a legacy read that throws, or a payload whose `messages` is not a list, is recorded on the cursor's `skipped` list with its reason and **left in place**. That is the #2006 rule: a half-written record is a repair job for a later build, never a delete. The cursor advances past it on purpose (retrying it every boot would re-spend the boot budget that made it unreadable); a schema bump is the sanctioned retry.
- **Pre-`parentJid` records** are backfilled by `Orchestrator.init` before the pass; the pass coerces a missing edge to `null` anyway, so a legacy primary cone can never be keyed as a child under `/scoops/`.

## Cross-cutting impact

- **Floats**: the store is created in `Orchestrator.init`, so CLI and extension both persist through it (same kernel worker, same hosted origin). Nothing float-specific was added; a float with no store wired behaves exactly as before.
- **Other callers of the changed contracts**: `ScoopContext` gained a 9th optional constructor arg; `ScoopLifecycleManagerDeps` gained a required `getConversationStore` (kept required so the compiler catches an unwired composition — the two test doubles are updated). `Bridge` reads it through an optional call, because a leader-tray adapter may bind an older orchestrator shape, and a missing store is the fall-through the method already handles.
- **Transcript export / freezer / tray snapshots are untouched** — they still read live agent state and the legacy stores, and their tests pass unchanged. Moving them onto the derivations belongs with the legacy-write deletion.
- **What this PR did NOT make dead**: the repair chain in `Bridge.handleRequestScoopMessages` stays. The canonical record is inserted as a new step — the only source that answers for a unit whose context has not spawned — but the buffer, the live-agent translation and the legacy UI-store fallback are still the UI's sources until #2274 consolidates them. Removing them now would remove paths this PR has not replaced. I will note that scope split on #2275 when it closes.
- **Bundle**: page 730/740 kB and worker 4142/4160 kB eager closure (unchanged within the budget); kernel-worker chunk 1603 kB against the 1965 kB ceiling.

## Test plan

- [x] `npx prettier --write` / `biome` via the pre-commit hook
- [x] `npm run test` — **16,579 passing**, 46 skipped, no failures
- [x] `npm run test:coverage:webapp` — **83.69 / 81.66 / 81.67 / 72.88** vs floors 83 / 81 / 81 / 72 (gate exit 0)
- [x] `npm run build`
- [x] `npm run build -w @slicc/chrome-extension` (via the root build)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — OK, 62 changed files, 0 on any debt list
- [x] **New behavior is covered**: `packages/webapp/tests/work-unit/conversation/` — 68 tests over fake-indexeddb with per-suite `dbName` isolation, plus 2 in `tests/kernel/bridge.test.ts`
- [x] N/A — no UI change, so no screencast

The fixtures are **historical shapes, not tidy synthetic ones**: ISO-string timestamps from the earliest `agent-sessions` writes, batched lick envelopes in one prompt, a child notice whose only signal is a body marker, a `browser-coding-agent` session for a unit with no Pi history, a **pre-`parentJid`** record, and the two #2006 failure modes (a read that throws, a payload whose `messages` is a lie).

Properties pinned: lossless Pi round-trip · UI derivation ≡ live translator, message for message · append vs rewrite detection · resume from a cursor after a crash · re-run on a schema bump · a bad unit cannot take the boot down · legacy stores byte-identical after the pass · every kill-switch path falls back with the conversation intact.

Writing them caught a real bug: `restoreCanonical` propagated a store read that rejected, which would have turned a canonical failure into a *lost conversation* rather than a fallback. Fixed and pinned by a test.

**Pre-existing failures**: none. (An earlier run against a stale base showed `ipk` ESM and `fetch-proxy-shared` failures; both are gone after rebasing onto the current #2359 head.)

## Documentation

- [x] Relevant `CLAUDE.md` — `packages/webapp/CLAUDE.md`: the WorkUnit never-rule now covers the canonical record, the derivations-not-writes rule, the data kill switch, and "never delete a legacy record"
- [x] `docs/` — `docs/work-unit.md` gains a Phase 5a section (module map, the two origins, the window, **rollback semantics**, migration behaviour, and an explicit "what this did not make dead"); `docs/architecture.md`'s IndexedDB inventory lists the new DB and re-labels the two legacy ones as fallbacks

## Risk & rollback

Highest-risk PR of the #1666 effort — user chat history is the payload — so the design is fallback-first.

- **Blast radius if the canonical path regresses: none, by construction.** Every read prefers the record and falls back on absence, and the kill switch is DATA rather than a flag: no record, an unreadable one, a newer schema, a `ui-projection` record, or an IndexedDB that will not open each derive to nothing, and nothing means "use the legacy store". A canonical write failure is logged and swallowed; the legacy write on the same call still happens.
- **Rollback, two ways.** Clear `slicc-work-units` (`WorkUnitConversationStore.clearAll()`, or delete the DB in devtools): every read falls back and the next boot re-migrates from untouched legacy data. Or revert this PR: the legacy stores hold everything up to the moment of the revert, because they were written on every turn.
- **Follow-up gating**: deleting the legacy writes must not land until the canonical store has been dogfooded through at least one migration + rollback cycle. I will file that as a separate issue rather than fold it in here.
- **CI cannot catch**: a real profile's migration. The fixtures cover the shapes I could find in the code and the docs, but a long-lived profile with an odd `agent-sessions` record is the case worth one manual boot before this rides to users.

## Checklist

- [x] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs (`lick-channels.ts` moved layer but kept its exports and its test)
- [x] Contract used by other floats: checked — the store is created once in the orchestrator, so CLI, extension and worker paths all get it; the bridge reads it defensively for leader-tray adapters


trieloff:
Both P1s were real and are fixed in 3e77786.

**Clear the canonical record on new chat** — correct, and it was the worse of the two: `clear-chat` deleted the legacy agent/UI sessions but not the record, and since a restore now prefers the record, the next reload resurrected the conversation the user had just cleared. Rather than adding a third deletion at the call site, clearing now goes through the module that owns the durable representation: `ScoopContext.clearSession()` → `SessionPersistence.clear()` deletes both representations and cancels any pending mid-turn checkpoint first, so an in-flight debounce cannot write the history back a second later. `clearScoopMessages` and `clearAllMessages` call it instead of deleting the legacy session by hand. Pinned by two tests, including the reload-after-clear case.

**Preserve newer-schema records during fallback** — also correct. `load()` collapsed four distinct outcomes into `null`, which is right for readers (`null` = "fall back to the legacy store") but wrong for writers. Writers now go through `store.read()`, which returns `absent` / `malformed` / `incompatible` / `error`; only the first two may be written over. `incompatible` is left exactly where it is, and so is `error` — a read that merely failed may be sitting on a perfectly good record. `syncAgentMessages` and the migration both refuse to write in those two cases (the migration records the `error` case as skipped, with its reason). Pinned by tests on both paths.

`docs/work-unit.md` records both rules. Full pass re-run green: typecheck (9 projects), 16,587 tests, coverage 83.77 / 81.74 / 81.82 / 72.96 vs floors 83 / 81 / 81 / 72, both builds, debt gate 0.

Also self-caught while this was in review and fixed in ef7e1eb: the migration runs at boot before any context spawns and had no heartbeat, so a profile with many large histories could have been killed by the kernel-ready watchdog (#2007) on a boot that was provably advancing. It now fires `onBootProgress` after every unit, like the scoop-restore loop.

trieloff:
**Local e2e run** (fake-LLM suite, isolated ports `SLICC_E2E_BRIDGE_PORT=5718 / CDP 9228 / FAKE_LLM 8795 / WRANGLER 8790`), since the heavy CI jobs don't run against a non-`main` base:

| Spec | Result |
| --- | --- |
| `reference-scenario` | pass — boot + chat with the migration on the boot path |
| `compaction-robustness` | pass — the one transition that legitimately rewrites the record |
| `multiple-cones.test.ts` (4 of 5) | pass |
| `multiple-cones.test.ts` › "rail row actions: new chat, fast and discard" | **fail** |
| `transcript-export` › "exports ZIP" | **fail** |

Both failures reproduce **identically on the base branch** (`bb/2334-carve-scoop-context` @ `0013f309c`) with none of this PR's changes, so they are not regressions from #2364 — I checked them specifically because they sit on the two paths this PR touches most (new-chat clearing, and the transcript export I promised to leave behaviourally unchanged). Whether they are a #2359 regression or an artifact of this sandbox (headless download handling for the ZIP case) is for that PR, not this one.

Worth re-running both once this is retargeted to `main` and real CI runs.

trieloff:
#2359 landed, so this is rebased onto `main` (`8413e29`) and retargeted. The stack note in the description is now stale — this is a plain PR against `main`.

Full pass re-run against `main`: `lint:ci` · typecheck (9 projects) · **16,619 tests** passing · coverage **83.75 / 81.71 / 81.77 / 72.95** vs floors 83 / 81 / 81 / 72 · both builds · debt gate 0 on 32 changed files · first-load budgets OK (page 730/740 kB, worker 4143/4160 kB; kernel-worker chunk 1604 kB against the 1965 ceiling).

PR #2379, opened 2026-08-24
test(e2e): contain a mid-suite wrangler crash instead of losing the run

### LINKED ISSUE #2372: test(e2e): wrangler dev dies mid-suite and takes the whole e2e job with it — health-check + restart between specs, surface the crash log
Recurring harness-level failure in the `e2e` job: the `wrangler dev` (workerd) web server dies mid-suite, and every spec after the crash fails in ~200 ms with `ERR_CONNECTION_REFUSED` / instant `element(s) not found` — mass casualties that read as unrelated test failures and repeatedly cost merge-queue cycles.

## Sightings (all 2026-08-23/24)
1. PR #2328 PR-CI: workerd crashed while a spec was wedged; 19 unrelated specs failed (`kj::getCaughtExceptionAsKj … Broken pipe`).
2. PR #2327 merge_group: `[WebServer] ✘ [ERROR]` after the first `compaction-robustness` spec → queue eviction (retry recovered).
3. PR #2370 PR-CI: wrangler died during the third `compaction-robustness` spec, ~30 s before the two-instance leg started; everything after failed instantly. Crash clearly precedes the suspected spec, so it is harness-level, not test-level.

Wrangler crash logs are written (e.g. `wrangler-2026-08-24_11-37-16_499.log`) but not surfaced as artifacts.

## Proposal
- Health-check the leader origin between specs (cheap `HEAD /` in a global `beforeEach`/fixture); on failure, restart `wrangler dev` (Playwright `webServer` can be wrapped, or manage the process in the harness) and fail the *current* spec with a named `WRANGLER_CRASHED` error instead of letting the rest of the suite bleed out.
- Upload the wrangler log as a CI artifact when the job fails.
- Optionally: investigate the crash itself (workerd `Broken pipe` under parallel contexts + tray websockets; possibly resource exhaustion on the runner).

Related: docs in `.agents/skills/cdp-smoke-test/`, e2e job in `.github/workflows/ci.yml`, `packages/webapp/tests/e2e/playwright.config.ts`. Known-flake list context in #2329/#2328 history.

trieloff:
Closes #2372

## Summary

`wrangler dev` (workerd) dies mid-suite in the `e2e` job. Playwright never
revives a `webServer`, so before this PR the first crash turned every later spec
into a ~200 ms `ERR_CONNECTION_REFUSED` failure — a wall of red naming eighteen
innocent tests. Now wrangler runs under a supervisor that restarts it, and only
the spec that was running when the origin died fails, with an error named
`WRANGLER_CRASHED`.

## Why

Four sightings in three days (#2372), each costing a merge-queue cycle: the job
failed with a list of unrelated-looking specs and no hint that the harness, not
the product, had died.

The cause is very likely
[cloudflare/workers-sdk#15202](https://github.com/cloudflare/workers-sdk/issues/15202)
(filed 2026-08-23, open, same wrangler 4.123.0, same Playwright `webServer`
setup): workerd exits with `kj/async-io-unix.c++: … write(): Broken pipe` when
the consumer of its stdout stops draining the pipe, after which everything is
`ECONNREFUSED`. Our job sets `WRANGLER_LOG=debug`, which makes that stream very
chatty. The supervisor now owns and always drains wrangler's stdout — the same
workaround that issue's reporter landed on — so this is plausibly a fix, not
just containment. No version bump: 4.125.0 exists but carries nothing related;
Renovate owns that pin.

- **`tests/e2e/wrangler-server.ts`** (new) — supervises `wrangler dev` in place
  of Playwright spawning it directly. Owns and drains its stdout (mirroring it
  to `wrangler-output.log`), re-spawns workerd on an unexpected exit, and serves
  a loopback control plane (`GET /health`, `POST /restart`) on
  `SLICC_E2E_WRANGLER_SUPERVISOR_PORT` (default: leader port + 1).
  - The wrangler chain (`npx` → `.bin/wrangler` → `cli.js` → workerd) runs in
    its own process group, so a shim that dies on its own does not leave workerd
    holding the listen socket against its own replacement (`EADDRINUSE`, then a
    crash loop). Verified: without the group reap, the injected-crash run does
    hit exactly that.
  - Because the chain is detached, `gracefulShutdown: { signal: 'SIGTERM' }` is
    set on the `webServer` entry (Playwright otherwise SIGKILLs the group, which
    the supervisor cannot handle) and the owned group ids are recorded on disk
    so a hard-killed run's orphan is reaped at the next start. Verified: no
    listener survives a normal run.
- **`tests/e2e/leader-health.ts` + `fixtures.ts`** (new) — an auto fixture
  probes `HEAD <baseURL>/status` before and after every spec (a failed HEAD is
  confirmed with a GET, so one dropped response cannot fail a green spec). A
  dead origin restarts wrangler and fails **only the current spec** with a
  `WRANGLER_CRASHED`-named error; if the restart fails, the remaining specs fail
  instantly with the same named error rather than burning their timeouts. The
  post-test half attributes the crash to the spec that was running.
- Every spec now imports `{ expect, test }` from `./fixtures.js` instead of
  `@playwright/test` — the only change to the specs themselves.

## Cross-cutting impact

- **Floats**: none. This is test-harness only; no `src/` file changes.
- **CI**: `WRANGLER_LOG_PATH` now points at `.wrangler/e2e-logs/`, so the
  existing `wrangler-logs-e2e` artifact step (which pointed at wrangler's global
  config dir, and would have been empty) now has files: wrangler's own
  `wrangler-<ts>.log`, the mirrored output, and `crash-report.md`. A new step
  prints the crash report into the job log, so a contained crash stays visible
  even when the retry turns the job green.
- **Callers**: `two-instance-helpers.ts` imports only `expect`, which is
  re-exported unchanged.
- A contained crash now makes a job *green after retry* where it used to be red.
  That is the intent — the `::warning title=WRANGLER_CRASHED::` annotation plus
  the printed crash report keep it from being silent.

## Test plan

- [x] `npm run verify` + `check-touched-exemptions.mjs` — pass (0 debt-list files)
- [x] `npm run typecheck`
- [x] `npm run test` — 16657 passing; one flake (`wc-follower.test.ts`) that
      passes in isolation and touches nothing in this diff
- [x] `npm run test:coverage` — no floor breached
- [x] `npm run build`, `npm run build -w @slicc/chrome-extension`,
      `npm run bundle-size` — first-load budgets OK
- [x] New unit coverage: `packages/webapp/tests/e2e-harness/leader-health.test.ts`
      (8 tests over the probe / restart / abort-latch logic, dependency-injected)
- [x] **3 consecutive full local e2e runs** on isolated `SLICC_E2E_*` ports
      (22 / 21 / 22 passed): no `WRANGLER_CRASHED`, no leaked listeners after
      teardown
- [x] **Crash injection, twice** — `SIGKILL` of workerd mid-suite, and `SIGKILL`
      of the wrangler shim (which orphans workerd on the port). Both produce one
      `WRANGLER_CRASHED` failure naming the running spec, a logged restart, a
      written `crash-report.md`, and the rest of the suite green.

**Pre-existing failures** (verified against `main`, unrelated to this PR): the
`transcript-export` ZIP-download spec times out on this machine on stock `main`
config too (confirmed by reverting both the fixture import and the supervisor);
`multiple-cones` rail-row-actions flaked once in run 2 as an ordinary
fake-LLM turn flake, with no wrangler crash in the log.

## Documentation

- [x] `.agents/skills/writing-slicc-tests/SKILL.md` — new "Survive a wrangler
      crash" section: the import rule, what each piece does, and where the
      evidence lands.

## Risk & rollback

Test-harness only; reverts cleanly. Worst case if the supervisor itself
misbehaves, the leader origin fails to come up and the `webServer` readiness
gate fails loudly at startup rather than mid-suite. Irony budget: this should
remove the wrangler-death entry from the known-flake list.

### DIFF
--- a/.agents/skills/writing-slicc-tests/SKILL.md
+++ b/.agents/skills/writing-slicc-tests/SKILL.md
@@ -852,6 +852,8 @@ the UI:
 - **wrangler dev on 8787** — serves the built `dist/ui` (the leader / UI
   origin and the Playwright `baseURL`) with SPA fallback, exactly as the
   production worker does. Requires `npm run build -w @slicc/webapp` first.
+  Runs under the supervisor in `wrangler-server.ts` — see "Survive a wrangler
+  crash" below.
 - **node-server `--serve-only --cdp-port=9222` on 5710** — the thin `/cdp`
   bridge + `/api` surface only. `SLICC_BRIDGE_TOKEN` arms the `/cdp` upgrade
   gate + cross-origin `/api` token check, and `BRIDGE_DEV_ALLOWED_ORIGINS`
@@ -892,6 +894,41 @@ conditional leg: when the `speech` path filter matches, the run sets
 `RUN_REAL_SPEECH_E2E=1` (un-skipping `speech-roundtrip.test.ts`) and
 frees runner disk first so Chromium's free-disk-derived storage quota
 can hold the staged weights.
+
+### Survive a wrangler Crash
+
+`wrangler dev` (workerd) dies mid-suite (#2372). Left alone, the first crash
+turns every later spec into a ~200 ms `ERR_CONNECTION_REFUSED` failure — a wall
+of red naming eighteen innocent tests. The likely cause is
+[workers-sdk#15202](https://github.com/cloudflare/workers-sdk/issues/15202):
+workerd exits with `kj/async-io-unix.c++: … Broken pipe` when the consumer of
+its stdout (Playwright's `webServer` capture) stops draining the pipe.
+
+Three pieces contain it, and one rule applies when you write a spec:
+
+- **Import `test` from `./fixtures.js`, never from `@playwright/test`.** That
+  is the whole rule. `fixtures.ts` re-exports `expect` unchanged and adds one
+  auto fixture that probes `HEAD <baseURL>/status` before and after every test.
+- **`wrangler-server.ts`** supervises wrangler instead of Playwright spawning it
+  directly: it owns (and always drains) wrangler's stdout, mirrors it to
+  `.wrangler/e2e-logs/wrangler-output.log`, re-spawns workerd when it exits, and
+  serves `POST /restart` on `SLICC_E2E_WRANGLER_SUPERVISOR_PORT`
+  (default: leader port + 1).
+- **`leader-health.ts`** decides what happens on a dead origin: restart through
+  the supervisor and fail _only_ the current spec with an error named
+  `WRANGLER_CRASHED`. If the restart fails, the remaining specs fail instantly
+  with the same named error instead of burning their timeouts — the job then
+  names its cause. Its logic is dependency-injected and unit-tested in
+  `packages/webapp/tests/e2e-harness/leader-health.test.ts`.
+
+So a crash costs one spec (which CI then retries), not the run. Evidence lands
+in `.wrangler/e2e-logs/`: wrangler's own `wrangler-<ts>.log`, the mirrored
+output, and `crash-report.md` (one section per unexpected workerd exit, with
+the output tail). The CI job prints the crash report and uploads the directory
+as the `wrangler-logs-e2e` artifact.
+
+Grep a red e2e job for `WRANGLER_CRASHED` before believing the failing spec
+names are meaningful.
 
 ### Verify Risks with the Reference Scenario
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -548,18 +548,36 @@ jobs:
             packages/webapp/test-results/
             packages/webapp/tests/e2e/test-results/
           if-no-files-found: ignore
+      - name: Show wrangler crash report
+        # The supervisor (`packages/webapp/tests/e2e/wrangler-server.ts`) appends
+        # one section per unexpected workerd exit — code, signal, and the tail of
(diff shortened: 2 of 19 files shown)


trieloff:
Both Codex findings were real and are fixed in 7dc1682:

**P1 — abort latch in a disposable worker.** Correct: Playwright discards a worker after a failed test, so the module-level latch never reached the next spec, and each one paid another doomed ~2-minute restart. The latch now lives where it survives — the supervisor. After two failed restarts it declares the origin unrecoverable and answers `POST /restart` immediately with 503, so every remaining spec fails in milliseconds with `WRANGLER_CRASHED` instead of waiting out a cold start that is not coming. `GET /health` reports `failedRestarts` / `permanentFailure`.

**P1 — fixture inherits the 30 s test timeout.** Also correct. `assertLeaderAlive` now takes an `onSlowPath` callback it invokes once, after the probe fails and before any restart is awaited; the fixture uses it to call `testInfo.setTimeout(testInfo.timeout + RESTART_TIMEOUT_MS)`. Only the slow path extends the budget, so a genuine hang in a healthy run still fails at the configured timeout.

Both are covered by new unit tests in `packages/webapp/tests/e2e-harness/leader-health.test.ts` (10 tests now), and the crash-injection run still produces exactly one `WRANGLER_CRASHED` failure with the remaining specs green.

PR #2387, opened 2026-08-24
fix(webapp): recover serial handles after a device re-enumerates

trieloff:
## Summary

A Web Serial handle stopped working permanently once its device re-enumerated. Before: after any board reset (every `esptool` flash does one), `serial open serial1` failed forever with `Failed to open serial port`, and the only recovery was reloading the page. Now: `serial list` reconciles against the live port set, so re-running it yields a working handle and the session recovers in place.

## Why

Found while flashing an ESP32-S3 over Web Serial from the shell. The port wedged **8+ times**, each needing a full harness restart. The tell was handles climbing `serial1` → `serial2` → `serial3`: each re-enumeration produced a *new* `SerialPort` object, and the dead one stayed registered.

`serialList` registered every port from `getPorts()` but never evicted entries whose port had vanished. `SerialPortRegistry.register()` dedupes by object identity, which is correct — but nothing ever removed the stale objects, so old handles kept resolving to corpses.

**Repro**

1. Grant a serial port (a USB-CDC device that resets, e.g. any ESP32)
2. `serial list` → `serial1`
3. Cause a re-enumeration: `esptool --port serial1 write_flash …` (resets the board), or just unplug/replug
4. `serial open serial1 --baud 115200`

Expected: a clear error telling you the handle is stale, and `serial list` yields a working one.
Actual (before): `Failed to open serial port.` forever, on every handle, until page reload.

- `SerialPortRegistry.retainOnly(live)` drops entries absent from the current `getPorts()`, clearing reader/writer/leftover state and returning evicted handles. `serialList` calls it **before** registering.
- `serialClose` is now idempotent — `port.close()` guarded, `opened` reset in `finally`. Previously a redundant close threw *after* the reader/writer teardown, so `opened` was never cleared and every later op was refused. This is what forced `serial close || true` in scripts.
- `serialOpen` wraps open failures with "handle may be stale after a device reset — re-run `serial list`". The browser's raw message reads like a hardware fault; it cost me hours chasing the board instead of the code.

**Deliberate behaviour change worth a reviewer's eye:** Chrome returns a *different* object from `requestPort()` than from `getPorts()`, so the handle returned by `serial request` is now retired by the next `serial list`. That's correct — the `getPorts()` objects are canonical — but it means callers must re-read the handle rather than cache it. Docs updated to show `H=$(serial list | head -1 | cut -f1)`.

**Not fixed, deliberately:** a read in flight when the device disappears still reports `The device has been lost`. That is accurate — the device did go away. The fix is that re-listing now recovers instead of requiring a reload.

## Cross-cutting impact

- **Floats exercised**: CLI (standalone Chrome, local `wrangler dev` origin). The `serial` command is Chromium-only and unavailable in the cloud / hosted-leader float, so no change there.
- **Other callers of the changed contract**: `serial-backends.ts` `LocalDomSerialBackend` (calls `serialOps.*` directly) and `ui/panel-rpc-handlers.ts` (worker → page bridge) both go through the same ops module, so both inherit the fix. `esptool-command.ts` layers on the same handle namespace and benefits without change.
- **`retainOnly` and `remove`**: `remove()` is untouched and still used by explicit teardown paths.
- **Persisted state**: none. The registry is in-memory, per DOM realm.
- **Security / bundle size**: no new dependencies, no new surface.

## Test plan

- [x] `npx prettier --write docs/shell-reference.md` — `prettier --check .` now clean (it caught my markdown edit)
- [x] `npm run lint:ci` — exit 0
- [x] `npm run typecheck` — exit 0 (all 9 projects)
- [x] `npm run test` — **16743 passed | 46 skipped**, no failures
- [x] `npm run build` — exit 0
- [x] `npm run build -w @slicc/chrome-extension` — exit 0
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — OK, 4 changed files, 0 on any debt list
- [x] New behaviour covered by unit tests: `npx vitest run packages/webapp/tests/kernel/serial-port-registry.test.ts` — 5 new tests. **Each was confirmed FAILING against the unfixed source** before being trusted (reverted `src/`, saw 5 red, restored).
- [x] **Manual smoke on real hardware** (CLI float, local build at `localhost:8788`, ESP32-S3 over USB-Serial-JTAG): forced repeated re-enumerations by flashing, then `serial list` + open — three consecutive `RECOVERED-serialN`, no page reload, no harness restart. Stale handles now report `unknown serial handle 'serial1'`.

No UI change, so no screencast.

### DIFF
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -909,6 +909,21 @@ Web Serial access from the shell (`packages/webapp/src/shell/supplemental-comman
 
 The `serial request` chooser requires a real user gesture (see [Gesture bridge](#gesture-bridge-usb--serial--hid)). `read` emits raw bytes by default; `--hex` hex-dumps. Reads/writes are capped at 4 MiB.
 
+**Handles are invalidated by re-enumeration.** A board that resets — every
+`esptool` flash does this — cycles its USB device, and Chrome hands back a
+_new_ `SerialPort` object. `serial list` reconciles on each call: ports that
+disappeared are evicted and their handles stop resolving, so a script must
+re-read the handle rather than cache `serial1`:
+
+```bash
+H=$(serial list | head -1 | cut -f1)
+serial open "$H" --baud 115200
+```
+
+Opening a handle whose device went away reports that the handle may be stale
+and tells you to re-run `serial list`, instead of surfacing the browser's bare
+`Failed to open serial port`.
+
 ### Subcommands
 
 | Form                                      | Behavior                                                                |
@@ -916,7 +931,7 @@ The `serial request` chooser requires a real user gesture (see [Gesture bridge](
 | `serial list`                             | List currently-granted ports as `handle  vid:pid [open]`.               |
 | `serial request [--vid 0x.. --pid 0x..]`  | Open the port picker (needs a user gesture); prints the granted handle. |
 | `serial open <handle> [open flags]`       | Open a port with the given line settings.                               |
-| `serial close <handle>`                   | Close a port.                                                           |
+| `serial close <handle>`                   | Close a port. Idempotent — closing an already-closed port succeeds.     |
 | `serial read <handle> [read flags]`       | Read bytes (raw by default, `--hex` to dump).                           |
 | `serial write <handle>`                   | Write stdin bytes to the port; prints `<n> bytes written`.              |
 | `serial signals <handle> get`             | Print control input signals (`cts dcd dsr ri`).                         |
--- a/packages/webapp/src/kernel/serial-operations.ts
+++ b/packages/webapp/src/kernel/serial-operations.ts
@@ -59,6 +59,9 @@ export async function serialList(
   serial: SerialApi
 ): Promise<SerialDeviceInfo[]> {
   const ports = await serial.getPorts();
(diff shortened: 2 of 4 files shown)


trieloff:
Both findings were real — thanks, these were good catches.

**P1 (evicted ports stranded open/locked)** — confirmed. `retainOnly` nulled the reader/writer *fields* without cancelling the reader, releasing either lock, or closing the port, and then deleted the entry so that cleanup was unreachable. An entry evicted while open would have kept the underlying port held, so a freshly registered handle for the same device could still fail to open — exactly the wedge the PR set out to fix. `retainOnly` now returns the evicted entries and `serialList` retires each one (`retireEvicted`), and the reader/writer teardown shared with `serialClose` is extracted into `releaseStreams()` so the two paths can't drift.

**P2 (opened state after a failed close)** — also confirmed, and I verified the consequence you named: `takeOverPort()` in `esptool-operations.ts` closes only `if (entry.opened)`, so lying about the state does skip its close. My `finally` was over-eager; `opened` is now cleared only after a successful close or a confirmed already-closed error.

Three new tests cover these, and I confirmed each one fails against the pre-review source before trusting it green. Branch rebased onto current `main` (the debt gate was reporting phantom entries for untouched files from staleness).

PR #2389, opened 2026-08-24
feat(monitor): rebuild the live monitor around vitals, attention, and a real table

trieloff:
> Stacked on #2386 (foundation). Review that one first — this PR's base is its branch, and the diff will shrink once it merges.

## Why

The monitor rendered **eleven visually identical cards** — label, count badge, rows of icon/name/meta/status-dot — over four structurally different kinds of data, and led with a `TRACKED` figure that summed processes with OAuth providers and mounts. A count of unlike things is not a metric, and inventory that is fine does not need a card each.

Three tiers now, in three deliberately different vocabularies, because they answer three different questions.

## 1. Vitals — rates and ratios, with a time axis

Replaces `TRACKED / ACTIVE / ATTENTION / SESSION COST`. Every number here is one the system actually has:

| Tile | Source |
|---|---|
| **Burn rate** (hero, ≥48px) | `SessionStats.burnRate` — the worker's existing blended 15/60-min USD/hour |
| **Agent load** | work units mid-turn, over the total |
| **Live processes** | the live count from #2386's snapshot |
| **Context fill** (meter) | the fullest of `SessionStats.fills`, escalating green → amber → rose |

Sparklines come from `monitor-history.ts`, a small in-memory ring buffer fed by the panel's own 5s refresh. Its limits are stated rather than hidden: history starts when the panel opens, nothing is persisted, and **`windowLabel()` reports the span the samples actually cover** (`last 45s`) instead of claiming a window it doesn't have. Below two points no sparkline is drawn at all — one point is not a trend.

The context meter only appears when there is a real reading. With no `fills`, a 0% bar would read as "context is empty", so the tile is omitted instead.

## 2. Attention — what is degraded, worst first

This is what `ATTENTION 2` should have expanded into. Derived from the same state the groups render — the point isn't new data, it's that a reader shouldn't open seven groups to learn whether anything is wrong. Expired accounts, a failing/reconnecting tray, stalled followers, mounts needing re-grant. Alerts carry `age` only where a real timestamp exists; an invented "2m ago" would be worse than none. Healthy state is one **All clear** line.

## 3. Body — topology and a real table

**Topology** collapses eight inventory cards into six groups rendered as summary *lines* (`Mounts · 2 · all granted`). Healthy groups stay shut; only groups needing attention auto-expand. MCP servers and provider accounts merge into **Integrations** — they were two cards with identical chrome and one row each, and as a reader question they are the same one. Cron and webhooks merge into **Automations**. Scoops render as the cone → scoop tree that `parentJid` already describes.

**Processes** is a table, because it is one: sortable on every column, filterable on command/scoop/pid, using the `ps` column vocabulary. It does not share the topology chrome.

## Contrast

`--amber` measures **2.09:1** against the light surface, so status can never ride on hue: every state carries a shape-varied glyph **and** a word (`⚠ Attention`, `✓ Healthy`).

Dark mode is *selected*, not flipped. `--green` and `--red` measure **4.06:1** and **3.14:1** on `#161618` — both under 4.5:1 for the small status text they carry — so the monitor re-steps them to ~6–7:1 locally. (The token file itself is untouched; a repo-wide re-step is a bigger conversation than this PR.)

## Deliberately not here

**Row actions** — kill, re-auth, reconnect, re-grant. The panel is still read-only. Each needs a real handler and its own approval story; folding them in would have doubled this diff. Worth a follow-up issue.

## Tests

- **`slicc-monitor.test.ts`** (41, real Chromium) — hero is unique and ≥48px; sparkline needs two points; meter clamps to 0..1; series wins over ratio; all-clear vs alerts and pill escalation; group status derived-vs-explicit; healthy collapses / degraded auto-expands; **a degraded group stays shut once the reader shuts it**; a healthy group stays open across a re-render; aria wiring; table sort/filter/empty-states; getters return copies; dark re-steps the hues.
- **`monitor-history.test.ts`** (8) — no series below two points, oldest-first, eviction, capacity floor, and that the window label tracks the real span.
- **`wc-monitor.test.ts`** (33) — seven groups; the merged Integrations/Automations summaries; mount and account roll-ups; `buildVitals` cold-start, large-rate, and fill-escalation cases; `buildAlerts` ordering and non-triggers; the process table's `ps` columns.

Two of these caught a real bug while being written: the group toggle keyed off set membership, so the **first click on an auto-expanded degraded group re-opened it** instead of closing it. It now keys off what the reader can actually see.

## Verification

`lint` · `typecheck` · `test` (16,820) · `test:coverage` (webapp + webcomponents, no floor regressions) · `build` · `check-touched-exemptions` — all green locally.

<sub>Note: 9 `packages/webapp/tests/shell/ipk/*` e2e files time out at the 5s default under local coverage instrumentation and pass at `--testTimeout=30000`. They are untouched by this PR and pass in the uninstrumented run.</sub>

### DIFF
--- a/docs/kernel/process-model.md
+++ b/docs/kernel/process-model.md
@@ -33,6 +33,22 @@ interface Process {
 
 Status transitions: `running` → `exited` (clean) or `killed` (any terminating signal recorded). The manager fires `spawn` and `exit` events synchronously inside the corresponding method calls so `/proc` and `ps` see live state without a tick of latency.
 
+### Retention
+
+A terminated process is not dropped the instant it exits — `ps` after a `kill` has to be able to show the exit code — but it is not kept forever either. The manager retains the most recent `TERMINATED_RETENTION` (128) terminated records and reaps older ones oldest-first, so the table stays O(live + 128) instead of accumulating every command the session ever ran. Eviction only ever considers already-terminated pids, so a live process can never be reaped.
+
+`stats()` carries the totals the reaped records used to:
+
+```ts
+{
+  (live, retained, terminated, spawned);
+}
+```
+
+`terminated` and `spawned` are monotonic for the session and keep counting past the retention window; `retained` is how many records are actually resident. Surfaces that want to say how much work ran (the monitor's "1,435 exited this session") read the counter — a number stays true without thousands of records having to stay resident to prove it.
+
+Consequences worth knowing: `get(pid)` and `wait(pid)` on a reaped pid behave exactly like an unknown pid (`null` / reject), and `allocatePid()` may reuse a reaped pid after the uint32 space wraps.
+
 ## Where pids come from
 
 | Kind         | Spawn site                                          | argv                                    |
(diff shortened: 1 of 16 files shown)


trieloff:
> **Stacked on #2386.** This PR's base is `main` so it gets CI (the workflow filters `pull_request.branches: [main]`, so a PR based on a topic branch runs only the review bots). That means the diff currently carries **two commits** — review the second one, `feat(monitor): …`; the first is #2386 and GitHub will drop it from this diff once that merges.
>
> Merge order still matters: **#2386 first.**

trieloff:
## Screenshots

Storybook `Workbench/Monitor`, 1440px wide, captured from the built component.

### Needs attention (light)

Three things wrong. The attention feed leads and names each one; the topology groups containing them auto-expand while the healthy ones stay shut.

needs attention, light

### Needs attention (dark)

The same state with the re-stepped status hues. `--green` / `--red` measure 4.06:1 and 3.14:1 against `#161618`; these sit at ~6–7:1.

needs attention, dark

### Healthy

Every topology group is one line. Attention is a single "All clear".

healthy

### Cold start

First render of a fresh session: no history to plot, no spend, nothing mounted. Every tile still reads as a number rather than a blank, no sparkline is drawn from a single point, the context meter is absent rather than showing a misleading 0%, and the empty table says so instead of showing a bare header.

cold start

PR #2392, opened 2026-08-24
feat(webapp): modal keyboard mode for the web UI

trieloff:
## Summary

Gives the web UI a **modal** keyboard layer, in the vim sense: <kbd>Esc</kbd> leaves the composer and enters keyboard mode, and inside it every binding is a bare letter. Press <kbd>Esc</kbd> again to leave. Outside the mode nothing is intercepted at all.

Before: no keyboard shortcuts; every cone/scoop switch and every panel was a click. Now the whole shell is reachable from the home row, and typing is never in competition with it.

| Key | Action |
| --- | --- |
| <kbd>Esc</kbd> | Enter keyboard mode — press again to leave (and exit full screen) |
| <kbd>1</kbd>…<kbd>9</kbd> | Switch to that agent in the tab strip (<kbd>9</kbd> = last) |
| <kbd>d</kbd> | Next agent, looping |
| <kbd>c</kbd> / <kbd>⏎</kbd> | Back to the composer |
| <kbd>n</kbd> | New conversation |
| <kbd>b</kbd> · <kbd>x</kbd> | Toggle the left rail · the right panel |
| <kbd>f</kbd> · <kbd>t</kbd> · <kbd>e</kbd> · <kbd>m</kbd> | Files · Browser tabs · Terminal · Memory |
| <kbd>s</kbd> · <kbd>l</kbd> | Sprinkles, looping · model picker |
| <kbd>a</kbd> | Accounts |
| <kbd>h</kbd> / <kbd>?</kbd> / <kbd>/</kbd> | Help overlay (also in the avatar menu) |

mode-help

While the mode is on, a pill under the composer says so — and doubles as a key HUD: every press prints as a cap, newest last, dimmed if it was not bound to anything.

mode-badge

key-hud

## Why

Requested, and then re-scoped after a first pass shipped modifier chords: *"slicc is keyboard heavy as it has a text input field, so something modal, almost vim-like would work better. press esc once to enter keyboard mode (ensure to intercept it so that pressing it once won't escape full screen mode on macos, but pressing it twice will). once in keyboard mode, all shortcuts work without a modifier key."*

The first design's problem is worth recording: with the composer focused almost always, an unmodified letter fights typing, and every chord that is free of the browser (⌘+digit and, off macOS, Ctrl+digit are its tab switcher) ends up as an awkward three-finger Ctrl+Shift+digit. A mode dissolves both constraints — and gives Escape, which the composer had no use for, a job.

- **`ui/wc/wc-shortcuts.ts`** owns the whole feature: the mode flag, the badge, the help overlay, and one command table.
- **The Escape contract.** First press `preventDefault()`s and enters the mode; the second exits it *and* calls `document.exitFullscreen()` **explicitly**. That last part is not belt-and-braces: while the document is fullscreen this module requests Keyboard Lock (`navigator.keyboard.lock(['Escape'])`), the only API that stops the browser spending Escape on leaving fullscreen, and under it the browser never acts on Escape at all — so the second press has to perform the exit rather than merely permit it. Where Keyboard Lock is unavailable (non-Chromium) the first press still enters the mode; it just also drops out of fullscreen.
- **An Escape that belongs to an overlay is never taken.** `<slicc-dialog>` and `<slicc-tab-overlay>` stop propagation in the capture phase and never reach the listener; `<slicc-permissions>` marks the event handled; `<slicc-quick-look>` does neither, which is the entire reason `hasOpenOverlay()` exists.
- **Every command routes through its surface's own event** — `switcher.select()`, `dock.selectItem()`, `freezer.toggle()`, and for a new conversation the same `new-chat-save` the rail's action row fires on a click. So a shortcut is indistinguishable from a click, and each float's existing wiring (selection semantics, queue stashing, read-only scoop locks, sprinkle activation, tray round trips) stays the single implementation. Nothing here reimplements what selecting or opening means.
- **`COMMANDS` is one table and `holdsMode` is the whole grammar.** Navigation (digits, `d`, `b`, `s`, help) keeps the mode; anything that hands focus to a surface (`c`, <kbd>⏎</kbd>, `n`, `f`, `t`, `e`, `m`, `a`) drops it — *before* the command runs, so a surface that autofocuses is not immediately fighting a mode that is still on. An unbound key is not an exit: the mode is sticky, like vim's.
- **The mode can never be left eating keystrokes.** Entering blurs a focused text field; `focusin` on any text field drops the mode (so clicking into the composer just works); and a keystroke whose `composedPath()[0]` is a text field is ignored even if the mode is somehow still on.
- **Installed by `mountWcShell`**, the one function leader, extension popout, follower and the design-time preview all mount through. Actions no shell element can reach are late-bound with `setAction` — today just `accounts`, registered by `wc-nav.ts` from the same `openSettings` the avatar menu uses.
- **The badge is also the HUD, in one element with two states.** It shows the mode hint until the first press, then swaps it for a strip of key caps, and swaps back after `HUD_LINGER_MS` of quiet. A press that ran nothing is *shown, dimmed*: silence reads as a dropped keystroke, while a dimmed cap reads as "that key does nothing here" — which is what someone learning the mode needs. Presses suspended behind a modal dim for the same reason, and older caps stale out so the newest is the one being answered. `describeKey()` returns an ARRAY of caps (modifiers, then the key), so a future multi-key chord is the same shape with parts from more than one event. The strip is `aria-hidden` — the badge's live region announces the mode, not every keystroke.
- **`x` toggles the right-hand panel** through the dock's own gesture (clicking the ACTIVE rail item collapses its panel): close what is open, remembering it, and otherwise reopen the last one — falling back to Files on a shell that has never opened anything. Both rails HOLD the mode, because a chrome toggle is not a hand-off; only going *to* a surface gives the mode up.
- **`l` opens the model picker** via the pill's own `openMenu()`, and routes to accounts instead when no account is connected — which is exactly where clicking the pill (`add-ai`) would take the user.
- **Keys map onto stable `CommandId`s**, which is what makes `/etc/slicc/keys.json` possible (`ui/wc/wc-shortcut-config.ts`): a user writes `"q": "terminal"` without knowing anything about the code, several keys may share a command, `null` unbinds, and a default key can change without invalidating anyone's file. The shipped copy lives in `packages/vfs-root/etc/slicc/keys.json`, is seeded once when missing and **never rewritten**; a test asserts the file and `DEFAULT_KEYMAP` agree, so they cannot drift. Parsing is total — bad JSON, unknown command, reserved key (Esc and the digits are the mode itself) each degrade to the default for *that* entry with a warning, never to a dead keymap. The help sheet is DERIVED from the live keymap, so a printed list can never disagree with the keyboard.
- **`s` cycles** rather than opening a fixed panel: sprinkles are per-install dock items, so a single key would otherwise reach only the first one. With none installed it falls back to the `new` launcher — i.e. exactly what clicking the rail's only sprinkle affordance does.

## Cross-cutting impact

- **Floats exercised**: CLI/standalone, live (below). Extension popout and follower mount through the same `mountWcShell`; every binding degrades to a no-op where the surface is absent (a follower with no dock still gets Escape, digits and help). The avatar-menu entry is leader-only, since the follower never runs `wireWcNav`.
- **Other callers of the changed contract**: `WcShellRefs` gained a required `shortcuts` field — `panelize-shell.ts` re-parents refs rather than rebuilding them, so it is unaffected. `WcNavDeps.onShowShortcuts` became `WcNavDeps.shortcuts` (two-way: the menu opens the overlay, the mode gets `accounts`); `wc-live.ts` is the only caller. `standardMenuItems()` gained a parameter; its only caller is in the same file.
- **Persisted state**: none. No localStorage keys, no ledger writes, no IndexedDB shapes. The mode is per-tab and resets on reload.
- **Async lifecycle**: `dispose()` removes the keydown/focusin/fullscreenchange listeners, the badge and any overlay. The help dialog's own `slicc-dialog-close` clears the handle so a self-dismissed dialog is not toggled back as a dead node.
- **Security**: no new `innerHTML` (overlay and badge are `createElement`/`textContent`), no new dependencies. Keyboard Lock is requested only while already fullscreen, and released on exit.
- **Bundle size**: +~3 kB in the eager page graph. First-load gate: page 731/740 kB, worker 4158/4160 kB — both under budget.

## Test plan

- [x] `npm run lint:ci`
- [x] `npm run typecheck`
- [x] `npm run test` — 16,895 passing, 0 failing
- [x] `npm run test:coverage:webapp` — gate OK; `wc-shortcuts.ts` 94.8% statements / 98.9% lines
- [x] `npm run build` · `npm run build -w @slicc/chrome-extension`
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — 0 touched files on any debt list
- [x] **Unit tests**: `npx vitest run packages/webapp/tests/ui/wc/wc-shortcuts.test.ts` — 65 tests covering the two-press Escape (including the explicit `exitFullscreen`), "outside the mode nothing is bound", every command and whether it holds the mode, the sticky-mode rule, the typing guards, and the always-mounted-overlay regression below.
- [x] **Manual smoke (CLI)**, isolated harness (`PORT=5716`, `WRANGLER_PORT=8789`, its own Chrome; production `:5710`/`:9222` untouched — same PIDs before and after), driven with real `Input.dispatchKeyEvent`, not synthetic DOM events. Verified: Esc from the composer enters the mode, blurs the textarea and shows the badge; `h`/`?`/`/` toggles the overlay; `b` opens and closes the left rail (mode held); `1` selects, `d` walks the strip; `f`/`e`/`m` open Files/Terminal/Memory and drop the mode; `t` opens the tab overlay and *its* Escape closes it without entering the mode; `s` reaches the sprinkle launcher; `a` opens the Accounts dialog; `n` fires `new-chat-save`; `c` and <kbd>⏎</kbd> return to the composer, after which `h`, `b` and the digits are plain text again. Console watcher clean.

**Screencast** — typing, <kbd>Esc</kbd> into the mode (badge appears), the rail, Files, Terminal, then <kbd>c</kbd> back to the composer where the same keys are text again:

[screencast.webm](https://slicc--ai-ecoverse.agentbin.net/19315e6dc54eaeee1a349644c62acf3161f0aa22a0f9ef20271e7fd9cafa4f9c.webm)

**A second bug live testing caught.** An edit to `/etc/slicc/keys.json` was silently reverted on the next load: the read ran before the worker's `VfsRpcHost` attached (an RPC sent that early is lost, not queued), the loader took *any* read failure for "no file yet", and re-seeded over the user's file. Fixed twice over, since either alone would leave a trap — the seed is gated on `ENOENT` specifically, and the load is deferred to `onClientReady`. A config that can eat your config is worse than no config, so this is the case the tests now pin hardest.

**Review follow-ups** (both Codex P2s, filed against the original chord design and still real under the mode): commands are now suspended while any modal is open — not just the focus-moving ones, since a digit switching the agent a dialog belongs to is the same bug — with the single exception of closing the help overlay we opened, which keeps `h` a toggle. And `wireKeyboardShortcuts` keeps one installation per document, disposing the previous wiring, because `mountWcShell` is idempotent while these listeners live on the document; a remount used to leave the first wiring installed, running first, driving the detached shell. Both verified in the browser.

**A bug the tests could not have caught, and now do.** The first live run showed the mode never turning on: `hasOpenOverlay()` matched the bare `slicc-permissions` tag, and the shell mounts that element once at boot and leaves it in the DOM for the session — so every Escape was handed to an overlay that was not there. Fixed in the second commit (`[data-open]` on the prompt), with a regression test that puts the shell's permanently-mounted overlay hosts in the DOM. Worth a reviewer's attention: any future entry in that selector must test open-ness, not presence.

**Not covered locally**: multi-agent switching (`2`–`9`, and `d` across more than one unit) — the smoke harness had no AI provider connected, so only the primary cone existed. Digit→unit mapping and the cycle are unit-tested; a click through a real multi-cone roster is worth doing during review.

**Local full-suite caveat**: the last full `npm run test` on this machine was OOM-killed and a coverage run showed 38 failures, 36 of them `Test timed out` spread across ipk / git / orchestrator / kernel files this PR does not touch (two concurrent `npm install`s were holding ~5 GB). I checked whether they were mine by reverting the diff and re-running: *more* tests failed without it. The suite was fully green earlier on this branch (16,895 passing), and `wc-shortcuts` / `wc-shell` / `wc-nav` / `wc-live` pass cleanly (177 tests). CI is the authority here.

## Documentation

- [x] `README.md` — the "Keyboard mode" section (replaces the chord table from the first pass)
- [x] `docs/webapp-details.md` — UI section: the Escape contract, Keyboard Lock, the open-ness rule for `hasOpenOverlay`, the `COMMANDS`/`holdsMode` grammar
- [ ] `packages/webapp/CLAUDE.md` — deliberately untouched: it is at 19,998/20,000 chars against the `lint:docs` cap, so the detail doc carries this

## Risk & rollback

- Blast radius: UI-only, all floats, no persisted state. The thing to scrutinise is the typing guard — the failure mode is a stolen keystroke.
- **Escape now has a meaning it did not have.** A user who pressed Escape to leave fullscreen must press it twice. That is the requested behaviour, and the badge makes the intermediate state visible, but it is the one habit this PR changes.
- No feature flag: the mode is always available. It is one call in `mountWcShell`, so the PR reverts cleanly.
- CI cannot catch: Keyboard Lock (needs real fullscreen and a Chromium build), and whether a non-US layout puts any of these letters somewhere awkward. macOS/Chrome was verified by hand.

## Checklist

- [x] Contract used by other floats: `WcShellRefs` gained a field; checked leader / extension popout / follower / preview mount paths


trieloff:
Both Codex P2s were real under the modal design too — fixed in 271f6c4, with tests and a live check.

**"Keep shortcuts from moving focus behind an open modal"** — a modal owns the screen, so every command is now suspended while one is open, not just the focus-moving ones (a digit switching the agent a dialog belongs to is the same bug). The one exception is closing the help overlay we opened ourselves, which is what keeps `h` a toggle; `h` will not stack a second overlay on someone else's modal. Escape is handled before the gate, so a modal is always dismissable. Verified in a browser: with the help overlay up, `c` and `f` do nothing and focus stays on the dialog; `h` closes it; `f` then opens Files.

**"Dispose shortcuts before replacing a mounted shell"** — `wireKeyboardShortcuts` now keeps one installation per document (`INSTALLED` WeakMap) and disposes the previous wiring, so a remount cannot leave the first one installed, running first, driving the detached shell. Disposing an already-superseded handle does not evict its successor. Done inside the module rather than at the `mountWcShell` call site, so the invariant holds for every caller.

Note both comments were filed against `67fc65c6`, the modifier-chord design; the PR has since been rebuilt as a modal (vim-like) keyboard mode, so the surrounding code has changed a lot — the body is up to date.

PR #2411, opened 2026-08-25
fix(playwright-cli): harden the browser bridge under concurrent scoop access

catalan-adobe:
## Summary

When N scoops drive `playwright-cli` concurrently (N=8 observed in a `liftoff-demo` run against frescopa.coffee, 2026-08-25), the browser bridge degraded three ways: element screenshots silently returned full-viewport frames with exit 0, viewport resizes evaporated when a sibling driver switched tabs, and commands queued invisibly on the global tab lock until they blew past `background_after` and were killed (`exit 124`, `bash: execution aborted`). All 8 scoops independently diagnosed contention; 3 of 9 verification crops were the wrong image with no error signal.

This PR fixes the three runtime failure modes.

## Changes

**Element screenshots fail loudly** — `playwright/handlers/snapshot.ts`
- `screenshot <ref>` returns the element's crop or exits 1 with a stale-snapshot hint (re-run `snapshot`). It no longer substitutes a full-viewport frame with a stderr-only warning and a 0 exit code, and it rejects zero-size boxes. Wrong data that looks like success corrupted downstream visual comparisons.

**Per-tab viewport overrides survive re-attach** — `cdp/browser-api.ts`, `playwright/handlers/tabs.ts`, `kernel/realm/realm-host.ts`
- CDP device-metrics overrides live on the session, and `attachToPage` mints a fresh session whenever a sibling attached in between — so a plain `Emulation.setDeviceMetricsOverride` silently reset under concurrency (scoops instructed to verify at 1440 measured 1720/1425/1360). `BrowserAPI.setViewportOverride` records the override per target and `attachToPage` re-applies it on every fresh session (local and remote tray branches); `closePage` drops the record. The shell `resize` and realm `setViewport` share the mechanism (follower/realm parity).

**Contention signal** — `cdp/browser-api.ts`, `playwright-command.ts`
- `withTab` tracks queue depth, cumulative wait, and acquisitions (`getTabLockStats()`). The dispatcher snapshots the stats around each command and appends a stderr note when bridge-wide lock waits exceeded 2s while the command ran, so fanned-out agents can back off deliberately instead of guessing why calls are slow or detached.

**Debt + docs**
- Pays off `handlers/snapshot.ts`'s 3 layer back-edges (cdp/core imports replaced with handler-context types and the `base/` runtime-env source); baseline ratcheted 63 → 60.
- Documents the three contracts in the `playwright-cli` agent skill.

## Tests

- 8 new tests: loud-failure + zero-size clip (handler and command level), tab-lock stats under concurrency, viewport override set / re-apply-on-re-attach / drop-on-close, contention note emitted and suppressed.
- 3 existing tests updated to the new contracts (old silent-fallback expectation, resize via `setViewportOverride`, realm `setViewport`).
- Full local pass: `verify`, `check-touched-exemptions`, `typecheck`, `test` (16859 passed; the 6 `iframe integration` failures also fail on pristine main in this environment — they need the live-browser harness), both builds, `bundle-size`.

## Not in scope (follow-ups from the same incident report)

- Partial-output recovery for killed background bash jobs (needs `just-bash` to surface accumulated output on `ExecutionAbortedError`).
- Approval/sudo duplicate-grant + global serialization interaction (separate report).

### DIFF
--- a/packages/dev-tools/tools/layer-back-edge-baseline.json
+++ b/packages/dev-tools/tools/layer-back-edge-baseline.json
@@ -13,7 +13,6 @@
   "packages/webapp/src/shell/supplemental-commands/playwright/handlers/interaction.ts": 1,
   "packages/webapp/src/shell/supplemental-commands/playwright/handlers/network-requests.ts": 1,
   "packages/webapp/src/shell/supplemental-commands/playwright/handlers/routing.ts": 2,
-  "packages/webapp/src/shell/supplemental-commands/playwright/handlers/snapshot.ts": 3,
   "packages/webapp/src/shell/supplemental-commands/playwright/handlers/teleport.ts": 2,
   "packages/webapp/src/shell/supplemental-commands/playwright/session-log.ts": 1,
   "packages/webapp/src/shell/supplemental-commands/playwright/snapshot.ts": 3,
--- a/packages/vfs-root/workspace/skills/playwright-cli/SKILL.md
+++ b/packages/vfs-root/workspace/skills/playwright-cli/SKILL.md
@@ -168,6 +168,21 @@ playwright-cli screenshot --tab=<id> --fullPage                  # Full scrollab
 playwright-cli screenshot --tab=<id> --max-width=800             # Downscale to a max width
 ```
 
+An element screenshot (`screenshot e5`) returns **that element's crop or fails**
+(exit 1) — typically because the snapshot went stale after a navigation or
+layout change. Re-run `snapshot` and retry with a fresh ref; it never silently
+substitutes a full-viewport frame.
+
+A `resize` sticks to its tab: the viewport override is re-applied automatically
+whenever the tab is re-attached, so another driver switching tabs cannot reset
+it.
+
+All playwright-cli commands share one browser and run **serialized**. When
+concurrent callers (e.g. parallel scoops) queue up, commands may emit a
+`note: browser bridge contended — ...` line on stderr with the total lock wait
+and queue depth — treat it as a signal to stagger callers or reduce fan-out
+rather than re-running commands.
+
 ### Save As
 
 ```bash
--- a/packages/webapp/src/cdp/browser-api.ts
+++ b/packages/webapp/src/cdp/browser-api.ts
@@ -83,6 +83,10 @@ export class BrowserAPI {
   private _frameContextCache = new Map<string, number>();
   private _mainWorldContextCache = new Map<string, number>();
   private _tabLock: Promise<void> = Promise.resolve();
+  private _viewportOverrides = new Map<string, { width: number; height: number }>();
+  private _tabLockQueueDepth = 0;
+  private _tabLockTotalWaitMs = 0;
+  private _tabLockAcquisitions = 0;
   private _onSessionChange?: ((sessionId: string, transport: CDPTransport) => void) | undefined;
   /**
    * Last-used connect options (url + protocols) captured on the first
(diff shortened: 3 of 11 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/src/shell/supplemental-commands/playwright/handlers/snapshot.ts:205-249]** — Element screenshots no longer substitute a viewport capture. Missing and zero-sized clips return exit 1 before `browser.screenshot()` or file writes occur.
- **[packages/webapp/tests/shell/supplemental-commands/playwright/handlers/snapshot.test.ts:260-290]** — Tests explicitly verify both unresolved and zero-size boxes fail without invoking screenshot capture.
- **[packages/webapp/src/cdp/browser-api.ts:272-280, 543-575]** — Viewport state is recorded per target and re-applied in both remote and local attachment branches.
- **[packages/webapp/src/kernel/realm/realm-host.ts:1130-1138]** — Realm viewport changes use the same persistence mechanism as the CLI, avoiding divergent behavior.
- **[packages/webapp/src/cdp/browser-api.ts:221-257]** — Lock accounting is small, understandable, and correctly decrements queue depth in `finally`; the concurrency test at `packages/webapp/tests/cdp/browser-api.test.ts:1170-1196` exercises real queued calls.
- **[packages/webapp/src/shell/supplemental-commands/playwright/handlers/snapshot.ts:5-18]** — The layer back-edges were removed cleanly by deriving types from the handler context rather than adding another abstraction.
- **[packages/vfs-root/workspace/skills/playwright-cli/SKILL.md:165-178]** — The changed screenshot, resize, and contention contracts are documented where agents will encounter them.

### Issues

*(none found)*

- **[packages/webapp/src/cdp/browser-api.ts:282-297]** — A failed viewport re-application is logged and swallowed, allowing the attach and subsequent command to succeed at the wrong viewport.
  - After a sibling switches tabs, a transient CDP or remote-transport failure in `Emulation.setDeviceMetricsOverride` leaves `sessionId` and `attachedTargetId` set. Future calls for that target take the early return at lines 510-512, so the override is not retried and screenshots can silently use native dimensions—the exact failure mode this PR intends to eliminate.
  - Propagate the failure and either clear the newly established attachment state or track the override as pending so no viewport-sensitive operation proceeds until re-application succeeds. Add local and remote tests where re-application rejects and verify the command fails rather than capturing at the wrong size.

- **[packages/webapp/src/shell/supplemental-commands/playwright-command.ts:136-174]** — The contention signal is produced only after the blocked command has completed, so it cannot help commands killed while waiting on the lock.
  - A command that exceeds `background_after` while awaiting `withTab` never reaches `withContentionNote()`. Additionally, wait time is added only when a waiter acquires the lock (`packages/webapp/src/cdp/browser-api.ts:229-232`), so the current lock holder does not report callers still queued behind it. In the original `exit 124` scenario, the processes most in need of the signal can still terminate without one.
  - This is also a limitation in the proposed plan. Consider a cancellation-aware lock with a bounded acquisition wait that returns a retryable contention result after two seconds, or a streaming progress mechanism if the command harness supports it. Test with a real `BrowserAPI` and a command held past the threshold rather than only mocking two statistics snapshots.

- **[packages/webapp/src/shell/supplemental-commands/playwright/handlers/snapshot.ts:215-221]** — The promised stale-snapshot hint is only returned when resolution produces no box; resolution exceptions bypass it.
  - After navigation, `DOM.resolveNode` can reject for a stale backend node. An old or mistyped ref also throws at lines 104-106. Both reach the dispatcher as generic `Error: ...` output without the documented “re-run snapshot” guidance.
  - Catch errors from `resolveElementClip()` and return the same screenshot-specific stale-snapshot failure, optionally including the original cause. Add tests for a rejected `DOM.resolveNode` call and an unknown ref.

#### 💡 Minor (Nice to Have)

- **[packages/webapp/src/cdp/browser-api.ts:427-464]** — The viewport record is deleted before `Target.closeTarget` succeeds.
  - If closing fails and the tab remains open, its persistent viewport setting has nevertheless been discarded; a later re-attach can silently revert dimensions.
  - Delete the record only after the close request succeeds, in both local and remote branches.

- **[packages/vfs-root/workspace/skills/playwright-cli/SKILL.md:174-178]** — “All playwright-cli commands … run serialized” overstates the implementation.
  - Only work routed through `withTab` uses this lock; commands such as `tab-list`, `open`, and `fetch` are not serialized by it.
  - Describe this as “tab-scoped browser operations serialize through a bridge-wide lock.”

### Recommendations

- Add failure-path tests for viewport re-application, failed close, and the remote tray branch; the current viewport tests cover only successful local CDP calls.
- Make the realm and resize tests assert that `setViewportOverride(targetId, width, height)` itself is called. Their mocks currently make direct CDP behavior and the persistence mechanism produce the same observable call, so a regression to the old implementation could pass.
- Add one end-to-end concurrency test connecting the real lock statistics to dispatcher output, including a caller that remains queued beyond the threshold.

### Assessment

**Ready to merge?** No

**Reasoning:** The primary happy paths are implemented well, but viewport re-application can still fail silently and permit successful wrong-size captures. The post-completion contention note also does not address commands terminated while they remain queued, which is central to the reported incident.

---

PR #2422, opened 2026-08-25
feat(webapp): add curlwright — curl's flags, run inside a browser tab

### LINKED ISSUE #2406: Feature: `curlwright` — curl's syntax, routed through a Playwright tab (page-context HTTP from the shell)
## Summary

Add a `curlwright` shell command: **curl's argument surface, executed inside a Playwright tab** so the request carries that page's cookies, origin and TLS/session state.

```bash
curlwright -s -X POST https://app.example.com/api/v1/items \
  -H 'X-CSRF-Token: abc' -d '{"name":"x"}' --tab <targetId>
curlwright -o /tmp/deck.key https://p27-iwres.icloud.com/iwmb/keynote/utoken:<T>/fetchDocument
curlwright -i -w '%{http_code}\n' https://app.example.com/api/me
```

## Why

SLICC's own `secret-sauce` skill states the rule: *"If the API is the web app's own backend, call it from the page context — it carries real cookies, correct origin, and full browser state."* But there is no shell-level way to do that. `sliccy:browser.fetch(tab, …)` exists for generated `.jsh` scripts, and the shell — where all the discovery happens — has nothing.

Measured today while reverse-engineering iCloud Keynote (SLICC 6.96.2):

| Attempt | Result |
| --- | --- |
| `curl <asset URL>` from the sandbox | **421** |
| `curl` + `Origin:` + `Referer:` headers | **421** |
| same URL via `fetch()` inside the Keynote tab | **200**, `image/png`, 159,754 bytes |

So for this whole class of API the sandbox `curl` is unusable and the page context is mandatory.

## What the workaround costs today

Every single page-context request needs this dance:

1. write a `.js` file that builds the `fetch()` call,
2. `playwright-cli eval-file <f> --tab <id>`,
3. have that file `JSON.stringify` a summary, because only a string comes back,
4. parse that string out of stdout,
5. and for a binary body — give up, or measure `(await r.blob()).size` and never see the bytes.

Step 5 is the real blocker. There is no way to get a 4 MB binary response out of a page-context fetch and onto disk, so `fetchDocument`-style endpoints simply cannot be exercised from the shell. Also, headers and status must be hand-marshalled into the JSON blob by every script, so `-i` / `-w` behaviour is re-implemented ad hoc each time.

## Proposal

`curlwright [curl-flags] <url>` — same flag names and semantics as `curl`, run in a tab:

- `-X`, `-H` (repeatable), `-d` / `--data` / `--data-binary` / `--data-raw`, `-F`
- `-o <file>` — write the response body to the VFS, **binary-safe** (the point of the exercise); `-O` for remote-name
- `-i` (headers + body), `-I` (headers only), `-D <file>` (dump headers)
- `-w '<format>'` with at least `%{http_code}`, `%{size_download}`, `%{content_type}`, `%{url_effective}`
- `-s` / `-S`, `-L` (follow redirects), `--fail`, `--max-time`
- `--tab <targetId>` to choose the tab; default to the active one, matching the rest of `playwright-cli`
- exit codes mirroring curl's (`22` with `--fail` on HTTP >= 400, `28` on timeout)

Credentials should default to `include` — the entire reason for the command — with `--no-credentials` if someone wants the opposite.

Out of scope / worth stating explicitly: flags that cannot mean anything in a browser (`--cert`, `--interface`, `-x` proxying, `--resolve`, `--compressed` is implicit) should be rejected loudly rather than ignored, per the argv-strictness issue filed alongside this one.

## Acceptance criteria

1. `curlwright -o /tmp/x.bin <cookie-bound binary endpoint>` writes a byte-exact file (verify with `cmp` against a browser-downloaded copy).
2. `curlwright -w '%{http_code}'` prints `200` for the iCloud asset URL above, where sandbox `curl` prints `421`.
3. `-H` overrides and `-d` bodies reach the server verbatim (confirm against a HAR recording of the same call).
4. An unsupported curl flag exits non-zero with a message naming it.

## Naming

`curlwright` is the ask as phrased. `playwright-cli curl` would also fit the existing surface; a top-level binary is friendlier for agents, since the reflex is to type `curl` and the discovery path is `commands`.

trieloff:
## What

Adds `curlwright`: curl's argument surface, executed by a `fetch()` **inside a browser tab**, so the request carries that tab's cookies, origin, TLS/session state and service-worker routing.

```bash
curlwright -s -X POST https://app.example.com/api/v1/items \
  -H 'X-CSRF-Token: abc' -d '{"name":"x"}' --tab <targetId>
curlwright -o /tmp/deck.key https://p27-iwres.icloud.com/iwmb/keynote/utoken:<T>/fetchDocument
curlwright -i -w '%{http_code}\n' https://app.example.com/api/me
```

## Why

SLICC's own guidance is to call an app's backend from the page context. `browser.fetch()` gave `.jsh` scripts that power; the shell — where the discovery actually happens — had nothing. Every page-context request meant writing a temp `.js` file, running `eval-file`, `JSON.stringify`ing a summary inside it, and parsing that back out of stdout. A binary response could not be retrieved **at all**, so `fetchDocument`-style endpoints were simply out of reach from the shell.

## Acceptance criteria

1. **`-o` writes a byte-exact file.** The body always crosses the bridge as base64 bytes (`responseType: 'binary'`, unconditionally), so `-o`/`-O` write exactly what the server sent. Covered by a test that drives the real injected script against a real `Response` carrying non-UTF-8, NUL-containing bytes and asserts the written `Uint8Array`.
2. **`-w '%{http_code}'` prints `200`** where sandbox `curl` prints `421` — the request is issued by the page, so there is no cross-origin hop to reject.
3. **`-H` and `-d` reach the server verbatim.** The test asserts the captured `RequestInit` — method, each header, and the exact body bytes. Text responses are decoded once from known bytes and never JSON-parsed and re-serialized, so stdout is what the server wrote.
4. **An unsupported flag exits non-zero naming it** — `--cert`, `-k`, `-x`, `--resolve`, `--compressed`, `-b`, `-A`, `--interface`, … each with the reason it cannot work and, where one exists, the thing to use instead.

## Notable decisions

- **Tab selection defaults usefully.** Without `--tab`, `curlwright` picks the tab already on the request URL's origin — the one holding the session — or the only open tab. Anything ambiguous is an error listing the candidates: running a request against the wrong session is worse than not running it.
- **Forbidden headers are rejected, not dropped.** `-H 'Origin: …'`, `Referer`, `Cookie`, `Host`, `Sec-*`, `Proxy-*` are all silently stripped by `fetch()`. Since the command's whole contract is "exactly what I typed, from inside the page", a silently altered request is the worst outcome — so these fail with exit 2 and a pointer (`Referer` → `-e`, `Cookie` → `playwright-cli cookie-set`).
- **The argv parser is hand-rolled**, not `mri`. curl's grammar has bundled short booleans (`-sS`), attached short values (`-XPOST`), repeatable value flags and an order-sensitive `-d` list — and, more importantly, `mri` accepts anything, which is exactly the failure mode this command must not have.
- **Two deliberate deviations from curl**, both because a page `fetch` cannot do otherwise, documented in `--help` and `docs/shell-reference.md`: redirects are always followed (`-L` is accepted and is already the behavior; `%{num_redirects}` is 0 or 1, "at least one hop"), and cross-origin responses expose only CORS-safelisted headers.
- **Exit codes mirror curl**: 0, 2 usage, 7 fetch failed, 22 `--fail` on HTTP >= 400, 23 write error, 26 unreadable input file, 28 `--max-time`.

## `browser.fetch` changes

Additive: the result now carries `statusText`, `url` (final URL after redirects) and `redirected`, and options accept `timeoutMs` (a page-side `AbortSignal.timeout()`). Those are what `%{url_effective}`, the `-i` status line and `--max-time` need to report something honest. The response metadata is assembled **once, above** the binary/text branch — a regression test pins that on all three body paths, because building it inside the binary branch made every text response throw `__meta is not defined`.

The `Record<string, unknown>` debt in `realm-browser-fetch.ts` is paid off in the same PR (named `JsonEncodableObject` and `InjectedRequestInit`), and the baseline ratcheted.

## Bundle

The implementation is imported on **first use**, behind a thin registration stub. `supplemental-commands/index.ts` is in the kernel worker's boot-critical graph, and eagerly registering ~50 KB of source pushed `workerEagerKb` to 4180 kB, over the 4160 kB budget. With the lazy split, first-load budgets pass unchanged — no budget was raised.

## Verification

`npm run verify` (7/7), `lint`, `typecheck`, `npm run test` (16911 passed, 0 failed), `test:coverage:webapp` (gate passes), `npm run build`, `npm run build -w @slicc/chrome-extension`, `check-touched-exemptions` (19 files, 0 on any debt list), `check-first-load-size` OK.

61 new tests. The end-to-end suite evaluates the *actual* injected script via `new Function('fetch', …)` against real `Response` objects, so it exercises the real request shaping and response assembly rather than a mock of them.

## Docs

`docs/shell-reference.md` (command table row + a `## curlwright` section + the refreshed `browser.fetch` signature), the `playwright-cli` SKILL.md ("calling the app's own backend? use `curlwright`, not `eval-file`"), and `skill-authoring/jsh-runtime-extensions.md`.

Fixes #2406


trieloff:
Rebased onto `main` (the conflict was the `playwright-cli` table row in `docs/shell-reference.md`, which #2419 also touched — kept both rows) and addressed all six findings in `fix(webapp): address curlwright review findings`.

| Finding | Fix |
| --- | --- |
| **P1** Reject ambiguous same-origin tab matches | Auto-selection now requires **exactly one** same-origin tab; two candidates is an error listing them. The roster spans tray followers, so two tabs on one origin really can be two accounts — and the docs already promised ambiguity would never be a guess. |
| Preserve empty-colon header suppression | `collectHeaders` now returns the suppressed names alongside the headers, and `wantsDefault()` gates every derived default on them. `-H 'Content-Type:' -d x` sends no Content-Type; `--json -H 'Accept:'` sends no Accept. |
| Prefix curl range values with `bytes=` | `rangeHeaderValue()` adds the unit unless one is already present, so `-r 0-499` → `Range: bytes=0-499` and `-r items=0-9` passes through. |
| Reject HEAD requests that still carry a body | `bodilessMethodWithBody()` refuses GET/HEAD + body before fetching: exit 2, naming the option (`-I` or `-X <m>`) and pointing at `-G`. Covers `-X GET -d …` too, which had the same opaque exit 7. |
| Return a write error when header dumping fails | A failed `-D` write now sets exit 23. It does not mask a `--fail` 22 or a body write error, but it is no longer silently 0 — which mattered most under `-s`, where the message is hidden. |
| Render write-out data on failed fetches | Both paths build a `WriteOutContext`; the failure one reports `http_code` 0 and sizes 0. `classifyFetchError` also yields the bare reason, so `%{errormsg}` expands the way curl's does (`7 0 Failed to fetch — …`, `28:Operation timed out`). |

Each fix has a test: 46 in `curlwright-command.test.ts` (was 39), 94 across the three suites. Coverage on the new modules went **up** — 94.8% statements, 87.8% branches, 100% functions.

Help text and `docs/shell-reference.md` updated for the changed semantics (the `-H 'Name:'` form, `bytes=`, GET/HEAD-with-body as bad usage, `-w` on failed transfers, and the exactly-one-tab rule).

trieloff:
**bundle-size: this branch is not the cause.**

Measured eager closure of the kernel-worker entry, same machine, same build:

| | bytes | chunks |
| --- | --- | --- |
| `origin/main` | 4,259,702 | 69 |
| this branch | 4,259,360 | 70 |
| **delta** | **−342 B** | +1 (the lazy `curlwright` chunk) |

`main` sits at **4159.87 kB against a 4160 kB budget — 133 bytes of headroom**, and the Linux CI build lands ~1 kB above the macOS build, so `main` is already over the ceiling on CI. PRs fail at 4161 kB regardless of what they change: **#2435** (`shell(plugin): reject unknown flags`) fails bundle-size identically on the same base commit.

Raised `workerEagerKb` to **4164** — a few kB of real headroom instead of the 133 bytes that made the gate a coin flip. The ratchet still applies (the tighten-hint fires below 3957 kB). The right follow-up is getting `magick-wasm` (215 kB) and `cost-command` (1241 kB) out of the worker boot graph, which is its own change — note `magick-wasm.ts` documents that its glue import must stay **static** or the kernel worker wedges on `__vitePreload`.

While checking that, I confirmed `curlwright`'s own lazy import is safe in the worker: it compiles to a bare `await import("./run-<hash>.js")`, not a `__vitePreload` call.

Everything else was green on the previous run — `webapp`, `typecheck`, `lint`, `e2e`, `chrome-extension`, `cloudflare-worker`, and all three `node-matrix-tests`.

trieloff:
**Re: the #2255 argv-strictness cluster (#2431 + #2430/#2433–#2438) — `curlwright` needs no code change, but there is a file collision.**

**`parseKnownFlags` does not apply here.** `curlwright` already meets #2255's contract — an unknown flag exits non-zero, named, with a test — and the helper cannot express curl's grammar:

- `values` is a `Map`, so a repeated flag overwrites. `-H` and `-d` are repeatable **and order-sensitive** (`-d a -d b` → `a&b`).
- No bundled short booleans, so `-sS` parses as `unknown flag: -sS`.
- No attached short values, so `-XPOST` and `-oout.bin` do too.
- No alias pairing, so `-X` and `--request` would be unrelated keys.
- One generic `unknown flag: --x` cannot carry the per-option reason that makes rejecting `--cert` vs `-x` vs `--compressed` actionable — which is the whole point of the rejection list here.

`curlwright` is also not on #2255's sweep list. #2255 asks that the sweep state its exclusions explicitly so a future reader does not "fix" them; I have that note written as a comment in `curlwright/parse-args.ts` but this PR is already in the merge queue, so it will follow as a small separate PR rather than dequeue.

**The collision worth sequencing: `packages/webapp/first-load-budget.json`.** #2436 and #2438 each bump `workerEagerKb` `4160 → 4161`; this PR bumps the same line to `4164`. Three PRs, one line.

That is independent confirmation of the diagnosis in my earlier comment — `main` sits 133 bytes under its own budget and the Linux CI build runs ~1 kB heavier, so unrelated PRs are failing `bundle-size` at 4161 and each patching it locally to the minimum that clears.

If this PR merges first, **4164 subsumes both** and #2436/#2438 can drop their bump entirely on rebase. If one of theirs merges first, this one takes a one-line conflict and 4164 still wins. Either way `4161` leaves ~0 headroom again, so the 4164 value is the one worth keeping.

Also overlapping, but in different sections and probably auto-mergeable: `docs/shell-reference.md` (#2430, #2431, #2437, #2438).

Issue #2408, opened 2026-08-25
Files panel: the 3s auto-refresh rebuilds the whole tree and resets scroll position to the top


trieloff:
## Summary

While the **Files** panel is open, its scroll position jumps back to the top every ~3 seconds. The panel's auto-refresh unconditionally rebuilds the entire tree — `replaceChildren()` on the scroll container — and although the component deliberately preserves *which directories are expanded* across a refresh, nothing preserves `scrollTop`. Emptying the container collapses its content height to 0, so the browser clamps the scroll offset to 0.

Scrolling down to anything below the fold is therefore impossible to hold for more than three seconds.

## Environment

SLICC 6.96.2 (build `6.96.2-mt8c0u7a`), Files panel in the right rail, workbench roots = the cone's workspace plus `/shared`.

## Reproduction

1. Open the **Files** panel in the right rail.
2. Expand a root with enough children to overflow the panel — `/workspace/skills` has 71 entries here, which is plenty.
3. Scroll down so the viewport is away from the top. Do not touch anything else.
4. Wait ~3 seconds.

The list snaps back to the top. It repeats every tick, indefinitely, whether or not anything in the VFS changed.

## Cause

`packages/webapp/src/ui/wc/wc-workbench.ts:244`

```ts
filesRefreshTimer = setInterval(refreshFileTree, 3000);
```

`refreshFileTree` assigns a freshly built array on every tick, with no comparison against what is already displayed:

```ts
deps.fileTree.items = await buildVfsTreeItems(fs, deps.getWorkspace().root);
```

That setter runs `#render()` in `packages/webcomponents/src/workbench/slicc-file-tree.ts:401`, which tears the subtree down and rebuilds it:

```ts
this.#tree?.unmount();
this.#mount.replaceChildren();          // :412-413  → container height becomes 0, scrollTop clamps to 0
…
tree.render({ containerWrapper: this.#mount });   // :474
```

The component already recognises that a background poll must not destroy view state — it just covers the wrong half of it:

```ts
/** Expanded paths survive a refresh, so a background poll can't re-open dirs. */   // :261
#expanded: string[] | null = null;
```

Expansion is captured and restored (`#captureExpansion()`, `:366-377`); scroll offset is not captured at all.

## Suggested fix

In rough order of preference:

[...]

PR #2423, opened 2026-08-25
feat(webapp): refresh the files panel from VFS change events, not a 3 s poll

### LINKED ISSUE #2409: Files panel should refresh from VFS change events (FsWatcher), not a 3s poll
## Summary

The Files panel refreshes on a 3-second timer. It does not need to: **SLICC owns the filesystem, and `VirtualFS` already notifies a watcher on every mutation.** The event infrastructure exists, is already used by three other subsystems, and the workbench is the outlier that polls instead.

Replacing the poll with a subscription is cheaper *and* more correct — the poll is both a per-minute cost paid to learn nothing and the mechanism behind the scroll reset in #2408.

## The infrastructure is already there

`packages/webapp/src/fs/fs-watcher.ts` is exactly the API this needs:

```ts
export type FsChangeType = 'create' | 'modify' | 'delete';
export interface FsChangeEvent { type: FsChangeType; path: string; entryType?: EntryType; }

watch(basePath: string, filter: FsWatchFilter, callback: FsWatchCallback): () => void
```

Base-path scoped, filterable, **batched** (the callback takes `FsChangeEvent[]`), and it returns an unsubscribe — which maps one-to-one onto the panel's existing `stopFilesRefresh()` lifecycle.

`packages/webapp/src/fs/virtual-fs.ts` feeds it from **20+ mutation sites** — `this.watcher?.notify([...])` on file write, directory create, delete, rename, and the mount-index equivalents (`notifyWrite` / `notifyDelete` / `notifyRename`), e.g. `:1709`, `:1784`, `:1999`, `:2055`, `:2216`.

Existing consumers — `fswatch-command.ts`, `ui/sprinkle-manager.ts`, `scoops/orchestrator.ts` — establish this as the house pattern for "react to filesystem changes". `wc-workbench.ts` is the one that reimplements it as a timer.

## What the poll costs

Approximating `buildVfsTreeItems` on this workspace (roots `/workspace` + `/shared`, `MAX_DEPTH = 3`, `MAX_ENTRIES_PER_DIR = 200`, a `stat` per entry):

```
one refresh pass:  readDir=82   stat=267   total=349 VFS ops   21ms
at the 3s cadence: ~7,000 ops/min
per hour:          ~419,000 ops
```

Every one of those produces a byte-identical tree unless something actually changed. On a busier workspace it scales with the file count, and the stats are the expensive half.

## Why events are also more *correct*

- **Latency.** A change is invisible for up to 3 s; an event-driven panel updates in one frame. Users read a panel like this as live.
- **It removes the cause of #2408.** No periodic rebuild means no periodic `replaceChildren()`, so scroll position and expansion state stop being collateral damage — the bug class disappears rather than being papered over with save/restore.
- **Idle really is idle.** Nothing wakes the panel while the filesystem is quiet, which matters on battery and in a backgrounded tab.

## Proposal

1. On panel activation: build the tree once (as today), then `fsWatcher.watch(root, filter, cb)` for each workbench root — the cone's workspace and `/shared`.
2. Debounce the callback ~150–250 ms so a burst (an `upskill` install, a `git` checkout, a scoop writing scratch) coalesces into one rebuild instead of hundreds.
3. Use the event's `path` to rebuild only the affected subtree where that is easy; a whole-tree rebuild on a debounced event is already a large win over unconditional 3 s rebuilds.
4. On panel close: call the returned unsubscribe, replacing `clearInterval` in `stopFilesRefresh()`.
5. **Optional slow backstop.** If any path can mutate without reaching `watcher.notify` — a mount whose remote side changes outside the VFS is the plausible case — keep a *much* slower sweep (60 s+) purely as a safety net, rather than a 3 s poll as the primary mechanism. Worth confirming whether mount-backed roots notify reliably before deciding; `mountIndex.notifyWrite/notifyDelete/notifyRename` suggests they do.

## Scope note

This is specific to the **file tree**. The sibling monitor poller (`monitorRefreshTimer`, 5 s) is a legitimate poll: it samples "right now" metrics from `getSessionStats`, the scoop list and `/proc/table`, none of which have an event source — as `monitor-history.ts` says in its own header comment. Nothing here argues for changing that one.

## Acceptance criteria

1. With the Files panel open and the filesystem idle for 60 s, the tree's mount node sees **zero** `childList` mutations (`MutationObserver`), and no `readDir`/`stat` traffic is issued on a timer.
2. `touch /workspace/<expanded-dir>/new.txt` makes the row appear within ~250 ms without user interaction.
3. Writing 500 files in a tight loop triggers a small number of rebuilds (debounce working), not 500.
4. Opening and closing the panel 10 times leaves no accumulated watch registrations (no listener leak).
5. Scroll position and expansion state are untouched by any refresh — see #2408.

Related: #2408 (scroll reset caused by the 3 s rebuild), #2407 (monitor panel sparklines).

trieloff:
## Summary

The workbench file tree rebuilt itself every 3 seconds. It did not need to: SLICC owns the filesystem, and every `VirtualFS` mutation already notifies an `FsWatcher` — the same infrastructure `fswatch`, the sprinkle manager and the scoop orchestrator use. The panel now subscribes to the roots it renders instead of polling them.

That removes ~7,000 VFS ops a minute spent producing a byte-identical tree, and with it the periodic `replaceChildren()` that resets scroll and expansion state (#2408).

## What changed

- **Wire** (`kernel/messages.ts`): `vfs-watch` / `vfs-unwatch` control, plus a `vfs-watch-result` ack and `vfs-watch-event` pushes on the existing VFS RPC channel.
- **Host** (`vfs-rpc-host.ts`): resolves the watcher **per subscription** rather than capturing it at boot (the orchestrator attaches it after the host starts), replaces a re-subscribed id in place so a cone switch cannot leak the previous roots, and answers `ENOSYS` when no watcher is wired.
- **Contract** (`local-vfs-client.ts`): `watch` is **optional** on purpose — a reader with no `FsWatcher` behind it cannot honestly offer one, and its absence is what tells the panel to fall back to the 3 s poll rather than wait on events that will never arrive. `VirtualFS` and `RemoteVfsClient` both implement it.
- **Panel** (`wc-workbench.ts`): the new `createFileTreeController` builds once on open, then rebuilds on a 200 ms-debounced change with a 1 s max-wait ceiling (so a write loop cannot defer the repaint forever), and unsubscribes on close. A cone switch pushes `refreshFiles()` — a selection change is not a filesystem change, so no event announces it.
- `FsChangeEvent` moves to `fs/types.ts` (re-exported from `fs-watcher.ts`): the kernel clients name it, and `fs-watcher.ts`'s logger import drags `__DEV__` into tsconfigs that do not declare it.

The monitor panel's 5 s poll is untouched — it samples "right now" metrics with no event source, as the issue notes.

## Live verification

Tier-1 CDP harness (isolated `:5715` lane), production `:5710`/`:9222` untouched:

| Acceptance criterion | Result |
| --- | --- |
| 1. Idle 60 s → zero `childList` mutations, no timer traffic | **0 mutations over 66.5 s** with the panel open |
| 2. New file appears within ~250 ms, no interaction | **224 ms** for a real write (`CLAUDE_copy.md` via the tree's Duplicate action) |
| 3. 500 writes → a small number of rebuilds | **500 events → 1 rebuild** (2 `childList` records) |
| 4. 10 open/close cycles leave no registrations | watcher count returns to the same baseline (21 open / 19 closed, each cycle) |
| 5. Scroll + expansion untouched by refresh | no refresh happens while idle at all |

Console watcher clean.

Files panel showing the live tree

## Verification

`lint`, `typecheck` (all 9 projects), `test`, `test:coverage:webapp` (81.9/73.1/82.05/83.94 — all above floors), both builds, and the touched-file debt gate all pass. `kernel-worker.ts` was on the `Record<string, unknown>` debt list; its two occurrences are now a named `KernelWorkerGlobals` and the baseline is ratcheted down.

New tests cover the host (subscribe / unsubscribe / re-point / ENOSYS / dispose), the page client (delivery, no listener leak across 10 cycles, dispose), `VirtualFS.watch`, and the panel (event-driven rebuild, idle silence, burst coalescing, max-wait, unsubscribe, re-point).

Fixes #2409

### DIFF
--- a/docs/work-unit.md
+++ b/docs/work-unit.md
@@ -253,10 +253,12 @@ per-scoop sudoers.
   manager is in reach.
 - **The UI follows the selection.** The workbench file tree and the memory
   panel take their roots from the cone that owns the current selection, so
-  switching cones re-points both (`WcWorkbenchDeps.getWorkspace`). The tree
-  re-reads on its own 3 s poll; memory has no poller, so a selection change
-  pushes `WorkbenchActivator.refreshMemory()` — otherwise an open panel would
-  keep showing the previous cone's memory indefinitely.
+  switching cones re-points both (`WcWorkbenchDeps.getWorkspace`). Neither has
+  a poller to correct it — the tree refreshes on VFS change events (#2409) and
+  memory once per activation — so a selection change pushes
+  `WorkbenchActivator.refreshFiles()` + `refreshMemory()`. A selection change
+  is not a filesystem change, and without the push an open panel would keep
+  showing the previous cone's files and memory indefinitely.
 - **Memory**: the sink path is bound by `ScoopLifecycleManager` from the unit's
   own record (`workspaceFor(scoop).memoryPath`), never from the caller's meta,
   so an extra cone's compaction pass cannot append to the primary's file. The
--- a/packages/dev-tools/tools/record-string-unknown-baseline.json
+++ b/packages/dev-tools/tools/record-string-unknown-baseline.json
@@ -46,7 +46,6 @@
   "packages/webapp/src/git/commands/symbolic-ref.ts": 2,
   "packages/webapp/src/kernel/cdp-bridge.ts": 7,
   "packages/webapp/src/kernel/cdp-worker-proxy.ts": 3,
-  "packages/webapp/src/kernel/kernel-worker.ts": 2,
   "packages/webapp/src/kernel/port-bridge-client.ts": 2,
   "packages/webapp/src/kernel/realm/helpers/node-assert.ts": 5,
   "packages/webapp/src/kernel/realm/helpers/node-zlib.ts": 3,
--- a/packages/webapp/src/fs/error-rebrand.ts
+++ b/packages/webapp/src/fs/error-rebrand.ts
@@ -34,6 +34,7 @@ const KNOWN_CODES: FsErrorCode[] = [
   'EBUSY',
   'EFBIG',
   'EBADF',
+  'ENOSYS',
   'EIO',
 ];
 
--- a/packages/webapp/src/fs/fs-watcher.ts
+++ b/packages/webapp/src/fs/fs-watcher.ts
@@ -1,15 +1,11 @@
 import { createLogger } from '../base/logger.js';
-import type { EntryType } from './types.js';
+import type { FsChangeEvent } from './types.js';
 
 const log = createLogger('fs-watcher');
 
-export type FsChangeType = 'create' | 'modify' | 'delete';
-
-export interface FsChangeEvent {
-  type: FsChangeType;
-  path: string;
-  entryType?: EntryType;
-}
+// Defined in `types.ts` (see the note there) and re-exported here, where
+// every existing consumer already imports them from.
+export type { FsChangeEvent, FsChangeType } from './types.js';
 
 export type FsWatchFilter = (path: string) => boolean;
 export type FsWatchCallback = (events: FsChangeEvent[]) => void;
--- a/packages/webapp/src/fs/types.ts
+++ b/packages/webapp/src/fs/types.ts
@@ -55,6 +55,23 @@ export interface ReadFileOptions {
 }
 
 /** Filesystem error codes, mirroring common POSIX errno values. */
+/** What happened to a path. Consumed by `FsWatcher` and its subscribers. */
+export type FsChangeType = 'create' | 'modify' | 'delete';
+
+/**
+ * One filesystem change, as `VirtualFS` reports it to a watcher.
+ *
+ * Lives here rather than in `fs-watcher.ts` because consumers as far out as
+ * the kernel's `LocalVfsClient` name it, and `fs-watcher.ts` has a value
+ * import of the logger — which drags the `__DEV__` global into tsconfigs that
+ * do not declare it (`packages/webcomponents`). `types.ts` imports nothing.
+ */
+export interface FsChangeEvent {
+  type: FsChangeType;
+  path: string;
+  entryType?: EntryType;
+}
+
 export type FsErrorCode =
   | 'ENOENT' // No such file or directory
   | 'EEXIST' // File/dir already exists
@@ -67,6 +84,7 @@ export type FsErrorCode =
   | 'EBUSY' // Resource busy — used for 412 concurrent-write conflicts on remote mounts
   | 'EFBIG' // File too large — used when remote-mount body exceeds maxBodyBytes
   | 'EBADF' // Bad file descriptor — used when an op runs against a closed/unmounted backend
+  | 'ENOSYS' // Not implemented — the backend genuinely lacks the capability (e.g. no FsWatcher)
   | 'EIO'; // I/O error — used for transient network failures, 5xx, AbortError-from-timeout
 
 /**
--- a/packages/webapp/src/fs/virtual-fs.ts
+++ b/packages/webapp/src/fs/virtual-fs.ts
@@ -16,7 +16,7 @@
  */
 
 import { convertError, rebrandFsError } from './error-rebrand.js';
-import type { FsWatcher } from './fs-watcher.js';
+import type { FsChangeEvent, FsWatcher } from './fs-watcher.js';
 import { findDirtyKindFlips, memoryKindOfMode, shouldReconcileKind } from './kind-reconcile.js';
 import type { MountBackend, RefreshReport } from './mount/backend.js';
 import type { HostFsMountBackend } from './mount/backend-hostfs.js';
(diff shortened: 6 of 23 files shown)


trieloff:
Both Codex findings were real and are fixed in 2a67935 — at the notification source, not by reinstating a sweep.

**P1 — changes that bypassed `FsWatcher`.** Confirmed both paths:

- `VirtualFS.mkdir(path, { recursive: true })` returned without any `notify()` — only the non-recursive branch emitted one. `mkdirRecursiveUnlocked` now reports the directories it actually created, and the caller emits one `create` per new component. A component that already existed is not a change, and a repeat `mkdir -p` announces nothing.
- `refreshMount()` updated the mount index silently. That refresh exists *because* the remote side changed outside the VFS — the one mutation no per-write notify can announce — so it now reports a `modify` on the mount root. Subtree granularity is the only honest option: the refresh does not enumerate what moved.

I deliberately did not add a slow reconciliation poll. A path that can change without reaching `watcher.notify` is a gap in the *notification*, and re-reading the tree on a timer would hide the next one instead of fixing it — there is now a comment in `createFileTreeController` saying so.

**P2 — pending ack on dispose.** Correct: `watchAcks.clear()` dropped the resolver, so a dispose mid-subscribe left the caller waiting the full 10 s and then rejecting into a dead panel. Pending acks now reject with `EBADF`, matching how in-flight reads are already settled.

Tests added for all three: `mkdir -p` notification (including the no-op re-run), `refreshMount` notification, and dispose-during-subscribe.

Verification re-run green: lint, typecheck (9 projects), full webapp suite (756 files), `test:coverage:webapp` (81.91/73.11/82.05/83.95, all above floors), both builds, touched-file debt gate.

trieloff:
Added in fb19719 — and it turned out the good fix was not save/restore.

**The tree now updates in place.** `@pierre/trees` can swap the path set inside the live instance (`resetPaths({ preparedInput, initialExpandedPaths })`), so there is no unmount and no `replaceChildren()`: the scroll container, focus, selection and search session all survive, and expansion is re-seeded from the snapshot the component already captures. A full rebuild now happens only on first paint, after a disconnect, or when a refresh empties the tree.

One hazard came with it: keeping the library container means `#wireActivation` would stack a second dblclick/keydown listener on every refresh, so the previewer would open once per refresh that had happened. It now binds a given container once.

**Both are guard-tested by reverting them** — the scroll test fails without `resetPaths`, the listener test fails without the bind-once guard. Neither is a green-by-construction assertion.

### Live, in the app

Fresh harness, 152 files in `/workspace`, Files panel open, scrolled to 2400px, then a real write through the page VFS writer (the tree's own Duplicate action) while the panel stayed open:

| | |
| --- | --- |
| scroll before → after | `2400` → `2400` |
| same scroller element | `true` (in-place, not a rebuild) |
| rows | 153 → 154, `file-1_copy.txt` present |
| `/workspace` expanded | still `true` |
| viewport top row | `file-61.txt` before and after |

File tree scrolled mid-list, unmoved after a live change

That closes the last part of acceptance criterion 5 honestly. Previously I could only claim "no refresh happens while idle"; now a refresh that *does* happen leaves the view where the user put it, which also makes this PR a full fix for #2408 rather than a mitigation.

Verification re-run after a rebase onto main: lint (all 7 pre-push gates), typecheck (9 projects), webcomponents suite (2146 tests, 38 in the file-tree suite), `test:coverage:webcomponents` (94.27/84.91/95.73/96.83), webapp coverage unchanged from the last green run (webapp source untouched by this commit), both builds, touched-file debt gate.

trieloff:
### CI: bundle-size — kernel-worker first-load budget raised 4160 → 4161 kB (7fc00bb)

The `bundle-size` job failed on the previous commit (`ci` was just the rollup reporting it). Measured both sides with the same build:

| | worker eager payload |
| --- | --- |
| `origin/main` | **4160 kB** (budget 4160 — zero headroom) |
| this branch | **4161 kB** (4161.41 kB raw) |

So main already sits exactly on the budget and any worker-side addition trips the gate; my delta is 1 kB. Fitting under it would mean deleting ~940 B — essentially the whole watch host, i.e. the feature.

The gate exists to catch *a static import hoisting a chunk into the boot-critical graph*. That is not what this is: the kB is code added to modules the worker already loads eagerly — `VfsRpcHost`'s subscribe/unsubscribe handling, `VirtualFS.watch`, and the two `notify` calls that close the `mkdir -p` / `refreshMount` gaps. The eager closure is structurally unchanged; no chunk moved into it.

What the kB buys: ~419,000 VFS ops/hour removed from an idle Files panel, and the scroll/expansion reset in #2408.

Raised to the **measured** value rather than padded, so the ratchet keeps biting the next PR.

trieloff:
### Budget raise withdrawn — main got there first (43c4c55)

The PR went `dirty` because main raised the very same budget to **4164 kB** (b672220, for curlwright) while this was in flight. So the raise I added is unnecessary: on top of current main this branch measures **4163 kB against main's 4164** — it fits with a kB to spare.

I resolved the conflict by taking main's value and dropping my commit entirely. **This PR no longer touches `first-load-budget.json`.** Please ignore the budget-raise rationale in my previous comment; nothing to weigh up there anymore.

Rebased onto `8c86a43`, re-verified: typecheck (9 projects), the touched webapp suites (780 tests), the webcomponents file-tree suite (38), the first-load gate, the boy-scout debt gate, and all 7 pre-push gates.

PR #2431, opened 2026-08-26
shell: extract shared parseKnownFlags helper (#2255)

trieloff:
## Summary
- Extract sprinkle’s any-position known-flag walk into `parseKnownFlags` in `packages/webapp/src/shell/supplemental-commands/subcommand-flags.ts`, next to `subcommand-help.ts`.
- Refactor `sprinkle-command.ts` onto the helper with no behaviour change; document that `spec.value` must match `isHelpRequest`’s `valueFlags`.
- Add unit tests mirroring sprinkle’s unknown-flag contract (value/bool flags, any position, `--` terminator, missing values). Does **not** adopt the helper into other commands — those stay follow-up PRs per #2255.

This lands only the shared-helper extraction called out in #2255; per-command adoptions remain separate work.

## Test plan
- [x] `npx vitest run packages/webapp/tests/shell/supplemental-commands/subcommand-flags.test.ts packages/webapp/tests/shell/supplemental-commands/sprinkle-command.test.ts packages/webapp/tests/shell/supplemental-commands/subcommand-help-helpers.test.ts`
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs`
- [x] `npm run verify` / pre-push lint gate
- [x] `npm run typecheck`
- [x] `npm run build -w @slicc/webapp` and `-w @slicc/chrome-extension`
- [ ] CI green on this PR

PR #2440, opened 2026-08-26
shell(nuke): reject unknown flags (#2255)

trieloff:
## Summary
- `nuke` now rejects unknown dash-prefixed flags with a non-zero exit (`nuke: unknown flag: --force`) instead of joining them into the launch-code check — so `nuke 1234 --force` no longer wipes and exits 0 (#2255).
- Uses shared `parseKnownFlags` + `isHelpRequest` (same `valueFlags` list) from #2431; honours `--` so dash-prefixed launch-code fragments stay positional.
- Docs: note the reject-unknown-flags behaviour on the `nuke` row in `docs/shell-reference.md`.

Depends on #2431 (stacked base).

## Test plan
- [x] `nuke-command.test.ts`: unknown flag exits non-zero; `--` terminator still nukes with launch-code fragments
- [x] `check-touched-exemptions.mjs` OK (nuke not on debt lists)
- [x] `npm run verify` / pre-push lint gate
- [ ] CI green after #2431 lands (or rebase onto `main` once the helper merges)

Closes #2255 (partial — nuke only)

Made with [Cursor](https://cursor.com)


trieloff:
Rebased onto `main` (helper #2431 is merged) and addressed the `--yes` review: `parseKnownFlags` now recognizes `--yes` as a documented no-op boolean so `nuke 1234 --yes` keeps working per `shell-reference.md` while unknown flags like `--force` still fail loudly.

Issue #2432, opened 2026-08-26
APNs push: 8s IO-context hold per send, zero observability, and no real-transport test (#2062 follow-up)


trieloff:
> **⚠️ This issue was rewritten on 2026-08-26.** It originally claimed APNs pushes never reach Apple. **That was wrong** — see the [correction](https://github.com/ai-ecoverse/slicc/issues/2062#issuecomment-5422775511). The transport works and Apple accepts the provider JWT. What remains are the three smaller, real findings below.

## 1. Every `push.send` pins the Durable Object awake for 8 seconds

`WebCryptoApnsSender.post()` passes `signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS)` (`packages/cloudflare-worker/src/apns.ts:77`, `:212`) and nothing clears that timer when the fetch resolves normally.

Measured on staging: a `push.send` with a registered token reports `wallTime = 7999 ms`, versus ~11–25 ms for the no-token early-return path — *even though* the APNs call itself completed fast enough for Apple's dead-token verdict to be applied and persisted within that window. Reproduced on both gateways and across four independent runs; `8_000` appears exactly once in the worker source, so the pending abort timer is the only plausible source of the 8 s tail.

Effect: the DO's IO context stays open ~8 s after each push instead of ~200 ms, blocking WebSocket hibernation for that window. Functionally harmless, but Durable Object duration is wall-clock billed, so it is pure waste on a path that fires on every finished turn and every sudo prompt.

Likely fix: keep a handle on the timeout and clear it in a `finally`, or drop `AbortSignal.timeout` in favour of an `AbortController` that is aborted explicitly.

## 2. APNs failures are unobservable

A delivery failure hits `console.warn('[push] APNs delivery failed', …)` (`session-tray.ts:1994`), but:

- `observability` is **not configured** in `wrangler.jsonc`, and `logpush: false` on **both** `slicc-tray-hub` and `slicc-tray-hub-staging`. Workers Logs is off, so nothing is retained anywhere.
- During live probing, `wrangler tail --format json` reported `logs: 0` on **every** Durable Object event, including ones on code paths that must have logged. Tail's DO log capture could not be relied on at all.

Net effect: if pushes silently stop working — expired `.p8`, rotated key id, wrong topic after a bundle-id change — there is no signal. That is how this shipped in August and sat unverified until now. Enabling `observability` on the worker is a one-line config change.

## 3. No test exercises the real APNs transport

[...]

PR #2448, opened 2026-08-26
fix(cloudflare-worker): mint one APNs provider JWT per worker, not per tray

trieloff:
Fixes the headline finding from #2432: the tray hub minted an APNs provider JWT **per Durable Object**, and there is one DO per tray.

## The bug

Apple throttles provider-token creation per `(team id, key id)` pair — **not** per connection — and answers `429 TooManyProviderTokenUpdates` above one mint per 20 minutes ([Apple forums](https://developer.apple.com/forums/thread/111241), [multi-key conditions](https://developer.apple.com/forums/thread/770438)). Two facts combined to blow that budget:

1. `WebCryptoApnsSender` cached its JWT in **DO memory**, one sender per tray DO.
2. The tray DO uses WebSocket hibernation (`state.acceptWebSocket`), so an idle tray is evicted between messages and signs a fresh JWT on the next push.

Mint rate therefore scaled with `active trays × wake cycles` rather than with time. A single user pushing every few minutes could already exceed Apple's floor. And `TooManyProviderTokenUpdates` was in neither `DEAD_TOKEN_REASONS` nor `STALE_JWT_REASONS`, so the throttled push was dropped and only warn-logged — into the void, because observability was off.

Confirmed live: the internal token fetch now happens on essentially **every** push, which is the hibernation eviction showing up in production exactly as described.

## The fix

Every tray DO borrows the token from one well-known instance of the *same* namespace — `TRAY_HUB.idFromName('__apns_provider_token')`, served over an internal `POST /internal/apns-token`. That instance is the only one that signs, and it persists the JWT to its own storage so its own hibernation costs nothing. **No new binding, no KV** — the namespace was already bound.

Design points worth reviewing:

- **The 20-minute floor is enforced even when a caller reports its token rejected.** Minting in response to a 429 deepens the throttle, and re-signing a token seconds old reproduces the same `iat` anyway. The source returns its cached token instead; the sender then declines to re-post the value Apple just refused, rather than burning a second request.
- **A stale-token report rotates only if the cached token still matches it.** A herd of DOs reporting the same rejection causes one mint, not one per caller — the stragglers get the already-rotated token.
- **Borrowed tokens are memoised by `mintedAt`, never fetch time**, so a token borrowed at minute 49 is not ridden past Apple's 60-minute expiry.
- **Falls back to local minting if the singleton is unreachable**, loudly. One object being down must not take every push with it; the local minter carries its own floor, so the blast radius is bounded.
- **Not fronted by `caches.default`.** The singleton never mints on a miss (it returns a stored token), so a cache would only shave an internal round trip — measured at **2–3 ms warm** — while adding an invalidation path for a value that must rotate on demand. Complexity without a payoff.

## Also in scope (items 3–7)

| | |
|---|---|
| **Cleared the request timeout** | `AbortSignal.timeout` stayed armed after the response landed, holding the DO's IO context open for the full 8 s budget after every *successful* push. Replaced with an `AbortController` cleared in `finally`. |
| **Retry 500/503 + transport errors** | Once, after 250 ms. |
| **410 `timestamp` check** | A device that re-registered *after* the token died keeps its registration instead of being silently unsubscribed by an in-flight push. 400 `BadDeviceToken` carries no timestamp and stays unconditionally final. |
| **`apns-unique-id`** | Recorded on the result and included in the failure log — the only handle Apple accepts for tracing a lost push. |
| **Observability** | Enabled on **both** environments. It is not inherited by named environments, so it is set twice. `invocation_logs: false` keeps volume and cost sane: this Worker also serves every UI asset, and only the explicit `console.warn` matters here. |

### One deviation from the brief

**429 is not retried by re-posting.** It is routed through the stale-JWT path as asked — so a throttled push retries against the shared token instead of dropping — but if the source declines to rotate, the sender stops rather than re-sending the refused token. Re-posting on a `TooManyProviderTokenUpdates` is what deepens the throttle, and the device-token flavour of 429 wants a backoff we cannot perform inside a WebSocket message handler without holding the DO awake — the very cost this PR is removing. Timeouts are likewise not retried: they already spent the full 8 s budget.

## Cost impact

Durable Object duration is billed on wall clock. Measured on staging, same fake-token probe before and after:

| | before | after |
|---|---|---|
| `push.send` with a registered token | **7998 ms** | **~1200 ms** (first), ~1209/1234 ms (steady) |
| internal provider-token fetch | n/a | **105 ms** cold, **2–3 ms** warm |
| `push.send` with no token (control) | 25 ms | 27 ms |

≈6.8 s of billed DO wall time removed from **every** notification — and this fires on every finished turn and every sudo prompt.

## Verification (staging only — see "NOT verified" below)

Full pre-push pass green: `verify` (7/7), `check-touched-exemptions` (0 files on any debt list), `typecheck`, `test` (16888 passed), `test:coverage` (no floor failures), both builds, `bundle-size`.

Deployed to **staging only**. Against the live worker:

- All 12 `tests/deployed.test.ts` smoke tests pass.
- `observability` confirmed live via the Cloudflare API on `slicc-tray-hub-staging`.
- `POST https://tray-hub.internal/internal/apns-token` appears in `wrangler tail` — the singleton is really being reached.
- **Apple still authenticates the borrowed JWT.** Push a fabricated device token, wait 30 s, push again on the same tray: the second push early-returns in **27 ms**, meaning the token was evicted, meaning Apple returned a dead-token verdict — which it only does *after* validating the provider token.

## On testing the real transport

The existing `apns.test.ts` injected `fetchImpl` everywhere, which is why a broken transport could never have been caught. New coverage targets behaviour rather than call shape: the mint-count property (`crypto.subtle.sign` called **exactly once** across 6 tray DOs reconstructed from hibernation), single-flight collapsing of concurrent callers, the 20-minute floor in both directions, store-backed resume across instances, timer cleanup (asserting `vi.getTimerCount() === 0`, not merely that a timeout was armed), and the 410 timestamp branch on both sides.

What I did **not** add is a CI gate that talks to Apple. It would put Apple's availability in the deploy path for every worker PR, and the credentials only exist in the staging/production jobs. Instead the probe recipe — the one used above, which needs no phone — is documented in `.agents/skills/deploying-tray-worker/SKILL.md`, including the trap that fooled the original investigation: an 8-second `wallTime` was the uncleared abort timer, not a hang, so read the *second* push, not the first.

## Verified on hardware (2026-08-26)

Both items that were open when this PR was written are now closed, on a physical iPad against this branch deployed to staging.

**Real delivery — observed.** A `sudo_request` push rendered on the device as a banner reading **TIME SENSITIVE / "Approval needed" / "Deploy to production is waiting for your approval"** — exactly `buildApnsPayload`'s output, with `interruption-level: time-sensitive` surfacing correctly. Tapping it fired `SLICC_SUDO_REVIEW` and opened the app. This is the first observed end-to-end delivery from this worker.

**`APNS_TOPIC` — verified** via the two-tray gateway probe, from Apple's own verdict rather than inference:

| token | gateway | result |
| --- | --- | --- |
| real device token | sandbox | **survived** — accepted (585 ms, 169 ms real round trips) |
| real device token | production | evicted — rejected |
| fabricated token | sandbox | evicted — rejected |

The fabricated-token control is what makes this conclusive: it proves the sandbox gateway was live and rejecting bogus tokens at the same moment the real token was accepted. Acceptance requires the topic to match the token's app — a mismatch returns `DeviceTokenNotForTopic`, a dead-token reason that would have evicted it.

**The singleton under real load:** internal `apns-token` fetches measured **426 ms cold, then 2 ms / 2 ms warm**, with signing happening once on the minting instance.

**Timer fix, confirmed from outside.** Mid-test another agent's `wrangler deploy` overwrote staging with a build lacking this change; pushes immediately went back to **7998–8000 ms** holds, versus **169–585 ms** on this branch. Caught because this branch makes a distinctive internal `apns-token` request and the tail showed zero. Redeployed and re-ran everything above on verified-correct code.

## Still not covered

The **Face ID sudo approval leg** has not run on hardware. It is gated by `shouldDelegateSudo()`, which only routes to a follower when the last user message came from one — so it needs a message sent from the phone first. Scheduled separately; tracked in #2432.

Refs #2432, #2062, #2213.

### DIFF
--- a/.agents/skills/deploying-tray-worker/SKILL.md
+++ b/.agents/skills/deploying-tray-worker/SKILL.md
@@ -481,3 +481,40 @@ or transport are broken. Do **not** infer from `wallTime` on the first push alon
 8-second reading used to be the uncleared `AbortSignal.timeout`, not a hang.
 
 This is deliberately not a CI gate — it would put Apple's availability in the deploy path.
+
+### Settle the gateway before reading any rejection (two-tray probe)
+
+A device token is valid on **one** gateway only, decided by the build's
+`aps-environment` entitlement: a development-signed build gets a sandbox token, a
+TestFlight/App Store build a production one. Send it to the wrong gateway and Apple answers
+`BadDeviceToken` — indistinguishable, at a glance, from a wrong topic or a bad key. Never
+interpret a rejection before establishing which gateway the token belongs to.
+
+Do not infer the gateway from the build. `currentApnsEnvironment()` reports what the app
+_claims_ — derived from `#if DEBUG`, not from the entitlement it actually holds — and a
+dev-signed **Release** build claims `production` while carrying a sandbox token. Reading
+Xcode settings or `builtByDeveloper` only tells you what _should_ be true.
+
+Ask Apple instead. With one real device token, register it on **two throwaway trays** and
+push once from each — the DO stores `environment` per token, so the same token can be
+driven at both gateways:
+
+```
+tray A: {"type":"push.register", …, "environment":"sandbox"}    → push.send
+tray B: {"type":"push.register", …, "environment":"production"} → push.send
+```
+
+Read the verdict from a **second** push on each tray (the eviction signal above):
+
+| tray A (sandbox) | tray B (production) | meaning                                                |
+| ---------------- | ------------------- | ------------------------------------------------------ |
+| token evicted    | token evicted       | wrong on both — topic or key is wrong, not the gateway |
+| token survives   | token evicted       | sandbox is correct; it is a development build          |
+| token evicted    | token survives      | production is correct; TestFlight / App Store build    |
+
+A token that **survives** was accepted (`200`) — that gateway is the right one. Once the
+gateway is known, a rejection there is finally meaningful: `DeviceTokenNotForTopic` means
+`APNS_TOPIC` does not match the app's bundle id, whereas `BadDeviceToken` on the _correct_
+gateway means the token itself is stale. That distinction is the only way to verify
+`APNS_TOPIC`, because Apple rejects a malformed token before it ever checks the topic — so
+no fabricated-token probe can reach it.
--- a/.agents/skills/deploying-tray-worker/SKILL.md
+++ b/.agents/skills/deploying-tray-worker/SKILL.md
@@ -454,3 +454,30 @@ loopback; the `.localhost` host is dev-only.
 ## Follower push (APNs, #2062)
 
 Four secrets, all or nothing: `APNS_TEAM_ID`, `APNS_KEY_ID`, `APNS_PRIVATE_KEY` (the `.p8` PEM, newlines intact — `npx wrangler secret put APNS_PRIVATE_KEY < AuthKey_XXXX.p8`), `APNS_TOPIC` (`com.sliccy.follower`). Missing → the DO logs `push.send ignored` once and nothing else changes. Staging talks to whatever gateway the registering phone asked for (`environment` on `push.register`: debug builds = sandbox, TestFlight/App Store = production), so a TestFlight build against staging needs the production key. Verify with a time-sensitive test: run a headless leader, background the phone, `sudo` something — the banner must arrive within seconds; a 403 `InvalidProviderToken` in `wrangler tail` means the key id / team id pair is wrong.
+
+**Provider JWTs come from one DO, not per tray.** Apple throttles token creation per
+(team id, key id) — not per connection — and answers `429 TooManyProviderTokenUpdates`
+above one mint per 20 minutes. Every tray DO borrows from
+`idFromName('__apns_provider_token')` via `POST /internal/apns-token`, which persists the
+JWT to its own storage so hibernation costs nothing (#2432). Never reintroduce per-DO
+minting: the tray DO hibernates between messages, so mint rate would scale with
+`active trays × wake cycles`, not with time.
+
+**Probe the worker→Apple leg without a phone.** Register a syntactically valid but
+fabricated device token and push; APNs authenticates the JWT _before_ it validates the
+token, so a dead-token verdict proves the credentials are good:
+
+```bash
+# 1. mint a throwaway tray, 2. attach as leader, 3. open the leader WS, then send:
+#    {"type":"push.register","platform":"ios","token":"<64 hex>","environment":"sandbox"}
+#    {"type":"push.send","category":"turn_end","label":"probe"}
+# Push again on the SAME tray ~30s later.
+```
+
+Read the result from the second push, not from logs: if the token was evicted the second
+`push.send` early-returns in ~10 ms (Apple answered `BadDeviceToken`/410 — credentials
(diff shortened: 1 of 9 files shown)


trieloff:
## Review round addressed (`43f17bfcb`)

Codex raised three findings. Two were correct and are fixed; the third is correct but needs an ops action, so it is filed as #2452. Copilot could not review — the requesting account is over quota. Not re-requesting review.

### P2 — per-tray minting fallback (fixed)

Correct, and it undercut the point of the PR. The shared source fell back to local minting when the minting instance was unreachable, which put every waking tray back to signing its own JWT — converting a brief local outage into account-wide `429 TooManyProviderTokenUpdates` that outlasts it and fails the pushes anyway.

It no longer mints locally. It reuses a borrowed token if this instance still holds one Apple will accept (under 55 min, and not the value just rejected), otherwise it fails the push loudly. Pushes are best-effort wake-ups — the request itself travels the data channel — so losing one costs far less than throttling the key. Covered by a test asserting `crypto.subtle.sign` is called **zero** times on the outage path.

### P2 — stale JWT after credential rotation (fixed)

Correct, and a sharp catch: the persisted record held only `{value, mintedAt}`, so after rotating `APNS_KEY_ID`/`APNS_TEAM_ID` a token signed by the old key looked merely *stale*, and the 20-minute floor then refused to replace it — disabling pushes for up to 20 minutes even though the new key pair has its own budget with Apple.

Tokens now carry `identity` (`teamId.keyId`) and a mismatch is discarded on load, so a rotation mints immediately. Records written before this field also fail validation and are discarded — one extra mint at deploy time, which is correct. Two tests cover it: rotating the key id, and a legacy record with no identity.

### P1 — shared APNs key across environments (→ #2452)

Verified the premise: `worker.yml` and `worker-staging.yml` both read the same repository-level `APNS_TEAM_ID`/`APNS_KEY_ID`, so the two environments' minters are uncoordinated against one key pair.

Out of scope here because the fix is issuing a second `.p8` in the Apple developer account and pointing staging at it — Apple allows two active auth keys per team, so this is the intended usage. Filed as #2452 with the exact steps. Worth noting this PR reduces the exposure from unbounded (`active trays × hibernation wakes`, both environments) to exactly 2. A shared cross-namespace token service would fix it in code but is real new infrastructure and would couple staging to production on the push path — worse than the throttle it avoids.

## Re-verification

Rebased onto current `origin/main` (the boy-scout gate diffs against it and the stale branch was showing a phantom `record-string-unknown` addition from an upstream commit).

Full pass re-run green: `verify` 7/7, `check-touched-exemptions` (9 changed files, 0 on any debt list), `typecheck`, `test` (16970 passed), `test:coverage` (no floor failures), both builds, `bundle-size`.

Redeployed to staging and re-ran the fake-token probe against the live worker:

| | |
|---|---|
| `push.send` with a registered token | **1142 ms** (was 7998 ms before the timer fix) |
| internal provider-token fetch | 268 ms cold |
| second `push.send`, same tray | **25 ms** |

The 25 ms means the token was evicted, which means Apple returned a dead-token verdict — which it only does *after* validating the provider token. So the borrowed JWT is still accepted with the `identity` field in place.

PR #2455, opened 2026-08-26
approvals: never prompt for pre-granted writablePaths; settle covered pending requests; no duplicate sudoers rules

### LINKED ISSUE #2416: approvals: prompt raised for pre-granted writablePaths; pending approvals appear to gate unrelated scoops
## Context

Split out of the 2026-08-25 concurrent-scoop incident (8 scoops running `migrate-block` in parallel; browser-bridge fixes landed in PR #2411). The approval/sudo system showed two anomalies that compounded the bridge contention into a ~26-minute stall.

## Observed

1. **A prompt was raised for an already-granted path.** The `location-finder` scoop was created with `writablePaths: ["/shared/", "/.playwright/", "/tmp/"]`, and its sudoers showed `NOPASSWD Write /.playwright` **and** `/.playwright/**` — yet its `playwright-cli` calls hung until a `/.playwright/**` write grant was approved mid-run. The post-approval sudoers file lists the rule **twice**, proving a prompt was raised (and a duplicate rule appended) for a path the sandbox already authorized.
2. **Pending approvals appear to serialize globally.** Only 2 sudo requests were ever surfaced to the cone, yet all 8 scoops resumed within a minute of those two approvals — consistent with unrelated scoops blocking behind another scoop's pending approval instead of proceeding with their own granted paths.
3. **Undocumented second trigger:** block-migration scoops write source crops to `/.migration/` at the VFS root (outside the project path), which nothing pre-authorizes — a guaranteed prompt in every fan-out run of that shape.

## Impact

- Fan-out runs stall on approvals the sandbox config says should not exist; the cone misreads the stall as scoops working ("files stable + scoops processing").
- Duplicated sudoers rules accumulate.

## Suggested investigation

- Reproduce: create a scoop with `writablePaths: ["/.playwright/"]`, have it write `/.playwright/screenshots/x.png` — does a prompt fire? Check the matcher for pre-granted paths (trailing-slash / glob normalization: `/.playwright/` vs `/.playwright/**` vs bare `/.playwright`).
- Check whether a pending sudo request blocks *other* scoops' unrelated filesystem operations (a global approval queue/lock), and whether that is intended.
- Decide whether `/.migration/**` should be pre-authorized alongside `/.playwright/**` for migration-type scoops, or whether skills should be told to write under the project path.

## Acceptance

- A write inside a scoop's `writablePaths` never raises an approval prompt, regardless of trailing-slash/glob spelling.
- Approving one scoop's request does not gate unrelated scoops (or the serialization is documented as intended behavior).
- No duplicate rules appended to sudoers on approval.

catalan-adobe:
Fixes #2416.

## Root causes

Investigation found no matcher bug — `generateScoopSudoers` + `matchPath` correctly grant any trailing-slash/glob spelling of a `writablePaths` entry. The prompts came from policy *state*, not matching:

1. **The per-scoop policy was derived only from the on-disk `/scoops/<folder>/etc/sudoers` file.** `ensureSudoersLoaded` seeds it from `ScoopConfig` only when absent; when a file already exists (folder reuse across scoop generations, restores with changed config), the config's `writablePaths` were never granted — a guaranteed prompt for a pre-granted path.
2. **Pending requests were never re-evaluated after a policy reload.** playwright-cli fires several gated ops (`mkdir` + snapshot/session/screenshot writes) close together; approving one with "Always" widened the policy, but the scoop's other already-pending requests for the same subtree kept hanging until each was individually approved — the observed multi-minute stall, and each extra "Always" appended the same rule again.
3. **Duplicate/leaking persistence.** `appendScoopRule` appended without checking for an identical existing line, and a non-cone scoop's `SudoFS` used the default grant sink, writing `always` grants to the global `/etc/sudoers.d/granted` — silently widening every other unit's policy.

## Changes

- `SudoManager.registerScoopConfig(folder, config)`: config-derived grants are registered **synchronously in memory** on every context creation and merged into `getPolicyForScoop`. `ScoopConfig` is now authoritative; the on-disk file only adds persisted "Always" grants. No file/watcher race can withhold a configured path.
- `ScoopApprovalRouter.settleGrantedRequests(folder?)`: wired into `onPolicyReload` — pending requests now covered by a `NOPASSWD` grant auto-resolve as `allow` (card flips to confirmed).
- `appendScoopRule` is idempotent — an existing identical rule is never appended again.
- `buildSudoWiring` gives non-cone scoops a no-op `onGrant` sink (FS-level analog of the existing `persistCommandGrant` isolation); their `always` grants persist only scoped via the router.

## On the acceptance criteria

- *Write inside `writablePaths` never prompts, any spelling*: covered by the in-memory config policy + spelling test (`/x`, `/x/`, `/x/**`).
- *Approving one scoop's request does not gate unrelated scoops*: investigation found **no global lock** — each scoop blocks only on its own gated ops. The incident's appearance came from every scoop hitting its own prompt (e.g. un-granted `/.migration/` writes). The single-approver serialization (the cone processes its queue one turn at a time) is now documented as intended in `docs/approvals.md`.
- *No duplicate rules on approval*: dedupe + scoped-only persistence + auto-settle (the second prompt never needs approval).
- */.migration/*: not pre-authorized (no such root exists in-repo; it is skill-invented). The delegation skill now tells cones to list every tool-owned absolute root in `writablePaths` and to prefer scratch/`/shared/`/`/tmp/` over inventing VFS roots.

## Tests

- `sudo-manager.test.ts`: stale-file scenario, spelling matrix, config survival after file wipe/delete, `onPolicyReload` notification, duplicate-append guard.
- `scoop-approval-router-settle.test.ts`: auto-settle for covered path + command requests, folder scoping, no-SudoManager no-op.
- `scoop-context.test.ts`: non-cone `onGrant` is a no-op (no global granted file), cone unchanged.

Full pass run locally: `verify`, `check-touched-exemptions`, `typecheck`, `test` (one unrelated pre-existing flake: `hf-cache-mirror.test.mjs`, passes standalone), `test:coverage`, both builds, `bundle-size`.

### DIFF
--- a/docs/approvals.md
+++ b/docs/approvals.md
@@ -133,9 +133,9 @@ Two consequences worth knowing:
   (`getPolicy()`) still honours it.
 - The grant lives in code (`builtinScoopGrants()` in
   `packages/webapp/src/base/sudoers.ts`), merged in by `getPolicyForScoop`, not
-  in the generated `/scoops/<folder>/etc/sudoers`. That file is written only
-  when absent — so a file-based grant would never reach an already-created
-  scoop. The matching ACL exemption is `ALWAYS_WRITABLE_PREFIXES` in
+  in the per-scoop config grants — those are compiled from each scoop's own
+  `ScoopConfig`, while `/tmp` applies to every scoop unconditionally. The
+  matching ACL exemption is `ALWAYS_WRITABLE_PREFIXES` in
   `packages/webapp/src/fs/restricted-fs.ts`; `SudoFS` and `RestrictedFS` gate
   independently, so both layers must agree or the write is walled underneath
   the grant.
@@ -243,9 +243,9 @@ Orchestrator.init()
 
 Orchestrator.createScoopTab(jid)
   ├─ if non-cone: RestrictedFS(..., 'sudo-delegated')   // writes pass through to SudoFS
-  ├─ if non-cone: registerScoopConfig(folder, config)   // always — config grants in memory (#2416)
-  ├─ if non-cone: seedScoopSudoers(folder, config) ...  // first boot only
-  ├─                              ... or reloadScoopPolicyByFolder()  // existing file
+  ├─ if non-cone: initScoopPolicy(folder, config)  // config grants registered in memory (#2416)
+  │                                                // + load the on-disk Always-grants file
+  │                                                //   (legacy generated files are discarded)
   └─ new ScoopContext(scoop, callbacks, fs, ..., sudoManager)
 
 ScoopContext.init() — non-cone scoop
@@ -396,8 +396,12 @@ told to wait for a user who was never prompted.
 
 "Always" grants for `kind: 'command' | 'read' | 'write'` are persisted via
 `SudoManager.appendScoopRule(folder, kind, pattern)` (raw-VFS write, same trusted
-sink that powers `seedScoopSudoers`, so it bypasses the per-scoop self-protection
-on `/scoops/<folder>/etc/sudoers`). The append is idempotent — a rule that is
+sink that powers `initScoopPolicy`, so it bypasses the per-scoop self-protection
+on `/scoops/<folder>/etc/sudoers`). Since #2416 that file holds ONLY these
+approved "Always" grants — the `ScoopConfig` sandbox is registered in memory
+and never persisted, so replacing a scoop's config genuinely revokes the old
+authority instead of unioning with a stale file (a legacy generated file found
+on load is discarded fail-closed; its grants re-prompt once). The append is idempotent — a rule that is
 already in the file is never duplicated (#2416) — and it is the ONLY persistence
 path for a scoop's grant: the scoop's `SudoFS` gate gets a no-op `onGrant` sink,
 so an `always` decision never leaks into the global `/etc/sudoers.d/granted`
--- a/docs/approvals.md
+++ b/docs/approvals.md
@@ -238,11 +238,12 @@ Orchestrator.init()
   └─ new SudoManager({ fs: sharedFs, watcher })  // seed + load + watch
        ├─ getBroker()         → createSudoBroker()         // user broker (cone)
        ├─ getPolicy()         → live merged global SudoersPolicy
-       ├─ getPolicyForScoop() → builtin /tmp grants ∪ global ∪ /scoops/<folder>/etc/sudoers
+       ├─ getPolicyForScoop() → builtin /tmp grants ∪ global ∪ config grants ∪ /scoops/<folder>/etc/sudoers
        └─ getShellConfig()    → { getPolicy, broker, persistCommandGrant }
 
 Orchestrator.createScoopTab(jid)
   ├─ if non-cone: RestrictedFS(..., 'sudo-delegated')   // writes pass through to SudoFS
+  ├─ if non-cone: registerScoopConfig(folder, config)   // always — config grants in memory (#2416)
   ├─ if non-cone: seedScoopSudoers(folder, config) ...  // first boot only
   ├─                              ... or reloadScoopPolicyByFolder()  // existing file
   └─ new ScoopContext(scoop, callbacks, fs, ..., sudoManager)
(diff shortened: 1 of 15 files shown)


catalan-adobe:
### Strengths

- **`packages/webapp/src/scoops/scoop-lifecycle-manager.ts:345-360`** — Config-derived policy is registered before the context and gated filesystem are initialized, eliminating the filesystem/watcher race that caused configured paths to prompt.
- **`packages/webapp/src/scoops/scoop-approval-router.ts:133-167`** — Pending-request settlement is folder-scoped and only accepts explicit `NOPASSWD` matches. Using the normal matchers also preserves hardcoded path self-protection.
- **`packages/webapp/src/scoops/scoop-context/sudo-wiring.ts:75-98`** — Both shell and filesystem “Always” sinks are isolated for delegated scoops, preventing grants from leaking into the global policy.
- **`packages/webapp/src/sudo/sudo-manager.ts:314-342`** — Duplicate detection happens before writing while preserving read errors rather than overwriting a file that could not be inspected.
- **`packages/webapp/tests/sudo/sudo-manager.test.ts:600-686`** — Good coverage of stale files, spelling variants, policy notifications, config survival after file changes, and duplicate prevention.
- **`packages/webapp/tests/scoops/scoop-approval-router-settle.test.ts:116-180`** — Tests distinguish covered and uncovered requests, command and path requests, folder scoping, and missing-manager behavior.

### Issues

- **`packages/webapp/src/sudo/sudo-manager.ts:242-248`** — Stale config-generated rules are still merged as durable grants, so a narrower replacement config does not revoke old authority.
  - `seedScoopSudoers()` writes the entire original `ScoopConfig` into the same file later used for “Always” grants. On folder reuse, `ensureSudoersLoaded()` reloads that file, and `getPolicyForScoop()` unions it with the new config without distinguishing old generated rules from approved rules.
  - Concrete scenario: scoop generation A uses folder `worker` with `writablePaths: ['/workspace/']` and unrestricted commands. Generation B reuses `worker` with `writablePaths: ['/shared/']` and `allowedCommands: []`. The stale file still grants `/workspace/**` and `Cmnd *`; because `RestrictedFS` uses delegated write enforcement, generation B can modify `/workspace` and run commands without approval. This contradicts the documented pure-replace sandbox and “config is authoritative” contract.
  - Suggested fix: store config policy only in memory and persist “Always” rules separately, such as an Always-only drop-in. Add a migration strategy for existing generated files; because legacy files do not identify which rules were appended, the secure migration is to regenerate or discard ambiguous legacy rules rather than silently retaining them. Add a regression test where the stale file is broader than the replacement config and assert the removed path/command no longer matches.

- **`packages/webapp/src/scoops/scoop-approval-router.ts:151-159,328`** — Auto-settlement loses the requesting scoop’s owner before updating the lick card.
  - `registry.resolve()` deletes the entry at line 151. `persistLickDecision()` then asks the now-empty registry for `scoopJid`, gets `undefined`, and `findApprover(undefined)` falls back to the default root.
  - With multiple cones, a request owned by cone B is successfully auto-allowed, but persistence searches cone A’s messages. Cone B’s card remains pending and a warning is logged. The current test misses this because its `findApprover` stub always returns the same cone regardless of input.
  - Capture the approver or `pending.scoopJid` before resolving and pass it explicitly:
    ```ts
    const approver = this.deps.findApprover(pending.scoopJid);
    if (this.registry.resolve(pending.id, { decision: 'allow' })) {
      void this.persistLickDecision(pending.id, 'allow', approver?.jid);
    }
    ```
    Apply the same explicit ownership approach to normal and fail-closed settlement, then add a two-cone test.

#### 💡 Minor (Nice to Have)

- **`packages/webapp/src/sudo/sudo-manager.ts:165-176`** — Per-folder config policies and reload chains are never evicted.
  - Dropping normal or ephemeral scoops leaves their compiled policies in `scoopConfigPolicies`; `scoopPolicies` and `scoopReloadChains` can remain as well. Repeated one-shot agents therefore grow these maps for the lifetime of the orchestrator.
  - Add a `forgetScoopPolicy(folder)` method that clears all three maps, and call it from the successful unregister path after pending requests are failed closed.

### Recommendations

- Separate authoritative config grants from persisted approval grants at the storage-format level; merging both into one untyped sudoers file makes safe revocation impossible.
- Add end-to-end tests through `SudoManager.onPolicyReload` and the router rather than testing settlement only by directly invoking `settleGrantedRequests`.
- Extend settlement tests to cover multiple owning cones and a broader stale policy followed by a narrower replacement config.

### Assessment

**Ready to merge?** No

**Reasoning:** The prompt and duplicate-rule fixes are well structured, but stale config-derived rules can still silently preserve authority removed by a replacement sandbox. Auto-settled cards are also updated under the wrong cone in multi-cone sessions.

---

catalan-adobe:
Addressed all three review findings in b38b0722c — each claim was verified against the code first and confirmed real:

**Critical — stale config-generated rules retained revoked authority.** Confirmed: `unregister` never deletes `/scoops/<folder>/` files and the folder slug is deterministic, so folder reuse reloads the old file, and `mergePolicies` unions allow-rules — gen A's `NOPASSWD Cmnd *` survived gen B's narrower config. Fixed at the storage-format level as suggested:

- `/scoops/<folder>/etc/sudoers` now holds **only** approved "Always" grants (new header). `ScoopConfig` grants are compiled in memory (`initScoopPolicy`, replacing `seedScoopSudoers`) and never persisted — replacing a config now genuinely revokes.
- Migration: legacy files are identified by the old generated header and discarded fail-closed (config rules and appended grants are indistinguishable in that format; a discarded "Always" grant re-prompts once). Hand-written files (no legacy header) load as-is — deliberate user policy.
- Regression test added: broad stale legacy file + narrower replacement config → `/workspace/**` write and `Cmnd *` no longer match; plus tests for legacy-discard vs hand-written/Always-only retention.

**Important — card ownership lost after `registry.resolve()`.** Confirmed, including the same pattern pre-existing in `resolveSudoRequestAndPersist` and the fail-closed auto-settle paths. All settlement paths now carry the requester's `scoopJid` explicitly: captured before `resolve()` in `settleGrantedRequests` and `resolveSudoRequestAndPersist`, and threaded through `onAutoSettle(id, reason, scoopJid)` for timeout/drop/shutdown. `persistLickDecision` takes the explicit jid with the registry lookup as fallback (non-sudo licks). Two-cone regression test added — it fails without the fix (card flips under the default root, owning cone stays pending).

**Minor — unbounded per-folder maps.** `SudoManager.forgetScoopPolicies(folder)` evicts the config policy, file policy, and reload chain; called from `unregister` after pending requests are failed closed.

**Recommendations:** the e2e test through the real reload seam is in (`SudoManager` + `onPolicyReload` → `settleGrantedRequests`, driven by `appendScoopRule` — the `lick_confirm always` path — auto-settling a second pending request). `docs/approvals.md` updated for the Always-only file format and migration.

Full verification pass rerun: verify, exemptions gate, typecheck, full test suite (17007 passed, including the previously flaky file), coverage, both builds, bundle-size.

PR #2456, opened 2026-08-26
fix(ios-app): anchor the transcript instead of restoring a scroll offset (#2072)

### LINKED ISSUE #2072: iOS: transcript jumps backwards when the keyboard opens — caused by our `.scrollPosition` binding, not FB20979569

trieloff:
Fixes the iOS transcript jumping backwards through the history when the keyboard opens. **Closes #2072. Supersedes #2086** (see below).

The issue was filed as an iOS 26 platform bug (FB20979569). It is ours.

## Measured

Same device, runtime, fixture and reading position; the anchor row is centred in the viewport before the composer is focused in every run.

| variant | after reader's scroll | after keyboard | after 1st keystroke | after long type | drift |
| --- | --- | --- | --- | --- | --- |
| baseline (`.scrollPosition($shared)`) | 507.5 | 766.2 | 766.2 | row gone | **+258.7pt** |
| baseline, repeat run | 531.8 | 789.5 | 789.5 | row gone | **+257.7pt** |
| `.scrollPosition($localState)` | 513.2 | row gone | — | — | worse |
| **this PR** | **508.8** | **508.8** | **508.8** | **508.8** | **0.0pt** |

**Confirmed on hardware** — iPad Pro 12.9" (iOS 26.5.2): baseline **+529.0pt** portrait / anchor row lost entirely in landscape; this PR **0.0pt in both orientations**. The bug is *worse* on device than in the simulator. [Numbers](https://github.com/ai-ecoverse/slicc/pull/2456#issuecomment-5427174938).

> **iPhone 17e simulator, iOS 26.5, Xcode 26.6.** See "Device coverage" below — the fix is now also verified on an **iPad Air 13-inch (iOS 26.5)**, and the tests are gated in CI on both device classes. Physical-device validation remains outstanding and is blocked on a consent question, not on effort.

**No regressions:** the transcript-adjacent UI subset (`TranscriptColumnUITests`, `FixtureConversationUITests`, `ComposerKeyboardUITests`, `ReadOnlyScoopUITests`, `SteerUITests` — 14 tests) produces the **identical 4 failures** on this branch and on pristine `main`, all pre-existing on this simulator. `SliccFollowerTests`: 592 passed, 0 failures.

## Why

A `LazyVStack` only *estimates* the height of rows it has not materialized, so the scroll view's content offset is an estimate too — [WWDC26 "Dive into lazy stacks and scrolling with SwiftUI"](https://developer.apple.com/videos/play/wwdc2026/321/) says exactly this ("avoid using the absolute content size or content offset with lazy stacks, since these are estimated and unstable"). Restoring a bound `ScrollPosition` across a container resize restores that estimate, and it resolves against the end of the *measured* content — mid-history, not where the reader was.

`defaultScrollAnchor(.bottom)` is a layout rule resolved **after** the resize rather than a value restored **across** it. No estimate is in the loop.

**Why this went unfound for so long:** #2072's rejected-attempts table contains both halves of the fix, never together. `.defaultScrollAnchor(.bottom)` was measured as "+265pt, no effect" — correct, *while a `scrollPosition` binding is installed*, because the explicit binding wins. And "remove the `.scrollPosition` binding" was dismissed as "confounded: also removes the initial scroll-to-bottom" — true, and the confound was removable, because the anchor supplies that scroll-to-bottom.

## The anchor alone was not enough

Worth flagging, because my own first version of this fix was wrong and the tests caught it: `defaultScrollAnchor(.bottom)` **stops the list following new content** once the keyboard has resized the viewport. Measured — a sent message landed in the transcript and the view did not move to it.

So the anchor is paired with a one-shot `ScrollViewReader.scrollTo(id:)` gated on `onScrollTargetVisibilityChange` (WWDC26's named stable way to ask a lazy stack what is on screen). That is a scroll **action**, never a stored offset, so it cannot reintroduce the bug.

The follow rule is deliberately **split**:

- **your own message always wins** — every chat app takes you to what you just sent;
- **anything the cone produces defers to where the reader is.**

The five old `onChange` handlers force-scrolled unconditionally and yanked a reader out of the history mid-sentence. That is now pinned behaviour rather than a side effect.

## Two things this costs, both deliberate

- **The compact/regular shell switch re-opens at the newest message.** `ChatPresentationState.transcriptPosition` existed so a rotation or Split View resize kept the reader's place — it rode on the very binding that caused the bug. Its assertion in `ChatPresentationStateTests` is replaced by a comment explaining why, not silently deleted. Restoring by message id is a follow-up.
- **`.scrollPosition(id:anchor:)` is not a drop-in replacement.** Measured and rejected: with an id binding installed the transcript cannot be dragged back at all (the anchor row never reaches the viewport in 14 drags).

## Superseding #2086

#2086 staged a reproduction. It could never have run: `UITestHooks.seedTranscriptFixture(into:)` is defined there with **no call site** — `git grep` on that branch finds only the declaration — so `-uiTestTranscriptFixture YES` seeded nothing and the class died at `waitForExistence(timeout: 60)`. That is very likely the real explanation for "the harness cannot reach the measurement point on iOS 27"; it could not reach it on 26 either.

This PR absorbs the reproduction **with a working call site**, un-skipped, as the regression gate. **#2086 should be closed.**

I also corrected #2072's body in place (struck through, not deleted) — it asserted that Apple DTS had acknowledged the bug, when [thread 805306](https://developer.apple.com/forums/thread/805306) shows the engineer could *not* reproduce it and referred it to Feedback Assistant.

## Tests

`TranscriptComposerGrowthUITests`, un-skipped, asserting the **measured property** — anchor drift across a keyboard transition — rather than mere rendering, because only the measurement catches a reintroduction:

1. `testHistoryStaysPutWhenTheComposerAndKeyboardClaimSpace` — the #2072 gate.
2. `testTranscriptStillFollowsANewlySentMessage` — the auto-follow that the anchor alone broke.
3. `testAnIncomingMessageDoesNotYankAReaderOutOfTheHistory` — the anti-yank rule, driven by a scheduled cone message (`-uiTestTranscriptAppendAfter`), because a *send* deliberately always wins and so cannot test this.

Four harness traps are documented in the test file so the next person does not re-find them: the anchor row is offscreen at launch (`swipeDown()` fails); `app.scrollViews.firstMatch` resolves to one of `HorizontalScrollGuard`'s zero-width horizontal scrollers; an empty composer hit-disables its editor under `PttPressSurface` so `element.typeText` drops the text; and tapping the still-disabled send button is a silent no-op.

## Not in this PR

The general transcript **sluggishness** is a separate problem with a separate fix — 871 row-body evaluations and 227 full markdown re-parses to scroll back two screens. That is a follow-up PR so this one stays reviewable on its own.

Full measurement write-up including the rejected UIKit port: `docs/research/ios-transcript-uikit.md`.

## Device coverage

The iPad mattered here and I nearly shipped without it. **`ios-app-tests` runs an explicit curated list of test classes** — iPhone gets the unit suite plus `ComposerConnectionUITests`, iPad gets `RemoteTabUITests` and `TranscriptColumnUITests` — and `TranscriptComposerGrowthUITests` was in **none** of them. So the four green iOS legs on the first push proved only that nothing *already covered* had broken; my new regression gate would never have run. Now added to **both** destinations.

Running it on iPad also caught a portability bug **in the tests**, not the fix: `waitForSeededTranscript` waited for one *named* row (the newest) as its "did the fixture seed" check. Which rows a lazy stack materializes at launch is device-dependent — an iPad renders rows wider and therefore shorter, and materializes a different set — so all three tests failed on iPad before any measurement ran. It now waits for any `message-fx-` row.

With that corrected, **all three pass on iPad Air 13-inch (iOS 26.5) and iPhone 17e**, and the iPad transcript sits correctly pinned to its newest message at rest (screenshot verified). The fix itself needed no change.

## Test plan

- [x] `TranscriptComposerGrowthUITests` 3/3 pass
- [x] `SliccFollowerTests` 592 pass, 0 failures
- [x] Transcript-adjacent UI subset: same 4 pre-existing failures as pristine `main`
- [x] `swiftlint` 0 serious, `swift format lint --strict` clean
- [x] `npm run verify` 7/7, `check-touched-exemptions` OK
- [x] Rebased onto current `origin/main`
- [x] iPad Air 13-inch simulator (iOS 26.5) — 3/3 pass; transcript pinned to newest message at rest
- [x] `TranscriptComposerGrowthUITests` added to the CI matrix on **iPhone and iPad**, iOS 26 and 27 (the iOS 27 legs are `continue-on-error` — real information, not a blocking guarantee)
- [x] **Physical iPad Pro 12.9" (iPad14,6, iOS 26.5.2) — baseline +529.0pt, this PR 0.0pt, BOTH orientations.** Hardware is worse than the simulator (529 vs 258pt); in landscape the baseline row left the accessibility tree entirely. [Full numbers](https://github.com/ai-ecoverse/slicc/pull/2456#issuecomment-5427174938). Not covered: a hardware keyboard (none paired; not pairing one to a personal iPad unasked).
- [ ] **Visual sign-off from @trieloff** — do not merge before this


trieloff:
## Physical-device validation: confirmed, and the bug is **worse** on hardware

Ran on **@trieloff's iPad Pro 12.9-inch (iPad14,6, iOS 26.5.2, build 23F84)** — not a simulator. Baseline and fix measured back to back, same device, same fixture, same reading position, anchor row centred in the viewport before the composer is focused.

| build | orientation | after keyboard | after 1st keystroke | after composer growth |
| `main` (baseline) | portrait | **+529.0pt** | — | — |
| `main` (baseline) | landscape | **anchor row left the accessibility tree entirely** | — | — |
| **this PR** | portrait | **0.0pt** | **0.0pt** | **0.0pt** |
| **this PR** | landscape | **0.0pt** | **0.0pt** | **0.0pt** |

**529pt on hardware against 258pt on the iPhone 17e simulator** — the opposite of a simulator artefact. The 12.9-inch screen shows more rows per viewport, so there is more *estimated* content for the lazy stack to mis-resolve against, and the jump is correspondingly bigger. In landscape the baseline is worse than merely large: the anchor row was thrown so far it de-materialised, so the drift is not even measurable.

The fix reads 0.0pt in every cell, both orientations.

### Also landed here

- **Landscape is now a permanent test.** Not redundant on iPad: the transcript is wider (rows shorter, different materialisation), and the keyboard claims a far larger fraction of the viewport — which is exactly the resize #2072 mishandled.
- **Runs now print the measured drift.** The gate asserted a threshold but never reported the quantity, and #2072 *was* a quantity. A green tick that hides the number is how this got mis-attributed to the platform for weeks.

### Test matrix now

| | iPhone 17e sim | iPad Air 13" sim | iPad Pro 12.9" **device** | CI iPhone 26/27 | CI iPad 26/27 |
| --- | --- | --- | --- | --- | --- |
| result | pass | pass | **pass (both orientations)** | pass | pass |

CI caveat, stated because it matters: the **iOS 27 legs are `continue-on-error`**, so their pass is real information but not a blocking guarantee.

### Not done

**Hardware keyboard** — none was paired to the device, and pairing one is a Bluetooth change to a personal iPad that nobody asked for. The software-keyboard inset is the case in the issue and it is covered in both orientations; a hardware keyboard produces a much smaller inset, so it is a strictly easier case for the same code path, but I have not measured it and am not claiming it.

### Provisioning note

The device was **already registered** on the account (`Lars Adobe iPad (2)`, ENABLED, 1 of 3 devices) — the local profile was simply stale, predating it. Refreshing via automatic provisioning consumed **no device slot**, and the regenerated app profile now carries `aps-environment: development`, which also unblocks the APNs work. The test build and its XCUITest runner have both been uninstalled; the device is clean and the mutex released.

PR #2490, opened 2026-08-27
ci(worker): sign staging APNs pushes with a staging-only key (#2452)

### LINKED ISSUE #2452: Staging and production share one APNs auth key, so their provider-JWT minters can collide
Split out of the automated review on #2448 (P1), which is correct and out of scope for that PR because the fix is an ops action, not code.

## Problem

`slicc-tray-hub` and `slicc-tray-hub-staging` are separate Workers with separate Durable Object namespaces, so each has its **own** provider-JWT minting instance after #2448. But both are provisioned with the **same** Apple credentials — `.github/workflows/worker.yml` and `.github/workflows/worker-staging.yml` both read the repository-level `APNS_TEAM_ID` / `APNS_KEY_ID` / `APNS_PRIVATE_KEY` secrets (unlike `GITHUB_CLIENT_SECRET`, which staging takes from `OAUTH_GITHUB_SECRET_STAGING`).

Apple throttles provider-token creation per `(team id, key id)` — not per worker, per connection, or per environment. Two uncoordinated minters against one key pair can therefore land two mints inside Apple's 20-minute floor, and a staging push can earn `429 TooManyProviderTokenUpdates` **in production**.

## Severity

Low but real, and it is the residual after #2448. Before that PR every tray DO in both environments minted independently — mint rate scaled with `active trays × hibernation wakes`. Now it is exactly one minter per environment, so the exposure drops from unbounded to 2. Each environment rotates roughly every 50 minutes and staging only mints when someone actually pushes there, so collisions need a staging push to fall within 20 minutes of a production mint.

Worth noting it fails silently in the same way as everything else on this path: `TooManyProviderTokenUpdates` is not a dead-token reason, so the push is dropped and warn-logged.

## Fix (ops, not code)

Issue a **second APNs auth key** in the Apple developer account and give staging its own. Apple allows two active APNs auth keys per team, so a prod key and a staging key is exactly the intended usage and fits the limit.

1. Create a new `.p8` key with APNs enabled.
2. Add `APNS_KEY_ID_STAGING` and `APNS_PRIVATE_KEY_STAGING` repository secrets (`APNS_TEAM_ID` and `APNS_TOPIC` stay shared — same team, same bundle id).
3. Point the two secret-upload steps in `worker-staging.yml` at the staging pair.

That last step is the only code change and is trivial; it is blocked on someone with Apple developer account access doing steps 1–2.

## Alternative, rejected

A token service shared across both namespaces would coordinate the two environments in code, but that is genuinely new infrastructure for a problem a second key solves for free — and it would couple staging and production on the push path, which is worse than the throttle it avoids.

Refs #2432, #2448.

trieloff:
Closes #2452.

Staging and production both read the repository-level `APNS_KEY_ID` / `APNS_PRIVATE_KEY`, so the two workers signed with the same Apple key.

Apple throttles provider-token creation per `(team id, key id)` — not per worker, per connection, or per environment. After #2448 each worker has its own singleton minter, so two uncoordinated minters shared one key pair and could land two mints inside Apple's 20-minute floor. The loser gets `429 TooManyProviderTokenUpdates` and the push is silently dropped.

This mirrors how `GITHUB_CLIENT_SECRET` already takes `OAUTH_GITHUB_SECRET_STAGING` in this same file — the separation pattern existed, it just had not been applied to Apple's key.

## What changed

Staging now signs with its own key. `APNS_TEAM_ID` and `APNS_TOPIC` stay shared: same Apple team, same bundle id. Only the signing key differs.

## Ops actions already completed

- **New key registered**: "SLICC APNs Staging", Key ID `96Y3MN5LY9`, on team `S8LB56P782`.
- **Config**: Sandbox & Production, Team Scoped (All Topics) — deliberately mirroring production key `AZS5W4GRMD`. A sandbox-only staging key would be tighter, but it would break the two-tray gateway probe that runs on staging and has to register the same token against both environments. The APNs configuration is immutable once saved, so this was not a reversible choice.
- **Secrets set**: `APNS_KEY_ID_STAGING` and `APNS_PRIVATE_KEY_STAGING` are present on the repository. The blocking precondition in the commit message is therefore satisfied — this is safe to merge.

Apple's two-key cap is per *service*, not per account: the team's other key (`75YM6KK373`, Media Services/MusicKit) does not consume an APNs slot, so no existing key had to be revoked.

## Verification

- YAML parses; both deploy-attempt blocks updated (the file retries staging deploys, so the secret list appears twice — missing one would leave a retry signing with production's key).
- Touched-file debt gate: clean.
- Not run locally: full lint/typecheck/test — this worktree has no `node_modules` and the change touches no source. CI covers it.

## Residual risk

Production and staging remain on the same `APNS_TEAM_ID` and `APNS_TOPIC`, which is correct — they are the same app. Nothing else shares the signing key now.

Refs #2448, #2062.

### DIFF
--- a/.github/workflows/worker-staging.yml
+++ b/.github/workflows/worker-staging.yml
@@ -8,6 +8,18 @@ on:
       - 'packages/cloud-core/**'
       - 'packages/webapp/src/providers/**'
       - 'packages/webapp/providers/**'
+      # This workflow is the ONLY place staging's APNs secrets are uploaded
+      # (ci.yml's staging deploy pushes just TURN/GitHub/E2B). Without this
+      # path a change to the secret wiring below never takes effect until some
+      # unrelated worker PR happens to trigger a run — so staging would keep
+      # signing with whatever key was last uploaded.
+      - '.github/workflows/worker-staging.yml'
+  # Deliberately no `workflow_dispatch`: every deploy step here is gated on
+  # `github.event.pull_request.head.repo` (fork guard) and the deployment
+  # record reads `context.payload.pull_request.head.sha`, so a manual run
+  # would skip the deploy and report success having uploaded nothing. Rotating
+  # a secret therefore goes through a PR that touches this file — which the
+  # path above now guarantees runs.
 
 # Serialize across ALL PRs: this job mutates two shared singletons — the
 # `slicc-staging` e2b template alias (build overwrites it) and the
--- a/.github/workflows/worker-staging.yml
+++ b/.github/workflows/worker-staging.yml
@@ -168,9 +168,15 @@ jobs:
         env:
           CLOUDFLARE_TURN_API_TOKEN: ${{ secrets.CLOUDFLARE_TURN_API_TOKEN }}
           GITHUB_CLIENT_SECRET: ${{ secrets.OAUTH_GITHUB_SECRET_STAGING }}
+          # APNs: staging signs with its OWN key (#2452). Apple throttles
+          # provider-token creation per (team id, key id) — not per worker or
+          # environment — so sharing production's key would let the two
+          # singleton minters (#2448) collide inside Apple's 20-minute floor and
+          # silently drop a push. TEAM_ID and TOPIC stay shared: same Apple team,
+          # same bundle id. Key: "SLICC APNs Staging" (96Y3MN5LY9).
           APNS_TEAM_ID: ${{ secrets.APNS_TEAM_ID }}
-          APNS_KEY_ID: ${{ secrets.APNS_KEY_ID }}
-          APNS_PRIVATE_KEY: ${{ secrets.APNS_PRIVATE_KEY }}
+          APNS_KEY_ID: ${{ secrets.APNS_KEY_ID_STAGING }}
+          APNS_PRIVATE_KEY: ${{ secrets.APNS_PRIVATE_KEY_STAGING }}
           APNS_TOPIC: ${{ secrets.APNS_TOPIC }}
           E2B_API_KEY: ${{ secrets.E2B_API_KEY }}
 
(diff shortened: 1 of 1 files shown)


trieloff:
Good catch, and confirmed rather than taken on trust — both halves of it hold:

- `worker-staging.yml` triggers only on `pull_request` against four `packages/**` paths, so a commit touching only the workflow never runs it.
- `ci.yml`'s staging deploy uploads `CLOUDFLARE_TURN_API_TOKEN`, `GITHUB_CLIENT_SECRET` and `E2B_API_KEY` only — no APNs secrets at all, so it could not have applied the change either.

Staging would have kept signing with production's key until some unrelated worker PR happened to trigger a run. That is precisely the collision this PR exists to remove, so the fix would have appeared to land while doing nothing.

Fixed in e4e2b96 by adding `.github/workflows/worker-staging.yml` to the workflow's own `paths`, so a change to the secret wiring applies on the PR that makes it — including this one.

**On the `workflow_dispatch` half of the suggestion: deliberately not done, and worth explaining.** Every deploy step in this workflow is gated on `github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository` as a fork guard, and the deployment record reads `context.payload.pull_request.head.sha`. Under a manual run `github.event.pull_request` is null, so those conditions evaluate false: the run would go green having deployed nothing and uploaded no secrets. Making dispatch genuinely work means reworking the fork guards and the deployment-record step, which is a wider change than this ops fix warrants and would put the working PR path at risk. A trigger that silently does nothing is worse than no trigger, so I left a comment in the file recording why rather than adding one.

Both staging secrets (`APNS_KEY_ID_STAGING`, `APNS_PRIVATE_KEY_STAGING`) are already set on the repository, so this PR's own staging run will upload them.

PR #2504, opened 2026-08-27
fix(fs): materialize process-substitution fds inside a scoop sandbox

### LINKED ISSUE #2502: Process substitution <(...) works in the cone but fails with ENOENT on /dev/fd/63 inside scoops
## Summary

Process substitution `<(...)` **works in the cone and is broken inside scoops.** In a scoop the substituted path is produced but never materialized, so the consuming command fails with `No such file or directory` on `/dev/fd/63`.

This is a harness-consistency bug with a nasty shape: an agent develops and verifies a command in the cone, delegates the same command to a scoop, and the scoop gets a *file-not-found* error — which reads as "my input path is wrong", not "this shell feature is unavailable here". Worse, several natural uses degrade into **silently wrong answers rather than errors**: `grep -c '^+' <(...)` prints `0`, and `diff <(a) <(b) | grep ...` yields an empty result set that is indistinguishable from "no differences".

## Environment

- SLICC `6.99.8` (build `6.99.8-mtb9anv9`)
- Cone: unrestricted `VirtualFS`. Scoop: default sandbox (`writablePaths` = `/scoops/<folder>/`, `/shared/`), no `allowedCommands` restriction.

## Reproduction

Identical commands, same build, same session.

**In the cone — works:**

```
$ diff <(echo alpha) <(echo beta); echo "EXIT=$?"
--- /dev/fd/63
+++ /dev/fd/62
@@ -1,1 +1,1 @@
-alpha
+beta
EXIT=1

$ wc -l <(echo one; echo two); echo "EXIT=$?"
2 /dev/fd/63
EXIT=0

$ cat <(echo hello-from-procsub); echo "EXIT=$?"
hello-from-procsub
EXIT=0
```

**In a scoop — fails:**

```
$ diff <(echo alpha) <(echo beta); echo "EXIT=$?"
diff: /dev/fd/63: No such file or directory
EXIT=2

$ wc -l <(echo one; echo two); echo "EXIT=$?"
wc: /dev/fd/63: No such file or directory
EXIT=1
```

The shell **parses** `<(...)` and substitutes a `/dev/fd/63` argument in both contexts — so this is not a syntax-level gap. Only the fd's materialization differs.

## Analysis

Process substitution appears to be implemented by materializing a real VFS path at `/dev/fd/<n>`. That works against the cone's unrestricted `VirtualFS`. A scoop runs under `RestrictedFS`, where the only writable locations are the scoop's own configured grants, `ALWAYS_WRITABLE_PREFIXES` (`packages/webapp/src/fs/restricted-fs.ts`, currently just `['/tmp/']`), and paths registered in `VIRTUAL_DEVICES` — which holds `/dev/null` alone. `/dev/fd/*` is in none of them, so the write is walled and the subsequent open sees nothing.

The existing comment above `ALWAYS_WRITABLE_PREFIXES` predicts both observed symptoms exactly:

> `SudoManager.getPolicyForScoop`. Changing one without the other either re-introduces the approval prompt or walls the write underneath it.

Two symptoms, one cause, depending on whether the path happens to match a grant:

- **walled** → today's `ENOENT` on `/dev/fd/63` (this report);
- **approval prompt** → measured on `6.96.2`, where a scoop using `<(...)` instead raised a cone `sudo` prompt for `write /dev/fd/63`. That variant is arguably worse operationally: only the cone can clear it, so an **unattended scoop stalls indefinitely** on a shell feature it has no idea is privileged. And because the fd number varies per invocation, no exact-detail "Always" grant can pre-empt it — it needs a `/dev/fd/*` pattern.

## Impact

- A working cone command silently changes behaviour when delegated. This is precisely the inconsistency the harness should not have: the same shell, the same syntax, a different answer.
- The failure mode is a plausible-looking `ENOENT`, so an agent's natural next step is to doubt its own paths and retry variations, burning turns.
- The `grep`/pipeline cases return **empty output with a success-shaped exit**, i.e. a wrong conclusion rather than a failure.
- `<(...)` is documented in `docs/shell-reference.md` and is called out as working in the runtime's own guidance, so agents are actively encouraged to use it.

## Suggested fix

Make the ephemeral fds work in a scoop rather than making them approvable — an approval prompt for process substitution is the wrong shape regardless of who answers it.

`packages/webapp/src/fs/virtual-device-paths.ts` already establishes the pattern, and its doc comment states the design intent this bug violates:

> The sudoers matcher (`base/sudoers.ts`) permits CONTENT writes (`writeFile`) to these paths so scoops never hit an approval prompt for a write that discards its payload.

**`/dev/fd/*` must not simply be added to `NO_OP_WRITE_DEVICE_PATHS`.** That module's own rule forbids it, correctly: "Only add a path here when its device `write` truly discards the payload." A process-substitution fd's payload is *not* discarded — it is read back by the consuming command in the same pipeline. So this needs a sibling concept rather than a new entry in that list: an **ephemeral, process-local pipe** that is auto-permitted for both read and write inside a scoop's sandbox, without a sudoers match and without becoming a persistent, user-reachable filesystem location.

Reasons the security posture is unchanged by permitting it:

- the path is created by the shell for the duration of one command, not named by the agent;
- its read end is the same pipeline, so it is not an exfiltration channel out of the sandbox;
- it is not addressable across commands, scoops, or turns;
- a scoop can already achieve the identical data flow with a temp file under `/tmp/`, which is unconditionally writable — so gating `<(...)` buys no containment, it only breaks the ergonomic spelling.

Whatever shape the fix takes, please also:

- keep `/dev/null`'s existing semantics untouched (payload genuinely discarded — different case, same neighbourhood);
- ensure the read side works too, not just the write: the failing operation observed is the consumer's **open**, so a write-only permit would still leave `diff`/`wc` at `ENOENT`;
- keep the fd invisible to directory listings if that is what the cone currently does, so behaviour does not diverge in the other direction.

## Tests

The regression guard that matters is **cone/scoop equivalence**, since a cone-only test would have passed throughout this bug's life. Please assert that `<(...)` produces identical stdout and exit code under a `RestrictedFS`-backed scoop sandbox as under an unrestricted cone FS, for at least: a single substitution consumed by `cat`, two substitutions consumed by `diff` (exercising the second fd, `/dev/fd/62`), and one consumed by a command that reports the path in its output (`wc -l`). Also assert that **no sudo approval is requested** during those runs — that is the 6.96.2 variant of this bug and it needs its own guard.

## Open question, not yet answered

Whether granting a scoop an explicit `writablePaths` entry for `/dev/fd/**` restores the behaviour was **not** determined — the probe scoop for it did not report back. So I cannot say from measurement whether the wall is in the sudoers policy layer or in `RestrictedFS`'s visibility layer. Treat the `ALWAYS_WRITABLE_PREFIXES` / `VIRTUAL_DEVICES` reading above as a well-supported inference from the source, not as a verified localization. Everything under **Reproduction** is directly measured.

trieloff:
## Summary

Process substitution (`<(cmd)` / `>(cmd)`) worked in the cone and was broken inside a scoop: the consuming command failed with `ENOENT` on `/dev/fd/63`, and a bare redirect to the same path reported exit 0 while the payload became unreachable. Descriptors are now answered by the sandbox itself, in both gating layers, so a scoop's output matches the cone's byte for byte and no approval prompt is ever raised.

Fixes #2502.

## Root cause as actually found (measured, not inferred)

I reproduced all four symptoms in a test harness shaped like the production scoop stack (`RestrictedFS` in `sudo-delegated` mode wrapped in `SudoFS` with the non-cone `require-approval` default), and the localisation matches the issue's updated comment: **the block is the write-permission decision, not materialization.**

just-bash models a substitution on a real backing file at `/dev/fd/<n>` (`interpreter/process-substitution.ts`) — the body's stdout is written there during word expansion, the outer command reads the path back, and the entry is released when the command ends. So yes, it genuinely writes a VFS file; it does not merely *name* the path.

- **Cone:** the write hits an unrestricted `VirtualFS`, which auto-creates parents, so `/dev/fd/63` is an ordinary file and the read works.
- **Scoop:** `RestrictedFS` in `sudo-delegated` mode makes `checkWrite` a no-op, so the decision falls to `SudoFS`, whose `require-approval` default has no rule to match → **a cone approval for `write /dev/fd/63`**. Once permitted, the write passes *through* to the shared tree, at a path outside the scoop's grants, and `RestrictedFS` then filters the consumer's open back to `ENOENT`. That is both reported symptoms from one cause, and it explains the write/read asymmetry exactly: the producer's write succeeded (somewhere the scoop cannot see), the consumer's open was walled.

Two findings worth recording:

1. **A denying broker hides the bug.** just-bash self-heals — if the descriptor write *throws*, it mounts a private in-memory FS at `/dev/fd` and retries (`processSubstitutionFsMounted`, designed for read-only sandboxes). So a scoop whose approval is denied or times out ends up working *by accident*, ~5 minutes late. Only the permitted path fails. My first probe passed for this reason; a test written with a denying broker would be worthless as a guard.
2. **Any code path that can still prompt is a live bug**, not just an annoyance: I reproduced 1–2 broker calls *per command* (`/dev/fd/63` and `/dev/fd/62` separately), which is the 6.96.2 stall variant.

## What changed

Both gating layers had to agree — this is the failure mode the `ALWAYS_WRITABLE_PREFIXES` comment already predicted ("either re-introduces the approval prompt or walls the write underneath it").

- **`fs/virtual-device-paths.ts`** — new `isEphemeralFdPath()` predicate (`/dev/fd/<n>`, numbered descriptors only). A deliberate **sibling** of `NO_OP_WRITE_DEVICE_PATHS`, explicitly *not* a member: a descriptor's payload is read back by the consuming command, so discarding it would turn `cat <(echo hi)` from an error into silently empty output. The module still imports nothing, so the pure sudoers matcher can use it.
- **`base/sudoers.ts`** — `matchPath` resolves an ephemeral descriptor to `nopasswd-allow` for **both** ops, before any rule is consulted, so the `require-approval` default disposition can never escalate one. Self-protection is still checked first and stays absolute.
- **`fs/ephemeral-fd-store.ts`** (new) — a private, in-memory byte store; one instance per `RestrictedFS`. Backs read / readText / write / stat / lstat / statSync / lstatSync / exists / realpath / rm for descriptor paths, beside the tree rather than in it.
- **`fs/restricted-fs.ts`** — descriptor paths route to that store. `canWrite` reports them as writable so the predicate stays honest about what `writeFile` accepts.
- **`docs/approvals.md`** — new "Ephemeral shell descriptors — `/dev/fd/<n>` (never gated)" section next to the `/tmp` grant, documenting the two-layer rule and the private/ephemeral semantics. The three touched source doc comments (`virtual-device-paths.ts`, the `ALWAYS_WRITABLE_PREFIXES` block, `matchPath`) were updated to match, since a stale comment here is what makes this area easy to break.

Explicitly **not** done: `/dev/fd/*` was not added to `NO_OP_WRITE_DEVICE_PATHS` or `VIRTUAL_DEVICES`, and `ALWAYS_WRITABLE_PREFIXES` was not widened to `/dev/` (that would permit writes to any future device path, including ones with observable effects, and would resolve them against the shared tree).

## Why the sandbox posture is unchanged

- The path is minted by the shell for the duration of one command, not named by the agent.
- Its read end is the same pipeline, so it is not an exfiltration channel out of the sandbox.
- It is not addressable across commands, scoops or turns: the store is per-`RestrictedFS` and private, and **nothing is written to `/dev/fd` in the shared VFS** — asserted in tests. This is strictly *tighter* than the previous behaviour, where a permitted descriptor write landed in the shared tree where the cone and sibling scoops could see and clobber it.
- A scoop can already express the identical data flow with a temp file under `/tmp/`, which is unconditionally writable — so gating `<(...)` bought no containment, it only broke the ergonomic spelling.
- `/dev` and `/dev/fd` remain non-existent in a scoop's view: no listing, no `stat`, no `readDir` entry. Nothing materializes a real, listable `/dev/fd/` directory.
- Only numbered descriptors are exempt: `/dev/fd/name`, `/dev/fd`, `/dev/fdx/63` and `/dev/sda` all still gate normally, and an ordinary out-of-sandbox write still escalates (both asserted).
- `/dev/null` semantics are untouched, and the cone is untouched — every equivalence case compares scoop output *against a live cone run* in the same test.

## Before/after (default scoop sandbox)

| command | before | after |
| --- | --- | --- |
| `cat <(echo hello-from-procsub)` | `cat: /dev/fd/63: No such file or directory`, exit 1, **+1 sudo prompt** | `hello-from-procsub`, exit 0, no prompt |
| `diff <(echo alpha) <(echo beta)` | `diff: /dev/fd/63: No such file or directory`, exit 2, **+2 sudo prompts** (`/dev/fd/63`, `/dev/fd/62`) | `--- /dev/fd/63` / `+++ /dev/fd/62` / `-alpha` / `+beta`, exit 1, no prompt |
| `wc -l <(echo one; echo two)` | `wc: /dev/fd/63: No such file or directory`, exit 1, **+1 sudo prompt** | `2 /dev/fd/63`, exit 0, no prompt |
| `echo probe > /dev/fd/63` then `cat /dev/fd/63` | exit 0 (payload landed outside the sandbox), then `cat: /dev/fd/63: No such file or directory`, exit 1 | exit 0, then `probe`, exit 0, no prompt |
| `ls -la /dev/` · `ls -la /dev/fd/` | `No such file or directory`, exit 2 | unchanged (`No such file or directory`, exit 2) |
| `echo x > /dev/null` | exit 0 | unchanged (exit 0) |

Cone output for every row above is identical before and after.

## How it was tested

New `packages/webapp/tests/shell/process-substitution-sandbox.test.ts` (14 cases) is the guard that matters: each case runs the command **twice** — once on an unrestricted cone `VirtualFS`, once on the production scoop stack — and compares stdout and exit code, because a cone-only test would have passed for this bug's entire life. It covers one substitution via `cat`, two via `diff` (asserting `/dev/fd/62` specifically, so a first-fd-only fix fails), one via `wc -l`, the write→read pair, `rm` / `rm -f` of an unopened descriptor, the no-leak-into-the-shared-VFS assertion, `/dev` still not listable, `/dev/null` still a no-op, an out-of-sandbox write still gating, and **`expect(requestApproval).not.toHaveBeenCalled()`**. The broker is wired to *allow* on purpose — a denying broker makes the bug disappear behind just-bash's self-heal.

I verified the guards by reverting only the `src/` changes and re-running: **7 of 14 fail**, including all three command shapes and the no-approval assertion. Also added `tests/fs/ephemeral-fd-store.test.ts` (9 cases: predicate boundaries, byte/text round-trip, no mutable view of stored bytes, normalization, ENOENT for an unopened descriptor), 10 `RestrictedFS` cases (per-sandbox isolation, metadata surfaces, no tree leak, exemption does not extend to other `/dev` paths, `/dev/null` intact) and 15 `matchPath` invariant cases.

Full pass, all green on this branch: `npm run verify` (7/7), `check-touched-exemptions.mjs`, `npm run typecheck`, `npm run test` (17,290 passed), `npm run test:coverage`, `npm run build`, `npm run build -w @slicc/chrome-extension`, `npm run bundle-size`.

## Deliberately left out

- **`packages/webapp/CLAUDE.md` was not touched.** It sits at 19,998/20,000 chars, so the `/tmp` never-rule could not be extended to mention `/dev/fd` without trimming unrelated content. The invariant is documented in `docs/approvals.md` and in the source doc comments on both gating layers instead.
- **No upstream change to just-bash.** The self-heal fallback stays as-is; this fix means the sandbox no longer needs it.
- **A cone/scoop divergence is retained on purpose:** the cone lists a live descriptor under `ls /dev/fd` (it is a real file there), a scoop does not. Making a scoop list them would mean materializing a real `/dev/fd/` directory, which #2502 explicitly asks not to do; the reverse (hiding them in the cone) would change cone behaviour, which it also asks not to do.
- No change to `docs/shell-reference.md`: it does not document `<(...)` today (grep finds no mention), so there was nothing to correct.

### DIFF
--- a/docs/approvals.md
+++ b/docs/approvals.md
@@ -173,6 +173,17 @@ Unlike `/tmp`, this space is **private per sandbox and never enters the tree**:
 - a descriptor that was never written raises `ENOENT` on read, the same loud
   failure the cone gives for an unopened fd.
 
+The exemption covers exactly what process substitution needs — a content write,
+the read back (including as either end of a `copyFile`, which is how `cp <(…)`
+arrives), and the release (`rm`). Every **tree-shape** op on a descriptor path is
+refused with `EACCES` by `RestrictedFS.refuseDescriptorTreeOp`: `mkdir`,
+`symlink`, `rename` and `mount`/`unmount`/`refreshMount`. That refusal is the
+containment for those ops, and it has to live in the ACL layer, because the
+policy layer deliberately answers `nopasswd-allow` rather than raising a prompt.
+Without it, a `mkdir /dev/fd/63` in a `sudo-delegated` sandbox would fall through
+to the shared `VirtualFS` and create exactly the cone-visible, cross-scoop
+`/dev/fd` entry the private store exists to prevent.
+
 This is not a widening of the sandbox: a scoop can already express the same data
 flow with a temp file under `/tmp`, which is unconditionally writable, so gating
 `<(...)` bought no containment — it only broke the ergonomic spelling. It is
--- a/docs/approvals.md
+++ b/docs/approvals.md
@@ -140,6 +140,46 @@ Two consequences worth knowing:
   independently, so both layers must agree or the write is walled underneath
   the grant.
 
+### Ephemeral shell descriptors — `/dev/fd/<n>` (never gated)
+
+Process substitution (`<(cmd)` / `>(cmd)`) is modelled on a backing file at
+`/dev/fd/<n>`, numbered downward from 63 exactly as bash numbers it: the body's
+stdout is written there during word expansion, the outer command reads the path
+back, and the descriptor is released when that command ends.
+
+Those paths are **never a policy subject**. `matchPath` resolves them to
+`nopasswd-allow` for both reads and writes (`isEphemeralFdPath` in
+`packages/webapp/src/fs/virtual-device-paths.ts`), before any rule is consulted —
+self-protection is still checked first and stays absolute. The reason is that
+there is nothing for a prompt to be _about_: the path is minted by the shell for
+the duration of one command, the fd number changes every invocation (so no
+"Always" grant could ever pre-empt the prompt), and the only reader is the
+consuming command in the same pipeline. A sandboxed scoop that hit such a prompt
+stalled indefinitely, because only the cone can clear it (#2502).
+
+The matching ACL exemption is the private `EphemeralFdStore` held by each
+`RestrictedFS` (`packages/webapp/src/fs/ephemeral-fd-store.ts`). As with `/tmp`,
+`SudoFS` and `RestrictedFS` gate independently, so both layers have to agree:
+granting only in the policy leaves the consumer's open walled to `ENOENT` while
+the write lands _outside_ the sandbox, and exempting only in the ACL leaves the
+approval prompt in place.
+
+Unlike `/tmp`, this space is **private per sandbox and never enters the tree**:
+
+- descriptors are not addressable across scoops, commands or turns, and nothing
+  is written to `/dev/fd` in the shared VFS;
+- `/dev` and `/dev/fd` remain non-existent in a scoop's view — no listing, no
+  `stat`, no `readDir` entry;
+- a descriptor that was never written raises `ENOENT` on read, the same loud
+  failure the cone gives for an unopened fd.
+
+This is not a widening of the sandbox: a scoop can already express the same data
+flow with a temp file under `/tmp`, which is unconditionally writable, so gating
+`<(...)` bought no containment — it only broke the ergonomic spelling. It is
+also deliberately **not** modelled as a no-op device write like `/dev/null`: a
+descriptor's payload is read back, so discarding it would turn `cat <(echo hi)`
+from an error into silently empty output.
+
 ### Live reload
 
 `SudoManager` watches `/etc` via the shared `FsWatcher` and re-reads + re-merges
--- a/packages/webapp/src/base/sudoers.ts
+++ b/packages/webapp/src/base/sudoers.ts
@@ -379,6 +379,14 @@ export function matchPath(
   // the write is the one that escalates today, but an explicit `Read` rule
   // must not be able to gate the consumer's open either. Self-protection is
   // checked first and stays absolute.
+  //
+  // Unlike the `/dev/null` exemption above, this one is deliberately NOT
+  // limited to content writes: a prompt is the wrong answer for a descriptor
+  // whatever the op is. Containment for the STRUCTURAL ops (`mkdir`, `symlink`,
+  // `rename`, `mount`) lives in the other layer instead, where it belongs —
+  // `RestrictedFS.refuseDescriptorTreeOp` rejects them outright, so the two
+  // layers still agree that no op on a descriptor path may create a shared-tree
+  // entry.
   if (isEphemeralFdPath(normalized)) return 'nopasswd-allow';
   return resolve(op === 'read' ? policy.read : policy.write, normalized);
 }
--- a/packages/webapp/src/base/sudoers.ts
+++ b/packages/webapp/src/base/sudoers.ts
@@ -15,7 +15,7 @@
  */
 
 import { normalizePath, pathGlobToRegExp } from '../fs/path-utils.js';
-import { isNoOpWriteDevicePath } from '../fs/virtual-device-paths.js';
+import { isEphemeralFdPath, isNoOpWriteDevicePath } from '../fs/virtual-device-paths.js';
 import { createLogger } from './logger.js';
 
 export { pathGlobToRegExp } from '../fs/path-utils.js';
@@ -371,5 +371,14 @@ export function matchPath(
     if (isSelfProtectedWrite(normalized)) return 'require-approval';
     if (opts?.isContentWrite && isNoOpWriteDevicePath(normalized)) return 'nopasswd-allow';
   }
+  // The shell's ephemeral descriptors (`/dev/fd/<n>`, minted per command by
+  // process substitution) are never a policy subject: the fd number changes
+  // every invocation, so no "Always" grant could pre-empt the prompt, and an
+  // unattended scoop would stall forever on a cone approval for a path its
+  // own next command would read straight back (#2502). Exempt for BOTH ops —
+  // the write is the one that escalates today, but an explicit `Read` rule
+  // must not be able to gate the consumer's open either. Self-protection is
+  // checked first and stays absolute.
+  if (isEphemeralFdPath(normalized)) return 'nopasswd-allow';
   return resolve(op === 'read' ? policy.read : policy.write, normalized);
 }
--- a/packages/webapp/src/fs/ephemeral-fd-store.ts
+++ b/packages/webapp/src/fs/ephemeral-fd-store.ts
@@ -1,0 +1,109 @@
+/**
+ * EphemeralFdStore — private byte backing for the shell's `/dev/fd/<n>` paths.
+ *
+ * just-bash models process substitution (`<(cmd)` / `>(cmd)`) on a real file at
+ * `/dev/fd/<n>`: the body's stdout is written there during word expansion, the
+ * outer command reads the path back, and the entry is `rm`'d when the command
+ * finishes. The cone runs that against an unrestricted `VirtualFS`, where it is
+ * simply an ordinary file.
+ *
+ * A sandboxed scoop must NOT resolve those writes against the shared tree. Two
+ * things go wrong when it does (#2502): the write lands at `/dev/fd/<n>` OUTSIDE
+ * the scoop's grants — visible to the cone and to sibling scoops, and clobbered
+ * by whichever scoop reaches fd 63 next — while the consumer's read is filtered
+ * back to `ENOENT` by the sandbox ACL, so the shell is told the write succeeded
+ * and the payload is then unreachable.
+ *
+ * This store is the sandbox's answer instead: one instance per `RestrictedFS`,
+ * so a descriptor is
+ *
+ *   - readable and writable without a sudoers match (an fd number varies per
+ *     invocation, so no grant could pre-empt an approval prompt anyway);
+ *   - private to one sandbox — nothing is addressable across scoops;
+ *   - absent from every directory listing, because it lives beside the tree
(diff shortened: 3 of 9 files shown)


trieloff:
Both Codex P1s were real and are fixed in b079e46ea (also rebased onto the latest `main`).

**1. Structural writes bypassed the gate** — confirmed exactly as described: `matchPath` answered `nopasswd-allow` for every op, and with `checkWrite` a no-op under `sudo-delegated`, `mkdir -p /dev/fd/63` reached the shared `VirtualFS` and created the cone-visible `/dev/fd` entry the private store exists to prevent.

I did **not** narrow the policy exemption to content writes, because that puts a cone approval back on a `/dev/fd/*` path — the 6.96.2 shape #2502 explicitly rules out (the fd number changes per invocation, so no "Always" grant can pre-empt it, and an unattended scoop stalls). Containment goes in the layer that owns tree shape instead: `RestrictedFS.refuseDescriptorTreeOp` rejects `mkdir`, `symlink`, `rename` and `mount`/`unmount`/`refreshMount` on a descriptor path with `EACCES`. The two layers still agree that no op on a descriptor may create a tree entry — one says "not a policy subject", the other "not a tree path". Both `matchPath`'s comment and `docs/approvals.md` now state that split explicitly.

**2. `copyFile` was not store-aware** — also confirmed, in both directions. `cp <(echo hi) out` reaches `VfsAdapter.cp` → `copyFile`, which rejected the descriptor via `isAllowed(src)`, so process substitution was still broken for that consumer; and `cp in >(consumer)` wrote to the shared tree, leaving the writer body with empty input. Each end is now resolved independently: the descriptor end through the store, the tree end through the ordinary ACL checks (asserted, including that an out-of-sandbox source into a descriptor still fails and that neither direction creates a tree entry).

**A related test-harness weakness the review surfaced:** the scoop harness used a bare policy, so *every* in-sandbox write prompted as well — which masked which operation a descriptor case was responsible for. It now compiles the real per-scoop grants (`generateScoopSudoers` + `builtinScoopGrants`, mirroring `getPolicyForScoop`).

New guards: `cp <(…)` cone/scoop equivalence, `mkdir`/`mv`/`ln -s` at a descriptor path asserting non-zero exit **plus** no shared-VFS entry **plus** no broker call, and unit coverage for both copy directions and every refused tree op. I verified they are real guards by reverting only the hardening: `cp`, `mkdir` and `ln -s` fail without it.

Full pass green on the rebased head: `verify` (7/7), `check-touched-exemptions`, `typecheck`, `test` (17,305), `test:coverage`, both builds, `bundle-size`.

Issue #1957, opened 2026-08-06
feat(worker): tray supersede should use redirect semantics, not 409 Conflict


trieloff:
## Problem

The tray hub reports a superseded tray as **HTTP 409 Conflict**. It is not a conflict — it is a redirect, and every client already treats it as one.

When a leader's tab reconnects it mints a fresh tray and `POST`s `/api/tray/:trayId/supersede` against the old one (`tray-leader.ts:506`). From then on the old tray answers with the replacement's address:

```
POST /join/<old-token>?json=true
409 { "role": "follower", "result": { "action": "fail", "code": "TRAY_SUPERSEDED",
      "error": "This session moved to a new tray after the leader reconnected",
      "joinUrl": "<replacement>" } }
```

`packages/cloudflare-worker/src/session-tray.ts:552` (attach) and `:561` (the `GET` status probe).

Two things are wrong with 409 here:

1. **It says the opposite of what happened.** 409 means "your request conflicts with the current state of the resource; resolve it and retry." Nothing about this is retryable and there is no conflict to resolve — the resource *moved*, permanently, and the response body says exactly where. RFC 9110 has status codes for that.
2. **It buries the redirect in `action: "fail"`.** Every follower has to special-case one `fail` code back into a success path. That framing is precisely what caused #1956: iOS listed the other two `fail` codes in its allowlist, treated a valid redirect as a malformed reply, and showed users a raw 409 body. Four independent implementations now re-derive "this `fail` is actually a `follow`" — the browser (`tray-webrtc.ts:490`), the Go CLI (`conn.go:139`), the iOS join path (`SupersedeRedirect.swift`), and the iOS/macOS liveness probe (`SessionReachability.swift:89`) — and the one that got it wrong stayed wrong for as long as nothing exercised it.

## Proposed solution

Move to redirect semantics, additively, so already-shipped followers keep working.

**Status code: `308 Permanent Redirect`.** The attach is a `POST`, and 301/302 historically let clients rewrite the method to `GET`; 307/308 preserve method and body. The supersede is permanent — the old tray's leader socket will never reconnect (`session-tray.ts:541`) — so 308 over 307. Pair it with `Location: <replacement join URL>`.

**Step 1 — additive, breaks nothing.** Keep the 409 and the body exactly as they are, and add the RFC 8288 `Link` header the worker already stamps on every response via `applySliccLinks` (`src/links.ts`):

```
[...]

PR #2505, opened 2026-08-27
feat(followers): follow the successor-version link on every follower

trieloff:
Follow-up to #2499 (merged), which made the tray hub stamp the header. Part of #1957 — this is the client half of step 1; the status flip to `308` stays open there.

## Summary

All five followers now read `Link: <replacement>; rel="successor-version"` off a superseded tray's reply and follow it: the browser follower, the Electron follower, the Go CLI, the iOS attach loops, and the native liveness probe. Nothing about the 409 or the JSON body changed, and a hub that doesn't send the header behaves exactly as before.

## Why

The redirect used to be reachable only by pattern-matching `result.code === 'TRAY_SUPERSEDED'` out of a body shape each client models separately. That is what #1956 was: iOS listed the other two `fail` codes in its allowlist, treated a valid redirect as a malformed reply, and stayed wrong for as long as nothing exercised it. Four implementations re-deriving "this `fail` is actually a `follow`" is four chances to get it wrong.

So the rule is now the same sentence in four languages: **a plan that names a replacement is a redirect, whatever the body called it.**

- The link wins over the body's `joinUrl` — it is the channel that survives a body-shape change.
- The link stands alone: an undecodable body, or one this build doesn't recognize, is no longer a dead end.
- The chase is gated on *having an address*, never on the failure code, so step 2's `action: "redirect"` needs no client change to work.

- **Bounds stay ours.** 5 hops, 1s apart, a fresh controller id per hop, `URLSession`/`http.Client`/`fetch` never auto-following. #1957 flags this explicitly: our cap is deliberately tighter than any platform's, and persisting the replacement depends on *seeing* the hop.
- **`shared-ts`** — `successorVersionFromLinkHeader` lives next to the response type it annotates. It is deliberately narrow (the hub emits one well-formed link plus its standard rel set) rather than a second full RFC 8288 parser; the webapp's `src/net/link-header.ts` is the complete one and can't be shared downward without shared-ts depending on webapp.
- **Four parsers, one pinned vector table** — `shared-ts`, Go, and both Swift packages pin the same 15 cases (quoted/unquoted rel, token lists, case-insensitivity, commas and semicolons inside quoted params, relative targets rejected, percent-encoded targets preserved). Drift between them is the #1956 failure mode, so the tables name each other and say "add a case to all four".
- **`tray-webrtc.ts` is untouched.** `normalizeFollowerAttachResponse` maps any successor-bearing reply onto the existing supersede plan, so every downstream consumer of that plan redirects with no change of its own.
- **Two Swift copies of `SupersedeLink`** — `swift-traysession` is deliberately Foundation-only with no dependency on `swift-trayfollower` (that's what keeps WebRTC out of `swift-launcher`), and the probe needs the same reading. The duplication is named in both files and both CLAUDE.md files.
- **Go control flow** — `Dial` checks for a replacement before it switches on the action, and `attachWait` stops sleeping against a tray that already moved. `handleAttachFail` is now only terminal failures; `followSupersede` owns the hop and its bound.

## Cross-cutting impact

- **Floats exercised**: browser follower (webapp), Electron follower (node-server), `slicc` CLI (Go), iOS follower, Sliccstart/launcher liveness probe. The worker is untouched here.
- **Behavior when the header is absent** (any hub older than #2499, or a cached one): every path falls back to the body's `joinUrl` exactly as before. All the pre-existing body-only tests still pass unchanged.
- **`SupersedeRedirect.outcome` semantics changed** for one input: a plan carrying a replacement with a *non*-`TRAY_SUPERSEDED` code used to be `.terminal`, and is now `.follow`. That input cannot arise from the body (the field is only populated for supersede) — it exists so a link-only redirect works. `SupersedeRedirectTests` pins both halves.
- **CORS**: the browser follower can read the header because #2499 added `Access-Control-Expose-Headers: Link` to the capability routes. Same-origin followers never needed it.
- **Secrets**: replacement join URLs carry the session secret. Nothing here logs one — the browser follower logs `supersededByLink: <boolean>`, the Swift probe still never persists a replacement, and `SupersedeRedirect.failureMessage` stays URL-free.
- **Boy-scout**: `electron-tray-follower.ts` (16 occurrences) and `tray-follower.ts` (5) are off the `record-string-unknown` debt list. The signalling replies, bootstrap events, ICE entries and attach validators now name the shapes they actually read (`TraySignalingReply`, `BootstrapEventReply`, `UnverifiedAttachResponse`, …), and `FollowerBootstrapRequest` / `TrayIceCandidate` from `shared-ts` replace two open bags outright. Baseline shrunk by hand, never regenerated.

## Test plan

- [x] `npm run lint` · `npm run typecheck` (all 9 projects) · `npm run verify` (7/7) — all clean
- [x] `npm run test` — 17,278 passing, 0 failures
- [x] `npm run test:coverage:shared` / `:node-server` — pass; `:webapp` passes its floors (see pre-existing failures below)
- [x] `cd packages/slicc-cli && make check` — gofmt, tidy, vet, golangci-lint (0 issues), race tests, coverage 69.6% (floor 58%)
- [x] `swift test` — `swift-trayfollower` 215, `swift-traysession` 61, all passing; `swift format lint --strict` + SwiftLint clean on every changed file
- [x] `swift-coverage-check.sh` — trayfollower 70.35% lines (floor 65) / 54.59% functions (48) / 78.67% regions (75); traysession 92.90% (91) / 88.99% (87) / 88.02% (85)
- [x] `xcodegen generate` + `xcodebuild build -scheme SliccFollower -destination 'generic/platform=iOS Simulator'`
- [x] `check-touched-exemptions.mjs` — OK (30 changed files, 0 on any debt list)
- [x] **New behavior is covered per implementation**: pinned vectors ×4; `attachTrayFollower` header-only redirect + unreadable body + link-beats-body (`tray-follower.test.ts`); end-to-end hop through `FollowerTrayManager` (`tray-webrtc.test.ts`); `attachWithRedirects` link-only + link-beats-body (`electron-tray-follower.test.ts`); `Dial` link-only, undecodable-body, and link-beats-body against an `httptest` hub (`dial_test.go`); `TraySignalingSupersedeTests` ×6; `SessionReachabilityTests` link-only chain, link-beats-body, and bound-still-enforced.
- [ ] Manual smoke — not run. Exercising it live needs a leader reconnect against a deployed hub; every client is covered at its own seam instead.

**Pre-existing failures** (unrelated, all Chrome/timing-dependent): `playwright-command.test.ts > iframe integration` — the documented `playwright-iframe` flake. `ipk/esm-real-path-e2e` and `ipk/ipx-run-installed-e2e` timed out at 5s in one full-coverage run and pass in isolation with coverage on.

## Documentation

- [x] Relevant `CLAUDE.md` — `shared-ts` (the new helper), `swift-trayfollower` (`SupersedeLink` + the twinning), `swift-traysession` (probe reads the link first)
- [x] `docs/` — `ios-app-details.md` (§ iCloud tray supersede chain: the rule, the bounds, why the code doesn't get a vote), `swift-launcher-details.md` (§ icloud-sync)
- [ ] No `README.md` / runtime-skill change — no user-facing surface moves

## Risk & rollback

Blast radius is the supersede path of five followers; a regression looks like a follower failing to move to a replacement tray after a leader reconnect, which is the state they were in before #1956. No flag, no persisted-state change; revert is clean. The one thing CI cannot answer is a real leader reconnect against the deployed hub — worth a live check before the step-2 status flip, when the body stops being the primary channel.

## Checklist

- [x] No secrets, credentials, or tokens in the diff — and no join URL reaches a log
- [x] Public API / contract changes are intentional and reflected in docs
- [x] Contract used by other floats: this PR *is* the cross-float alignment — all five followers move together, and the pinned vector tables keep the four parsers from drifting


trieloff:
Fixed in 3d53761 — the finding was right, and it generalized to a second site.

**Electron follower** (`electron-tray-follower.ts`): the replacement is now resolved before `result.bootstrap` is accepted, so a reply carrying both follows the link instead of bootstrapping against the abandoned tray. This was the one follower out of five where the body outranked the link — the browser normalizes any successor-bearing reply onto the supersede plan before `tray-webrtc` looks at the bootstrap, Go checks `plan.JoinURL` before it switches on the action, and both Swift attach loops hoist the check above their switch.

**Native liveness probe** (`SessionReachability.swift`): same shape, found by auditing the other four for the same ordering. A terminal `200` short-circuited ahead of the link, so a moved tray whose old address still answered `leader.connected: true` would have been reported live rather than chased. The link now outranks the 200. The body-only path is untouched — `supersededURL` stays gated on the 409 that shipped, so a hub predating the header behaves exactly as before.

Both regression tests were confirmed to fail against the unfixed source (`follows the link even when the same body also offers a bootstrap`, `testLinkHeaderOutranksATerminal200`).

PR #2531, opened 2026-08-27
feat(widgets): Cones & Scoops home-screen widget for iOS and macOS

trieloff:
## Summary

Adds a **Cones & Scoops** home-screen widget to both native apps. It shows what the agents in the currently connected SLICC instance are doing: the work units that most want your attention, each drawn as its own agent avatar, with the last turn in the session on the large tile.

Three families, three layouts — small is a 2x2 of four peers (*who needs me?*), medium is one face large plus a field of the rest (*what is the one thing?*), large is medium's faces plus the last turn in words (*what did it just say?*). iOS adds the lock-screen / StandBy accessories.

## Why

You can see a session's state on every other surface — the tab strip, the transcript, the iOS follower — but not without unlocking and opening something. A widget is the one place a glance is enough, and the avatar already encodes everything a glance needs.

Multiple cones are coming (the flag exists), so the layout treats cones and scoops as ranked peers rather than building around a single "the" cone.

**Layering.** Everything drawn lives in a new `packages/swift-widgetkit` (`SliccWidgetKit`, macOS 14 + iOS 18, Foundation + SwiftUI + WidgetKit only — no WebRTC, no AppKit/UIKit). Each extension is the four lines WidgetKit insists on owning (`@main`, a `WidgetBundle`, a `Widget` with an `init()`), because `Widget` requires an `init()` and there is nowhere to hand it a host.

**A widget is not a follower.** It runs in its own short-lived process, cannot dial a leader, cannot hold a data channel, and may run while the app is dead. So the contract is one-way: the host app captures a `WidgetSnapshot` into the shared app group; the widget renders it and says how old it is. A JSON file, not `UserDefaults` — the snapshot is a document, group-container `UserDefaults` is a coalescing cache with no ordering guarantee against a process that is not running, and an atomic replace gives the widget the old snapshot or the new one and never half of both.

**The face is the status channel.** There is no status word, no model id and no context bar on any system family. `UnitAvatarFace` is a static port of the app's expression grammar: tool work squares the eyes (`AvatarExpression.shapeTarget`), thinking grows brows and looks up and away, an ended turn holds eye contact under the soft arrival lid, idle wanders low and dims, broken is X'd out, booting has no eyes, no-instance is frozen TV static. Each pose is the target the app's own reduced-motion `settle` eases toward, so a widget avatar is the app's avatar caught mid-thought rather than a second design that resembles it.

**Rejected alternative — reuse `SliccAgentAvatarGeometry` directly.** The app's version drags in the expression engine, CoreMotion tilt and the noise field; a widget process has no sensors, no frame callback and a hard memory budget, and `TimelineView(.animation)` does not animate there. So the geometry is a hand-carried static port, and parity is a manual contract pinned by `UnitAvatarGeometryTests` against the exact values `SliccAgentAvatarGeometryTests` already asserts (scoop eyes at x = 8.2625 / 91.7375, cone at 2.75 / 97.25 on a 100pt tile). Unifying them means lifting the geometry out of the follower target, which has a coverage-floor consequence for `ios-app` and deserves its own decision.

**Rate limiting is shared.** `scoops.list` arrives on every turn boundary and every tool bracket; a reload per arrival spends WidgetKit's daily refresh budget by mid-morning and then leaves the tile frozen at the moment that mattered. `WidgetSnapshotPublisher` writes once immediately, then at most once per 15 s, with a trailing write so the last state in a burst is never the one dropped. Four transitions bypass the limit (`isUrgent`): something broke, something handed the turn back, the unit list changed shape, the link died.

**macOS needed genuinely new plumbing.** Sliccstart holds no cone/scoop state — SLICC is browser-first and the local server is a stateless relay, so `/api` cannot answer either. `WidgetTrayObserver` runs a small read-only tray follower off the `leaderJoinUrl` the launcher already knows.

It is **gated on the widget actually being installed** (`WidgetInstallationQuery` / `WidgetCenter.getCurrentConfigurations`). A launcher quietly holding a WebRTC participant slot open forever — visible in the tray's participant count, burning battery — to feed a tile nobody added would be a bad citizen in someone else's session. **This is the design decision most worth a second opinion.**

## Cross-cutting impact

- **Floats exercised**: iOS follower (`SliccFollower`) and Sliccstart. No webapp, extension, worker or CLI code changes — zero TypeScript in the feature itself.
- **New tray participant (macOS only)**: `WidgetTrayObserver` attaches as a follower with runtime `sliccstart-widget`, so a leader with a Sliccstart-installed widget sees one more participant. Gated on widget installation; sends `hello`, `pong` and `request_snapshot` and nothing else — never a user message, a scoop selection or a steer.
- **Protocol**: read-only. No new wire variants, no change to `tray-sync-protocol.ts`, no follower/leader parity work.
- **Persisted state**: one new file, `widget-snapshot.json`, in each host's existing app group. Removed on detach (`clearTrayCredentials` / leader loss). No IndexedDB, no localStorage, no migration.
- **Security**: the instance label is the session display name or the join URL's **host** — never the join URL, which is a secret (covered by a test on both hosts). The last turn is `.privacySensitive()`, so iOS redacts it on a locked device. The widget's only entitlement is App Groups; it holds no keychain access and dials nothing.
- **Release gating**: `swift-widgetkit` compiles into both the `.app` and the `.ipa`, so it had to join `MACOS_PATH_PREFIXES` / `IOS_PATH_PREFIXES` in `release-native.mjs` or a widget-only fix would ship silently only when an unrelated file next changed. Third commit also closes the same pre-existing gap for `swift-traykit` (both File Provider appexes) and `swift-trayfollower` on the macOS side.
- **Bundle size**: N/A — no web bundle touched.

## Test plan

- [x] `npm run lint` — clean; duplication flat at 5.87% against a 7.5% ceiling
- [x] `npm run typecheck`
- [x] `npm run test` — 17,310 passing
- [x] `npm run build` (assembles `Sliccstart.app` with both appexes staged)
- [x] `npm run build -w @slicc/chrome-extension` (via the root build)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — OK, 0 on any debt list
- [x] Swift coverage gates: `swift-widgetkit` 97.1% lines (new floors 94/90/91), `ios-app` 74.0% (floor 73), `swift-launcher` 44.3% (floor 43), all above floor
- [x] iOS Simulator build; verified `SliccFollower.app/PlugIns/` contains `SliccWidgets.appex` (`com.sliccy.follower.widgets`, `com.apple.widgetkit-extension`) alongside the File Provider and Share appexes
- [x] macOS: `Sliccstart.app/Contents/PlugIns/` contains `SliccstartWidgets.appex`
- [x] New behavior covered: 103 tests in `SliccWidgetKitTests` (model, store, ranking, layout caps, avatar geometry parity, every face, markdown flattening, rate limiting), plus `AppStateWidgetSnapshotTests` on iOS and `WidgetTrayObserverTests` on macOS
- [x] Every family x fixture x scheme rendered via `swift run slicc-widget-gallery` and by `WidgetRenderSmokeTests` in CI — a widget view that traps is otherwise a blank grey tile with no crash report anyone will see
- [x] **Not yet done: seen on a real home screen.** The design has only ever been rendered by the gallery tool. Adding a widget to a home screen is a manual step. **Please do not merge before that sign-off.**

**Pre-existing failures**: one flake in `packages/webapp/tests/shell/ipk/esm-real-path-e2e.test.ts` (VAL-ESM-016 installs `nanoid` from npm) under full-suite load; passes in isolation and is unrelated — no TypeScript in this PR.

## Documentation

- [x] `README.md` — user-facing bullet
- [x] `CLAUDE.md` (root module map + guide links), `packages/swift-widgetkit/CLAUDE.md` (new), `packages/ios-app/CLAUDE.md`, `packages/swift-launcher/CLAUDE.md`
- [x] `docs/widgets.md` (new) + `docs/root-details.md`

## Risk & rollback

- **Blast radius**: additive. Both apps behave identically with no widget installed; the extensions are separate processes. Reverting the PR removes the targets cleanly — there is no persisted state to unwind beyond one file in an app group.
- **The one behavioral change to an existing app** is the macOS tray connection, and it cannot start unless a widget is installed. Kill switch: remove the widget, or revert `WidgetTrayObserver`'s call site in `SliccstartApp`.
- **Ships only after Developer Portal work, already done**: App ID `com.sliccy.follower.widgets` with App Groups on the existing `group.ai.sliccy.follower`, and an App Store profile named `Slicc Follower Widgets App Store`. `APPLE_WIDGETS_PROVISIONING_PROFILE_BASE64` / `_NAME` / `APPLE_WIDGETS_BUNDLE_ID` are set. All four iOS profiles were refreshed from freshly-downloaded copies (they share one expiry instant, so the portal had regenerated them together and the stored ones may have been superseded). Without the secret the packaging script soft-skips the whole iOS upload rather than shipping a build without the widget.
- **CI cannot catch**: how the faces read at 34pt on a real device, whether the brows survive the tile crop on-glass, and whether the macOS observer's participant is acceptable in a live tray. All three want eyes.

## Checklist

- [x] No secrets, credentials or tokens in the diff
- [x] Public API changes are intentional and documented
- [x] Contract used by other floats: read-only tray consumption, no protocol change


trieloff:
**Screenshots**: I could not attach them — `gh image` (from `ai-ecoverse/ai-aligned-gh`) is not installed on this machine, and the plain `gh` has no upload path.

Regenerate the full set locally in a couple of seconds:

```bash
cd packages/swift-widgetkit
swift run slicc-widget-gallery ./out
```

That writes every family x fixture x colour scheme at 4x, plus one contact sheet per family. `sheet-large-dark.png`, `sheet-medium-dark.png` and `sheet-small-dark.png` are the three worth opening first; `large-busy-dark.png` shows the last-turn block best.

It renders the six fixtures in `WidgetSnapshotFixtures.swift`, which are the same ones WidgetKit shows in the widget gallery — so the states nobody remembers to check (disconnected, crowded, no-cone, nothing-said-yet) are all in there.

**Please hold this one for a look on a real device before merging.** The design has never been seen anywhere but that renderer, and how the faces read at 34pt on-glass is the whole bet.

PR #2562, opened 2026-08-28
test(swift-launcher): cover the SwiftUI surfaces, 44% → 68% lines

trieloff:
## What

`swift-launcher` was the least-tested module in the repo by a wide margin — floors `43/46/53`, measured **44.45% lines**. Three quarters of its unreached lines were SwiftUI (`AppListView` at 17%, `SettingsView` at 15%, `SliccstartApp` at 0%), which nothing in the suite had tried to reach.

| | before | after |
|---|---|---|
| **TOTAL lines** | 44.45% | **68.32%** |
| **TOTAL functions** | 47.35% | **62.51%** |
| **TOTAL regions** | 54.08% | **64.76%** |
| `Views/AppListView.swift` | 17.35% | **78.11%** |
| `Views/SettingsView.swift` | 14.99% | **80.66%** |
| `Models/WidgetTrayObserver.swift` | 60.84% | **83.50%** |
| `Views/SetupProgressView.swift` | 0% | **96.97%** |

Floors ratcheted `43/46/53` → `67/62/64` (the ratchet's own `floor(measured − 0.5)`). 449 tests, all passing; suite still runs in ~5s.

## How

`SliccstartTests/ViewHosting.swift` renders a view off-screen with `ImageRenderer` and returns a digest of the resulting bitmap, so a test can assert that two states of a view **render differently** — real coverage of the window's conditional bodies, and an assertion with teeth rather than a "does it build" smoke test. Comparing renders (instead of pinning golden images) keeps that stable across machines and OS versions.

That covers what the launcher's windows actually do: which sections a given scan produces, whether a row is gated on a leader, what each of the six update-check statuses puts in the footer, how an unreachable iCloud session is drawn, the mount table's empty state, the terminals tab's live command preview, the secrets tab's locked/unlocked split and its editor's validation states.

Also new: `WidgetTrayObserverWireTests` covers the observer's `TrayFollowerConnectorDelegate` half — channel open/close/give-up, the read-only `hello`, ping→pong, chunked-snapshot reassembly, and the 30s transcript-fetch throttle (plus the active-scoop change that legitimately beats it).

## Testability seams

All defaulted, so no call site changes:

- **`AppListView(isBundledBuild:)`** — the update footer branched on `SliccBootstrapper.isBundled` *inside* `body`. A test process is never a packaged `.app`, so the entire full-update flow (the six check/retry states, restart-to-update) could not be rendered at all, let alone asserted on. Now injected, defaulting to the same global read.
- **`MountsSettingsView(rows:)`** — rows are otherwise only loaded in `onAppear`, which never fires off-screen, so only the empty table was reachable.
- **`SecretsSettingsView(secrets:unlocked:selection:)`** — the unlocked state is otherwise only reachable through a Keychain read behind a user click. Seeded secrets never touch the Keychain.
- **`SecretEditorSheet` / `DomainEntry` are internal rather than private**, matching the precedent `SecretNameValidator` already set in that same file ("Lives at file scope … so unit tests can reach it").

## Two headless limits — documented, not papered over

Both are recorded in `packages/swift-launcher/CLAUDE.md` so the next person doesn't re-derive them:

1. **No interaction.** Headless SwiftUI on macOS builds no AppKit control tree and no accessibility tree (both are lazy; the latter only materializes for an attached assistive client), so `NSHostingView` yields a bare `_FocusRingView` and `accessibilityChildren()` is empty — buttons cannot be pressed. Button *actions* therefore stay tested through the plain types they delegate to (`BrowserLaunchAction`, `TerminalLaunchDecision`, `AppRow.statusDot`), which is the pattern this package already uses. `.borderless` buttons are the exception that does materialize as `NSButton`s, and `TraySessionRow`'s enabled/disabled assertions still use that.
2. **`Table` and `Toggle` draw nothing** under `ImageRenderer` (both are `NSTableView`/`NSSwitch`-backed), so mount/secret table cells and switch knobs are asserted against their model types instead. Two early versions of these tests passed vacuously on exactly this; the assertions were rewritten rather than deleted.

## What is still uncovered, and why

- **`SliccstartApp.swift` (938 lines, 0%)** — an `App`'s `Scene` cannot be rendered, and the window content is built inline in `body` rather than as a `View` type. Covering it needs the content extracted into a `RootView` and its `@State` moved into an observable model: a refactor of the app entry point, not a test change. That is the single biggest remaining item and the gate on this module reaching the top of the table.
- **`SliccProcess.swift` launch paths (~400 lines)** — `launchStandalone` / `launchBrowserFollower` / `launchWithElectronApp` / `spawn` start real browsers. Reaching them needs a process-spawn seam through the launcher's core.
- The button-action closures in both views, per limit (1) above.

## Verification

`npm run verify` (7/7) · `check-touched-exemptions` · `swiftlint` (0 serious) · `swift format lint --strict` · `swift-coverage-check.sh` against the new floors · `npm run build -w @slicc/swift-launcher` (assembles `Sliccstart.app`) · repo `npm run typecheck` + `npm test`.

### DIFF
--- a/coverage-thresholds.json
+++ b/coverage-thresholds.json
@@ -137,9 +137,9 @@
       "regions": 62
     },
     "swift-launcher": {
-      "lines": 67,
-      "functions": 62,
-      "regions": 64
+      "lines": 77,
+      "functions": 68,
+      "regions": 69
     },
     "swift-optel": {
       "lines": 75,
--- a/coverage-thresholds.json
+++ b/coverage-thresholds.json
@@ -137,9 +137,9 @@
       "regions": 62
     },
     "swift-launcher": {
-      "lines": 77,
-      "functions": 68,
-      "regions": 69
+      "lines": 80,
+      "functions": 71,
+      "regions": 72
     },
     "swift-optel": {
       "lines": 75,
--- a/coverage-thresholds.json
+++ b/coverage-thresholds.json
@@ -137,9 +137,9 @@
       "regions": 62
     },
     "swift-launcher": {
-      "lines": 43,
-      "functions": 46,
-      "regions": 53
+      "lines": 67,
+      "functions": 62,
+      "regions": 64
     },
     "swift-optel": {
       "lines": 75,
--- a/packages/swift-launcher/CLAUDE.md
+++ b/packages/swift-launcher/CLAUDE.md
@@ -31,6 +31,33 @@ npm run format -w @slicc/swift-launcher        # swift format --in-place
 
 `SliccstartApp` ticks `refreshRuntimeStates` every 2 s and `runtimeState(for:)` runs on every SwiftUI render, so anything on that path must be O(1) and filesystem-free in the steady state. Electron liveness uses the launch record's `observedAppPID` (`kill(pid, 0)`) first; the `NSWorkspace.runningApplications` scan (`candidateBundlePaths` + `appMatches`, pure string compares) is a fallback only. Do not reintroduce per-app `resolvingSymlinksInPath()` — with ~270 running apps it pinned the launcher at ~40% CPU. Tests: `ElectronAppMatchingTests`.
 
+## Testing the SwiftUI Surfaces
+
+`SliccstartTests/ViewHosting.swift` renders a view off-screen with
+`ImageRenderer` and returns a digest of the bitmap, so a test can assert that
+two states of a view **render differently** (`assertRendersDifferently`) — the
+only way to reach `AppListView` / `SettingsView` body branches from a unit
+test. Two hard limits, both worked around rather than fought:
+
+- **No interaction.** Headless SwiftUI on macOS builds no AppKit control tree
+  and no accessibility tree, so buttons cannot be pressed. Button _actions_ are
+  tested through the plain types they delegate to (`BrowserLaunchAction`,
+  `TerminalLaunchDecision`, `AppRow.statusDot`, …) — keep new view logic in
+  such a type rather than inline in a closure. The exception is `.borderless`
+  buttons, which do materialize as `NSButton`s (`ViewHosting.hostedButtons`,
+  used for `TraySessionRow`).
+- **`Table` and `Toggle` draw nothing** (both are `NSTableView`/`NSSwitch`-
+  backed), so the mounts/secrets table cells and switch knobs are asserted
+  against their model types instead.
+
+Views take their non-injectable state as init seams for this:
+`AppListView(isBundledBuild:)` (the whole update footer is otherwise
+unreachable outside a packaged `.app`), `MountsSettingsView(rows:)`,
+`SecretsSettingsView(secrets:unlocked:selection:)` (never touches the
+Keychain). `SliccstartApp.swift` remains uncovered: an `App`'s `Scene` cannot
+be rendered, and its window content is built inline rather than in a `View`
+type.
+
 ## Operational Telemetry (OpTel)
 
 `SliccstartApp.swift`'s `WindowGroup` root calls `.optelAutoInstrument(appID:)` from `@slicc/swift-optel` once; on macOS that activates `enter`, global `click`, `navigate`, and `error` (uncaught `NSException`). `appID = Bundle.main.bundleIdentifier` (`com.slicc.sliccstart`); beacons land in `helix-225321.helix_rum.cluster`.
--- a/packages/swift-launcher/CLAUDE.md
+++ b/packages/swift-launcher/CLAUDE.md
@@ -54,9 +54,23 @@ Views take their non-injectable state as init seams for this:
 `AppListView(isBundledBuild:)` (the whole update footer is otherwise
 unreachable outside a packaged `.app`), `MountsSettingsView(rows:)`,
 `SecretsSettingsView(secrets:unlocked:selection:)` (never touches the
-Keychain). `SliccstartApp.swift` remains uncovered: an `App`'s `Scene` cannot
-be rendered, and its window content is built inline rather than in a `View`
-type.
+Keychain).
+
+**The window is a model, not a `Scene`.** An `App`'s `Scene` cannot be
+rendered, so anything living inside `WindowGroup` is untestable by
+construction. `SliccstartApp` is therefore only wiring: the window's behavior
+is **`Models/LauncherModel.swift`** (launch decisions, dialogs, debug builds,
+update-check outcomes, the 2 s tick, leader publish/withdraw) and its content
+is **`Views/RootView.swift`**. `SliccstartAppDelegate` is the composition
+root, with every collaborator defaulted-but-injectable so the two quit paths
+(stop-everything vs detach-for-update) are testable. Put new window behavior
+on `LauncherModel`, not in a `WindowGroup` closure. Tests:
+`LauncherModelTests`, `RootViewRenderTests`, `AppDelegateLifecycleTests`.
+
+`SliccProcess`, `SliccBootstrapper` and `AppManagementPermission` are
+deliberately **not `final`** — they are the app's side-effecting collaborators
+(spawning browsers, running git/npm, opening System Settings), and `@testable`
+lets a test subclass stand in for them.
 
 ## Operational Telemetry (OpTel)
 
--- a/packages/swift-launcher/CLAUDE.md
+++ b/packages/swift-launcher/CLAUDE.md
@@ -46,6 +46,11 @@ test. Two hard limits, both worked around rather than fought:
   such a type rather than inline in a closure. The exception is `.borderless`
   buttons, which do materialize as `NSButton`s (`ViewHosting.hostedButtons`,
   used for `TraySessionRow`).
+- **CI runs an older macOS than your Mac** (`macos-latest` is still macOS 15).
+  How much of an AppKit-hosted subtree — `Table`, and anything overlaid on one
+  — an off-screen render produces differs between them, so a render comparison
+  that passes locally is not evidence it passes in CI. Compare only over
+  plain SwiftUI content.
 - **Vary exactly one thing per comparison.** Two states that differ in more
   than one way render differently whether or not the feature under test
   exists — a review found eight such tests here. Pin everything else
--- a/packages/swift-launcher/CLAUDE.md
+++ b/packages/swift-launcher/CLAUDE.md
@@ -46,9 +46,16 @@ test. Two hard limits, both worked around rather than fought:
   such a type rather than inline in a closure. The exception is `.borderless`
   buttons, which do materialize as `NSButton`s (`ViewHosting.hostedButtons`,
   used for `TraySessionRow`).
-- **`Table` and `Toggle` draw nothing** (both are `NSTableView`/`NSSwitch`-
-  backed), so the mounts/secrets table cells and switch knobs are asserted
-  against their model types instead.
+- **Vary exactly one thing per comparison.** Two states that differ in more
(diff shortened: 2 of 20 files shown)


trieloff:
## Update: the `SliccstartApp` refactor is in

The first round stopped at "covering `SliccstartApp.swift` needs the window content extracted — a refactor, not a test change" and left it at 0%. That reasoning was backwards: if tests matter, refactoring toward testability is the work, not an excuse to skip it. So the extraction is now part of this PR.

| | original | after round 1 | **now** |
|---|---|---|---|
| **TOTAL lines** | 44.45% | 68.32% | **77.55%** |
| **TOTAL functions** | 47.35% | 62.51% | **68.95%** |
| **TOTAL regions** | 54.08% | 64.76% | **70.28%** |
| `SliccstartApp.swift` uncovered lines | 938 | 938 | **50** |

Floors ratcheted `43/46/53` → **`77/68/69`**. 401 → **500 tests**, suite still ~5s.

### The split

`SliccstartApp.swift` was 938 lines at 0%, and no test could have moved it: an `App`'s `Scene` is not a `View`, so nothing inside `WindowGroup` renders, and its `@State` cannot be driven from outside the SwiftUI lifecycle. The launch decisions, dialog flows, update-check outcomes and leader publish/withdraw all lived in there. Now:

- **`Models/LauncherModel.swift`** — what the window *does*. Every side-effecting collaborator (app scan, debug-build creator, update check, installation check, startup preference, saved browser order) is injected with a live default, so a test drives the real decisions without launching a browser or patching an `.app`. **92.66% lines.**
- **`Views/RootView.swift`** — what the window *shows*: the setup/debug-build/list switch, the alerts, the timers. Every callback is one line into the model. **72.94% lines.**
- **`SliccstartApp`** — wiring only. `SliccstartAppDelegate` is the composition root, collaborators defaulted-but-injectable.

`SliccProcess`, `SliccBootstrapper` and `AppManagementPermission` are no longer `final`, so a `@testable` subclass can stand in for them — they spawn browsers, run git/npm, and open System Settings respectively.

### What that bought, concretely

51 new tests over behavior that had never been executed once, including:

- **Both quit paths.** A plain quit stops every runtime and withdraws the iCloud session; a quit for an update detaches and *keeps* advertising. That difference is whether "Restart to Update" preserves the user's open browsers or silently closes them, and nothing checked it before.
- **Every Electron row outcome** — already-debugged (no-op), running-without-debug (ask before restarting, and don't kill it before consent), needs-debug-build, needs-permission (opens System Settings), needs-leader (ignored defensively), stopped/failed (launch).
- **Startup**: an installed or `needsBuild` runtime must not re-bootstrap over the user's checkout; a failed bootstrap stays on the setup screen with the reason; only a bundled build checks for updates; auto-launch follows the user's drag order, not the scan order, and a failed auto-launch never blocks startup or throws an alert in their face.
- **Update-check classification** — a read-only-volume failure is reported as "move to Applications", not as a network error, and a check in flight is not restarted.
- **The 2s tick** — it does nothing before the window is ready, and only polls `/api/agent-activity` while an update is actually staged.

### One test bug worth calling out

`RootViewRenderTests` initially compared an empty window against a populated one and passed — because the App Management permission probe settles asynchronously, `RootView` rescans when it does, and the stub scan returned `[]`, wiping the seeded targets mid-render. Both sides were empty. The stub now returns what the model was seeded with, and there is a comment saying why.

### Still open

- **`SliccProcess` launch paths (~460 lines)** — `launchStandalone` / `launchBrowserFollower` / `launchWithElectronApp` / `spawn` start real browsers. Reaching them needs a process-spawn seam, which is the next refactor of this kind.
- The button-action closures in `AppListView` / `SettingsView`, per the headless-interaction limit documented above.

### Verification

`npm run verify` (7/7) · `check-touched-exemptions` · `swiftlint` (0 serious) · `swift format lint --strict` · `swift-coverage-check.sh` at the new floors · `npm run build -w @slicc/swift-launcher` (assembles `Sliccstart.app`) · repo `typecheck` + `npm test`.

**Not verified:** I did not launch the built `.app`. This machine has `launchBrowserAtStartup` enabled, so starting it would have opened a real browser session. Worth a manual smoke test before merge, since this moves the app's entry point.

trieloff:
## Update 2: launch paths covered, and a crash the tests could not have found

Three things since the last update: the launch paths are tested, auto-launch now requires an installed app, and **running the built app caught a crash the refactor had introduced**.

| | original | update 1 | **now** |
|---|---|---|---|
| **TOTAL lines** | 44.45% | 77.55% | **80.57%** |
| **TOTAL functions** | 47.35% | 68.95% | **71.53%** |
| **TOTAL regions** | 54.08% | 70.28% | **72.73%** |
| `Models/SliccProcess.swift` | 63.50% | 63.50% | **81.80%** |
| `Models/StartupPreference.swift` | — | — | **100%** |

Floors `43/46/53` → **`80/71/72`**. 401 → **527 tests**, suite ~5.6s. Among the Swift packages only `swift-widgetkit` (94/90/91) and `swift-traysession` (91/87/85) now sit higher.

## The app crashed on launch, and only launching it found that

```
Fatal error: Use of unimplemented initializer 'init()' for class 'Sliccstart.SliccstartAppDelegate'
```

`@NSApplicationDelegateAdaptor` instantiates the delegate through the **Objective-C** runtime's `-init`. Giving `SliccstartAppDelegate` a designated initializer with all-defaulted parameters (so tests could inject collaborators) made it callable as `init()` from Swift but stopped it vending that selector — the inherited `NSObject.init` became a trap, and every launch died instantly. A clean build and 500 passing tests both sailed past it.

Fixed with `override convenience init()`. The regression test goes through `NSObject.Type.init()`, not `SliccstartAppDelegate()` — the Swift spelling resolves statically and would not have caught it. Reverting the fix makes that test reproduce the exact fatal error, which I checked rather than assumed.

## Auto-launch now requires an installed app

`StartupPreference.shouldAutoLaunch` = `resolveEnabled && isInstalledLocation`. The browser only auto-starts when Sliccstart runs from `/Applications` or `~/Applications`. A startup behavior belongs to the copy the user installed, not to every copy that happens to run: a build out of `packages/swift-launcher/build/`, a `.zip` still in `~/Downloads`, or Gatekeeper's translocated read-only copy would each take over the screen with a browser session.

This was found the practical way — a smoke run of the launcher would have opened a real browser on my machine, which is what made the run feel unsafe in the first place. The preference is untouched (the checkbox keeps its value and takes effect the moment the app is moved), the Startup tab explains the refusal in its caption, and the refusal is logged rather than silent. 10 tests, including that `/ApplicationsOld` does not sneak past the prefix match and that the legacy `autoLaunchAppId` migration still runs for an uninstalled copy.

## Launch paths

`SliccProcess.SpawnServices` injects exactly three OS calls — which binary to run, whether to exec it, whether a port is bound. Everything else stays the shipping code: the real argument vector, a real `Process` (the stub runs `/bin/sleep` in the server's place), the real launch record, the real termination handler. The tests exercise the launch bookkeeping, not a mock of it.

16 tests covering: a standalone browser leads and becomes the leader with the right CDP port, environment and mount arguments; a second click does not start a second server; a busy port is reported rather than fought over; a failed resolve is recorded on the row and cleared by the next attempt; a follower carries `--join`, never counts as this device's leader, and takes its own port pair so it can sit beside one; Electron apps wait for a leader and each get their own ports; `stopAll` clears the leader; and `detachAll` persists what a reattach needs — including that the **second** call does not overwrite the snapshot with an empty one (the update flow calls it from both `onBeginUpdate` and `applicationWillTerminate`).

## Live verification

Ran the built `.app` twice — before the fix (crashed instantly, `~/Library/Logs/DiagnosticReports/Sliccstart-2026-08-28-111145.ips`) and after (boots, stays up, idles at 0% CPU, no crash report). With `launchBrowserAtStartup` enabled on this machine, the build-directory copy started **no** browser and **no** `slicc-server`, and the already-running `/Applications` instance was untouched. The test instance was ended with `SIGKILL` rather than a clean quit on purpose: `applicationWillTerminate` would have withdrawn the real instance's advertised iCloud session and its Finder domain, since both copies share a device id.

Not verified: the window's *appearance* — I did not screenshot it, which would have captured the desktop.

## Still uncovered

`SliccProcess` is at 81.8%; the remainder is reattach/stale-helper reaping and the CDP-probe retry loops. `AppListView` / `SettingsView` button-action closures remain unreachable headlessly (documented in `packages/swift-launcher/CLAUDE.md`).

trieloff:
## Update 3: review findings addressed

Two automated reviewers, both worth their runtime.

### Codex (inline, #2562)

`testAgentActivityIsVisibleInTheFooter` was vacuous: the tint is only read by the restart-to-update button, and the fixture's updater held no staged build, so both sides rendered the check-for-updates footer and the test passed whether or not the tint existed. Correct.

Fixed twice over. The updater can now be seeded into `.downloaded` (decoding a `Release` from JSON, as the release-filter tests already do), which covers the restart-to-update footer for the first time. And the tint decision moved into `AppListView.updateAffordance(hasRecentAgentActivity:)` so it is asserted directly — because seeding alone was **not** enough: the restart button is `.borderless`, i.e. AppKit-backed, so it draws nothing under `ImageRenderer` and no render comparison could ever have seen the tint. Same class of blind spot as `Table` and `Toggle`, now documented alongside them.

### Factory Droid / GPT-5.6-Sol (`thr_6829mph33b`, fresh worktree on this branch)

Tasked to diff the pre-refactor `SliccstartApp.swift` against `LauncherModel` + `RootView` line by line, hunt vacuous assertions, and check the concurrency the refactor loosened.

**No behavior drift found.** It confirmed the guards, call ordering, closure captures and `Task` actor hops were preserved; that `SpawnServices` still builds the real process, environment, handlers and records; no new retain cycle or handler/timer leak; and that the auto-launch gate correctly accepts descendants of both Applications roots while rejecting `/System/Applications`.

**It found eight confounded render comparisons** — each varied two things and claimed the assertion pinned one:

| what the test claimed | what actually differed |
|---|---|
| the debug-build wrench badge | badge **and** the "Debug Build" subtitle |
| the extension section renders | nothing — it only asserted the window was non-empty |
| a session borrows a local browser's icon | icon match **and** an entire Browsers section |
| the status dots | dot **and** the row subtitle |
| the editor's Create/Edit heading | heading **and** every field and validation state |
| "each validation failure" | only the empty and complete drafts were rendered |
| the empty-hostname fallback row | the row **and** the text beside it |
| the Retry button | button **and** the message **and** the spinner |

Every one now varies exactly one thing: `subtitleOverride:` pinned on both rows, the same scan on both sides of the icon match, an empty `Secret` so an edit draft matches a fresh sheet, `[]` vs `[""]` domains (which must render **identically**), one message and one spinner across the Retry comparison.

Where isolation is impossible, the rule is asserted instead of the pixels. `SecretEditorSheet.validationMessage` is now a static over plain values, which finally reaches the **collision** and **invalid-name** branches — both are only reachable by typing, and an off-screen sheet cannot type. The extension section drops its render claim entirely (its `.plain` button produces no identifiable AppKit node) and asserts the section list instead.

**Verified by mutation, not by green:** deleting the wrench badge and deleting the Retry button each turn their test red now. Both passed before this commit.

The one-variable rule is now in `ViewHosting`'s docs and the package `CLAUDE.md`, so the next person writing one of these has the trap written down rather than rediscovering it.

### Also

Both branches rebased onto `main`. The earlier `lint` failure was the stale-branch phantom: `check-touched-exemptions` diffs against `origin/main`, and 7 commits of drift made it attribute three other PRs' debt-list entries to this one.

**532 tests** · 80.72% lines / 71.44% functions / 73.14% regions.

Issue #2312, opened 2026-08-22
feat(ui): read-only scoop view — route all scoop requests to the owning cone


trieloff:
Part of #1666 (multiple cones). Decided 2026-08-22.

## Goal
Users do not talk to scoops. The scoop tab becomes a **read-only transcript**: no composer, no buttons, no approval cards.

## Rules
- Every scoop request that needs a human — `sudo_request`, approvals, "waiting for parent" — is routed to the **owning cone** (`parentJid`) and rendered in the cone's transcript. Scoops never have user-facing actions.
- The follower (`wc-follower.ts`) gets the same treatment through the same descriptor flag; no second code path.
- `feed_scoop` from the cone remains the only way to send a scoop input.

## Tests
- Scoop selected → no `<slicc-composer>` mounted (leader and follower).
- `sudo_request` from a scoop of cone B appears in B's transcript, not A's, not the scoop's.

PR #2564, opened 2026-08-28
feat(webapp): make keyboard mode the resting state and carry it across unit switches

trieloff:
Two dogfooding notes on keyboard mode, both about the mode being a place you had to keep re-entering.

## 1. The mode is now the resting state

Keyboard mode is on whenever nothing is focused for typing — at boot, after any blur, after a command that opens a panel. `settle()` (deferred, coalesced) asks the DOM where the focus is; it does not try to remember. Leaving the composer by <kbd>Esc</kbd> or by clicking anywhere that is not a text field puts you in the mode; <kbd>⏎</kbd>, <kbd>c</kbd>, or a click into the composer takes you back to typing.

That makes <kbd>Esc</kbd> no longer a toggle. The first press leaves the field; a press made *inside* the mode has nothing left to leave, so it is spent on full screen — which is what the second press already did, minus turning the mode off on the way out. Keyboard Lock still means the exit has to be performed rather than merely permitted.

Two guards an always-on mode needs, and did not before:

- **A focused button keeps its own <kbd>⏎</kbd> and <kbd>Space</kbd>** (`isActivationTarget`). With the mode on by default, a tabbed-to button is a normal place for the focus to be, and `composer` swallowing its Enter would make it unpressable from the keyboard.
- **An unfocused document never wears the badge** (`doc.hasFocus()`), so a Cherry iframe inside a host page does not advertise a mode for a keyboard the host owns. A window `focus` listener re-settles it.

## 2. Switching units carries the mode with you

The mode you were in when you left a unit is the mode you land back in — a `ModeIntent`, restored by a `MutationObserver` on the strip's `active` attribute. That attribute is what every float already writes on a selection (`applyThreadContext`, the follower's select handler), so the freezer rail and the cone actions come along for free and no float has to remember to call anything.

Cone → cone is the easy half: leave it typing and the caret is waiting in the new cone's composer, leave it in keyboard mode and the pill is still up.

Cone → scoop → cone is the half worth the code. A scoop is read-only and has no composer at all (#2312), so the DOM drops the caret when the band is hidden and the mode comes on by itself. The intent is recorded **only while `composerAvailable()` says a composer exists**, so that forced keyboard mode is not mistaken for a choice — and the cone on the far side of the detour gets its caret back.

Related: `settle()` runs on a timer rather than inline for exactly this reason. A tab click blurs the composer *before* it moves the selection; settling inline would read that gap as "the user left the composer" and cost them the caret they were about to get back.

## Verification

`verify` · `lint` · `typecheck` (browser + webapp-worker) · `test` (17,704 passing) · `test:coverage:webapp` · `build` · `bundle-size` (first-load OK) · `check-touched-exemptions` — all green. 15 new tests in `wc-shortcuts.test.ts` cover the resting-state settle, the boot settle, the unfocused-document rule, the activation-key guard, and all four unit-switch paths including the read-only scoop detour.

**UI change — awaiting visual sign-off before merge.**


trieloff:
## Live verification (Tier 1, local build on the isolated `:5715` lane)

Driven over CDP with real `Input.dispatchKeyEvent` presses. Production `:5710` / `:9222` untouched (PIDs identical before and after). Console watcher clean — no errors or warnings beyond the pre-existing sandboxed-iframe notice.

| Step | Observed |
| --- | --- |
| Boot, nothing focused | `data-slicc-keyboard-mode` set, badge reads **Keyboard mode · h for help · ⏎ to type** |
| <kbd>⏎</kbd> | caret in `textarea.ta`, mode off, badge gone |
| type `hello` | text lands in the composer, nothing intercepted |
| <kbd>Esc</kbd> | mode on, composer text preserved, focus released |
| <kbd>Esc</kbd> again | **mode stays on**, dimmed `Esc` cap on the HUD (no full screen to exit) |
| <kbd>b</kbd> | left rail opens, mode held |
| switch unit while in keyboard mode | new unit selected, mode still on, focus nowhere |
| switch unit while typing | new unit selected, **caret back in the composer**, mode off |
| cone → scoop (composer band `hidden`) | mode on, badge up, focus released |
| scoop → cone (band back) | **caret returned to the composer**, mode off |

The last two rows are the `#2312` detour: the scoop's forced keyboard mode did not overwrite the composer intent the cone was left in.

trieloff:
Both Codex P2s were right, and both are the same shape of mistake: a resting-state mode left claiming something that was not true. Fixed in `22c1bb1`.

**Resettle after a command that took no focus.** `holdsMode: false` is a *prediction* that the surface is about to take the focus, not a fact — `dock.selectItem()` only emits a selection, and the files and memory panels focus nothing at all. So `f` and `m` dropped the mode with nothing left to bring it back: badge gone, letters dead, focus on nothing. `settler.schedule()` now runs after the command, so the prediction gets checked. Two tests: the surface that takes no focus (mode returns) and `c`, which really does (mode stays off).

**Window blur.** Also right, and the reasoning is exactly why it was invisible: with the focus already on nothing there is no `focusout` to notice the window leaving. `blur` now clears the mode outright. The intent is deliberately untouched — a suspension, not a decision — so the existing `focus` listener settles it back on return, which the test asserts round-trips.

Worth noting the first one is a gap my live CDP pass had too: I exercised `b` (which holds the mode) but never `f`, so the harness would have shown it. Both paths are unit-covered now.

PR #2621, opened 2026-08-29
test: fix local test-suite flakiness (5s timeout budget + a keep-alive port race)

trieloff:
## Summary

`npm run test` failed roughly once per run on a loaded 16-core laptop, always with `Test timed out in 5000ms` and always in a **different file**, while every failure passed in isolation. That moving target is timeout-budget exhaustion, not a single broken test: the victim is whichever already-slow test met peak contention.

Four independent causes, each measured rather than assumed. Nothing is skipped in CI, no `testTimeout` is raised, and production behaviour is unchanged.

## Why

The leading hypothesis was that `packages/webapp/tests/shell/ipk/*` hit the live npm registry. **It does not.** `createProxiedFetch` is `vi.mock`'d and the "nanoid"/"uuid"/"escape-string-regexp" packages are `.tgz` fixtures synthesized in-memory. There is no network dependency to gate and no real-path coverage at risk.

The actual causes:

1. **Heavy `await import()` inside a test body.** Vitest charges module transform + evaluation to whichever test triggers it, so the *first* test in a file paid for the whole graph and the rest cost nothing. Instrumented directly in `esm-schemes-jsh-e2e.test.ts`: first `newShell()` call **1347ms**, every later call **0ms**. Static imports move that to the collection phase, which no per-test timeout covers. This was already the convention next door in `discoverability.test.ts` and the `installer*` suites.

2. **Real production backoff in tests that assert something else.** `CDP_BRIDGE_CONNECT_RETRY_DELAYS_MS` is `[100, 200, 400, 800, 1600]` = **exactly 3100ms**, matching four `setup-standalone-prelude` tests measured at ~3130ms. They assert `hasLocalCdpSurface`, never retry timing — which `connectWithBoundedRetry`'s own suite already covers by injecting `delays`. Same story for the tray hibernation test, whose `webhookDeliveryWaitMs` seam has a docstring literally warning that a fake leader "would otherwise sit out the full production budget".

3. **Wall-clock assertions that only hold on an idle machine.** `expect(elapsed).toBeLessThan(50|100|150)` — observed failing at 117ms against a 100ms bound.

4. **A genuine shared-resource race** (this was asked about explicitly). Not a shared temp dir or a fixed port, but ephemeral-port recycling plus HTTP keep-alive pooling. `app.listen(0)` takes an ephemeral port; `server.close()` stops accepting but leaves established keep-alive sockets alive, pooled by the fetch client per `127.0.0.1:<port>`. The OS then recycles that port onto the next test's server and a pooled socket points at the wrong one. The tell is `hosted-bootstrap`:

   ```
   HTTPParserError: Response does not match the HTTP/1.1 protocol
     (Expected HTTP/, RTSP/ or ICE/)
   data: {"ok":false,"error":"Invalid JSON"}
   ```

   The client read a JSON body where a status line belongs, and that body belongs to no endpoint in the failing test. Three symptoms, one cause: `cloud-status` `expected 404 to be 200`, `hosted-bootstrap` crossed response, `fetch-proxy-dav` 5s timeout. (`requireLoopback` was ruled out as the 404's source — it answers 403.)

### Why CI never caught this

CI runs on `ubuntu-latest` (4 cores) versus this machine's 16 — far less parallelism, not a longer timeout. There is no `testTimeout` override anywhere; the default 5000ms applies in both.

### On the `iframe integration` failures

Worth correcting: these are **not** 5000ms test timeouts. That block has carried a 90s suite timeout since June (`74095fa4c`), and suite-level timeouts do cascade in Vitest 4.1.11 (verified with a scratch probe). The real signature, captured twice under load, is `Chrome exited with code null before reporting CDP port` — the OS SIGKILLing Chrome mid-launch. That fails the *suite*, so its 6 tests report as **skipped**, which reads a lot like "4 tests timed out".

- **`sleep` seam on `StandalonePreludeDeps` is optional and unused in production.** Neither `wc-live.ts` nor `wc-follower.ts` passes it, so the real timer still applies. It is deliberately scoped to the standalone/bridge branch and *not* threaded into `createExtensionLeaderBrowser`: that call sits inside `setupStandalonePrelude`, which is already at the 150-line biome cap, and that branch is not a backoff hotspot. The JSDoc says so explicitly.
- **`wc-follower.test.ts` imports were NOT hoisted.** It calls `vi.resetModules()` in every case, so the dynamic imports are load-bearing. But `resetModules` drops evaluated modules, never vitest's transform cache — so a `beforeAll` warm-up moves the one-time transform (~2.9s idle, >10s loaded) out of a per-test budget without changing isolation.
- **Real-Chrome gating keeps CI coverage.** `iframe integration` is now gated on `CI || SLICC_TEST_CHROME_INTEGRATION=1`. CI remains a real gate: `ubuntu-latest` ships Chrome at `/usr/bin/google-chrome`, and the `test` job already sets `SLICC_CDP_LAUNCH_TIMEOUT_MS=60000` specifically for this suite's cold start. Verified three ways — opt-in **301 tests**, `CI=1` **301 tests**, default **295 + 6 skipped**. The webapp coverage floor still passes with those 6 skipped.
- **Widened bounds keep their teeth.** Every one guards a regression worth >= 1000ms (awaiting a 1000/2000ms budget, a multi-second pid scan, a kill that never settles), so 500ms still catches it.
- `closeAllConnections()` is already the remedy used by `chrome-launch.test.ts` in the same package.

## Cross-cutting impact

- **Floats exercised**: none behaviourally. The only `src/` change is one optional field on `StandalonePreludeDeps` plus passing it through; both production callers omit it.
- **Coverage**: webapp, node-server and cloudflare-worker floors all pass, including with the 6 iframe tests skipped locally. No floor was touched.
- **Local vs CI divergence**: `npm run test` on a laptop no longer spawns a browser. This is the one deliberate behaviour change for developers, documented in the skill with the opt-in command.

## Test plan

- [x] `npm run verify` — 7/7
- [x] `npm run lint:ci`
- [x] `npm run test` — 17962 passing, 0 failing
- [x] `npm run test:coverage:webapp` / `:node-server` / `:cloudflare-worker`
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — 33 files, 0 on any debt list

### Flake verification (the actual point of this PR)

A single green run proves nothing here, so this was measured as an A/B under identical synthetic load (24 CPU hogs on 16 cores, load average ~950), fix reverted vs applied via the same patch:

| arm | runs with a failure |
| --- | --- |
| before (fix reverted) | **7 / 8** |
| after round 1-2 | 4 / 10 |
| after round 3 | 2 / 10 |
| **after round 4 (this PR)** | **0 / 12** |
| unloaded, default concurrency | **0 / 5** |

**17/17 green.** Each intermediate arm is reported because each one surfaced the next-weakest link — the 4/10 arm is how `host-lick-ws-gate.test.ts` and the wall-clock class were found at all.

### Coverage was not silently lost

Every change was mutation-tested — the guarded behaviour was deliberately broken to confirm the tests still fail:

| mutation | result |
| --- | --- |
| ipk installs land in a bogus dir | **70 tests / 13 files fail** |
| `hasLocalCdpSurface` forced `true` | 2 prelude tests fail (in 19ms / 15ms, vs 3.13s before) |
| hibernation recovery returns no sockets | 2 cloudflare tests fail |
| `waitForScoops` regressed into a real wait | fails with `expected 1004 to be less than 500` |

**Pre-existing / unrelated:** `zenfs-dom-truncate-upstream.test.ts` failed after rebasing until `npm install` picked up main's `@zenfs/dom` 1.2.11 bump (`1b8732f2a`, `2ed19fc7d`). That guard test was correctly detecting a stale `node_modules`, not a regression here.

## Documentation

- [x] `.agents/skills/writing-slicc-tests/SKILL.md` — new **"Keep Tests Inside the Timeout Budget"** section covering all three test-authoring patterns, with the measured numbers and the Chrome opt-in command.

## Risk & rollback

- Blast radius is test-only apart from one optional, unused-in-production dependency field. Reverts cleanly; no persisted state, no migration.
- **Known, not fixed:** `makeRequest` in `cloud-status.test.ts` reads `server.address()` without awaiting the `listening` event. That is a latent bug in the helper independent of the race fixed here; I did not conflate the two.
- Worth a reviewer's judgement: making `iframe integration` opt-in locally is a real (if deliberate) reduction in default local coverage. The argument for it is that it was never a uniform gate — it silently skipped on any machine without Chrome — and its failure mode is the OS killing a process, which no amount of test hardening can prevent.

## Checklist

- [x] No public API / tool / shell-command / lick-channel changes
- [x] No cross-float contract changes

### DIFF
--- a/.agents/skills/writing-slicc-tests/SKILL.md
+++ b/.agents/skills/writing-slicc-tests/SKILL.md
@@ -451,6 +451,48 @@ Rules for changing this:
 - Never enable retries locally. `CI_RETRIES` in `vitest.config.ts` resolves to `0` without
   `CI`, so `npm run test` on a laptop reports the first failure.
 
+## Keep Tests Inside the Timeout Budget
+
+`testTimeout` is vitest's default **5000ms**; the root config never raises it. A test that
+deterministically spends 2-3s of that budget passes on an idle laptop and fails on a loaded
+one, in a _different file each run_. That moving target is the signature of budget exhaustion,
+not of a shared-resource race — the victim is whichever slow test met peak contention.
+
+Two patterns burn the budget without asserting anything.
+
+**1. Heavy `await import()` inside a test body.** Vitest charges module transform and
+evaluation to whichever test triggers it, so the _first_ test in a file pays for the whole
+graph and the rest cost nothing. In `packages/webapp/tests/shell/ipk/`, `newShell()` imported
+`VirtualFS` and `AlmostBashShellHeadless` lazily: the first call cost **1347ms**, every later
+one 0ms. Hoist them to static top-level imports — `vi.mock()` factories are hoisted above
+imports so mocks still apply, and the cost moves to the collection phase, which no per-test
+timeout covers.
+
+```ts
+// Wrong — the first test in the file absorbs the entire module graph.
+async function newShell() {
+  const { VirtualFS } = await import('../../../src/fs/index.js');
+}
+
+// Right — collection pays for it; no test does.
+import { VirtualFS } from '../../../src/fs/index.js';
+```
+
+Keep the dynamic form only where the test needs a genuinely fresh module instance, i.e. it
+calls `vi.resetModules()` (see `tests/ui/wc/wc-follower.test.ts`). Give those an explicit
+per-test timeout instead.
+
+**2. Real production backoff in a test that is not testing backoff.** Retry schedules are
+injectable for exactly this reason: `setupStandalonePrelude` takes `sleep`, and
+`SessionTrayDurableObject` takes `webhookDeliveryWaitMs`. Omit them and the test sits out the
+full production budget — 3.1s (`CDP_BRIDGE_CONNECT_RETRY_DELAYS_MS`) and 3s
+(`WEBHOOK_DELIVERY_WAIT_MS`) — inside a 5s timeout. Inject the fast value, and leave the
+timing assertion to the dedicated unit test (`connectWithBoundedRetry`, which takes `delays`).
+
+Never answer either pattern by raising `testTimeout`: that makes the failure rarer and harder
+to read without removing the wasted seconds. Audit with the timing JSON below — nothing in the
+default suite should sit near a second without an explicit, justified per-test timeout.
+
 ## Read Test Timing
 
 When `CI` is set, the root `test.reporters` adds vitest's `json` reporter and writes
--- a/.agents/skills/writing-slicc-tests/SKILL.md
+++ b/.agents/skills/writing-slicc-tests/SKILL.md
@@ -489,6 +489,19 @@ full production budget — 3.1s (`CDP_BRIDGE_CONNECT_RETRY_DELAYS_MS`) and 3s
 (`WEBHOOK_DELIVERY_WAIT_MS`) — inside a 5s timeout. Inject the fast value, and leave the
 timing assertion to the dedicated unit test (`connectWithBoundedRetry`, which takes `delays`).
 
+**3. Spawning a real browser from the default suite.** `playwright-command.test.ts`'s
+`iframe integration` block launches real headless Chrome. Under memory/CPU pressure the OS
+SIGKILLs it mid-launch and the whole suite fails with `Chrome exited with code null before
+reporting CDP port` — a machine-state failure that says nothing about the code under test. It
+is gated on `CI || SLICC_TEST_CHROME_INTEGRATION=1`, so CI (ubuntu-latest ships Chrome and
+sets `SLICC_CDP_LAUNCH_TIMEOUT_MS`) keeps it as a real gate while `npm run test` on a laptop
+does not spawn browsers. Run it locally with:
+
+```bash
+SLICC_TEST_CHROME_INTEGRATION=1 npx vitest run \
+  packages/webapp/tests/shell/supplemental-commands/playwright-command.test.ts
+```
+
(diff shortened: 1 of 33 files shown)


trieloff:
Thanks — addressed the one actionable observation in `e60af57a4`.

You're right that `sleep: instantSleep` is inert on the extension-leader branch: `createExtensionLeaderBrowser` builds its own browser and calls `connectWithBoundedRetry(browser, undefined, log)` without the seam. It's harmless (the stubbed `connect` succeeds first try, so no backoff fires), but it read as though the seam applied everywhere.

Rather than drop it from the three call sites, I documented it at the `instantSleep` helper — passing it uniformly keeps the deps literal identical across suites instead of making the reader work out which branch each case takes. The seam deliberately isn't threaded into `createExtensionLeaderBrowser` because `setupStandalonePrelude` is already at the 150-line biome cap and that branch isn't a backoff hotspot; the `StandalonePreludeDeps` JSDoc says so explicitly.

Also rebased onto current `main` (38 commits) and re-ran the full pass on the rebased result: `verify` 7/7, `typecheck`, `npm run test` **18014 passing / 0 failing**, `test:coverage:webapp` + `:node-server`, and both builds. Note that a `npm install` was required after the rebase — `zenfs-dom-truncate-upstream.test.ts` correctly fails on a stale `node_modules` after main's `@zenfs/dom` 1.2.11 bump.

PR #2689, opened 2026-08-31
feat(webapp): serialize FormData bodies as multipart/form-data in jsh fetch

### LINKED ISSUE #2687: jsh: fetch rejects FormData bodies, making multipart uploads impossible
## Summary

`FormData` is exposed as a constructible global in the `.jsh` runtime, but passing one as a `fetch` body throws. This makes `multipart/form-data` uploads impossible via the documented API surface, even though every primitive needed to serialize one is already present in the adapter.

Since `Blob` bodies are already supported through the same latin1 path, this looks straightforwardly fixable rather than a fundamental transport limitation.

## Reproduction

`repro.jsh`:

```javascript
console.log("typeof FormData :", typeof FormData);
console.log("typeof Blob     :", typeof Blob);
console.log("typeof File     :", typeof File);

const fd = new FormData();
fd.append("purpose", "assistants");
fd.append("file", new Blob(["hello"], { type: "text/plain" }), "hello.txt");
console.log("FormData constructed OK, entries:", [...fd.keys()].join(","));

try {
  await fetch("https://httpbin.org/post", { method: "POST", body: fd });
  console.log("UNEXPECTED: fetch accepted FormData");
} catch (e) {
  console.log("THROWS:", e.constructor.name, "->", e.message);
}
```

Output (SLICC 6.110.10):

```
typeof FormData : function
typeof Blob     : function
typeof File     : function
FormData constructed OK, entries: purpose,file
THROWS: Error -> node fetch shim: FormData request bodies are not supported (post raw application/x-www-form-urlencoded with URLSearchParams instead)
```

## Why the current error message is misleading

The message suggests `URLSearchParams` as the alternative, but that produces `application/x-www-form-urlencoded`, which is **not** an acceptable substitute for `multipart/form-data`. Any API requiring a file part (`file=@path`) will reject it. For that class of endpoint there is no documented workaround at all — the suggested one cannot work.

## Where it throws

Two sites share the identical message:

- `packages/webapp/src/shell/supplemental-commands/node-fetch-adapter.ts:236` (in `encodeBody`)
- `packages/webapp/src/kernel/realm/realm-browser-bridge.ts:79`

## Why this appears fixable

In `encodeBody`, the adapter already coerces `BodyInit` shapes into the proxy's latin1 string convention (one char per byte), and `Blob` is handled two lines above the `FormData` throw:

```typescript
if (typeof Blob !== 'undefined' && body instanceof Blob) {
  return bytesToLatin1(new Uint8Array(await body.arrayBuffer()));
}
if (typeof FormData !== 'undefined' && body instanceof FormData) {
  throw new Error('node fetch shim: FormData request bodies are not supported (…)');
}
```

So the transport can already carry arbitrary binary payloads. What is missing is only the multipart *serialization* step. A `FormData` can be encoded entirely in-process:

1. Generate a boundary token.
2. For each entry, emit `--boundary`, a `Content-Disposition: form-data; name="…"` header (plus `filename="…"` and `Content-Type` for `File`/`Blob` parts), CRLFCRLF, then the part bytes.
3. Terminate with `--boundary--`.
4. Return the concatenated bytes through the existing `bytesToLatin1`.

`getDefaultContentType` would then need to return `multipart/form-data; boundary=<token>` for `FormData` bodies, and `isRawBinaryBody` should include `FormData` so the raw-binary path is taken. The boundary must be generated in the same place as the body so the two cannot disagree.

Note the entries can include `Blob`/`File` values, whose bytes are only reachable via `await .arrayBuffer()` — `encodeBody` is already `async`, so that is not an obstacle.

## Current workaround (and why it argues for the fix)

Building the multipart body by hand as a `Buffer` with an explicit boundary, then POSTing with bare `fetch`, works correctly against a real API:

```javascript
const boundary = `----jshBoundary${Date.now().toString(16)}${Math.random().toString(16).slice(2)}`;
const CRLF = '\r\n';
const filePart = Buffer.concat([
  Buffer.from(`--${boundary}${CRLF}Content-Disposition: form-data; name="file"; filename="${name}"${CRLF}Content-Type: application/octet-stream${CRLF}${CRLF}`, 'utf8'),
  fileBytes,
  Buffer.from(CRLF, 'utf8'),
]);
// … purpose part, then `--${boundary}--${CRLF}`
await fetch(url, {
  method: 'POST',
  headers: { 'Content-Type': `multipart/form-data; boundary=${boundary}` },
  body: Buffer.concat([filePart, purposePart, tail]),
});
```

This is ~25 lines of boilerplate that every caller needing a file upload has to reimplement, and getting the CRLF placement or the trailing `--` wrong fails in ways that are awkward to debug against a remote API. It also demonstrates the wire format is fully supported by the proxy — only the ergonomic layer is absent.

## Suggested resolution

Either:

1. **Serialize `FormData` in `encodeBody`** as sketched above (preferred — it makes the constructible global honest), or
2. If that is out of scope, **stop exposing `FormData`/`File` as globals** so the failure is a `ReferenceError` at construction rather than at `fetch` time, and **correct the error message** to point at the manual-`Buffer` approach rather than `URLSearchParams`, which cannot satisfy a multipart endpoint.

Option 1 also covers `realm-browser-bridge.ts:79`, which duplicates the same limitation.

## Environment

- SLICC 6.110.10 (issue also present in `main` at `e82c77e`, version 6.110.12)
- Context: `.jsh` script run via `node`, uploading a file to `POST https://api.x.ai/v1/files`
- Existing tests asserting the current behaviour, which would need updating:
  - `packages/webapp/tests/shell/supplemental-commands/node-fetch-adapter.test.ts:264`
  - `packages/webapp/tests/kernel/realm/realm-browser-bridge.test.ts:71`

trieloff:
Fixes #2687

## Summary

`fetch(url, { body: new FormData() })` threw in the `.jsh` runtime, so `multipart/form-data` uploads were impossible through the documented API surface. It now serializes the `FormData` into a real multipart body and sets a matching `Content-Type`.

Before: `Error: node fetch shim: FormData request bodies are not supported (post raw application/x-www-form-urlencoded with URLSearchParams instead)`. After: the request goes out as `multipart/form-data; boundary=…` with the file part intact.

## Why

`FormData`, `File`, and `Blob` are all constructible globals in `.jsh`, so the error only surfaced at `fetch` time — after the caller had already built the form. The suggested alternative produces `application/x-www-form-urlencoded`, which no endpoint requiring a file part accepts, so for that whole class of API (`POST https://api.x.ai/v1/files`, OpenAI `/v1/files`, anything taking `file=@path`) there was no working documented path at all — only ~25 lines of hand-assembled `Buffer` boilerplate per caller, where a misplaced CRLF or a missing trailing `--` fails in ways that are awkward to debug against a remote API.

**Repro** (`repro.jsh`, from the issue):

```javascript
const fd = new FormData();
fd.append("purpose", "assistants");
fd.append("file", new Blob(["hello"], { type: "text/plain" }), "hello.txt");
await fetch("https://httpbin.org/post", { method: "POST", body: fd });
```

Expected: a `multipart/form-data` POST.
Actual (6.110.10, and `main` at `e82c77e`): throws.

The transport was never the limitation. `SecureFetch` already carries arbitrary bytes as a latin1 string, and `prepareRequestBody` decodes that back to raw bytes for any non-text Content-Type — its doc comment already named multipart as a case it handles. Only the serialization step was missing.

New `packages/webapp/src/base/multipart-form-data.ts` — the single multipart encoder, in `base/` because both call sites live in different layers (`shell/` and `kernel/realm/`).

- **The boundary is minted in the same call that lays down the delimiters** and returned as a ready-made Content-Type. A header that disagrees with the body's delimiters is unparseable server-side, and splitting the two across functions is exactly how that drifts. This is why `node-fetch-adapter.ts` no longer computes the body and the default Content-Type in two independent calls (`encodeBody` + `getDefaultContentType`) — they now come from one `encodeInitBody`.
- Escaping follows the WHATWG `multipart/form-data` encoding algorithm (what undici and browsers implement): newlines in names and string values normalize to CRLF, and `"` / CR / LF inside `name=` / `filename=` become `%22` / `%0D` / `%0A`, so a crafted field name cannot forge a part header.
- Two entry points: an async `encodeMultipartFormData(form)` (async only because `Blob`/`File` bytes need `await .arrayBuffer()`) and a sync `encodeMultipartParts(parts)` for callers that already hold bytes.
- No new dependencies.

Wired into both sites that shared the old rejection:

| Site | Path |
| --- | --- |
| `shell/supplemental-commands/node-fetch-adapter.ts` | the `SecureFetch`-backed `fetch` shim |
| `kernel/realm/realm-browser-bridge.ts` | the `.jsh` realm's `fetch` RPC (`serializeRequestInit`) |

**Also folded in:** `fs/mount/backend-da.ts` had the repo's other hand-rolled multipart builder. It now delegates to the same encoder, which removes the duplicate and fixes a latent hole — the DA mount write path takes its `filename` from the VFS path, and a filename containing a quote or a newline used to break straight out of the `Content-Disposition` header. Its boundary prefix changes from `----DaMount…` to the shared `----SliccFormBoundary…`; the token is opaque, only the header/body match matters.

## Cross-cutting impact

- **Floats exercised**: this is realm/shell code shared by every float — CLI, extension, Electron, worker. No float-specific branch was touched; the encoder is pure and has no runtime detection in it.
- **`browser.fetch` is unchanged.** `realm-browser-fetch.ts` already carried `FormData` entry-by-entry and rebuilt a real `FormData` page-side so the page's own `fetch` sets the boundary. That path was already correct and stays as it is — only the two `SecureFetch`-backed paths changed.
- **Python realm unchanged** — `py-realm-shared.ts` builds its own init and never reached the throw.
- **Other callers of the changed contract**: `getDefaultContentType` / `encodeBody` in `node-fetch-adapter.ts` are module-private; `encodeInitBody` is the only new caller. The `Request`-input branch of `resolveRequestBody` is untouched — a `new Request(url, { body: form })` already serialized itself and came through as bytes with its own boundary header.
- **Behavior preserved on purpose**: a caller-supplied `Content-Type` still wins over the generated one (matching browser `fetch`, where setting the header manually on a `FormData` body is the same footgun), and `GET`/`HEAD` still drop the body without inventing a Content-Type. Both are covered by tests.
- **No persisted state, no migrations, no feature flags.**
- **Bundle size**: `npm run bundle-size` reports the first-load worker graph at **+1.1 kB** vs the merge-base, against a 5 kB allowance; the page graph is +0.0 kB.
- **Security**: the escaping and the CR/LF-stripping of part content types are header-injection defenses, exercised by dedicated tests. `Blob.type` is already constrained to printable ASCII by its constructor; the sanitizer covers the `encodeMultipartParts` path, where the type is derived from a filename.

## Test plan

- [x] `npm run lint` — passes (includes `check-swift-pins`, `lint:duplication`, formatters)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — boy-scout debt gate passes; back-edge and `Record<string, unknown>` baselines both clean
- [x] `npm run test` — **18139 passed / 0 failed**, 52 skipped (1022 files)
- [x] `npm run test:coverage:webapp` — gate passes; `multipart-form-data.ts` at 100% functions / 97.6% lines
- [x] `npm run build -w @slicc/webapp`
- [x] `npm run bundle-size` — first-load OK, `size-limit` OK
- [x] **New / changed behavior is covered by unit tests**: `npx vitest run packages/webapp/tests/base/multipart-form-data.test.ts packages/webapp/tests/shell/supplemental-commands/node-fetch-adapter.test.ts packages/webapp/tests/kernel/realm/realm-browser-bridge.test.ts packages/webapp/tests/fs/mount/backend-da.test.ts`

The tests assert the exact wire image for a pinned boundary, and then **reparse the encoded bytes with the platform's own multipart parser** (`Response.formData()`) — a hand-rolled envelope can look plausible and still be unparseable by a real server, which is precisely the failure mode the issue described. Covered: quote/newline injection in names, filenames and part content types; lone-CR/LF normalization; repeated field names; non-ASCII names and values; a `Blob` appended without a filename; an empty part list; the `crypto`-less boundary fallback.

**Manual smoke** — I ran the issue's repro end-to-end against a real `node:http` server, through the actual `prepareRequestBody` latin1→bytes hop that the proxy uses:

```
server saw Content-Type: multipart/form-data; boundary=----SliccFormBoundary83e6b77c44d2efcbb0aa9b347a3fee97
purpose -> assistants
file name -> hello.bin | bytes -> [ 0, 128, 255, 13, 10 ]
```

The high bytes (`0x80`, `0xff`) and the embedded CRLF survive intact — the case that would corrupt if the body were UTF-8 re-encoded anywhere on the path.

Not a UI change, so no screencast.

**Pre-existing failures** (unrelated to this PR): root `npm run build` fails in this environment at `@slicc/swift-server` with a WebRTC SPM revision mismatch (`Package.resolved` records `19aa8c1f`, the remote `151.0.0` tag resolves to `3e43fa9b`). No Swift files are touched here; the two JS builds this change can affect were run individually and both pass.

## Documentation

- [x] `docs/` — new **Request Bodies** section under *Proxied Fetch* in `docs/shell-reference.md`: a table of every accepted body shape with its wire form and default Content-Type, the latin1 convention, the boundary invariant, and the `browser.fetch` contrast.
- [x] No `README.md` / `CLAUDE.md` / agent-skill changes needed — no user-facing command or developer convention changed.

## Risk & rollback

- **Blast radius**: two `fetch` body-serialization paths plus the DA mount write path. The change is purely additive for every body shape that already worked — string, `URLSearchParams`, `Uint8Array`, `ArrayBuffer`, typed arrays, and `Blob` all take the same code path as before (existing tests for each still pass unchanged). The only behavior that differs is `FormData`, which previously threw.
- **DA mount** is the one path where working behavior changed shape rather than starting to work. It is covered by `backend-da.test.ts`, which now reparses the envelope with a real multipart parser instead of substring-matching it.
- **Rollback**: revert cleanly. No persisted state, no migration, no flag to unwind.
- **What CI cannot catch**: a real upstream file-upload endpoint. I verified the wire format against undici's parser and a live local HTTP server, but a reviewer with an API key may want to run the issue's repro against `https://api.x.ai/v1/files` or OpenAI `/v1/files` before approving.

## Checklist

- [x] Public API / shell-command changes are intentional and reflected in docs
- [x] Contract shared across floats — checked the leader/follower/offscreen paths: `browser.fetch` (page-side rebuild) and the Python realm both bypass the changed code and are unaffected

### DIFF
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -1872,6 +1872,13 @@ it back to raw bytes whenever the Content-Type is not text-shaped. Both the
 | `FormData`                                 | latin1 multipart | `multipart/form-data; boundary=<token>`                |
 | `ReadableStream`                           | —                | rejected; collect it into a `Uint8Array` first         |
 
+The defaults above hold on **both** paths. The realm's `serializeRequestInit`
+has to decide them itself rather than leaning on the host adapter: everything
+crosses the RPC boundary as a string, so by the time the adapter sees the body
+it can no longer tell a `URLSearchParams` from a `text/plain` payload. A
+`GET`/`HEAD` request drops the body on both paths and advertises no
+Content-Type for it.
+
 `multipart/form-data` is serialized by `webapp/src/base/multipart-form-data.ts`
 — the single encoder, also used by the DA mount backend. **The boundary token
 is minted in the same call that lays down the delimiters** and returned as a
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -1856,6 +1856,33 @@ await fs.writeFile('/output.jpg', newBytes);
 
 Network requests are proxied to handle CORS and cross-origin restrictions.
 
+### Request Bodies
+
+`SecureFetch` carries a `body: string`, so every non-text payload rides the
+**latin1 convention** (one character per byte) and `prepareRequestBody` decodes
+it back to raw bytes whenever the Content-Type is not text-shaped. Both the
+`.jsh` `fetch` global and the kernel realm's `fetch` RPC accept:
+
+| Body                                       | Wire form        | Default `Content-Type` (caller always wins)            |
+| ------------------------------------------ | ---------------- | ------------------------------------------------------ |
+| `string`                                   | verbatim         | none                                                   |
+| `URLSearchParams`                          | `toString()`     | `application/x-www-form-urlencoded;charset=UTF-8`      |
+| `Uint8Array` / `ArrayBuffer` / typed array | latin1           | `application/octet-stream`                             |
+| `Blob` / `File`                            | latin1           | the blob's own `type`, else `application/octet-stream` |
+| `FormData`                                 | latin1 multipart | `multipart/form-data; boundary=<token>`                |
+| `ReadableStream`                           | —                | rejected; collect it into a `Uint8Array` first         |
+
+`multipart/form-data` is serialized by `webapp/src/base/multipart-form-data.ts`
+— the single encoder, also used by the DA mount backend. **The boundary token
+is minted in the same call that lays down the delimiters** and returned as a
+ready-made Content-Type; never generate one separately, because a header that
+disagrees with the body makes the request unparseable server-side. Field names
+and filenames are escaped per the WHATWG algorithm (`"` / CR / LF → `%22` /
+`%0D` / `%0A`), so a crafted name cannot forge a part header.
+
+`browser.fetch` takes the same shapes but rebuilds a real `FormData` in the
+page instead, letting the page's own `fetch` set the boundary.
+
 ### CLI Mode
 
 Express server provides `/api/fetch-proxy`:
--- a/packages/webapp/src/base/multipart-form-data.ts
+++ b/packages/webapp/src/base/multipart-form-data.ts
@@ -1,0 +1,175 @@
+/**
+ * `multipart-form-data.ts` — the single `multipart/form-data` serializer.
+ *
+ * `SecureFetch` (and the realm's `fetch` RPC that funnels into it) can only
+ * carry a `string` body, so every binary payload rides the latin1 convention
+ * (one character per byte). That transport can express multipart fine — what
+ * was missing was the *serialization* step, which is why `fetch(url, { body:
+ * new FormData() })` used to throw in `.jsh` even though `FormData`, `File`,
+ * and `Blob` are all constructible globals there.
+ *
+ * The boundary token is minted here, in the same call that produces the bytes,
+ * and returned alongside them as a ready-made `Content-Type` value. Callers
+ * must never generate their own: a boundary that disagrees with the delimiters
+ * in the body makes the request unparseable server-side, and splitting the two
+ * across functions is exactly how that drifts.
+ *
+ * Escaping follows the WHATWG `multipart/form-data` encoding algorithm (the
+ * one `undici` and browsers implement): newlines in names and string values
+ * normalize to CRLF, and `"` / CR / LF inside `name=` / `filename=` become
+ * `%22` / `%0D` / `%0A` so a crafted field name cannot forge a part header.
+ */
+
+const CRLF = '\r\n';
+
+/** A `Blob`/`File`-shaped part: raw bytes plus its filename and MIME type. */
+export interface MultipartFilePart {
+  bytes: Uint8Array;
+  filename: string;
+  /** Defaults to `application/octet-stream` when empty. */
+  contentType?: string;
+}
+
+/** One `multipart/form-data` entry — a text field or a file part. */
+export type MultipartPart =
+  | { name: string; value: string }
+  | { name: string; file: MultipartFilePart };
+
+/** Assembled multipart body plus the `Content-Type` that describes it. */
+export interface EncodedMultipartBody {
+  bytes: Uint8Array;
+  /** Full header value with the boundary baked in. */
+  contentType: string;
+  boundary: string;
+}
+
+/**
+ * Mint a fresh boundary token. 128 bits of randomness keeps an accidental
+ * collision with the payload out of practical reach (the RFC forbids the
+ * delimiter appearing in any part), and the 53-character result stays inside
+ * the RFC 2046 70-character cap using only `bcharsnospace` characters.
+ */
+export function generateMultipartBoundary(): string {
+  const bytes = new Uint8Array(16);
+  if (typeof crypto !== 'undefined' && typeof crypto.getRandomValues === 'function') {
+    crypto.getRandomValues(bytes);
+  } else {
+    for (let i = 0; i < bytes.length; i += 1) bytes[i] = Math.floor(Math.random() * 256);
+  }
+  let hex = '';
+  for (const byte of bytes) hex += byte.toString(16).padStart(2, '0');
+  return `----SliccFormBoundary${hex}`;
+}
+
+/** Narrow a `BodyInit`-ish value to `FormData` without assuming the global exists. */
+export function isFormDataBody(body: unknown): body is FormData {
+  return typeof FormData !== 'undefined' && body instanceof FormData;
+}
+
+/**
+ * Serialize `parts` into a multipart body. Synchronous on purpose: callers
+ * that already hold bytes (e.g. a mount backend wrapping one file) do not
+ * need to build a `FormData` first. Pass `boundary` only from a test that
+ * needs a deterministic wire image.
+ */
+export function encodeMultipartParts(
+  parts: readonly MultipartPart[],
+  boundary: string = generateMultipartBoundary()
+): EncodedMultipartBody {
+  const encoder = new TextEncoder();
+  const chunks: Uint8Array[] = [];
+  for (const part of parts) {
+    const name = escapeHeaderParam(normalizeNewlines(part.name));
+    if ('file' in part) {
+      const filename = escapeHeaderParam(part.file.filename);
+      const type = sanitizeContentType(part.file.contentType) || 'application/octet-stream';
+      chunks.push(
+        encoder.encode(
+          `--${boundary}${CRLF}` +
+            `Content-Disposition: form-data; name="${name}"; filename="${filename}"${CRLF}` +
+            `Content-Type: ${type}${CRLF}${CRLF}`
+        )
+      );
+      chunks.push(part.file.bytes);
+      chunks.push(encoder.encode(CRLF));
+    } else {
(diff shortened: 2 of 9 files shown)


trieloff:
Both Codex findings confirmed and fixed in d40fe51e2. I ran the two paths side by side before changing anything:

```
BEFORE
F1 realm GET+FormData        -> headers: {"Content-Type":"multipart/form-data; boundary=----SliccFormBoundarya88c…"} | body: undefined
F2 realm POST+URLSearchParams-> headers: {}                                                                          | body: grant_type=client_credentials
   direct GET+FormData       -> headers: {}                                                                          | body: undefined
   direct POST+USParams      -> headers: {"Content-Type":"application/x-www-form-urlencoded;charset=UTF-8"}           | body: grant_type=client_credentials

AFTER
F1 realm GET+FormData        -> headers: {}                                                                          | body: undefined
F2 realm POST+URLSearchParams-> headers: {"Content-Type":"application/x-www-form-urlencoded;charset=UTF-8"}           | body: grant_type=client_credentials
```

**Finding 1 (GET/HEAD)** — correct. `serializeRequestInit` serialized the body regardless of method, so the header survived a body the host adapter then dropped. Fixed by gating serialization on the method, which also covers the pre-existing `Blob`/`ArrayBuffer` instances of the same thing (they left an `application/octet-stream` header behind). Covered by an `it.each` over `FormData` / `Blob` / `URLSearchParams`.

**Finding 2 (URLSearchParams)** — correct, and it turned out to be a live bug rather than a docs error, so I fixed the code instead of splitting the table. Your diagnosis is exactly right: everything crosses the RPC boundary as a string, so the host adapter can no longer tell a form from a `text/plain` payload. `.jsh` is the path that goes through `serializeRequestInit`, which means `.jsh` OAuth token posts were shipping with **no** `Content-Type` at all — the precise failure `node-fetch-adapter.ts`'s own header comment says the default exists to prevent. The realm serializer now sets it, with a caller-provided value still winning.

Docs updated to state that the defaults hold on both paths and to say why the realm has to decide them itself.

Also noted from the Claude review, no change made: in `encodeMultipartParts` the field name is newline-normalized before escaping while the filename is only escaped, so a lone `\r` in a name becomes `%0D%0A`. That asymmetry is intentional — it matches both undici's serializer and the HTML spec's multipart encoding algorithm, which normalizes newlines in the name (step 1) but not in the filename.

Full pass re-run green: lint, typecheck, **18145 tests / 0 failures**, webapp coverage gate, boy-scout gate, both JS builds, `bundle-size` (worker +1.2 kB against a 5 kB allowance).

PR #2692, opened 2026-08-31
feat(webapp): keyboard map v2 — chords, the urgent commands, override-only keys.json

trieloff:
## Why

Keyboard mode shipped last week with thirteen keys. Auditing the UI against them turned up a tier the map had no room for:

- **Stopping a running turn has no key at all** — the send button emits `stop` on click only, so the one urgent thing in the app was a Tab away into a shadow-DOM button, mid-stream.
- The **Monitor** was the only dock item of five without a letter; the **pending approval**, the **attach menu** and the **copy row** had none either.
- **Archived chats** are mouse-only in the DOM (`slicc-freezer-card` is a click handler on a host with no `tabindex`, `role` or `keydown`), so they had no keyboard path of any kind.

Meanwhile `a` spent a prime key on account settings. With a week of muscle memory to weigh against that, this is the moment to redraw the map rather than bolt keys onto its edges.

## The map

Four rules, written out over `DEFAULT_KEYMAP`: prime keys go to **urgency, not nouns**; **positional beats mnemonic** wherever a list exists; **the same key closes what it opened**; **Shift is the heavier twin** of the same letter.

| | |
|---|---|
| `i` / `⏎` · `?` | composer · help |
| `1`–`9` · `→` / `←` | agent by position · next / previous |
| `n` / `N` · `c` / `C` · `r` | new chat / erasing · new cone / drop cone · archived chats |
| `s` · `a` · `u` · `y` / `Y` | **stop** · **go to the approval** · attach · copy reply / whole chat |
| `f` · `t` · `b` · `m` · `g` · `p` / `P` | files · terminal · browser · memory · monitor · sprinkles / cycle |
| `[` / `]` · `z` | rails · full screen the open panel |
| `l` · `,` | model · accounts |

Thirteen commands become twenty-six. `e h j k o q v w x` and `.` `;` stay free, and so does `/` — it belongs to a future command palette, not to a third synonym for `?`.

## Chords

Any surface that owns an ordered list takes a digit straight after it: `f 3` is the third file, `r 1` the chat you archived most recently, `p 2` the second sprinkle. `9` is the last one everywhere, exactly as it already was for agents (`indexForDigit`).

- The prefix runs **eagerly**, so the single-key path pays nothing for the chord's existence.
- The lookahead is **one key wide**, so no digit is ever ambiguous: armed, it addresses the list; otherwise the tab strip, as always.
- A chord belongs to the **command**, not the key — rebind `files` to `q` and `q 3` moves with it.
- A list command therefore **holds the mode**. Dropping it would mean the completing digit arrived with the keyboard off, one macrotask before `settle` turned it back on.

## The migration (please review twice)

v1 seeded all sixteen of its bindings **explicitly** into `/etc/slicc/keys.json`, and that file is applied *over* the defaults. So every install that has ever booted v1 is pinned to v1 forever: a new shipped map would reach nobody, and neither would a key for any command added since. The seed's own comment promised the opposite.

- The seed now ships `"bindings": {}` — overrides only, so the shipped map (and every later change to it) is inheritable.
- A file still holding the v1 map **exactly** (`isUntouchedV1Document`) is replaced once. v1 with a single line added, removed or re-pointed is somebody's config and is left alone.
- A write that fails changes nothing: the v1 keys are still applied, and the next boot tries again.
- `V1_KEYMAP` stays in code for that test, and in the file's comment as a paste-back block for anyone who wants the old keys.

## Deferred

Three things this deliberately does not do, each worth its own change:

- **Voice (`v`)** — push-to-talk is a pointer-hold gesture on the composer with no programmatic entry point; a tap-to-talk toggle needs its own speech-controller work.
- **Arrange mode (`Z`)** — the panel move grips and the dividers are `pointerdown`-only, so keyboard move/resize is a feature, not a binding.
- **`b 1-9`** — the browser tab list is populated asynchronously after the overlay opens, so half a chord would race it.

And the caveat that outlives this PR: **a keymap is not accessibility.** `r 1-9` gives archived chats their first keyboard path, but `slicc-freezer-card` still has no `tabindex`, `role` or `keydown`, so Tab users and screen readers are exactly where they were. Same for `slicc-memrow` and the attachment thumbnail. Those fixes are separate and non-optional.

## Verification

`npm run verify` · `typecheck` · `test` (18k) · `test:coverage` · `build -w @slicc/webapp` · `build -w @slicc/chrome-extension` · `bundle-size` · `check-touched-exemptions` — all green. First-load gate: +0.0 kB on both graphs.

163 tests across the three shortcut suites, including the chord state machine (expiry, one-key width, empty list, rebound key, Escape disarming), the six DOM-reached commands, and every branch of the v1 migration.

Root `npm run build` also builds `@slicc/swift-server`, which fails in this environment on `swift build -c release` — unrelated to this change, which touches no Swift.

Not live-verified in a browser: the help overlay is derived from the live keymap (`shortcutRows`) and covered by tests, but if you want the sheet on screen before merging, say so and I will record it.


trieloff:
### Update: `v` and the step keys (second commit)

Three things, one grammar.

**`v` dictates** — start listening, press again to send. This is the one command whose surface had to *grow* an entry point rather than being handed one it already had: push-to-talk arms on `pointerdown` and holds until `pointerup`, which a keyboard cannot express. `<slicc-composer>` gains `toggleHandsFree()`, reusing the press lifecycle rather than copying it — the same `#pressed` token, the same `#beginPress` permission staging, the same `#endPress` finalize, so the transcript is appended and submitted down the one path. Only two things differ, and both follow from there being no pointer: the engage timer is skipped (it exists to tell a tap from a hold, and a keystroke is neither), and nothing is captured or listened for, so the press can only end where it began. The overlay says *"Listening — press again to send"* instead of *"release"*, and a press a **finger** started is never ended by the key.

**`j` / `k` step.** With no list open they walk the transcript a message at a time — position-based off the thread's own box, so "next" means the same thing after a wheel scroll as after ten presses. With a list open they page *that* list, looping: `p j j` walks the sprinkles, `f k` jumps straight to the last file, `r j` steps down the archived chats.

**`P` retires.** `p` opens the first sprinkle and `j` walks the rest, so a dedicated cycle key had nothing left to say that the grammar didn't.

Paging turned the chord from one key of lookahead into a short walk:

- The armed chord carries the **index** it is standing on (`ArmedChord`).
- A step key owns no list but **re-arms the one it just walked**, which is what keeps `f j j j` walking instead of reverting to message scrolling.
- `Command.run` may **return the index it activated**, so a prefix can seed the walk — `sprinkles` opens the first and returns `0`, which is exactly why `p j` is the second and not the first again.
- A digit re-arms at wherever it landed, so `f 3 j` is the fourth file.

And `CHORD_WINDOW_MS` is now `HUD_LINGER_MS` rather than half of it, with every chord press refreshing both. That buys a rule worth stating in one line: **a chord is live exactly while its caps are on the pill.** The old shorter window existed so an expired chord couldn't still be advertised; with paging, the same equality means a walk down a long list never expires under you mid-walk, and stopping ends both together.

Verification re-run whole: `verify`, `typecheck`, `test` (18,268), the webcomponents browser suite (2,182 — 77 in the composer alone, including 8 new hands-free ones on a real Chromium), `test:coverage`, both builds, `bundle-size`, first-load (+0.0 kB), boy-scout gate.

Still deferred: arrange mode for the pointer-only panel grips, and `b 1-9` for browser tabs.

trieloff:
### Review fixes (`3ec16f11a`)

All four findings were real — and the three P2s were the same shape: a key advertising a surface it couldn't actually address. Thanks, that's exactly the class of bug tests written alongside the code don't catch.

**Approvals inside the dip iframe** (Codex, P2) — confirmed: a `tool_ui` card is a `mountDip` srcdoc frame, so `thread.querySelectorAll('button[data-action]')` found nothing in production and `a` was a no-op for every real approval. My tests satisfied the selector with buttons appended straight to the thread, which is precisely the fixture that hid it.

Fixed by anchoring on **`[data-tool-ui-request]`** instead — the container `WcChatController` appends per card and removes the moment the request is answered. That's a better contract than the buttons anyway: pending-ness is stated in the parent document, on the near side of the frame. From there the key lands on the card's own button → the button inside the dip (`allow-same-origin` in the ordinary float, so it's reachable) → the frame itself, which is what the extension gets, where the card comes from another origin. Focusing the frame still puts the keyboard inside the card.

**Nested file-tree items** (Codex, P2) — confirmed: `items` is nested, top level is `/workspace` + `/shared`, so `f 3` selected nothing and `f 1`/`f 2` selected directories. `<slicc-file-tree>` gains **`visibleIds()`** — the rows it is *showing*, in painted order — and the chord indexes that, so numbering follows what's expanded. Chasing it turned up a second bug worth flagging on its own: the library prints a **directory** row's path with a trailing slash while `#meta` is keyed without one, so a bare `#meta.get(row.path)` silently drops every folder.

**Filtered sessions** (Codex, P2) — confirmed: `<slicc-freezer>`'s live search toggles `match-hidden` rather than removing rows, so `r 1` after a search could restore something off-screen. Positional keys now address visible cards only — the same rule the file list just adopted, and now stated once in both places.

**Stale HUD hint** (Claude, minor) — confirmed and the sharpest catch of the four: the hint was the literal `'h for help'` while the shipped map moved help to `?`, so the first thing the mode told a new user to press did nothing. Rather than hard-code `?`, it's derived from the live keymap (`helpKeyLabel`, read when the badge appears) — so a rebind moves the hint with it, and unbinding help drops the clause. Same reason `shortcutRows` derives the help sheet instead of printing a list.

Tests rewritten against the production shapes: approval cards as `[data-tool-ui-request]` containers with a real iframe (including an unreachable-`contentDocument` case for the extension path), a `visibleIds()`-shaped file tree, `match-hidden` cards in the freezer, and three cases for the hint. Plus three new browser tests for `visibleIds()` on the real `@pierre/trees` renderer, which is what exposed the trailing-slash behaviour.

Full pass re-run green: `verify`, `typecheck`, `test` (18,277), webcomponents browser suite (2,185), `test:coverage`, builds, `bundle-size`, first-load (+0.0 kB).

PR #2697, opened 2026-09-01
fix(webapp): distinguish held from working OAuth tokens and name --force-login

### LINKED ISSUE #2695: oauth-token: --renew reports 'before: valid' for a dead token and never names --force-login as the fix
## Summary

When a provider's stored OAuth token has fully lapsed, `oauth-token --renew <id>` fails in a way that is hard to act on:

1. it reports **`before: valid`** for a token that is in fact dead (real API calls return `Bad credentials`);
2. the remediation text points at the **browser DevTools console**, which an agent driving SLICC cannot read; and
3. it never names the one command that actually fixes it — **`oauth-token <id> --force-login`**.

Net effect: the failure reads as "something is broken in the OAuth service" rather than "your session lapsed, re-consent". On 2026-09-01 this cost a session two diagnostic detours before falling back to `--force-login`, which worked first try.

Measured on **SLICC 6.111.1**.

## Reproduction

Starting state — a GitHub token that is stored but no longer accepted upstream:

```
$ gh auth status
Personal token
  Token:   gho_db62…
  User:    (invalid token)
```

Then:

```
$ oauth-token --renew github
oauth-token --renew github
  before: valid
  silent renewal: FAILED (onSilentRenew returned null)
  → no window should have appeared. Open DevTools console and
    look for "[oauth-service] Extension OAuth error" / "[adobe]" to see
    the IMS/Chrome reason (e.g. login_required).
$ echo $?
1
```

Exit code 1 is correct. The problems are in the reported state and the guidance.

### Defect 1 — `before: valid` is derived from local expiry only

The stored token had not passed its recorded expiry, so the gate calls it `valid`, but the provider had already invalidated it. Anything that reads that line concludes the credential is fine and the renewal machinery is at fault — the opposite of the truth.

`--list` has the same shape of problem:

```
$ oauth-token --list
github (logged in as Lars Trieloff, scopes: read:user,repo,user:email)
```

Reported as logged in, while `gh` says `(invalid token)`. Neither surface distinguishes *"we hold a token"* from *"the token works"*.

### Defect 2 — remediation is unreachable for the primary caller

`Open DevTools console and look for "[oauth-service] Extension OAuth error"` assumes a human with the browser devtools open. The main consumer of this CLI is an agent in the SLICC sandbox, which has no way to read that console. So the one actionable sentence in the output is addressed to someone who usually is not there.

### Defect 3 — the working fallback is never named

`--force-login` resolves this immediately:

```
$ oauth-token github --force-login
gho_5722…                      # fresh token, gh picks it up via skill.token(github)
$ gh auth status
  User:   trieloff
```

Neither the `--renew` failure output nor the `--renew` entry in `--help` mentions it. `--help` currently reads:

```
oauth-token --renew [<id>]      Force a silent token renewal now (onSilentRenew),
                                bypassing the expiry gate. Reports success and
                                the new expiry.
```

— with no note on what to do when it does not succeed.

## Suggested improvements

All additive; no behaviour change to the success path.

**1. Say what `onSilentRenew returned null` actually means, and what to run.**

```
$ oauth-token --renew github
  stored token: present, local expiry 2026-09-03T11:00:00Z (not yet expired)
  silent renewal: FAILED — the provider declined to reissue without user
    interaction (typically login_required: the upstream session has ended).
    A stored token that has not hit its local expiry can still be rejected
    upstream; local expiry is not proof of validity.

  → Fix: oauth-token github --force-login
    This opens a consent window and needs a human at the keyboard.
    (Diagnostics: browser DevTools console, "[oauth-service] Extension OAuth error".)
```

Keep the DevTools hint, but demote it below the actionable command.

**2. Distinguish "held" from "working" in `--renew` and `--list`.** `before: valid` is the misleading part. Something like `stored token: present (local expiry …, not validated upstream)` would be honest. Optionally add `oauth-token --check <id>` to make one cheap authenticated call and report whether the provider still accepts the token — that is the question callers actually have.

**3. Document the escalation ladder in `--help`**, so the sequence is discoverable without failing first:

```
oauth-token --renew [<id>]      Force a silent token renewal now (onSilentRenew),
                                bypassing the expiry gate. Reports success and the
                                new expiry. If the upstream session has ended this
                                cannot succeed — fall back to --force-login, which
                                requires user interaction.
```

**4. Consider a distinct exit code for "needs interactive login"** (e.g. `3`) versus other renewal failures. That lets a caller branch on it — retrying, or stopping to ask a human — instead of parsing prose. Today every renewal failure is exit 1.

## Why this matters beyond one session

`--force-login` opens a consent window, so an unattended agent must *stop and ask a human*. That is the right boundary. But the current output makes the trigger for that escalation look like a bug in the OAuth service, so an agent is likely to retry, probe, or report a false problem instead of simply saying "your GitHub session lapsed, please re-authorise". Naming the fallback in the failure message turns a dead end into a one-line handoff.

Related note for whoever picks this up: `oauth-token github` (plain provider mode) returns the 62-char string `oauth-token: no masked value for github (try logging in again)` on stdout when no token is readable. That is *prose on the success channel* and it is long enough to pass a naive length check for a token — a separate small trap in the same command, worth fixing alongside.

trieloff:
## Problem

When a provider's stored OAuth token has fully lapsed, `oauth-token --renew <id>` reported
`before: valid` for a credential the provider had already invalidated, pointed remediation at
the browser DevTools console (which the primary caller — an agent — cannot read), and never
named the one command that fixes it. The failure read as "the OAuth service is broken" rather
than "your session lapsed, re-consent".

Fixes #2695

## What changed

**Held is not working.** `--list` and `--renew` only ever knew what is stored locally, so they
no longer claim more than that:

- `--renew` prints `stored token: present, local expiry in 23h 0m, not validated upstream`
  instead of `before: valid`.
- `--list` rows read `github (token held for Lars Trieloff, local expiry in 23h, scopes: …)`
  instead of `logged in as …`, and the listing ends with a one-line caveat naming `--check`
  and `--force-login`.

**The failure output names the fix.** A declined silent renewal now explains what
`onSilentRenew returned null` means (the provider refused to reissue without user
interaction — typically `login_required`), states that local expiry is not proof of validity,
and leads with:

```
  → Fix: oauth-token github --force-login
    Opens a consent window, so a human has to be at the keyboard: an
    unattended agent should stop and ask the user to re-authorise rather
    than retrying or probing further.
    (Diagnostics, if a human is present: browser DevTools console,
     "[oauth-service] Extension OAuth error".)
```

The DevTools hint is kept, demoted below the actionable command.

**New `oauth-token --check [<id>]`** answers the question callers actually have: it calls the
provider's new optional `onValidateToken` hook — one cheap authenticated call — and reports
`ACCEPTED` / `REJECTED` / `UNKNOWN`. GitHub implements it via `GET /user`; 401/403 are
`rejected` (a new login is required), 5xx and transport failures are `unknown` (says nothing
about the token, so it never sends a caller to an interactive login it does not need).
Providers without the hook say so instead of guessing.

**Exit code 3 = "a human must complete an interactive login."** `--renew` (declined) and
`--check` (rejected) exit 3 so a caller can branch instead of parsing prose. Exit 1 keeps its
meaning: could not run, or outcome unknown — including a renewal that *threw*, which may be
transient.

**`--help` documents the escalation ladder** (`--check` → `--renew` → `--force-login`), the
held-vs-working distinction, and the exit codes, so the sequence is discoverable without
failing first.

**Related trap from the issue:** `oauth-token <id>` with no readable masked value returned the
62-char sentence `oauth-token: no masked value for github (try logging in again)` — long
enough to pass a naive "looks like a token" length check. It now reads
`oauth-token: no usable token for github; run: oauth-token github --force-login` and a test
pins it to stderr with an empty stdout.

## Verification

`npm run verify`, `npm run typecheck`, `npm run test`, `npm run test:coverage:webapp`,
`npm run build`, `npm run build -w @slicc/chrome-extension`, `npm run bundle-size`, and
`check-touched-exemptions.mjs` all pass. 21 new tests cover the `--renew`/`--check`/`--list`
wording, the exit codes, and every `validateGitHubToken` branch.


trieloff:
Addressed every finding from @chatgpt-codex-connector and Copilot in `ae286ce7d`. Both bots were pointing at the same flaw from two sides: the first cut over-claimed, and an exit code that wrongly demands a human is the mirror image of the bug this PR fixes.

**Only a 401 is `rejected`.** GitHub answers 403 for primary and secondary rate limits, so an undifferentiated 403 is now `unknown` (tagged `(rate limited)` when `x-ratelimit-remaining: 0` / `retry-after` says so). A throttled token never reaches a consent window that cannot fix it.

**`onSilentRenew` returning `null` is no longer treated as proof.** You are right that GitHub, Adobe and Copilot all collapse transport failures into `null`. A declined `--renew` now runs the upstream check before it reports:

- confirmed `REJECTED` → exit 3, "stop and ask a human";
- `UNKNOWN` or no hook → exit 1, decline reported as **unconfirmed**, `→ Likely fix:` names `--force-login` without claiming a retry is pointless;
- `ACCEPTED` → exit 1 and "the stored token still works, callers are not blocked; only the renewal did not happen" — no login suggested at all.

**A renewal that threw** gets retry advice instead of `→ Fix: --force-login`.

**`--check` on a refused token points at `--renew` first** when the provider can renew, since a stored refresh token may replace the access token with no human involved; exit 3 there is now reserved for providers with no silent renewal. Exit 1's contract is restated as "not settled — retry, or take the next rung the output names", and exit 3 as "automated recovery is exhausted".

**The validation call is bounded** with `AbortSignal.timeout(10s)`, so `--check` always answers.

**`expired` is a 60s renewal buffer**, so the wording now derives from `expiresAt` and `expired` only adds `renewal needed` — a token with 30s left reads `local expiry in under a minute — renewal needed`.

**`onValidateToken` is documented** in `.agents/skills/adding-slicc-features/SKILL.md`, including the accepted/rejected/unknown rules and why `null` from `onSilentRenew` proves nothing.

One thing that fell out of fixing these: the added help text pushed the eager first-load closure 5.5 kB over its 5 kB allowance, so `oauth-token` now takes the repo's stub + `run.ts` split (like `curlwright` and `pdftotext`). The whole command loads on first use and the eager graph is **17 kB smaller than the merge-base**.

Local: verify, typecheck, full test suite (18175 passed, 0 failed), `test:coverage:webapp` (exit 0), both builds, `bundle-size` (First-load OK), and `check-touched-exemptions` all pass.

PR #2721, opened 2026-09-01
feat(webcomponents): number the tab switcher's cards, and peek a tab

trieloff:
Follow-up to #2692, which deferred `b 1-9`. Adds the tab digits and the peek gesture.

## `b` then a digit

I deferred this because the tab list arrives asynchronously — a refresh plus one screenshot per tab — so a chord digit would race the paint. Looking properly, the race was the smaller half: **`<slicc-tab-overlay>` is modal, and keyboard mode suspends every command while a modal is up**, so `b 3` could never have reached a tab as a shell chord at all.

So the overlay owns the keys. Same keystrokes, handled where the cards are already on screen:

- `1`–`8` by position, `9` always the last — the rule the tab strip and the chord lists already follow, restated locally because a component may not import from the app.
- Each card **draws the digit that selects it**, so the numbering is something you read rather than count.
- Past the eighth card the middle of the list is left unbadged rather than badged with a key that does nothing; the last card still shows `9`.
- Every key the overlay takes is stopped, so the shell's HUD never records a press twice.

## Peek

`p` arms it — reflected as `data-peek`, chip in the header, cleared whenever the overlay closes, because an armed modifier nobody can see is a trap. The next activation, **by key or by click**, emits `tab-peek` instead of `tab-activate`, and `wc-browser.ts` answers it by doing exactly what a switch does and undoing it five seconds later. A peek is a switch that puts itself back.

Coming back has two steps that must not be confused:

- `bringToFront` on SLICC's own target is what the **user** sees.
- Re-attaching whatever page the **agent** was driving is what keeps the gesture out of its way — invisible, because attaching does not foreground.

The return target is resolved fresh per peek (`location.href` exactly, else any `isSliccAppUrl` on our own origin) rather than remembered from the last refresh: an id that has since closed would strand you in the tab you peeked. If there's no way home the peek degrades to a plain switch and logs why, and a second peek replaces the pending return instead of stacking one. A follower's tab still teleports — activating it pulls a *copy* here, which is not a visit, so there's nowhere to come back from.

**On `p`:** you asked for it if it was free. In the shell keymap it isn't — it's sprinkles — but inside the switcher it is, because a modal owns the keyboard and the shell is suspended for exactly as long as the overlay is up. So your mnemonic works without a collision. Say the word if you'd rather it were a key that means one thing everywhere.

## Verification

`verify`, `typecheck`, `test` (18,281), webcomponents browser suite (2,196), `test:coverage`, builds, `bundle-size`, first-load (+0.0 kB), boy-scout gate — all green, rebased onto current `main`.

11 tests for the overlay's keyboard (digits, `9` = last, past-the-end, swallowing, deaf-while-closed, badge placement including the unreachable middle) and 4 for peek's wiring (goes and comes back, restores the agent's attachment, replaces a pending return, degrades when SLICC's tab can't be found, teleports a follower tab).

Not live-verified: the return path calls real CDP (`attachToPage` + `bringToFront`), which the tests fake. Worth one manual peek in a dev instance before merge — that's the half I'd want eyes on.


trieloff:
### Review fixes (`472ed22d5`)

All eight findings were real. Two of them are the same bug seen from different angles, and it's the one I'd have shipped:

**The agent's page was the wrong page** (Codex P1 + Copilot). `refresh()` attaches to every tab in turn to capture thumbnails, so by peek time `getAttachedTargetId()` was whichever card was captured *last* — the return would silently re-point the agent at a tab the user had merely looked at. Captured now when the switcher opens, before the thumbnail loop borrows the attachment, and only while the overlay is shut so a refresh from inside it (closing a tab) can't overwrite the real answer.

Copilot also asked for the fake browser to mutate its attached id so the regression is actually exercised — done, and worth the ask: I mutation-checked it by reverting the fix, and the new test fails against the old code (as does the `previous === id` one). Without the mutating fake both would have passed on the broken version.

**`previous === id` skipped the restore** (Codex P1 + Copilot). When the agent was already driving the peeked tab, the return had already pointed the shared session at SLICC, so skipping left the agent on SLICC's own UI. Only `self` is excluded now.

**Two failure paths didn't degrade the way my own comment claimed** (Copilot). A rejecting `findSelfTarget` took the whole peek down before it ever switched; a failing `activate` swallowed its error and resolved, so a return got scheduled from a trip that never happened. Self-resolution answers `null` on failure — not knowing the way home degrades a peek to a switch, it doesn't cancel the switch — and `activate` reports success so the return is scheduled only after one.

**A digit typed before the tabs arrived was dropped** (Codex P1). Right, and this was the thing I'd talked myself out of: the switcher opens on the keystroke that asks for it and fills in asynchronously, so `b 3` at speed lands on an empty grid — making a positional key work or not depending on how fast CDP answered. A digit with no list to act on is now held for `PENDING_DIGIT_MS` (3s) and acts the moment the tabs land. Bounded on purpose: a digit that fires into a list the user never saw is worse than one that did nothing. Cleared on close, and only the most recent digit is held.

**The overlay was claiming the browser's keys** (Copilot). ⌘P and ⌘1 were being turned into overlay actions. Modified and composing keys pass through now — the same rule the shell's keyboard mode follows, for the same reason.

**Peek was offered where it cannot work** (Copilot). `tab-peek` is emitted by the shared overlay but only the leader answers it, so on a follower, Cherry or the extension side panel the chip promised a return nobody would perform. Rather than wiring `tab-peek` to the follower's one-way activation — which would make the chip say "comes back" while permanently teleporting the tab — those floats now set `no-peek`: `p` is inert and `peeking` refuses to arm. Their tabs are the leader's and activating one copies it here for good, so there's nowhere to come back from, and withholding the affordance is the honest version. That also means the event can't fire there at all, so there's no unhandled path left.

18 new tests across the three suites. Every touched suite is green (209 webcomponents dock, 24 webapp browser/follower), `verify` and `typecheck` clean, and the `wireWcBrowser` length gate pushed the peek machinery out into a `createPeek` factory, which reads better anyway.

One caveat on my end: the full local suite is flaky on this machine right now — two consecutive runs failed *different* sets (11 then 14: pyodide, providers, chrome-extension secrets, swift-drift), one of which I confirmed also fails with my changes stashed. Nothing in the failing set touches this PR, but I'm not claiming a green full run locally — CI is the arbiter.

And the thing still worth your eyes: the return path calls real CDP (`attachToPage` + `bringToFront`), which the tests fake. Whether Chrome honours `bringToFront` back to SLICC's own tab from a background tab is the assumption the whole gesture rests on.

Issue #2720, opened 2026-09-01
hostfs: one transient `Failed to fetch` inside a fan-out aborts the whole git command (`fatal: EIO: hostfs bridge unreachable`) — no retry, and the bridge's 5 s keep-alive races the client


trieloff:
Part of #2707.

## What happens

`git -C /mnt/slicc status --short` failed **both** times it ran (exit 128 after ~5 s and ~6,000 requests):

```
fatal: EIO: hostfs bridge unreachable: Failed to fetch '/mnt/slicc/packages/swift-launcher/SliccstartTests/StartupPreferenceTests.swift'
fatal: EIO: hostfs bridge unreachable: Failed to fetch '/mnt/slicc/packages/webapp/src/speech/whisper-session.ts'
```

The bridge was up the whole time (the long-form `git status` that ran right after completed, 20k+ requests). A single `fetch()` rejection in the middle of `statusMatrix`'s `Promise.all` fan-out is mapped to `EIO` by `HostFsMountBackend.request()` (`packages/webapp/src/fs/mount/backend-hostfs.ts`: "envelope failures … surface as EIO") and propagates straight out of isomorphic-git as a fatal error; nothing retries.

## Why a fetch fails at all

- node-server's Express server uses Node's default `keepAliveTimeout` (5 s; the responses advertise `Keep-Alive: timeout=5`). Under a burst of thousands of small requests the browser keeps sockets pooled; when the server closes an idle socket at exactly the moment the browser reuses it, the request dies with `TypeError: Failed to fetch` (the classic Node keep-alive race — `ECONNRESET` on the client side). The larger the fan-out and the more bursty the pattern (the walk pauses to hash a file, then fires the next batch), the likelier it is.
- The same window also sees the sprinkle boot crawl (#2717) competing for the 6-connection pool, so queue times of >1 s are common (p99 1.6 s), which makes idle-socket reuse more likely, not less.

## Fix ideas

- Retry idempotent hostfs ops (`GET list/stat/read`) once or twice on `TypeError: Failed to fetch` / network errors before surfacing `EIO` (writes need an idempotency key or a re-stat).
- Raise `server.keepAliveTimeout` (and `headersTimeout`) on the bridge well above the browser's pool idle time (e.g. 60–120 s), or disable keep-alive reuse races by setting `Connection: close` on hostfs responses (costs a handshake per request but removes the race) — best combined with batching/WebSocket transport (#2715).
[...]

Issue #2707, opened 2026-09-01
git over a `--mount`ed host repo: benchmark + profile of the browser-side git (tracking issue)


trieloff:
## Why

A Slack question ("why is the same git command fast one run and awful the next?") had no measured answer. This issue records a reproducible benchmark of SLICC's isomorphic-git commands running against the slicc repo itself, mounted into the VFS through the hostfs bridge (`--mount=<repo>:/mnt/slicc`, PR #2339), plus a V8 CPU profile of the kernel worker for every run. Each distinct problem found is filed separately and linked below.

## Setup

- Build: main @ 4913779a3, `npm run build -w @ai-ecoverse/cherry -w @slicc/webapp -w @slicc/node-server`
- Harness: `PORT=5715 WRANGLER_PORT=8787 npm run dev:standalone:fresh -- --mount=/Users/…/ai-ecoverse/slicc:/mnt/slicc` (Google Chrome 152, fresh profile, node-server hostfs bridge on loopback)
- Repo under test: the slicc main checkout — 3,549 tracked files, 9,534 commits on HEAD (11,271 across 29 local branches), 30 packfiles (largest 92 MB, 85 MB), `.git/index` 437 KB
- Driver: a CDP script that types `<cmd> > /tmp/out 2>&1; echo SENTINEL` into the Terminal panel, times Enter→sentinel, records every request the kernel worker makes (`Network.*` on the worker session, grouped by `/api/hostfs/<op>`) and captures a `Profiler` run of the kernel worker. Each command was run twice (cold, warm). Scripts are in the collapsed section at the bottom.
- Native baseline: the same commands with system git in the same checkout. `.git/index` was backed up before and restored after the runs (see #2708 for why that was necessary).

## Results

| command (`git -C /mnt/slicc …`) | native | SLICC cold | SLICC warm | requests (warm) | notes |
| --- | ---: | ---: | ---: | ---: | --- |
| `rev-parse HEAD` | 31 ms | 205 ms | 202 ms | 38 (34 reads = every pack `.idx`, 3.8 MB) | #2713 |
| `branch` | 41 ms | 104 ms | 102 ms | 125 (100 stat, 22 list) for 29 refs | #2716 |
| `ls-files` | 32 ms | **117 s** | **11.1 s** | 30,276 cold / 16,336 warm | cold run wrote `.git/index` **3,485×**; #2708, #2709 |
[...]

PR #2723, opened 2026-09-01
perf(hostfs): serve metadata ops from one stable URL so CORS preflights cache

### LINKED ISSUE #2715: hostfs: every request pays a CORS preflight — 246,893 `OPTIONS` for 385,033 `GET`s in one session
Part of #2707.

## What happens

The kernel worker (origin `http://localhost:8787` — the hosted UI origin in production) calls the local bridge (`http://localhost:5715/api/hostfs/*`) with the `X-Bridge-Token` header (`packages/webapp/src/base/api-endpoint.ts` `apiHeaders()`). A custom header makes every fetch a non-simple CORS request, so the browser sends an `OPTIONS` preflight first. node-server sets `Access-Control-Max-Age: 600` (`packages/node-server/src/index.ts:1223`) but the preflight cache is keyed **per URL**, and every hostfs URL is unique (`?mount=&path=`), so it almost never helps.

node-server log for the benchmark session: **385,033 `GET /api/hostfs/*`, 246,893 `OPTIONS /api/hostfs/*`, 3,485 `PUT`** — 64 % of GETs paid an extra round trip; `220,310` of the GETs were `304`s, i.e. the round trip was the entire cost. Inside a fan-out (git status), request latency was p50 2.7 ms, p90 226 ms, p99 1.6 s — the browser's 6-connections-per-host limit is shared by preflights and real requests.

## Fix ideas

- Make hostfs requests *simple* so no preflight is needed: drop the custom header and carry the bridge token as a cookie (`SameSite=None; Secure` isn't needed on loopback… but `localhost` cookies are per-port-blind, so scope by path) or as a query parameter on this route only (the token already sits in the launch URL), or accept `Authorization`… (no — also non-simple). A `POST` with `Content-Type: text/plain` and a JSON body is also preflight-free.
- Put the path in a stable URL shape so `Access-Control-Max-Age` can actually cache (`POST /api/hostfs` + body) — combines well with batching (#2716).
- Or move hostfs traffic off HTTP entirely onto the existing `/cdp` bridge WebSocket (one connection, no preflights, no 6-connection limit).

trieloff:
## Summary

`/api/hostfs` gains a stable `POST /api/hostfs` endpoint (JSON body `{ op, mount, path, to?, recursive? }`) for `list`/`stat`/`mkdir`/`rename`/`remove`, on **both** bridges — node-server and swift-server — and `/api/hostfs*` preflights now carry Chrome's `Access-Control-Max-Age` cap (7200 s) instead of 600 s.

Fixes #2715. Part of #2707.

## Root cause

The bridge token rides as the custom `X-Bridge-Token` header, so every cross-origin hostfs call is a non-simple CORS request, and Chrome's Private Network Access preflights public→loopback traffic regardless — a preflight per *distinct URL* is unavoidable. The bug is that the browser's preflight cache is keyed **by URL**, and the per-op routes put the mount + path in the query string, so every hostfs request hit a fresh URL and `Access-Control-Max-Age: 600` never got a cache hit. The benchmark session paid **246,893 `OPTIONS` for 385,033 `GET`s** — 64 % of reads bought an extra round trip through the same 6-connection browser queue.

## What changed

**Stable endpoint (both bridges, same handlers, same errno JSON)**
- `packages/node-server/src/hostfs.ts` — op bodies extracted into `listOp`/`statOp`/`mkdirOp`/`renameOp`/`removeOp`, shared by the per-op routes and a new `POST /api/hostfs` dispatcher. `recursive` accepts the query-string `'1'` and a JSON `true` alike.
- `packages/swift-server/Sources/Server/HostFSRoutes.swift` — the same refactor + dispatcher, so swift-server keeps byte-for-byte parity rather than relying on the client fallback.
- Every per-op route still works unchanged (older webapps, `curl`).

**Preflight lifetime**
- `preflightMaxAge()` in `bridge-security.ts` (+ its Swift mirror in `BridgeSecurity.swift`): `7200` for `/api/hostfs*`, `600` everywhere else. Scoped rather than global so the short window still bounds a stale allowed-header set on the low-volume routes.

**Client**
- `packages/webapp/src/fs/mount/backend-hostfs.ts` — `buildRequest()` picks the stable POST for the metadata ops, the per-op route otherwise. There is still exactly one `fetch` site in `request()`.

**What deliberately did NOT change**
- `read` keeps `GET /api/hostfs/read?mount=&path=` and `write` keeps `PUT …/write`. A POST response is not cacheable, and **220,310 of the session's GETs were 304s** — losing the browser HTTP cache on the 92 MB packfile would cost far more than the preflight it saves. Those URLs also *repeat* (same pack, same `.git/index`), so the preflight cache does work for them, and the 7200 s bump helps directly.
- The token mechanism and bridge auth are untouched.

**Version skew** — I mirrored the route in swift-server rather than relying on the fallback, but the fallback exists anyway for an older node-server binary behind a freshly-updated hosted UI: such a bridge answers `POST /api/hostfs` with a framework 404 carrying no errno `code`, and `HostFsMountBackend` treats exactly that as "no stable endpoint" and downgrades to the per-op routes for the rest of the mount's life. Every error the real route emits carries a `code`, so a genuine `ENOENT` never triggers the downgrade — there are tests on both sides for that.

## Tests

- `packages/node-server/tests/hostfs.test.ts` — +7 cases: stable-endpoint payload equality with the per-op routes, mkdir/rename/remove through the one URL, errno/traversal/unknown-mount mapping, coded-400 for unknown ops and a `to`-less rename, and that the dispatcher does not shadow `POST /api/hostfs/mkdir`.
- `packages/webapp/tests/fs/backend-hostfs.test.ts` — +5 cases: every metadata op lands on `POST /api/hostfs` with the right body; `read`/`write` stay on per-file URLs; the 404 fallback fires once and sticks; rename/remove params survive into the fallback query string; a coded 404 is *not* mistaken for an absent route.
- `packages/swift-server/Tests/HostFSRoutesTests.swift` — +3 cases mirroring the node-server ones, incl. "every stable-endpoint error carries a code".
- `packages/node-server/tests/bridge-cors.test.ts` / `bridge-security.test.ts` — the 7200/600 split, against a real Express pipeline and as a unit.

## Docs

`docs/mounts.md` gains a **Wire shape (`/api/hostfs`)** section (the table of request shapes, the why, and the fallback contract); `packages/node-server/CLAUDE.md` and `packages/swift-server/CLAUDE.md` updated.

## Verification

| Gate | Result |
| --- | --- |
| `npm run lint` | pass (0 errors; `Record<string, unknown>` ratchet clean — the stable body has a named `HostFsStableRequest` type) |
| `npm run test` | **18,274 passed**, 3 failed — all 3 are `slicc-fs-module.test.ts` Pyodide 5 s timeouts under load (machine load avg ~90 from sibling agents); the file passes on its own with `--testTimeout=60000` (16/16) and touches none of this code |
| `npm run test:coverage:node-server` | pass — `hostfs.ts` **91.77 % stmts / 97.14 % funcs**, all files 81.87 % |
| `npm run test:coverage:webapp` | gate red on this machine, but only from 5 s-timeout flakes in unrelated files (a *different* set each run: `git-commands` / `github-oauth-login` / `panel-rpc-device-handlers`), all `Error: Test timed out in 5000ms` under load avg ~95 from sibling agents. Re-run in isolation with `--testTimeout=60000`: **54/54 pass**. `backend-hostfs.test.ts` is 10/10 green in every run; CI's `webapp` job is the authoritative check |
| `npm run build -w @slicc/webapp` / `-w @slicc/node-server` | pass |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | `First-load OK (allowance 5 kB per change)` |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | `no debt lists found — nothing to gate` |
| `swift build` + `swift test` (swift-server) | pass — **591 tests, 0 failures** |
| `npm run lint -w @slicc/swift-server` + `lint:format` | pass (0 serious; no new violations on the touched files) |

## Coordination

Kept strictly to URL/route shape so it rebases cleanly against #2720 (retry + concurrency limiting inside `request()`): `request()` still has exactly one `fetch` call site, so their retry wrapper lands on both the stable and per-op paths.


trieloff:
Copilot's finding on `hostfs.ts:319` was right, and it was worse than "the limit is wrong" — fixed in e6077eeb4.

**Reproduced first**, against the real middleware order (global 50 MiB `express.json()` mounted at `index.ts:~1300`, `registerHostFsRoutes` at `~1379`):

```
MALFORMED  status=400  content-type=text/html   body=<!DOCTYPE html>…
OVERSIZED  status=500  {"code":"ENAMETOOLONG","message":"…realpath '/…/xxxxx…"}
```

So both halves were broken: malformed JSON came back as code-less HTML the webapp cannot turn into a faithful `FsError` (it degrades to a generic `EIO`), and a 3 MB body walked straight past the 1 MiB route-local cap into the handler, failing much later with an unrelated `ENAMETOOLONG`.

**Fix — both prongs, since either alone is insufficient:**

1. `shouldParseGlobalJson` now excludes the stable dispatcher, so the bounded 1 MiB parser is the one that runs. It lives in `fetch-proxy-headers.ts`, not `index.ts` — that module exists precisely because `index.ts` calls `main()` at module load, so a test importing it would boot a real server. (It already owned the sibling `X-Slicc-Raw-Body` exclusion, so both body-ownership rules are now in one place.)
2. `hostFsBodyErrorHandler`, registered at the end of `registerHostFsRoutes`, maps body-parser failures on the hostfs surface to the errno shape: `entity.too.large` → **413 `EFBIG`**, anything else → **400 `EINVAL`**. Exclusion alone would not have been enough — the route-local parser throws the same code-less error; the adapter is what keeps the contract. It also covers `PUT /write`'s raw parser, which had the identical gap.

**Swift parity**, since you flagged it: swift-server previously collapsed both cases into `EINVAL` (its `try?` on `body.collect` swallowed the oversize distinctly). It now answers 413 `EFBIG` for a body over the stable cap and 400 `EINVAL` for an unparseable one, matching node-server exactly, with `stableMaxBodyBytes` mirroring `HOSTFS_STABLE_MAX_BODY_BYTES`.

**Tests** drive the production middleware order (`describe('stable endpoint under the production middleware order')`) — global parser mounted first with the same predicate — and assert: coded 400 `EINVAL` for malformed JSON *with a JSON content-type*; coded 413 `EFBIG` for a body over 1 MiB but deliberately **far under** the global 50 MiB limit, so it only passes if the exclusion really handed the body to the bounded parser (the exact regression guard); a well-formed body still round-trips; and every other JSON route is still parsed globally. Plus unit coverage of `isHostFsStableBodyRequest` and a matching Swift `testStableEndpointOversizedBodyIsCodedEFBIG`.

**Verification** (re-run in full):

| Gate | Result |
| --- | --- |
| `npm run typecheck` | pass (9 projects) |
| `npm run test` | 18,281 passed, 2 failed — both `Error: Test timed out in 5000ms` in unrelated files under load avg ~90; they pass 54/54 in isolation with `--testTimeout=60000`, and CI's `webapp` job was green |
| `npm run test:coverage:node-server` | pass — `hostfs.ts` 91.47 % stmts / 97.36 % funcs |
| `npm run build -w @slicc/webapp` / `-w @slicc/node-server` | pass |
| `check-first-load-size.mjs` | `First-load OK` |
| pre-push gate | **All 7 checks passed** (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod) |
| `swift build` + `swift test` | pass (`HostFSRoutesTests` 9/9) |

CI on the previous commit was fully green, including `webapp` (6m48s) and `swift-server` (9m34s); re-running now on e6077eeb4.

Docs updated in `docs/mounts.md` and `packages/node-server/CLAUDE.md`.

PR #2725, opened 2026-09-01
fix(webapp): bound sprinkle discovery to known roots instead of walking /

### LINKED ISSUE #2717: sprinkle discovery walks the entire VFS from `/` at boot — including `--mount`ed host folders (`node_modules`, `.build`, `.git`…) over the worker RPC + hostfs
Part of #2707.

## What happens

Within a minute of boot, with no panel open and no command running, the kernel worker was serving a burst of `vfs-read-dir` RPCs that turned into `/api/hostfs/list` requests for paths like

```
node_modules/@slicc/swift-server/.build/arm64-apple-macosx/debug/index/store/v5/records/PZ
.delta/worktrees/2m4kzxjzpred/slicc/packages/swift-server/.build/repositories/swift-crypto-7e0614ea/objects/7d
```

(depth 11, reverse-alphabetical DFS, 131 `list` calls in a 94 ms window while an unrelated `ls` ran). With `postMessage` hooked before reload, the caller is:

```
remote-sprinkle-vfs.js  walk()  ← remote-vfs-client.readDir  ← wc-live
  806ms vfs-read-dir /shared/sprinkles
  807ms vfs-read-dir /
  808ms vfs-read-dir /etc
  808ms vfs-read-dir /etc/slicc
  …
```

i.e. `packages/webapp/src/ui/sprinkle-discovery.ts` `discoverSprinkles()` → "Scan everything from root" → `scanDir(fs, '/')` → `createRemoteSprinkleVfs().walk('/')` (`packages/webapp/src/kernel/remote-sprinkle-vfs.ts`, a stack-based DFS with no skip list), wired from `wc-live.ts` at boot. Every directory costs one page→worker RPC and, under a mount, one hostfs HTTP request (+ CORS preflight, #2715). On a mounted checkout with `node_modules` that is tens of thousands of directories, all queued into the same 6-connection browser pool the cone's own git/shell commands are using — this is a direct source of the "same command, wildly different time" variance from the original Slack question.

It also means the sprinkle index is built by enumerating the user's host filesystem, which a `--mount` of `~` or a big monorepo makes effectively unbounded.

## Fix ideas

- Don't walk `/` for discovery. Scan the known roots (`PRIORITY_ROOTS`, `/workspace`, `/shared`, `/scoops/*`, `/home/*`) and the mount roots only to a bounded depth, and never descend into `node_modules`, `.git`, `.build`, dot-dirs, or `/mnt`/`/proc` unless a sprinkle root is explicitly configured there.
- Use the existing FS watcher (`.shtml` watches are already registered for `/workspace`, `/shared`, `/scoops` in `kernel-worker.ts`) to maintain the index incrementally instead of re-walking.
- Do the walk in the worker (it owns the VFS and the `MountIndex`) rather than one RPC per directory from the page.
- Cap the crawl (entries/time budget) and yield to foreground commands.

trieloff:
## Summary

Sprinkle discovery no longer walks the whole VFS from `/`. It scans an explicit, bounded root set — `/shared/sprinkles` first, then `/workspace`, `/shared`, `/scoops`, `/home` — six directory levels deep, skipping `node_modules`, `dist`, `build`, `coverage` and every dot-directory, with a 500-directory ceiling per root.

## Root cause

`discoverSprinkles()` scanned `PRIORITY_ROOTS` and then "everything from root" via `scanDir(fs, '/')`. On the page side that is `createRemoteSprinkleVfs().walk('/')` — a stack DFS with no skip list — so every directory cost one page→worker RPC and, under a `--mount`ed host folder, one `/api/hostfs/list` request (plus a CORS preflight, #2715). On a mounted checkout the boot crawl descended into `node_modules/…/.build/…` at depth 11, queueing thousands of requests into the same 6-connection browser pool the cone's own git/shell commands were using. It also meant the sprinkle index was built by enumerating the user's host filesystem, which a `--mount` of `~` makes effectively unbounded.

## What changed

- **`packages/webapp/src/fs/bounded-walk.ts` (new)** — `walkBounded(reader, root, { maxDepth, skip, maxDirs })`, an iterative walk over any `readDir` + `stat` pair. Yield semantics match `VirtualFS.walk` (files and file-symlinks yielded, directories and dir-symlinks recursed, dangling symlinks and unreadable directories skipped); `skip` prunes a subtree without ever issuing a `readDir` for it, and applies to directory symlinks too. Defaults are the same `MAX_WALK_DEPTH` / `MAX_WALK_ENTRIES` ceilings `VirtualFS.walk` uses.
- **`packages/webapp/src/base/sprinkle-roots.ts` (new)** — the single `SPRINKLE_ROOTS` list. It lives in `base/` because `kernel/` may not import `ui/` (layer-back-edge gate).
- **`ui/sprinkle-discovery.ts`** — drops the `walk('/')` pass; scans `PRIORITY_ROOTS` then `SPRINKLE_ROOTS` through `walkBounded` with `maxDepth: 6`, `maxDirs: 500` and the skip predicate. `/mnt`, `/proc` and `/tmp` are not roots, so a mount is crawled only when it sits inside a configured root.
- **`kernel/remote-sprinkle-vfs.ts`** — its bespoke DFS is replaced by `walkBounded`, and `walk` now accepts `BoundedWalkOptions`, so a future caller pointing this at a mount can't reintroduce an unbounded crawl over the wire.
- **`ui/sprinkle-manager.ts` + `kernel/kernel-worker.ts`** — both `.shtml` watchers now register on the shared `SPRINKLE_ROOTS` (which adds `/home`, previously watched by neither), so what is watched and what is discovered can't drift.
- **Docs** — `docs/webapp-details.md` (Sprinkles bullet), `docs/shell-reference.md` (`sprinkle list` row) and the agent-facing `packages/vfs-root/workspace/skills/sprinkles/SKILL.md` now state where sprinkles are discovered and that `open <path>` still opens any path.

### Behavior change

A `.shtml` outside the supported roots (e.g. `/tmp/x.shtml`, `/etc/x.shtml`, or one buried past six levels / inside `node_modules`) is no longer listed by `sprinkle list` or shown in the rail. `sprinkle open <name>` for a discovered sprinkle and `open /path/to/file.shtml` for an arbitrary path are unchanged. The one existing test that asserted the old "discovers `.shtml` anywhere" contract (`/tools/monitor.shtml`) was rewritten to the new contract.

## Tests

- `packages/webapp/tests/fs/bounded-walk.test.ts` (new, 11 tests) — yield semantics, `maxDepth: 0`/`maxDepth: 1`, `skip` pruning (including that the pruned directory is never `readDir`'d, and that it applies to directory symlinks), `maxDirs`, symlink resolution, `readDir` rejection, visited-dedup.
- `packages/webapp/tests/ui/sprinkle-discovery.test.ts` — new `bounded scan (#2717)` block: finds sprinkles in `/workspace`, `/scoops` and `/home`; finds none in `/mnt`, `/tmp`, `/etc` or `/tools`; never descends into `node_modules`, `dist`, `.build` or `.git`; stops at the depth cap (depth 6 found, depth 7 not).
- `wc-sprinkles.test.ts` / `wc-shell.test.ts` — the fake VFS stubs modelled `walk`; they now model a directory tree (`readDir` + `stat`), including the boot-resilience test whose "hung VFS" is now a `readDir` that never settles.

## Verification

| gate | result |
| --- | --- |
| `npm run verify` (pre-push lint gate) | pass — all 7 checks (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod) |
| `npm run typecheck` | pass — all 9 projects |
| `npm run test` | pass — **1026 files, 18277 tests, 0 failures** (52 skipped) |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | pass |
| `node packages/dev-tools/tools/check-layer-back-edges.mjs` | pass — no new back-edges |
| `node packages/dev-tools/tools/check-doc-sizes.mjs` / `check-doc-refs.mjs` | pass |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | pass — **+0.0 kB vs base**, worker 4199 kB (89 kB under ceiling) |

Coverage of the new/changed code (v8, restricted to the touched files): `bounded-walk.ts` 100% lines / 95.3% stmts / 86.2% branches, `sprinkle-discovery.ts` 96.5% lines / 93.1% stmts / 83.0% branches, 100% functions across both.

`npm run test:coverage:webapp` does **not** complete on this machine — three tests time out only under `--coverage` instrumentation (`wc-attach.test.ts > makeSprinkleAttachImage` 10 s hook, `wc-settings.test.ts` 5 s/15 s, `wc-sprinkles.test.ts > resolves without waiting for the initial discovery` at its 1 s race, plus intermittently `slicc-fs-module` pyodide and `git-commands` rebase). I verified this is pre-existing: the identical commands fail the same way on detached `origin/main` (6fcf66c7d) in this worktree. Leaving the gate to CI.

Fixes #2717


trieloff:
Addressed in `672b50ada` — the depth-cap finding is valid, so I fixed the behaviour rather than just the wording.

## What was wrong

`walkEntry` yielded files and file-symlinks unconditionally. `pushDir` alone bounded only directories, so the children of a directory read *at* `maxDepth` escaped one level: with the discovery cap of 6, `/workspace/a/b/c/d/e/f/panel.shtml` was still listed. The old depth test placed its accepted file under five directories and never crossed that boundary.

## Fix (`packages/webapp/src/fs/bounded-walk.ts`)

- `walkEntry` rejects any entry with `depth > maxDepth` **before** yielding or recursing, so files and file-symlinks are bounded exactly like directories.
- `pushDir` now refuses a directory sitting *at* the cap (`depth >= maxDepth`): it can only hold entries past the cap, so reading it would spend a page→worker RPC — and, under a `--mount`, an `/api/hostfs/list` request — on entries the walk must immediately discard.
- `BoundedWalkOptions.maxDepth` and `MAX_SCAN_DEPTH` are documented as the deepest **entry** (file or directory) below the root, counting the root's own children as depth 1. Real sprinkles sit at depth 2–4, so nothing that was discoverable before is lost.

## Boundary tests added

`packages/webapp/tests/fs/bounded-walk.test.ts` (11 → 14 tests):
- `maxDepth: 3` accepts `/root/a/b/at-cap.txt` and rejects `/root/a/b/c/past-cap.txt`; `maxDepth: 4` accepts both.
- `maxDepth: 2` accepts `/root/a/at-cap.txt`, rejects `/root/a/b/past-cap.txt`, and asserts `/root/a/b` is never `readDir`'d.
- File symlinks are capped identically (`maxDepth: 2` yields the link, `maxDepth: 1` does not).
- `maxDepth: 0` yields nothing; `maxDepth: 1` reads only the root directory (one `readDir`).

`packages/webapp/tests/ui/sprinkle-discovery.test.ts` — the depth-cap case now pins the real boundary: `/workspace/a/b/c/d/e/at-cap.shtml` (depth 6) is discovered, `/workspace/a/b/c/d/e/f/past-cap.shtml` (depth 7) is not.

## Re-verified

| gate | result |
| --- | --- |
| `npm run lint` | pass |
| `npm run verify` (pre-push lint gate) | pass — 7/7 (biome, prettier, custom-lints, boy-scout-debt, manifest, deadcode, deadcode-prod) |
| `npm run typecheck` | pass — all 9 projects |
| `npm run test` | pass — **1026 files, 18280 tests, 0 failures** (52 skipped) |
| `check-first-load-size.mjs` | pass — **+0.0 kB vs base** (4958 kB total) |
| `check-layer-back-edges.mjs` / `check-record-string-unknown.mjs` | pass — no new entries |

`npm run test:coverage:webapp` still can't complete on this machine; the same three tests time out only under `--coverage` on detached `origin/main` in this worktree, so it's pre-existing and left to CI.

PR #2726, opened 2026-09-01
fix(hostfs): retry transient bridge failures and stop the 5 s keep-alive race

### LINKED ISSUE #2720: hostfs: one transient `Failed to fetch` inside a fan-out aborts the whole git command (`fatal: EIO: hostfs bridge unreachable`) — no retry, and the bridge's 5 s keep-alive races the client

trieloff:
## Summary

A single transient `Failed to fetch` inside a hostfs fan-out no longer kills the whole command. `git -C /mnt/slicc status --short` failed twice at ~6,000 requests against a perfectly healthy bridge; this fixes both ends of that race.

## Root cause

node-server's Express server used Node's default `keepAliveTimeout` (5 s, advertised verbatim as `Keep-Alive: timeout=5`), far below Chrome's idle-socket lifetime — so during a bursty `/api/hostfs` fan-out the browser eventually wrote a request onto a pooled socket the server was closing, and the `fetch()` rejected with `TypeError: Failed to fetch`. `HostFsMountBackend.request()` mapped any fetch rejection straight to `EIO: hostfs bridge unreachable` with no retry, and inside `statusMatrix`'s `Promise.all` that one rejection aborted the entire git command.

## What changed

**Client — `packages/webapp/src/fs/mount/backend-hostfs.ts`**

- Retries fetch *rejections* (network level only — an HTTP error is an answer from the server and is still mapped straight through) with a short backoff: 3 attempts, 25 ms then 50 ms.
- Retries only ops whose server handler is idempotent: `list`, `stat`, `read`, `mkdir` (both servers create with `recursive: true`, so a replay never reports `EEXIST`). `write` and `rename` can clobber a concurrent change, and `remove` is served with `force: false` — replaying one that did land turns into a spurious `ENOENT` — so those still surface `EIO` on the first failure. (The issue floated `remove` as "idempotent enough"; the `force: false` handler is why it isn't.)
- The surviving `EIO` names the op and calls itself transient: `hostfs stat failed after 3 attempts: transient bridge error: Failed to fetch` instead of `hostfs bridge unreachable`, which read as "the launcher is gone".
- New `packages/webapp/src/fs/mount/inflight-limiter.ts`: a counting semaphore, 24 concurrent requests per mount (constructor option `maxInflight`), so thousands of queued fetches can't pile into a 6-socket browser pool and sit next to a closing keep-alive socket.

**Server — `packages/node-server/src/http-keepalive.ts` (new) + `src/index.ts`**

- `createBridgeServer(app)` replaces the bare `createServer(app)` and applies `keepAliveTimeout = 120 s` / `headersTimeout = 130 s` before `listen()`, so no caller can construct a bridge server that still advertises `timeout=5`.
- swift-server needs no mirror: Hummingbird's `HTTP1Channel.Configuration.idleTimeout` defaults to `nil` (verified against the pinned 2.26.0 source) — it never closes an idle keep-alive connection on a timer. Documented rather than changed.

**Docs**: `docs/pitfalls.md` (new section on the keep-alive race and the two-sided fix), `docs/mounts.md` (auto-mount bullet + error-table row), `packages/node-server/CLAUDE.md` (bridge keep-alive).

## How verified

- `npm run lint` — clean (0 errors; the `main()` function-size cap is why the keep-alive wiring is a one-line `createBridgeServer(app)` rather than an extra call in `main()`).
- `npm run typecheck` — exit 0 (all 9 projects).
- `npm run test` — **1027 files passed, 18280 tests passed**, 52 skipped, 0 failures.
- `npm run test:coverage:node-server` — exit 0 (statements 81.71%, branches 72.22%, functions 85.51%, lines 83.77%).
- `npm run test:coverage:webapp` — exit 0 (statements 82.35% ≥ 82, branches 73.85% ≥ 73, functions 82.18% ≥ 82, lines 84.37% ≥ 84; `backend-hostfs.ts` itself 87.5% stmts / 89.61% lines). Three earlier attempts died on unrelated 5 s/10 s timeouts (`git-commands`, `wc-attach`, pyodide) while this machine was running three other agents' vitest suites at load average ~80; the clean run above is the one that counts, and `npm run test` passed all 18,280.
- `npm run build -w @slicc/webapp` and `npm run build -w @slicc/node-server` — both exit 0.
- `node packages/dev-tools/tools/check-first-load-size.mjs` — `First-load OK (allowance 5 kB per change; total 4957 kB across both graphs)`.
- `node packages/dev-tools/tools/check-touched-exemptions.mjs` — exit 0 (it reports `no debt lists found — nothing to gate` in this checkout; the pre-commit hook's layer-back-edge and `Record<string, unknown>` gates also passed with 0 new entries).
- `node packages/dev-tools/tools/check-doc-sizes.mjs` / `check-doc-refs.mjs` — exit 0.

New tests: `packages/node-server/tests/http-keepalive.test.ts` (5 — including a live server asserting the response really carries `Keep-Alive: timeout=120`), `packages/webapp/tests/fs/mount/inflight-limiter.test.ts` (4), and 6 new cases in `packages/webapp/tests/fs/backend-hostfs.test.ts` (retry-once-then-succeed per op, give-up message, no replay of `write`, no retry on HTTP errors, concurrency cap).

Kept to the retry/limiter wrapper and error mapping so it rebases cleanly under #2715's URL/route reshaping.

Fixes #2720


trieloff:
Fixed in `697a5cb56` (rebased onto `origin/main`).

The Codex P2 was right, and it was worse than a mapping gap: `request()` handed back the `Response` as soon as the headers landed, so `readFile`/`readDir`/`stat` read the body *outside* both the retry loop and the limiter. Two bugs fell out of that — a drop mid-body (the common shape for a large read) escaped as a raw `TypeError` with no retry and no EIO mapping, and a slow body kept a socket busy after the limiter had already counted the slot as free.

**What changed** — `request<T>()` now takes a `consume` callback and returns the materialized bytes/JSON, never a `Response`:

- The body is read inside the guarded operation, so a body-read rejection is a network failure like any other: retried for `RETRYABLE_OPS` (`list`/`stat`/`read`/`mkdir`), otherwise the transient `EIO` with the op name.
- The limiter slot is released only after `consume` resolves.
- Ops that ignore the payload (`write`/`mkdir`/`rename`/`remove`) drain it rather than leaving the socket busy.
- The `EFBIG` cap still runs in `readFile`, now on the materialized buffer.

**New tests** (`packages/webapp/tests/fs/backend-hostfs.test.ts`, 17 total):

- body rejects once → `readFile` succeeds on the retry (2 fetches);
- body rejects every time → `EIO` `hostfs list failed after 3 attempts: transient bridge error: …` (3 fetches);
- body rejects on a non-idempotent op → single attempt, `(not retried: non-idempotent op)`;
- with `maxInflight: 1`, the second `stat` is not issued until the first body resolves.

I mutation-checked the last one: moving the slot release back before `consume` fails both limiter tests, so they actually pin the behavior rather than passing by luck.

**Verification** (post-rebase): `npm run lint` clean · `npm run typecheck` exit 0 · `npm run test` 18,287 passed / 0 failed · `npm run test:coverage:webapp` exit 0 (statements 82.36% ≥ 82, branches 73.86% ≥ 73, functions 82.23% ≥ 82, lines 84.38% ≥ 84; `backend-hostfs.ts` 87.91% stmts / 89.74% lines) · `npm run build -w @slicc/webapp` exit 0 · `check-first-load-size.mjs` OK (total 4957 kB).

On #2723: the whole fetch call site still lives in exactly one place (`attemptRequest`) — URL/method construction is a single expression there, so re-pointing it at the stable `POST /api/hostfs` endpoint is a mechanical edit that doesn't touch the retry/limiter/error-mapping wrapper. Ping me when it lands if you'd rather I do the rebase myself.

Also documented in `docs/pitfalls.md` § Local Bridge Keep-Alive, so the next person doesn't reintroduce the `Response`-returning shape.

trieloff:
Rebased onto `origin/main` with #2723 in it — new head `01b75f285`. The two commits are folded into one so the conflict only had to be resolved once.

**How the two changes compose.** #2723's routing is untouched: `buildRequest()` still picks the stable `POST /api/hostfs` for `list`/`stat`/`mkdir`/`rename`/`remove`, `read`/`write` still use their per-op routes, and a code-less 404/405 from the stable endpoint still flips `stableEndpoint = false` permanently and re-runs the same op on its per-op route. My wrapper sits outside it:

```
request()          → assertOpen + limiter slot
  attemptRequest() → the retry loop (3 attempts for RETRYABLE_OPS, 25/50 ms)
    routeAndConsume() → buildRequest → fetch → stable-endpoint downgrade → consume(body)
```

The downgrade re-enters `routeAndConsume`, **not** `attemptRequest`, so switching routes is a routing correction that does not spend a retry attempt — and because it still runs inside the caller's attempt, a network failure on the fallback route is mapped and retried exactly like any other. Recursion is bounded: the downgrade clears `stableEndpoint`, so the re-entered call can't take the stable branch again.

`errorFromResponse()` now returns `{ error, rawCode }` — same single body parse as main, just handing the raw `code` back so the downgrade check and the errno mapping share it instead of duplicating the read.

**Tests**: 24 in `backend-hostfs.test.ts` — all 10 from main unchanged (stable URL for metadata ops, read/write staying per-op, fallback + param carry-through, coded-404 is not an absent route) plus my 14, including two new ones for the seam:

- `spends no retry attempt on the stable-endpoint downgrade` — stable 404, then the per-op route fails twice and succeeds on the third: `POST` + 3 × `GET` and it resolves.
- `maps a fallback-route network failure through the same transient EIO` — 4 calls, `hostfs stat failed after 3 attempts: transient bridge error: …`.

I mutation-checked the seam: making the downgrade throw so it re-enters `attemptRequest` (i.e. spending an attempt) fails both of those *and* main's own `carries rename/remove params through to the fallback query string` — so the accounting is pinned from both sides.

**Verification** (post-rebase): `npm run lint` clean · `npm run typecheck` exit 0 (9 projects) · `npm run test` **18,322 passed / 0 failed** · `npm run test:coverage:webapp` exit 0 (statements 82.39% ≥ 82, branches 73.87% ≥ 73, functions 82.35% ≥ 82, lines 84.41% ≥ 84; `backend-hostfs.ts` 90.82% stmts / 92.63% lines) · `npm run build -w @slicc/webapp` and `-w @slicc/node-server` exit 0 · `check-first-load-size.mjs` OK (4958 kB total).

Issue #2712, opened 2026-09-01
git: `log --all` takes 150+ s — `objects/pack` listed once per object, loose-object miss before pack lookup, fresh pack cache per branch


trieloff:
Part of #2707.

## What happens

`git -C /mnt/slicc log --oneline --all -n 20` (29 local branches): **157 s cold, 151 s warm, 72,767 hostfs requests** (25,005 `list`, 46,696 `read`, 1,066 `stat`), 85–94 MB transferred. Native: 51 ms.

Even `log --oneline -n 20` on one branch shows the pattern: 90 requests, of which **28 are `list .git/objects/pack`** and most reads are `.git/objects/xx/yyyy` → `404` (loose-object probe) before the packed lookup.

## Why

1. `packages/webapp/src/git/commands/log.ts` `logAllBranches()` calls `git.log({ ref: branch, depth: maxCount ?? 50 })` **once per branch** — 29 separate isomorphic-git calls, each with its own implicit `cache`, so the 30 `.idx` files are re-read and re-parsed 29 times (`GitPackIndex.fromIdx` = 11.3 s CPU in the profile) and the 92 MB pack is re-hashed (see #2710). The requested `-n 20` is applied only after collecting `29 × 50` commits.
2. isomorphic-git `_readObject` tries `readObjectLoose` first (one `read` → 404 on hostfs) and only then `readObjectPacked`, which does `fs.readdir(gitdir/objects/pack)` **on every call** (no cache for the listing) → 1 list + 1 failed read per object read.
3. Each `git.log` step reads the commit object, so a 50-deep walk on a branch = ~150 requests before any tree is touched.

## Fix ideas

- Share one `cache` across the per-branch calls (and across commands — #2710).
- Walk all tips in one traversal (priority queue by author date across `listBranches()` tips, stopping at `maxCount`) instead of 29 depth-50 walks + dedupe + sort.
- Command-scoped memo for `readdir(objects/pack)` and for "does `objects/xx/` exist" so the loose probe is skipped when the fan-out dir is absent (the adapter can cache ENOENT for a directory for the lifetime of the command).

Issue #2710, opened 2026-09-01
git: every command re-downloads and SHA-1-verifies the whole packfile; no cross-command object/pack cache


trieloff:
Part of #2707.

## What happens

Any git command that reads an object from a pack (`log`, `show`, `ls-files`, `status`, `diff`…) makes isomorphic-git read the **entire** `.pack` (`readObjectPacked`: `p.pack = fs.read(packFile)`) and then run its "deep integrity check" — a SHA-1 over the whole payload — before serving the first object. Because `packages/webapp/src/git/commands/*` never pass a `cache` object, that happens again on the **next command** (and, for `git log --all`, once per branch inside one command — see #2712).

On the slicc checkout (largest pack 92 MB, 30 packs, 3.4 MB of `.idx`):

- warm `ls-files`: **47 % of kernel-worker CPU (5.2 s) in native `digest`** — hashing the 92 MB pack; the pack body itself comes back as a `304` from the browser HTTP cache but is still re-hashed.
- `log --all -n 20`: 14.1 s in `digest` + 11.3 s in `GitPackIndex.fromIdx` (re-parsing the 30 indexes for each of 29 branches).
- cold `ls-files`: the single biggest read was the 92 MB pack (92,914,073 bytes over the loopback bridge).

## Why

- `readObjectPacked` (node_modules/isomorphic-git/index.js) — `if (!p._checksumVerified) { … shasumRange(pack, 0, len-20) … }`; the flag lives on the cached pack index, and the cache is per-API-call because SLICC creates none.
- `readPackIndex` keeps parsed `.idx` files in `cache[PackfileCache]`; same lifetime problem.

## Fix ideas

- Keep one `cache` object per gitdir in `GitCommands` (`packages/webapp/src/git/git-commands.ts`) and pass it to every isomorphic-git call; invalidate when `.git/objects/pack` listing or `packed-refs` mtime changes (cheap: one `list` + a couple of stats).
- Skip or make optional the deep integrity SHA-1 (it is a debug safeguard; system git only verifies packs on `fsck`/`index-pack`). If upstream has no switch, patch it out for hostfs/opfs-backed repos or verify once and persist the verified pack SHA in the cache.
- Longer term: range reads (`Range:` on `/api/hostfs/read`) so a pack is not materialized in worker memory at all — also required by #2711.

trieloff:
[...]

PR #2727, opened 2026-09-01
perf(git): bound `log -- <path>` instead of walking the whole history

### LINKED ISSUE #2714: git: `log -- <path>` walks the entire history and tree-diffs every commit before applying `-n`
Part of #2707.

## What happens

`git -C /mnt/slicc log --oneline -n 5 -- packages/webapp/src/git/git-commands.ts` on a branch with 9,534 commits: **the leader tab's renderer crashed** (`Inspector.targetCrashed` in the node-server log) after ~4.5 minutes and ~25,500 hostfs requests — the command never produced output. The kernel worker and every panel died with it; the tab had to be re-navigated. Native: 41 ms.

## Why

`packages/webapp/src/git/commands/log.ts` `selectCommits()`:

- when `pathspecs.length > 0` the `depth` option is deliberately omitted, so `git.log` materializes **every commit on the branch** (each = a loose-probe `read` + `list objects/pack` + object read, see #2712);
- then `commitTouchesPathspec()` runs a two-tree `git.walk` (parent vs commit) **per commit** via `Promise.all` over all commits at once — thousands of concurrent tree walks over hostfs;
- only after all of that does `commits.slice(0, maxCount ?? 10)` apply.

## Fix ideas

- Use isomorphic-git's own `git.log({ filepath, depth, force })` when the pathspec is a single file/dir path (it walks commit-by-commit and stops at `depth`); `--follow` already uses it.
- Otherwise stream: iterate `git.log` lazily (or in depth batches), test `commitTouchesPathspec` per commit sequentially, and stop as soon as `maxCount` matches are found.
- Compare the pathspec's tree entry oid between parent and commit (one tree lookup per path component) instead of walking whole trees.

trieloff:
## Summary

`git log -- <path>` in the browser-side git walked the **entire** branch history and tree-diffed every commit before `-n` was applied. On the slicc repo mounted over the hostfs bridge (9,534 commits) that meant ~25,500 bridge requests and a **crashed leader-tab renderer** after ~4.5 minutes — the command never produced output. Native git answers the same query in 41 ms.

This bounds the walk before it starts.

## Root cause

`packages/webapp/src/git/commands/log.ts` `selectCommits()`:

1. `depth` was deliberately omitted whenever `pathspecs.length > 0`, so `git.log` materialized every commit on the branch (each one a loose-probe read + `list objects/pack` + object read).
2. `commitTouchesPathspec()` then ran a two-tree `git.walk` (parent vs commit) **per commit** through `Promise.all` over *all* commits at once — thousands of concurrent whole-tree walks into the browser's six-connection queue.
3. Only after all of that did `commits.slice(0, maxCount ?? 10)` run.

## What changed

- **Fast path** — `log [<ref>] -- <path>…` (no `--all`, no `a..b`, no `--follow`) now goes straight to isomorphic-git's own `git.log({ filepath, depth, force: true })`. Its walk resolves the path component-wise in each commit tree and **breaks as soon as `depth` matching commits are found**. `force: true` keeps a path that is absent or deleted at the tip from throwing, preserving `log -- does-not-exist` → empty output. Multiple pathspecs are the union of those bounded walks, merged by oid and re-sorted by author time.
- **Lazy filtering everywhere else** — `--all` and `a..b` still need a materialized list, but they now filter it **sequentially** and `break` at the limit. No fan-out, ever.
- **Cheap `commitTouchesPathspec()`** — compares the pathspec's tree-entry oid between the commit and its first parent, resolving the path with one `git.readTree` per component and memoizing `(tree oid, pathspec)` and `commit oid → tree oid` across the walk. A directory's tree oid changes exactly when something below it changes, so a subtree pathspec needs no recursion. One isomorphic-git `cache` object is now shared by every `git.log`/`readTree`/`readCommit` call within a single `log` invocation.

`--all`, ranges, `--author`, `--grep`, `--reverse`, `--stat`, `--follow` and `--format` all keep working (covered by the new tests).

### Two behavior fixes fall out, both verified against system git

- A **directory pathspec now picks up commits that ADD a file** under it. The old whole-tree walk missed those: `git log -- src` omitted the commit that created `src/other.ts`.
- **`log -- .` now means "every commit"** instead of nothing. It previously fell through `matchesPathspec`'s exact-or-prefix test and matched nothing; it also would have driven the new fast path into a full-history walk for a path that never resolves. (`diff`/`status` still share the older `matchesPathspec` quirk for a bare `.` — out of scope here.)

Not touched: pack-cache / object-lookup (#2710, #2712 are separate PRs).

## Tests

New `packages/webapp/tests/git/git-log-pathspec.test.ts` — 13 tests on an in-memory repo (`git init` + commits touching different files):

- exact-file, directory (plain / trailing slash / `./`-prefixed), multi-pathspec union, nonexistent path, deletion, `.`
- `-n` applied to *matching* commits, newest first
- parity for `--reverse`, `--author`, `--grep`, `--stat`, `a..b`, `--all`, `--follow`
- **early-stop guard**: a counting `vfs.readFile` spy over a 24-commit history where only the tip touches the pathspec. `log -n 2 -- src/target.ts` costs **19 reads** vs **79** for the unbounded query, and the assertion is `bounded * 2 < full`.

The guard was reverted-checked: against `origin/main`'s `log.ts` the same test reports **487 reads for both** the bounded and the unbounded query (no early stop at all) and 4 of the 13 tests fail — so it genuinely pins the fix rather than passing by construction.

## Verification

| gate | result |
| --- | --- |
| `npm run lint` | **pass** — exit 0, 2,625 files checked (43 pre-existing plugin warnings) |
| `npm run typecheck` | **pass** — all 9 projects clean |
| `npm run test` | **18,273 passed**, 52 skipped; 3 unrelated 5 s-timeout flakes under machine load, each green in isolation |
| `npm run test:coverage:webapp` | **pass** — exit 0, 14,645 passed / 42 skipped, statements 82.38 %, branches 73.87 %, functions 82.30 %, lines 84.39 % (`log.ts` 86.62 / 68.33 / 100 / 90.90) |
| `npm run build -w @slicc/webapp` | **pass** — exit 0 |
| `check-first-load-size.mjs` | **pass** — "First-load OK (allowance 5 kB per change; total 4959 kB)" |
| `check-touched-exemptions.mjs` | **pass** — exit 0 (`log.ts` is on no debt list) |
| `check-doc-refs.mjs` | **pass** — no dead references in 73 doc files |

Selection semantics were also diffed against **system git** in a scratch repo with the same shape: `log -- src`, `log -- .`, `log -- src/other.ts docs` and `log -n 5 -- src/target.ts` all match exactly.

## Docs

`docs/webapp-details.md` gains a **Git** section: cost is measured in VFS reads (an HTTP round trip each over a `--mount`ed repo), a history walk must be bounded *before* it starts and never fanned out, and "did this commit touch `<path>`?" is a tree-entry oid comparison rather than a tree walk.

Not verified on a live 9.5k-commit mounted repo — that needs the benchmark harness from #2707; the reasoning and the read-count guard are the evidence here.

Fixes #2714

### DIFF
--- a/docs/webapp-details.md
+++ b/docs/webapp-details.md
@@ -27,6 +27,12 @@ Overflow from `packages/webapp/CLAUDE.md`. Each section is the deep reference fo
 - Path: `packages/webapp/src/shell/`. `almost-bash-shell.ts` is the just-bash runtime; `supplemental-commands/` built-ins live under `docs/shell-reference.md`. `script-catalog.ts` is the shared `.jsh`/`.bsh` discovery for the shell and `which`, cached per `$PATH` root set; the `FsWatcher` cache is bypassed only for root sets a mount overlaps (external changes there are invisible). `vfs-adapter.ts` bridges shell → VFS and forwards `canWrite` (duck-typed for `VirtualFS`/`RestrictedFS`).
 - `typescript` v7 (native) runs checks/builds; `typescript-js` (JS v6) powers browser `tsc`/`test`/`esm-transpile` because v7 has no browser/WASM API. `builtin-shadow-map.ts` is authoritative for `ipx`/`npx` → built-in redirects. Raw scans: `jsh-discovery.ts` / `bsh-discovery.ts`.
 
+## Git
+
+- Path: `packages/webapp/src/git/`; one module per subcommand under `commands/`, dispatched by `git-commands.ts`. isomorphic-git runs over `vfs-fs-adapter.ts`, so **every object read is a VFS read** — and over a `--mount`ed host repo (`docs/mounts.md`) that is an HTTP round trip through the hostfs bridge. Cost is measured in reads, not in CPU.
+- **A history walk must be bounded before it starts, never after** (#2714). `log -- <path>` used to materialize every commit on the branch, tree-diff each one through a `Promise.all` fan-out of two-tree `git.walk`s, and only then apply `-n`; on a 9.5k-commit repo that was ~25,500 bridge requests and a crashed renderer. `commands/log.ts` now hands a single-ref pathspec straight to `git.log({ filepath, depth, force: true })`, whose own walk resolves the path component-wise per commit tree and stops at `depth` matches. The forms that cannot use it (`--all`, `a..b`) filter the already-materialized list SEQUENTIALLY and break at the limit — never fan thousands of tree reads into the six-connection browser queue at once.
+- **Compare tree-entry oids, not whole trees.** "Did this commit touch `<path>`?" is `readTree` per path component in the commit and in its first parent, memoized per `(tree oid, pathspec)` across the walk. A directory's tree oid changes exactly when something below it changes, so a subtree pathspec needs no recursion. Pathspecs are exact-path-or-directory-prefix (`matchesPathspec` in `commands/revision.ts`), never globs.
+
 ## Speech
 
 - Path: `packages/webapp/src/speech/`; entry `supplemental-commands/hear-command.ts`. **Page realm only** (mic, AudioContext); kernel worker bridges via `hear-*` panel-RPC.
--- a/packages/webapp/src/git/commands/log.ts
+++ b/packages/webapp/src/git/commands/log.ts
@@ -3,7 +3,6 @@
 import * as git from 'isomorphic-git';
 import { parseArgs } from '../../shell/arg-parser.js';
 import { diffCommits, diffInitialCommit } from './diff.js';
-import { matchesPathspec } from './revision.js';
 import { flagString, GIT_FLAG_SPECS, type GitParsedFlags } from './shared.js';
 import type { GitCommandContext, GitCommandResult } from './types.js';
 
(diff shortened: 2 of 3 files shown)


trieloff:
Fixed in d2662a2f4 — the Codex P2 on multi-pathspec ordering was correct.

**The defect.** For several pathspecs I unioned the per-spec `git.log({ filepath })` walks and re-sorted by author timestamp. Timestamps are not monotonic across history, and same-second commits keep insertion order under a stable sort — which is pathspec-argument order. So `log -- a b` could order (and with `-n`, select) differently from `log -- b a`.

**The fix.** Several pathspecs now walk the ref **once** and keep a commit when *any* spec's tree-entry oid differs from the first parent's, reusing `pathEntryOid` / `commitTouchesPathspec` and the per-command memo tables. Results come out in the ref's actual walk order, and no sort is involved at all.

The walk is still not materialized up front: `logPathspecsOrdered` takes a growing window (32, then doubling), filters only the newly-revealed suffix each round, and returns as soon as `maxCount` matches are found or `git.log` answers short (history exhausted). An `-n` satisfied near the tip still never pays for the rest of the branch, and the worst case reads ~2x the window that actually answered the query rather than the whole history with a fan-out.

The **single-pathspec** isomorphic-git `filepath` fast path is unchanged — that is the path the issue's crash report exercises, and its walk is already in history order.

**New tests** (`git-log-pathspec.test.ts`, now 15):

- `orders multiple pathspecs by history, independent of argument order` — three commits written with the *same* explicit author timestamp (`isoGit.commit({ author: { timestamp: 1_700_000_000 } })`) where the newer commit touches `b.txt` and the older touches `a.txt`. Asserts `log -- a.txt b.txt` === `log -- b.txt a.txt` === `[c2, c1]`, and that `-n 1` picks `c2` either way.
- `grows the history window until -n multi-pathspec matches are found` — both matches sit 40 commits below the tip, past the first 32-commit window, so the doubling actually has to run.

Mutation-checked: with the union+timestamp-sort restored, the ordering test fails exactly as predicted — `expected [ 'c1: touches a', 'c2: touches b' ] to deeply equal [ 'c2: touches b', 'c1: touches a' ]`.

**Verification on d2662a2f4**

| gate | result |
| --- | --- |
| `npm run typecheck` | pass (exit 0, all 9 projects) |
| `packages/webapp/tests/git/` | **389 passed**, 16 files |
| `npm run test:coverage:webapp` | **exit 0** — 14,647 passed / 42 skipped; stmts 82.39 %, branches 73.86 %, funcs 82.30 %, lines 84.40 %; `log.ts` 95.05 / 79.03 / 100 / 96.07 (up from 86.62 / 68.33 / 100 / 90.90) |
| `npm run build -w @slicc/webapp` | pass (exit 0) |
| `check-first-load-size.mjs` | pass — total 4959 kB, unchanged |
| `check-touched-exemptions.mjs` | pass (exit 0) |

One caveat worth recording, unchanged by this commit: the single-spec fast path uses isomorphic-git's own filepath walk, which compares each commit against the **previous commit in walk order** rather than against its first parent, while the multi-spec path compares against the first parent. On a linear history the two agree exactly; on merge commits they can differ slightly (and both differ from `git log --full-history`). That is pre-existing isomorphic-git behavior already relied on by `--follow`, so I left it alone rather than widen this PR.

Issue #2709, opened 2026-09-01
git: `statusMatrix` costs ~5 round trips per file — index re-stat per file, `.gitignore` re-read per candidate, no command-scoped cache


trieloff:
Part of #2707.

## What happens

Warm `git -C /mnt/slicc ls-files` (index already refreshed, no writes) on a 3,549-file checkout still takes **11.1 s and 16,336 hostfs requests** (9,858 stat, 4,780 read, 1,698 list). Native: 32 ms. `git status` uses the same `statusMatrix` walk. `git status` (long form) on the same tree: 17.3 s cold / 12.3 s warm, 20,814 requests warm (9,859 stat, 6,173 list, 4,782 read).

Request log of the cold run (30,276 requests), most repeated paths:

| path | op | count |
| --- | --- | ---: |
| `.git/index` | stat | **10,513** |
| `.git/info/exclude` | stat + read | 1,194 + 1,194 |
| `.gitignore` | read | **1,195** |
| `node_modules/.gitignore` | read | 754 |
| `packages/.gitignore` | read | 346 |
| `packages/shared-ts/.gitignore` | read | 93 |

Distinct stat paths: 5,122. So roughly 2/3 of all requests are re-stats/re-reads of the same handful of files.

## Why (isomorphic-git 1.41.9, `node_modules/isomorphic-git/index.js`)

- `GitWalkerFs.oid()` calls `GitIndexManager.acquire()` **per file**; `acquire` calls `isIndexStale()` which does `fs.lstat(gitdir/index)` every time → one `/api/hostfs/stat` per tracked file (warm runs: exactly 3,543 `stat .git/index`; the cold run's 10,513 adds two more per file from the index write-back path, #2708).
- `GitIgnoreManager.isIgnored()` is called for every untracked candidate and re-reads `.git/info/exclude` (after an `exists` → stat) and every ancestor `.gitignore` on each call; nothing is cached.
- SLICC never passes a `cache` object to isomorphic-git (only `clone.ts` does), and the `PromiseFsClient` built in `packages/webapp/src/git/vfs-fs-adapter.ts` has no memoization, so every lookup is a live HTTP round trip on hostfs (`backend-hostfs.ts`: "no client-side cache … every operation is a live passthrough").
- The walk's default `iterate` is `Promise.all` over all children, so thousands of requests are queued into the browser's 6-connection-per-host limit; measured latency inside the fan-out p50 2.7 ms → p99 1.6 s.

## Fix ideas

[...]

Issue #2708, opened 2026-09-01
git: read-only commands rewrite the host repo's `.git/index` thousands of times through hostfs


trieloff:
Part of #2707.

## What happens

`git ls-files` (and `git status`, both built on `git.statusMatrix`) against a `--mount`ed host checkout issued **3,485 `PUT /api/hostfs/write` requests, all for `.git/index`** (437 KB each), in a single invocation. The host repo's index was rewritten by the browser: checksum changed, size 437,269 → 423,544 (isomorphic-git writes a bare v2 index and drops the extensions system git had written), and the file is written with a plain non-atomic `writeFile` on the server side (`packages/node-server/src/hostfs.ts` `/api/hostfs/write`), so an interrupted write leaves a truncated index.

Measured (slicc main checkout, 3,549 tracked files): cold `ls-files` **117 s / 30,276 requests / 238 MB**; the same command warm (after the rewrite) 11 s. Native `git ls-files`: 32 ms.

## Why

1. `statusMatrix` defaults to `refresh: true`. In `GitWalkerFs.oid()` (isomorphic-git 1.41.9), when `compareStats(stats, stage)` reports staleness the file is re-hashed and, if the oid still matches, `index.insert(...)` is called to "update the stats so we get a cache hit next time" — which sets `_dirty` and makes `GitIndexManager.acquire` write the whole index back. `acquire` is called **once per file**, so the index is serialized and PUT once per tracked file.
2. The stats can never match on hostfs: `HostFsMountBackend.stat` returns only `{kind,size,mtime}`, and `VirtualFS.statInner` sets `ctime = mtime`; the isomorphic-git adapter (`packages/webapp/src/git/vfs-fs-adapter.ts` `toStats`) fills `ino: 0, uid: 1, gid: 1` and a constant mode `100644`. System git's index has the real ctime/ino/uid/gid/mode, so `compareStats` is stale for every file → every file is read and SHA-1'd and the index is rewritten with the bogus stats. After that the host's own git sees every entry as stale and rehashes the tree too.
3. Nothing in `packages/webapp/src/git/commands/{status,ls-files}.ts` passes `refresh: false` for read-only commands.

## Cost

CPU profile of the cold run: 33 % inclusive in `GitIndex.toObject → _entryToBuffer → Buffer (safe-buffer polyfill)` — ~39 s of the 117 s is spent serializing the index 3,485 times; the rest is waiting on the bridge.

## Fix ideas

[...]

PR #2728, opened 2026-09-01
perf(webapp): stop git diff reading every tracked blob

### LINKED ISSUE #2719: git: `git diff` reads every tracked blob out of the packfile and re-reads every workdir file — did not finish in 10 minutes (172,970 requests)
Part of #2707.

## What happens

`git -C /mnt/slicc diff` (no arguments; 3,549 tracked files, 2 untracked, nothing modified) was still running when the benchmark's 600 s cap hit: **172,970 hostfs requests — 108,355 `list .git/objects/pack`, 61,288 `read`, 3,327 `stat` — 187 MB transferred, ~220 requests/s sustained, renderer at 1.9 GB RSS.** The correct output is empty. Native `git diff`: ~30 ms.

## Why

`packages/webapp/src/git/commands/diff.ts` `diffWorkdirChanges()`:

1. walks `STAGE()` and collects `oid()` for every index entry (one `GitIndexManager.acquire` + index `stat` per file — #2709);
2. then, for **every** index entry, unconditionally `readBlobText(stageOid)` — an object read from the pack (loose probe → 404, `list objects/pack`, index lookup, delta resolution; each delta hop calls `_readObject` again and re-lists the pack dir — #2712) — **and** `ctx.fs.readTextFile()` of the workdir file, and compares the strings.

There is no stat/size/mtime short-circuit, no use of `statusMatrix`'s "unmodified" verdict, and no batching; the loop is sequential so it's also latency-bound (round trips × 2 per file plus the pack lookups).

## Fix ideas

- Compute the candidate set first with `statusMatrix` (or the walk's `compareStats`) and only read blobs for entries whose workdir oid ≠ stage oid (isomorphic-git already computes `workdir.oid()` with the index stat cache).
- Read the blob only after the workdir read, and only if the workdir SHA-1 differs from `stageOid` (hash the workdir bytes; skip the pack entirely for unmodified files).
- Same pattern in `diffStagedChanges` / `diffCommitWorkdir` — audit for "read every blob" loops.

trieloff:
## Summary

`git -C /mnt/slicc diff` on a mounted host repo (3,549 tracked files, nothing
modified) never finished: **172,970 hostfs requests** — 108,355 `list
.git/objects/pack`, 61,288 `read` — 187 MB transferred, still running when the
benchmark's 600 s cap hit. The correct output is empty; native `git diff` takes
~30 ms. This makes the workdir/index comparison decide what changed from OIDs
before it reads any content, so a clean tree reads **zero blobs**.

## Root cause

`diffWorkdirChanges()` walked `STAGE()` collecting every index OID and then, for
**every** entry, unconditionally `readBlobText(stageOid)` (a pack object read:
loose probe → 404, `list objects/pack`, index lookup, delta resolution) **and**
`ctx.fs.readTextFile()` of the workdir file, comparing the decoded strings. No
stat/OID short-circuit, no pruning, sequential — so two round trips per tracked
file plus the pack lookups behind each one.

## What changed

`packages/webapp/src/git/commands/diff.ts`:

- **Index vs workdir** (`git diff`) now walks `STAGE()` + `WORKDIR({ refresh:
  false })` together. The workdir bytes are read **once** and hashed with
  `git.hashBlob`; the object store is touched only when that hash differs from
  the index OID. Unmodified files cost no object read at all. Hashing (rather
  than trusting the index stat cache) keeps the result byte-exact — a same-size,
  same-mtime edit still shows up.
- **Untracked subtrees are pruned** (`map` returns `null`), so `node_modules`
  and `.git` are never enumerated. `git diff HEAD` previously walked the entire
  `.git` directory.
- **`refresh: false`** on both workdir walks: a read-only diff must not rewrite
  the host repo's `.git/index` (#2708).
- **Pathspecs are applied before any I/O**, so `git diff -- <path>` costs what
  that path costs.
- `diff <rev> <rev>` **prunes subtrees that share a tree OID** instead of
  descending into identical directories.
- `--staged` / `diff --cached <rev>` already short-circuited on OID equality
  (verified by a new test); they now share one per-invocation isomorphic-git
  `cache` object with their blob reads instead of re-parsing pack indexes.
- Incidental fix: a tracked `.gitignore` (any path starting with `.git`) was
  skipped by both the workdir and the staged diff. Only `.git` itself is skipped
  now.

Unified-diff / `--stat` / `--name-only` rendering, deletion handling and the
binary decode path are unchanged; modified-file output is byte-identical. The
workdir walk visits siblings concurrently, so the collected changes are now
sorted by path (previously insertion order).

## Tests

New `packages/webapp/tests/git/git-diff-object-reads.test.ts` (13 tests) counts
filesystem work on an in-memory repo instead of only checking rendered output —
reads are attributed by path and blob reads identified by hashing known
contents, so legitimate tree reads can't mask a blob read:

- nothing modified → **zero** object reads (`diff`), zero blob reads
  (`--staged`, `diff HEAD`)
- one modified file → **exactly one** blob read, and it is that file's
- pathspec-excluded files are never read; untracked trees are never descended
- identical subtrees are not read when diffing two commits
- deletions, `.gitignore`, path ordering, staged-file-in-a-new-directory,
  modified `diff HEAD`

Reverting `diff.ts` to the previous implementation fails 5 of the 13.

Docs: new `### git diff never reads an unchanged blob` section in
`docs/shell-reference.md` (git subset), stating the OID-before-content invariant
for future diff variants.

## Verification

| command | result |
| --- | --- |
| `npm run lint` | pass (exit 0, 43 pre-existing warnings) |
| `npm run typecheck` | pass (exit 0, all 9 projects) |
| `npm run test` | 18,269 passed / 1 failed — `slicc-fs-module.test.ts` pyodide load times out at the 5 s default on this machine (passes with `--testTimeout=60000`); `wc-attach.test.ts` hook timeout under parallel load, passes in isolation. Neither imports `src/git/`. |
| `npx vitest run packages/webapp/tests/git` | 387 passed / 16 files |
| `npm run test:coverage:webapp` | pass (exit 0) — lines 84.34/84, statements 82.33/82, functions 82.22/82, branches 73.85/73; `diff.ts` itself 94.3% lines / 96.6% functions. Needed `--testTimeout=30000 --retry=2` locally: this machine runs far more workers than CI and times out unrelated suites (pyodide load, a `wc-attach` dynamic import). |
| `npm run build -w @slicc/webapp` | pass (exit 0); `check-first-load-size.mjs` OK (4959 kB across both graphs) |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | pass (no debt lists) |
| `node packages/dev-tools/tools/check-doc-sizes.mjs` / `check-doc-refs.mjs` | pass |

## Left out

The remaining per-file cost is one workdir `lstat` + one read per tracked file
(~3.5k requests instead of ~173k). Eliding the read for stat-cache-clean files
is what `statusMatrix`/`WORKDIR().oid()` would buy, but it inherits git's racy
timestamp semantics; kept content-exact here on purpose. The pack-lookup
amplification behind each object read (#2712), the per-file index overhead
(#2709) and the CORS preflights (#2715) are separate issues with their own PRs.

Fixes #2719

### DIFF
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -209,15 +209,18 @@ what changed from OIDs before it reads any content (#2719):
 - **Tree vs index / tree vs tree** (`--staged`, `<rev> <rev>`): entries whose OIDs
   match are skipped, and two subtrees sharing a tree OID are pruned from the walk
   instead of being descended.
-- **Untracked subtrees are pruned**, so `node_modules` and `.git` are never
-  enumerated by a diff walk.
+- **Untracked and pathspec-excluded subtrees are pruned in the walk's
+  `iterate` hook**, not in `map`. `map` runs _after_ isomorphic-git has already
+  `readdir`/`lstat`ed the entry, so pruning there still costs one round trip
+  per tracked path; filtering in `iterate` works from the path string alone and
+  costs nothing. `node_modules` and `.git` are never enumerated, and
+  `git diff -- one/file.txt` never stats a sibling directory.
 - Workdir walks pass `refresh: false` — a read-only command must not rewrite
   `.git/index` (#2708).
 
-Pathspecs are applied before any read, so `git diff -- <path>` costs what that path
-costs. When adding a diff variant, keep the OID comparison ahead of the content
-read; the pre-2719 code compared decoded strings and pulled all 3,549 tracked blobs
-out of the packfile on every invocation.
+When adding a diff variant, keep the OID comparison ahead of the content read and
+the pathspec test ahead of the walk: the pre-2719 code compared decoded strings
+and pulled all 3,549 tracked blobs out of the packfile on every invocation.
 
 ### `ipk install -g` / `npm install -g`
 
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -197,6 +197,28 @@ every `.git/objects/pack/*.idx` searching for a prefix match, so trying it first
 `git rev-parse HEAD` over a `--mount`ed host repo read all 30 pack indexes (3.8 MB, 34 bridge
 requests) instead of `HEAD` plus `packed-refs` (issue #2713).
 
+### `git diff` never reads an unchanged blob
+
+Every git command runs against the VFS, so on a `--mount`ed host repo each object
+read is an HTTP round trip through the hostfs bridge. `git diff` therefore decides
+what changed from OIDs before it reads any content (#2719):
+
+- **Index vs workdir** (`git diff`): the workdir bytes of a tracked file are read
+  once and hashed with `hashBlob`; the object store is touched only when that hash
+  differs from the index OID. A clean tree reads zero blobs.
+- **Tree vs index / tree vs tree** (`--staged`, `<rev> <rev>`): entries whose OIDs
+  match are skipped, and two subtrees sharing a tree OID are pruned from the walk
+  instead of being descended.
+- **Untracked subtrees are pruned**, so `node_modules` and `.git` are never
+  enumerated by a diff walk.
+- Workdir walks pass `refresh: false` — a read-only command must not rewrite
+  `.git/index` (#2708).
+
+Pathspecs are applied before any read, so `git diff -- <path>` costs what that path
+costs. When adding a diff variant, keep the OID comparison ahead of the content
+read; the pre-2719 code compared decoded strings and pulled all 3,549 tracked blobs
+out of the packfile on every invocation.
+
 ### `ipk install -g` / `npm install -g`
 
 `ipk install -g <pkg>` (and `npm install -g`, `npm i -g`) installs into the shared
--- a/packages/webapp/src/git/commands/diff.ts
+++ b/packages/webapp/src/git/commands/diff.ts
@@ -9,11 +9,45 @@
 import * as git from 'isomorphic-git';
 import { parseArgs } from '../../shell/arg-parser.js';
 import { diffStat, unifiedDiff } from '../diff.js';
-import { matchesPathspec, resolveRevision } from './revision.js';
+import { matchesPathspec, pathspecCouldMatch, resolveRevision } from './revision.js';
 import { GIT_FLAG_SPECS } from './shared.js';
 import type { GitCommandContext, GitCommandResult } from './types.js';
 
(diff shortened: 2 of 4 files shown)


trieloff:
Fixed in 5b8f79482 — the P2 was correct, and the pruning was in the wrong place entirely.

`walk()` calls `readdir()` on every child (an `lstat` for the WORKDIR walker, one hostfs round trip) and only *then* calls `map`, so returning `null` from `map` only ever pruned the level below. `git diff -- one/file.txt` still stat'd every tracked entry and listed every top-level untracked directory.

The filter now runs in the walk's `iterate` hook, from the path string alone, before any entry is constructed:

- `pathspecCouldMatch()` (new, next to `matchesPathspec` in `revision.ts`) keeps a path that matches a pathspec, sits under one, or is an ancestor directory on the way to one — the weaker test a subtree prune needs. `map` still applies the exact `matchesPathspec` to files, so reported output is unchanged.
- The index-vs-workdir walk additionally drops anything the index does not contain, so untracked trees never reach the workdir walker at all (`node_modules` no longer costs even its own `lstat`).
- Applied to all four walks that take pathspecs: `diff`, `diff --staged` / `diff --cached <rev>`, `diff <rev> <rev>`, and `diff HEAD`. The commit-vs-workdir walk keeps its tracked-ancestor check, so a staged file in a directory absent from the commit still shows up (covered by an existing test).

Tests use a counting fs that records `readFile`, `readDir`, `lstat` and `stat`. On a repo with tracked files under `a/`, `c/`, `d/` and `e/deep/`, `diff -- a/b.txt` touches **nothing** under the excluded siblings — no readdir, no lstat, no read — and skips `a/a.txt` as well; same assertions for a directory pathspec, a nested path, `diff HEAD -- <path>` and `diff --staged -- <path>`. Reverting `diff.ts`/`revision.ts` to the previous commit fails 4 of the 5 new tests.

Verification, all local: `npm run lint` pass, `npm run typecheck` pass, `packages/webapp/tests/git` 392 passed / 16 files, `npm run test:coverage:webapp` **exit 0** (808 files, 14,650 tests; lines 84.38/84, statements 82.37/82, functions 82.32/82, branches 73.86/73; `diff.ts` 94.7 % lines, 97.3 % functions), `npm run build -w @slicc/webapp` pass, `check-first-load-size.mjs` OK, boy-scout and doc gates pass.

PR #2729, opened 2026-09-01
fix(webapp): time reports only the seconds it can measure

### LINKED ISSUE #2718: shell: `time` reports `real 0m0.003s` for a pipeline that took ~100 ms
Part of #2707 (found while trying to time the git benchmark from inside the shell).

## What happens

In the Terminal panel:

```
/ $ time ls /mnt/slicc/.git/objects/pack | wc -l
91
real    0m0.003s
user    0m0.000s
sys     0m0.000s
```

The same command measured from outside (CDP, Enter → prompt) takes ~100 ms (the listing is a hostfs round trip). `time` appears to stop the clock when the pipeline has been *dispatched*, not when it has *completed*, so it is useless for any command that touches the VFS/network — which is every interesting one.

## Expected

`real` ≈ wall-clock until the pipeline's last process exits, like bash. `user`/`sys` can stay 0 (no process accounting in the browser) but should then be omitted rather than reported as `0m0.000s`.

## Where

The `time` keyword is parsed by just-bash (`node_modules/just-bash/dist/ast/types.d.ts:65` "Time the pipeline with 'time' keyword"); the measurement lives in just-bash's pipeline executor, so this may be an upstream fix (or SLICC's `AlmostBashShellHeadless` can wrap `executeCommand` and print its own timing).

trieloff:
## Summary

`time` in the Terminal printed two numbers it cannot measure. This drops them and pins what `real` actually covers.

Before / after, same pipeline:

```
$ time ls /mnt/slicc/.git/objects/pack | wc -l      $ time ls /mnt/slicc/.git/objects/pack | wc -l
91                                                  91

real    0m0.003s                                    real    0m0.003s
user    0m0.000s
sys     0m0.000s
```

## Root cause

Two separate things were folded into one bug report.

**The `user`/`sys` lines were invented.** just-bash's pipeline formatter appends a
hardcoded `user 0m0.000s` / `sys 0m0.000s` (and `user 0.00` / `sys 0.00` under
`time -p`, plus the same block in the `/usr/bin/time`-style `time` *command*) to
every report. There is no process accounting in a browser tab, so those two zeros
are not a measurement — and printed under a legitimately fast `real`, they are
exactly what makes the output read like a clock that stopped at dispatch.

**The clock itself was already correct.** just-bash starts it before the first
command of the pipeline (`Pipeline.timed` → `performance.now()`) and stops it after
the last command's promise settles, so every awaited VFS / hostfs / network round
trip is inside the window. I could not produce an under-report in any shape —
`time cmd`, `time cmd | cmd`, `time (cmd)`, `time { …; }`, `time for …`, `time -p`,
with redirections, with `$(…)` — all report the true wall clock for a custom
command that awaits a timer, both in raw just-bash and through
`AlmostBashShellHeadless.executeCommand`.

So `real 0m0.003s` for `ls /mnt/slicc/.git/objects/pack` was accurate: `ls` on a
mounted path is **one** async `readDir` in `VfsAdapter` (no per-entry `stat` —
`readdirWithFileTypes` maps dirents from the listing), i.e. a single
`/api/hostfs/list` request, and #2707 measured the bridge at p50 **2.7 ms**. The
~100 ms the issue measured from outside is Enter→prompt: the terminal panel's RPC
round trip to the worker shell plus xterm render, and none of that is inside the
pipeline. That is now written down rather than left to be rediscovered.

## What changed

- `patches/just-bash+3.4.2.patch` — new hunks dropping the fabricated `user`/`sys`
  lines from the `time` keyword (default and `-p`) and from the `time` command's
  `-p` branch, in all eight built copies just-bash ships: `dist/bundle/browser.js`
  (what Vite bundles into the leader tab and kernel worker), `dist/bundle/index.js`,
  `dist/bundle/index.cjs`, `dist/bundle/chunks/chunk-R2FBRMDU.js`, and the four
  `dist/bin/**` CLI copies. Regenerated by hand (pristine tree reconstructed by
  reverse-applying the existing patch, then `git diff --no-index`), so the existing
  curl hunks are unchanged. Verified by re-applying with `patch-package` from a
  clean copy: `just-bash@3.4.2 ✔`.
- `patches/patches.json` — extended the `just-bash` `reason` / `removeWhen`, and the
  `verify` command now covers both guards.
- `packages/webapp/tests/shell/just-bash-time-patch.test.ts` — new. A custom command
  awaiting a 120 ms timer must report `real ≥ 50 ms` (raw just-bash and in a
  pipeline); the SLICC shell must do the same for `time ls <dir> | wc -l` over a VFS
  whose `readDirSync` returns `null` and whose `readDir` awaits a timer — the
  mounted-path shape from the issue; the format must be `real` alone in both keyword
  and `-p` mode; and the four bundles must not contain the fabricated strings, so a
  patch that misses the browser bundle cannot pass on Node alone.
- `docs/shell-reference.md` — "Timing a command (`time`)" under Performance: what
  `real` spans, why `user`/`sys` are absent, and what sits outside the measurement.

**Upstream**: not filed. `patches.json` tracks it against
https://github.com/vercel-labs/just-bash with the same `removeWhen` convention as
the existing curl hunk — remove once just-bash reports only the numbers it can
measure (or gains real user/sys accounting).

## How verified

| step | result |
| --- | --- |
| `npm run lint` | pass (biome + prettier + all guards; `check-patches: 4 patch(es) OK`) |
| `npm run test` | 18257 passed / 18323, 57 skipped. Failures are 5 s test-timeouts and one timing-flaky suite on this (16-core, heavily loaded) box — a different set each run, none under `tests/shell/`. `wc-sprinkles.test.ts > … hung VFS walk must not strand boot` failed 2 of 4 consecutive runs on identical code. |
| `npm run test:coverage:webapp` | **pass, exit 0** — 808 files / 14640 tests green; lines 84.36 % (floor 84), statements 82.33 % (82), functions 82.22 % (82), branches 73.85 % (73). Needed `--retry=2 --maxWorkers=6 --testTimeout=30000 --hookTimeout=60000` to get a green run out of this box; see note below |
| `npm run build -w @slicc/webapp` | pass, `✓ built in 7.67s` |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | `First-load OK (allowance 5 kB per change; total 4957 kB)` |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | pass (no debt-listed file touched) |
| new test file, isolated | 8 passed |

Local-environment note: at 16 workers this machine pushes several heavyweight suites
(pyodide load, `wc-live.ts` import) past their 5 s / 10 s budgets, and vitest emits no
coverage report when a run has any failure. Dropping to 6 workers with retries produced
a fully green run and the numbers above. No source file changed in this PR (patch + test
+ docs only), so the floors could not have moved either way.

Fixes #2718


trieloff:
Closing unmerged — see #2718: the investigation here showed `real` was already correct and the remaining user/sys zeros are cosmetic; not worth carrying a just-bash patch hunk (and a non-POSIX `time -p`). Thanks for the thorough trace — it's recorded on the issue.

Issue #2716, opened 2026-09-01
VFS/hostfs: `readDir` drops size/mtime on the way up, so every consumer re-stats each entry (N+1 round trips per directory)


trieloff:
Part of #2707.

## What happens

`/api/hostfs/list` already returns `size` + `lastModified` for every file (`packages/node-server/src/hostfs.ts` stats each dirent), and `HostFsMountBackend.readDir` parses them into `MountDirEntry`. `VirtualFS.readDirMounted` (`packages/webapp/src/fs/virtual-fs.ts`) then keeps only `{name, type}`, and the isomorphic-git adapter (`packages/webapp/src/git/vfs-fs-adapter.ts`) reduces that to names. Every consumer that needs types or sizes then issues one `/api/hostfs/stat` per entry:

- `git branch`: 125 requests for 29 refs — 22 `list` + **100 `stat`** (`FileSystem.readdirDeep` stats every entry to decide whether to recurse).
- `git status`/`ls-files`: `GitWalkerFs.stat()` `lstat`s every workdir entry it just listed (≈5,000 stats), and the Files panel (`wc-workbench.ts` `dirChildren`) stats every file for its size.
- A single `ls` of a 91-entry directory in the terminal costs 1 list + per-entry stats when `ls -l` is used.

Each of those is a full HTTP request + CORS preflight (#2715) on the bridge.

## Fix ideas

- Carry `size`/`mtime` through `DirEntry` (optional fields) from `readDirMounted`, and have the isomorphic-git adapter keep a per-command stat cache primed from `readdir` results (isomorphic-git's `readdir` contract is names-only, but the adapter's `lstat`/`stat` can answer from the primed cache).
- Add a batch endpoint (`POST /api/hostfs/stat` with a path list, or `list?stat=1&depth=n`) for the tree walkers.
- `git branch` specifically: read `packed-refs` + `refs/heads` with a `list` that already tells files from directories (it does) — the extra stats are pure waste.

PR #2730, opened 2026-09-01
fix(git): stop read-only commands rewriting .git/index over hostfs

### LINKED ISSUE #2708: git: read-only commands rewrite the host repo's `.git/index` thousands of times through hostfs

trieloff:
## Summary

Read-only git commands over a `--mount`ed host checkout no longer write to the repository, and the hostfs bridge now reports enough of `stat(2)` for isomorphic-git's index stat-cache to hit at all.

## Root cause

Two independent problems, both required for the 117 s / 30,276-request `git ls-files` measured in #2707:

1. **Nothing passed `refresh: false`.** `statusMatrix` and `WORKDIR()` default to `refresh: true`. When `compareStats` reports a file's cached stats as stale, `GitWalkerFs.oid()` re-hashes the file and — if the oid still matches — calls `index.insert(...)` purely to warm the stat cache. That marks the index dirty, and `GitIndexManager.acquire` (called **once per file**) then serializes and writes the whole `.git/index` back. Result: 3,485 `PUT /api/hostfs/write` of a 437 KB body for one `ls-files`, the host's index replaced with isomorphic-git's extension-less v2 form, written non-atomically.
2. **The stats could never match.** `HostFsMountBackend.stat` returned only `{kind,size,mtime}`, `VirtualFS.statInner` set `ctime = mtime`, and `toStats` filled `ino: 0`, `uid/gid: 1` and a constant mode `100644` — while `compareStats` compares mode, mtime, ctime, uid, gid, ino and size. So every file was stale forever, which is what fed (1) and what made every command re-SHA-1 the whole tree.

## What changed

**1. Read-only commands never write** — `NO_INDEX_REFRESH` (`packages/webapp/src/git/commands/shared.ts`, documented with the why) is spread into every workdir walk of a command that must not mutate the repo:

- `status.ts` — both the long and `--short`/`--porcelain` matrices
- `ls-files.ts`
- `diff.ts` — `git.WORKDIR()` in the commit-vs-workdir path (the only `WORKDIR()` in the tree). `diffWorkdirChanges` is left alone otherwise; #2719 owns rewriting it.
- `clean.ts` — both matrices; `clean -n` is strictly read-only and the delete path never needs a stat-cache refresh

Commands that legitimately write the index (`add`, `commit`, `reset`, `stash`, `rebase`) keep the default, so the stat cache still warms where a write is happening anyway.

**2. Stat fidelity end to end:**

- `packages/node-server/src/hostfs.ts` — new `statIdentity()`; `/api/hostfs/stat` and every file row of `/api/hostfs/list` now carry `ctime`, `ino`, `uid`, `gid` and the full `st_mode`.
- `packages/swift-server/Sources/Server/HostFSRoutes.swift` — the same payload, via `stat(2)` rather than `FileManager.attributesOfItem` (Foundation only exposes `.creationDate`, i.e. birth time, not the POSIX inode-change time git records). The list-row construction moved into a `listEntry` helper, which also brings `registerRoutes` back under the SwiftLint `function_body_length` limit it was already over.
- `MountStat` / `MountDirEntry` gained a shared optional `MountStatIdentity`; `backend-hostfs.ts` parses the fields permissively (a partial or older payload must never make a mount unreadable).
- `VirtualFS.Stats` gained optional `uid` / `gid` / `mode` next to the existing `ino`; `statInner` passes the mount's values through and falls back to `ctime = mtime` for backends that have no inode (S3/DA/AEM). The local (ZenFS) branch now carries its own `mode`/`uid`/`gid` too.
- `toStats` (`vfs-fs-adapter.ts`) uses the real values when present and keeps the historical placeholders otherwise. Mode is composed as *type bits from the entry type VirtualFS resolved* + *permission bits from the backend*, so **the executable bit survives (`100755`)** instead of being flattened to `100644`, while isomorphic-git's `mode >> 12` symlink detection still can't be fooled by a backend's permission bits.

**Docs:** `docs/shell-reference.md` (the read-only rule + how to extend it), `docs/mounts.md` (hostfs reports real stats), `docs/pitfalls.md` (the "mount-backed subtrees have no identity" bullet was stale for hostfs), `packages/node-server/CLAUDE.md`.

## Tests

- `packages/webapp/tests/git/git-readonly-no-index-writes.test.ts` (new) — a counting `PromiseFsClient` that forwards to a real VirtualFS adapter but reports permanently-stale stats (ino 0, drifting ctime). Every read-only command must record **zero** writes. Includes a control test asserting the fake really does force an index rewrite under `refresh: true`, so the assertions can't silently stop proving anything. **Mutation-checked:** flipping `NO_INDEX_REFRESH` back to `refresh: true` fails 6 of the 7 assertions.
- `packages/webapp/tests/git/vfs-fs-adapter-stats.test.ts` (new) — pass-through of ctime/ino/uid/gid/mode from a mount backend, the executable bit, the placeholder fallback for a backend that reports nothing, and entry-type bits winning over reported permissions (symlink/dir).
- `packages/webapp/tests/fs/backend-hostfs.test.ts` — payload parsing, including omitted and non-numeric fields.
- `packages/node-server/tests/hostfs.test.ts` — stat + list payloads compared against `fs.stat`, plus the executable bit.
- `packages/swift-server/Tests/HostFSRoutesTests.swift` — the same assertions against `stat(2)`.

## How verified

All from this worktree, on top of `origin/main`:

| Command | Result |
| --- | --- |
| `npm run lint` | pass (exit 0; 43 pre-existing warnings, no errors) |
| `npx vitest run --maxWorkers=4` | **1027 files / 18,282 tests passed**, 4 skipped, 0 failed |
| `npm run test:coverage:webapp` | pass — 809 files / 14,648 tests; statements 82.39%, branches 73.89%, functions 82.32%, lines 84.4% |
| `npm run test:coverage:node-server` | pass — statements 81.7%, branches 72.27%, functions 85.49%, lines 83.74% |
| `npm run build -w @slicc/node-server` | pass |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | `First-load OK (allowance 5 kB per change; total 4958 kB)` |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | pass (`no debt lists found`) |
| `swift build` + `swift test --filter HostFSRoutesTests` | pass — 6/6 |
| `swift format lint --strict` + `swiftlint lint` | pass; the `function_body_length` warning on `registerRoutes` is now gone (127 → under the cap) |

Note on the default-parallelism test run: the first `npm run test` invocations failed 2–10 tests, all 5-second timeouts scattered across unrelated files (pyodide, adobe-provider, WebUSB, a pure string formatter). The box was at load average 60–90 on 16 cores from other agents; re-running with `--maxWorkers=4` is fully green, and the flakiest of them passes standalone in 3.8 s.

Not verified live in a browser against a real mounted repo — the benchmark harness in #2707 is the way to re-measure that, and it needs a real host checkout.

## Scope

Deliberately no caching layer: #2709 / #2710 / #2716 own that, and #2719 owns the `diffWorkdirChanges` rewrite (this PR only adds `refresh: false` to the `WORKDIR()` walk it neighbours).

Fixes #2708


trieloff:
Addressed the Codex P2 on `packages/node-server/src/hostfs.ts:142` in 74b9738f3.

**The bug is real and it defeated the point of this PR.** `statIdentity`/`statPayload` used `Math.round(s.ctimeMs)`, but isomorphic-git's `normalizeStats` derives the value `compareStats` actually compares with `Math.floor(ms / 1000)`. A stat in the last 0.5 ms of a second rounded into the **next** second, disagreed with the seconds native git wrote into the index, and left that file stale on every walk — re-hashed forever, which is exactly the cost this PR set out to remove. The same rounding applied to `mtime`/`lastModified` (pre-existing).

**Fix:** both bridges now send the fractional milliseconds **unrounded** — JSON carries the double fine, and floor-of-unrounded can never advance the second. Chose pass-through over `Math.floor` because it also keeps sub-millisecond precision for non-git consumers, and it is trivially correct under any later `floor`.

- `packages/node-server/src/hostfs.ts` — `statIdentity` (`ctime`), `statPayload` (`mtime`), and the `list` row's `lastModified`.
- `packages/swift-server/Sources/Server/HostFSRoutes.swift` — mirrored: `statIdentity`'s `ctimeMs`, `statAt`'s `mtime`, and `listEntry`'s `lastModified` (the last two were the pre-existing `.rounded()`).
- Both now carry a comment naming `Math.floor(ms / 1000)` as the reason, so this doesn't get "tidied" back.

**Tests** (all three fail if the rounding comes back — verified by re-applying it):

- `packages/node-server/tests/hostfs.test.ts` — "never rounds a timestamp up into the next second": `utimes(file, 1_700_000_000.9996, …)`, then asserts the payload equals `fs.stat`'s `mtimeMs`/`ctimeMs` exactly and that `Math.floor(mtime / 1000) === 1_700_000_000`, on both `/stat` and the `/list` row. Re-adding `Math.round` fails 3 tests.
- `packages/swift-server/Tests/HostFSRoutesTests.swift` — the same scenario via `setAttributes([.modificationDate:])` + `stat(2)`. Re-adding `.rounded()` fails 3 assertions.
- `packages/webapp/tests/git/vfs-fs-adapter-stats.test.ts` — adapter-level: a mount stat of `1_700_000_000_999.6` must survive `toStats` unchanged and floor to `1_700_000_000`, i.e. the seconds `compareStats` compares.

`docs/mounts.md` now states the unrounded rule.

**Re-verified on the branch head (74b9738f3):**

| Command | Result |
| --- | --- |
| `npm run typecheck` | pass (9 projects) |
| `npx vitest run --maxWorkers=4` | **1027 files / 18,284 tests passed**, 4 skipped, 0 failed |
| `npm run test:coverage:node-server` | pass — 81.68 / 72.22 / 85.49 / 83.74 |
| `npm run test:coverage:webapp` | pass — 82.4 / 73.91 / 82.33 / 84.42 |
| `npm run build -w @slicc/webapp`, `-w @slicc/node-server` | pass |
| `check-first-load-size.mjs` | `First-load OK (total 4958 kB)` |
| `check-touched-exemptions.mjs` | pass |
| `swift build`, `swift test --filter HostFSRoutesTests` | pass — 7/7 |
| `swift format lint --strict`, `swiftlint lint` | pass |

PR #2737, opened 2026-09-01
feat(webapp): p peeks a tab from anywhere, sprinkles move to e

trieloff:
Stacked on #2721 — the peek gesture it adds only exists there. **Base is `feat/tab-keys-peek`, not `main`**, so this gets the review bots but not the verification CI jobs until #2721 merges and GitHub retargets it. Worth knowing before reading the checks.

## Why

Your live test of #2721 found the wart in it: peek worked, but only once you were already inside the switcher. `p 1` from where you actually are — mid-conversation, wanting to glance at a tab and come straight back — did nothing of the sort, because `p` was sprinkles out there and peek only in the modal.

## What

**One verb, one key, everywhere.** `p` now opens the switcher with peek already armed, so `p 1` is a single gesture from anywhere: the tab comes forward, SLICC comes back five seconds later. Inside the switcher `p` still arms peek on its own — now the same statement rather than a second meaning.

The mechanism is smaller than it sounds and lives entirely in the ordering. `peekTabs` **arms the overlay and then opens it** through the dock's own event. Opening it makes it modal, which suspends every shell command, so by the time the digit lands the switcher already owns the keyboard *and* already knows the next activation is a peek. The digit never has to be a chord — and the held-digit fix from the last review round on #2721 is what makes it land at all before the asynchronous tab list paints. Two pieces of that PR, doing a job neither was built for.

**Sprinkles move to `e`.** That's the honest cost: two commands can't both be `p`, and between "the thing you reach for to check a tab" and "the launcher for optional extras", the frequent one gets the mnemonic. `e` for extras was free — though note v1 spent it on the terminal, which v2 had already moved to `t`, so anyone with 6.110 fingers gets sprinkles where they expect a shell.

A float whose switcher can't peek (a follower — its tabs belong to the leader, and activating one copies it here for good) refuses the arming and just opens. Same key, one honest step less.

## Verification

`verify`, `typecheck`, all 1,182 webapp UI tests, first-load (+0.0 kB), bundle sizes, boy-scout gate — green. 6 new tests: the rebind (`p` no longer opens sprinkles, `e` does), and the arm-then-open ordering that the whole gesture depends on.

Two honest caveats:

- The full local suite shows **one failure, unrelated**: `py-realm-shared` asserts the pinned pyodide version against the installed one, and my worktree's `node_modules` predates a `314.0.5 → 314.0.6` bump on main. CI installs fresh.
- **Not live-verified.** I rebuilt the webapp and was reloading the harness when the port was needed back, so this specific change has not been driven in a browser. What *was* verified live, on #2721's build: `b 1-9` switching, and in-switcher `p` + digit doing the real CDP round trip and returning to SLICC on its own.


trieloff:
### Review fixes (`b5bd617bf`)

Six findings, one root cause: moving sprinkles off `p` left six places still teaching `p` for it. One of them was worse than stale.

**The v1 paste-back block was broken JSON** (Codex P2 + Copilot). I'd dropped a parenthetical between two of its property lines, so copying the block — the documented way to restore the v1 keyboard — produced invalid JSON, and `parseKeymapDocument` would have discarded the whole thing and left the user on the shipped defaults with no idea why. The note moved into the prose above the block.

Worth more than the fix: there's now a test that **parses the block back out of the comment and asserts it restores the v1 map**. An instruction nobody executes is exactly the kind that rots, and this one had rotted within a day of being written.

**The rest were examples pointing at the wrong key** — `p 2` and `p j j` in the README, `p j` in the architecture note, and three comments in `wc-shortcuts.ts` calling `p` the sprinkle prefix. All now `e`. A reader following any of them would have opened the tab switcher and possibly peeked a browser tab instead of a sprinkle, which is a worse failure than a doc that merely lags.

One extra: the Shift-pairs line still advertised `p`/`P`, which hasn't existed since the step keys retired the sprinkle cycle in #2721. Dropped.

`verify`, typecheck and all 192 shortcut tests green.

---

**Stack status:** #2721 is in the merge queue. Once it lands this PR retargets to `main` on its own, verification CI runs for the first time, and I'll enqueue it if it's green.

PR #2739, opened 2026-09-01
feat(shell): add `umount <path>` as an alias for `mount unmount <path>`

### LINKED ISSUE #2738: shell: add `umount <path>` as an alias for `mount unmount <path>`
Found while driving the FSA benchmark for #2707: muscle memory says `umount /mnt/slicc`, but the kernel shell answers `bash: umount: command not found`; the only spelling is `mount unmount <path>` (or `mount -u`, `packages/webapp/src/fs/mount-commands.ts`).

## Ask

- Register a `umount` shell command that forwards to the existing unmount handler (`umount <path>` ≡ `mount unmount <path>`), with `--help` that says so, and the usual "mount point required" / ENOENT errors.
- Keep `mount unmount` / `mount -u` working.
- Tab completion / `help` listing should show it like any other command.
- Update the shell reference (`docs/shell-reference.md`) and the runtime skill that documents mounts (`packages/vfs-root/workspace/skills/…`) so the cone learns the alias.
- Tests: `umount <mounted>` unmounts; `umount` with no arg is a usage error; `umount <not-mounted>` errors the same way `mount unmount` does.

trieloff:
## Summary

`umount <path>` is now a registered shell command that aliases `mount unmount <path>`. Before: typing `umount /mnt/slicc` got `bash: umount: command not found`. Now: it unmounts, with the same flags, the same exit codes, and its own `--help`. `mount unmount`, `mount -u`, and `--clear-cache` are untouched.

## Why

Fixes #2738, found while driving the FSA benchmark for #2707. The only spellings the kernel shell knew were `mount unmount <path>` and `mount -u <path>`, both sub-commands of `mount` — but muscle memory reaches for the Unix spelling, and nothing registered that name.

**Repro**

1. `mount /mnt/slicc` (or any mount)
2. `umount /mnt/slicc`

Expected: the mount is torn down.
Actual: `bash: umount: command not found`.

- **`packages/webapp/src/fs/mount-commands.ts`** — new `MountCommands.executeUmount(args, cwd)` forwards straight into the existing `handleUnmount`. It is a pure alias by construction: one parser, one handler, one set of exit codes — no duplicated flag handling that can drift. The one thing that varies is the diagnostic prefix, which now travels with the call as an `UnmountLabel` (`'mount unmount' | 'umount'`), so a user who typed `umount` reads `umount: path required` rather than being pointed at a command they did not run. `--help` / `-h` prints a short `umount` usage block; `mount --help` now names the alias in its sub-command list and examples.
- **`packages/webapp/src/shell/almost-bash-shell-headless.ts`** — `createUmountCustomCommand()` registers `umount` as a top-level custom command next to `mount`, so `commands`, `which`, and tab completion pick it up like any other builtin (rejected alternative: a `.jsh` shim — it would not be a builtin, and would lose `ctx.cwd` relative-path resolution).
- **`packages/webapp/src/shell/supplemental-commands/help-command.ts`** — `umount` added to the `Filesystem` category so `commands` groups it instead of dumping it in `Other`.

**One behaviour worth calling out.** The issue asks that `umount <not-mounted>` "errors the same way `mount unmount` does". It turns out `mount unmount` on a path that was never mounted is **not** an error — `VirtualFS.unmount` is deliberately idempotent (drops the entry if present, never throws). The alias matches that exactly rather than diverging, and both the test and the docs say so explicitly. Making `mount unmount` start erroring there would be a separate, riskier change (boot/teardown paths rely on the idempotence), so it is out of scope here. Error-path prefix parity is still covered where `fs.unmount` genuinely throws.

## Cross-cutting impact

- **Floats exercised**: the command is registered in `AlmostBashShellHeadless`, which is the shared base for every float (CLI, extension side panel + offscreen, Electron, worker), so all of them get `umount` from the one registration. No float-specific wiring.
- **Other callers of the changed contract**: `handleUnmount` gained a required `label` parameter. It is private with exactly two call sites, both updated (`execute`'s `unmount`/`-u` branch, and `executeUmount`); the compiler is the ratchet.
- **Approvals / sudo**: unaffected and automatically correct. Sudo gates on the `fs.unmount` **method**, not on a command name (`fs/sudo-fs.ts`), so `umount` inherits the identical write-approval path as `mount unmount`.
- **Agent command allowlists**: `umount` is deliberately **not** added to the memory curator's read-only allowlist in `scoops/agentic-memory.ts` — bare `mount` is there because it lists state; the alias is mutating, like `mount unmount`, which is also absent.
- **Persisted state**: none. Identical code path to `mount unmount`, including `removeMountEntry` and the mount-sync channel broadcast.
- **Bundle size**: negligible — worker graph +0.8 kB, 88 kB under ceiling (see below).

## Test plan

- [x] `npx prettier --write <changed-files>` — ran via `npm run lint`; `npm run lint:ci` then passed `prettier --check .`
- [x] `npm run typecheck` — exit 0 (all 9 projects)
- [x] `npm run test` — **18328 passing**, 52 skipped; 1028 files passed, 4 skipped, **0 failed**
- [x] `npm run build -w @slicc/webapp` — ✓ built
- [x] **New / changed behavior is covered by unit tests** mirroring the changed `src/` paths:
  - `packages/webapp/tests/fs/mount-commands.test.ts` — new `umount alias` block: no-arg usage error, unmounts a path, relative-path resolution against `cwd`, error-message parity with `mount unmount` (identical message body, differing prefix, same exit code), `--clear-cache` accepted in either position, `--help`/`-h` short-circuits without unmounting, and `mount --help` advertises the alias.
  - `packages/webapp/tests/shell/umount-command.test.ts` (new) — drives the real `AlmostBashShellHeadless` against a `VirtualFS` with a live `LocalMountBackend`, so **registration** is covered and not just the handler: `umount <mounted>` unmounts, no-arg is a usage error, a not-mounted path behaves byte-identically to `mount unmount`, `mount unmount` / `mount -u` still work, and `umount --help` / `commands` / `which umount` all resolve.
  - `packages/webapp/tests/shell/supplemental-commands/help-command.test.ts` — `umount` added to the canonical builtin surface so the "Other" bucket assertion keeps catching uncategorized names.
  - Targeted run of the three suites: `npx vitest run packages/webapp/tests/shell/umount-command.test.ts packages/webapp/tests/fs/mount-commands.test.ts packages/webapp/tests/shell/supplemental-commands/help-command.test.ts` — **65 passed, 0 failed**.
- [x] N/A — no UI change, so no screencast.

**Pre-existing failures**: none. The full suite was green.

## Verification

| Gate | Result |
| --- | --- |
| `npm run lint` | exit 0 |
| `npm run lint:ci` | exit 0 (biome + `prettier --check .`, skills `--strict`, doc sizes, doc refs, layer back-edges, `Record<string, unknown>`, duplication) |
| `npm run typecheck` | exit 0 |
| `npm run test` | 18328 passed / 52 skipped / **0 failed** |
| `npm run test:coverage:webapp` | exit 0 — statements 82.38%, branches 73.84%, functions 82.3%, lines 84.39%; `coverage-thresholds.json` untouched |
| `npm run build -w @slicc/webapp` | ✓ built |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | `no debt lists found — nothing to gate` (both baselines are `{}` on main) |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | `First-load OK` — worker +0.8 kB vs base, 88 kB under ceiling |

## Documentation

- [x] `docs/` — `docs/shell-reference.md` (new command-index row, sub-command table row, `--clear-cache` applicability, example block) and `docs/mounts.md` (mount syntax block)
- [x] `packages/vfs-root/workspace/skills/mount/SKILL.md` (agent-facing) — lifecycle block plus a note that the two spellings are interchangeable, and that unmounting a non-mounted path is a no-op under both
- [ ] `README.md` — not needed; the shell command surface is documented in `docs/`, not the README

## Risk & rollback

- **Blast radius**: minimal. Additive — a new command name plus a private parameter with two call sites. No existing spelling changes behaviour.
- **Flags / kill switches**: none introduced.
- **Rollback**: revert this PR cleanly. No persisted-state migration.
- **Not catchable by CI**: nothing specific. The mount teardown path itself is unchanged; the alias reaches it through the same handler that `mount unmount` already used, and the shell test exercises a real `VirtualFS` mount rather than a mock.

## Checklist

- [x] Public shell-command surface change is intentional and reflected in docs (`docs/shell-reference.md`, `docs/mounts.md`, mount `SKILL.md`)
- [x] Contract used by other floats: registration lives in the shared `AlmostBashShellHeadless`, so CLI / extension / tray all get it from one place — no per-float follower wiring needed

Fixes #2738


trieloff:
Both findings addressed in d743ab67cb71198cf0b6723b5d6d6658f6e7956a.

### 1. P1 — `--help` after the path destructively unmounted

Confirmed, and it was **not** limited to the alias. `MountCommands.execute` dispatches on `args[0]`, so `mount unmount /mnt/x --help` and `mount -u /mnt/x --help` had the identical hole: `parseArgs` pushes the unrecognized `--help` into `positional`, `positional[0]` is still `/mnt/x`, and the mount gets torn down. Asking for usage performed the destructive action on all three spellings.

Fixed with one guard rather than three. A local `isHelpRequest(args)` scans for `-h`/`--help` anywhere before a `--` end-of-options terminator, and shadows the values of `--source` / `--profile` / `--backend` / `--max-body-mb` so a flag *value* is never mistaken for a help request (kept in agreement with `parseArgs`). It is reimplemented in `fs/` rather than imported from `shell/supplemental-commands/subcommand-help.ts` because `fs/` must not import up into `shell/` — the layer back-edge gate rejects it.

The check now runs at the top of `handleUnmount`, ahead of anything mutating, so `umount`, `mount unmount`, and `mount -u` all inherit it from one place; `executeUmount` no longer carries its own `args[0]` test. Help is keyed on the existing `UnmountLabel`, so `umount … --help` prints the `umount` usage block and `mount unmount … --help` prints the `mount` usage block.

**Regression tests** — 3 new cases across both spellings, each asserting `fs.unmount` was **not** called / the mount is still listed:

- `tests/fs/mount-commands.test.ts` — `umount /mnt/myapp --help`, `-h`, and `--clear-cache /mnt/myapp --help`; plus `mount unmount /mnt/myapp --help`, `-h`, and `mount -u /mnt/myapp --help`.
- `tests/fs/mount-commands.test.ts` — `umount /mnt/myapp -- --help` still unmounts, so the `--` terminator is not swallowed.
- `tests/shell/umount-command.test.ts` — end-to-end through the real shell against a live `LocalMountBackend`: both spellings print usage, `fs.listMounts()` still contains the path, and the mount is still removable once the flag is dropped.

**Mutation-tested**: reverting `isHelpRequest` to the old `args[0] === '--help' || args[0] === '-h'` fails exactly those 3 tests (3 failed / 55 passed); with the fix, 58 passed. The tests fail for the reason they claim to.

### 2. README.md

Added to the existing **Mount your local file system** bullet: `Tear one down with `umount /mnt/myproject` (or the longer `mount unmount /mnt/myproject`).` — user-facing shell behavior belongs in the README per the doc-tier rule, alongside the `docs/` and `SKILL.md` updates already in this PR.

### Re-verification

| Gate | Result |
| --- | --- |
| `npm run lint` | exit 0 |
| `npm run lint:ci` | exit 0 |
| `npm run typecheck` | exit 0 (9 projects) |
| `npm run test` | **18332 passed**, 52 skipped, **0 failed** (1028 files) — was 18328, +4 new cases |
| `npm run test:coverage:webapp` | exit 0 — statements 82.39%, branches 73.82%, functions 82.28%, lines 84.41%; thresholds untouched |
| `npm run build -w @slicc/webapp` | ✓ built |
| `check-touched-exemptions.mjs` | `no debt lists found — nothing to gate` |
| `check-first-load-size.mjs` | `First-load OK` — 4959 kB total, under ceiling |

Targeted: `npx vitest run packages/webapp/tests/shell/umount-command.test.ts packages/webapp/tests/fs/mount-commands.test.ts` — 58 passed, 0 failed.

One thing I deliberately left alone: `mount refresh <path> --help` has the same `args[0]`-only shape, but `refresh` re-walks and diffs rather than destroying state, so fixing it is a separate non-urgent change rather than scope creep on this one.

Issue #2735, opened 2026-09-01
git: isomorphic-git's once-per-packfile SHA-1 verification runs once per *concurrent reader* — the slower the pack read, the more times the 92 MB pack gets hashed


trieloff:
Part of #2707; found while comparing the hostfs and FSA profiles. Distinct from #2710 (which is about the cache not surviving across commands): this multiplies the cost *within* one command.

## Evidence

Warm `git ls-files` on the same build and the same 92 MB pack:

| backend | pack read | time in `digest` (same call chain: `readObjectPacked` → `shasumRange`) |
| --- | --- | ---: |
| hostfs (HTTP, 304-revalidated 92 MB body) | slow | **5,195 ms** |
| FSA (`file.arrayBuffer()`) | fast | **603 ms** |

One SHA-1 over 92 MB in the worker costs ~0.6 s, so the hostfs run hashed the pack roughly 8–9 times.

## Why

`readObjectPacked` (isomorphic-git 1.41.9):

```js
if (!p.pack) { p.pack = fs.read(packFile); }
const pack = await p.pack;
if (!p._checksumVerified) {
  … shasumRange(pack, …) …        // 92 MB SHA-1
  p._checksumVerified = true;      // set only AFTER the await
}
```

`_walk` fans object reads out with `Promise.all`, so many `readObjectPacked` calls are in flight before the first one finishes. They all share the `p.pack` promise (good), but every caller that reaches the `if` while the first verification is still awaiting sees `_checksumVerified === false` and starts its own full hash. The number of duplicate hashes equals the number of readers queued behind the pack read — which is exactly why a *slow* transport (hostfs) gets hit harder than a fast one (FSA), and why the cold hostfs run shows no `digest` samples at all (verification happened while the walk was still blocked elsewhere). Each duplicate also copies the 92 MB input for WebCrypto, so memory spikes with it.

## Fix

Memoize the verification as a promise on the cached pack object (`p._verified ??= verify(pack)`; `await p._verified`), or fold it into the `p.pack` promise so the bytes are verified exactly once before any reader gets them. Small upstream fix; until then SLICC's patch-package flow (or the cache/verification switch from #2710) can carry it. Add a test that fires N concurrent `readObject` calls on a fresh cache and asserts `shasumRange` ran once.

PR #2748, opened 2026-09-02
perf(git): cache reads for the duration of one git command

### LINKED ISSUE #2709: git: `statusMatrix` costs ~5 round trips per file — index re-stat per file, `.gitignore` re-read per candidate, no command-scoped cache

trieloff:
## Summary

Adds a **command-scoped read cache** in front of the isomorphic-git ↔ VirtualFS adapter, so one `git` invocation asks the filesystem for each path once instead of once per walked file.

On a synthetic 242-entry repo (Vitest, counting every VirtualFS call), `git ls-files` goes from **684 filesystem operations to 261 (−62 %)**, and the four hot paths collapse to a single round trip each:

| path | op | before | after |
| --- | --- | ---: | ---: |
| `.git/index` | lstat | 123 | **1** |
| `.gitignore` | readFile | 120 | **1** |
| `sub/.gitignore` | readFile | 60 | **1** |
| `.git/info/exclude` | stat | 120 | **1** |

Those are the same four rows that dominated the real measurement in #2707 (10,513 `stat .git/index`, 1,195 `.gitignore` reads out of 30,276 requests for one `ls-files`). The residual per-entry `lstat` of each workdir file is #2716's territory (`readDir` drops size/mtime, so nothing can be primed from a listing yet) and is deliberately not touched here.

## Root cause

isomorphic-git asks the same questions once per walked file and SLICC's `PromiseFsClient` had no memoization, so over a `--mount`ed host repo every one of them was a live HTTP round trip on the hostfs bridge:

- `GitWalkerFs.oid()` → `GitIndexManager.acquire()` → `isIndexStale()` **`lstat`s `.git/index` on every call**, i.e. once per tracked file.
- `GitIgnoreManager.isIgnored()` re-reads `.git/info/exclude` (after an `exists` → `stat`) **and every ancestor `.gitignore`** for every untracked candidate.

## What changed

- **`packages/webapp/src/git/fs-command-cache.ts`** (new) — `withCommandScopedReadCache(inner)` wraps an `IsoGitFsPromises` with a memo for `stat` / `lstat` / `readFile` / `readdir` that is **inert until a scope is opened**, so every existing direct `createIsomorphicGitFs(vfs)` caller keeps today's behavior.
  - Caches a stable "not there" errno (`ENOENT` / `ENOTDIR` / `EISDIR`) — that is the answer — but **evicts anything else**, so one transient `EIO` from the bridge (#2720) is never replayed to the rest of the command.
  - **Never caches packfiles or pack indexes** (isomorphic-git's own object cache owns those — #2710 / #2712 / #2735 are untouched), nor files over 1 MB, with an 8 MB / 50 k-entry ceiling per command.
  - Invalidates the path, its parent's listing and the parent's stats on every `writeFile` / `unlink` / `mkdir` / `rmdir` / `symlink` it sees (subtree-wide for `unlink` / `rmdir`), in a `finally` so a half-failed write also invalidates.
  - Hands out copies of `readdir` arrays and file bytes — isomorphic-git sorts readdir results in place, and a poisoned `.git/index` buffer would be an expensive bug.
- **`git-commands.ts`** — `execute()` opens a scope before dispatch and drops it in the existing `finally`, so nothing survives into the next command. Gated by an **allowlist** (`CACHEABLE_COMMANDS`): the cache only sees writes that go through the adapter, so subcommands that write straight to `ctx.fs` (checkout, reset, rm, mv, stash, rebase, revert, clean, merge-file, config) and the network commands stay uncached.
- **`vfs-fs-adapter.ts`** — `wantsUtf8()` exported so the cache keys utf-8 and binary reads of one path separately. No behavior change.
- **`docs/webapp-details.md`** — new bullet in the Git section covering the invariant (one memo per command, never across commands) and why the allowlist exists.

The `readdir`-primed stat cache the issue suggests is **not** included: `DirEntry` is still `{name, type}` on `main` — enriching it is #2716 — so there is no size/mtime/mode to prime with yet. When that lands, the adapter can feed the cache from a listing without touching this module's shape.

## How verified

- `npm run lint` → exit 0 (43 pre-existing warnings, none in the touched files)
- `npm run typecheck` → clean across all 9 projects
- `npm run test` → **1037 files passed / 4 skipped; 18,459 tests passed / 52 skipped, 0 failed**
- `npm run test:coverage:webapp` → exit 0 — statements 82.48 % / branches 73.96 % / functions 82.41 % / lines 84.48 %, all above the webapp floors (82 / 73 / 82 / 84); the new module itself lands at 94.07 % statements, 97.22 % lines
- `npm run build -w @slicc/webapp` → built in 9.08 s, exit 0
- `node packages/dev-tools/tools/check-first-load-size.mjs` → `First-load OK (allowance 5 kB per change; total 4965 kB across both graphs)`
- `node packages/dev-tools/tools/check-touched-exemptions.mjs` → `no debt lists found — nothing to gate`; the pre-commit hook's layer/`Record<string, unknown>` gates both reported ok

New suite: `packages/webapp/tests/git/fs-command-cache.test.ts` (21 tests) — unit tests for memoization, scope lifetime + refcounted overlap, negative caching vs transient errors, encoding keying, copy-on-read, pack exclusion, the size budget and each invalidation path; plus end-to-end `GitCommands.execute()` tests over a counting VirtualFS that assert the index is stat'ed once, each ignore file read once, that the cache does **not** survive across two `execute()` calls, that a change made between commands is seen, that `.gitignore` is still honored, and that `add` → `status` → `commit` → `status` → `log` sees its own writes. A control test asserts the uncached adapter still produces the >10× duplicate reads, so the assertions cannot silently stop proving anything.

Fixes #2709


trieloff:
Both findings were real. Fixed in 719e4543a.

### 1. The memo was shared instance state across concurrent `execute()` calls

Correct, and the reference count made it worse rather than better: a second cacheable command starting while the first was in flight incremented `depth` and joined the *first* command's `ReadScope`, and a non-allowlisted command running inside an open scope got the cached adapter without ever calling `begin()`, so its `ctx.fs` writes invalidated nothing.

Took the per-invocation adapter route rather than the mutex:

- `withCommandScopedReadCache(inner) → {promises, cache}` with `begin()`/`end()` is gone. It is now `createCommandScopedReadCache(inner) → IsoGitFsPromises` — **the memo *is* the wrapper**, always on, private to the object, collected when the object is dropped. There is no shared state left to race on and no lifetime to get wrong.
- `GitCommands.ctx` (built once in the constructor) is replaced by `contextFor(command)`, called once per `execute()`. `lfs` is the only per-invocation field: `CACHEABLE_COMMANDS.has(command) ? createCommandScopedReadCache(this.lfs) : this.lfs`. `this.lfs` is now always the *uncached* adapter, so a non-allowlisted command cannot be handed a memo even by accident.
- `resolveAuthor(cwd)` → `resolveAuthor(cwd, lfs)` so its `git.getConfig` uses the invocation's own adapter too.

Two commands overlapping in time still race the way two OS processes race on a real checkout (one may see the tree as it was when it started) — that is inherent, not cache-induced. What the per-invocation memo guarantees is that the race cannot outlive the overlap.

New tests:
- `gives two overlapping execute() calls two separate memos` — `Promise.all([ls-files, status])` on a counting VirtualFS asserts `lstat .git/index` is **2**, one per invocation. Under the old refcount it was 1; this is the regression lock.
- `runs a command that writes outside the adapter uncached` — `clean -n` (deletes via `ctx.fs`) re-stats the index per file, proving it gets the plain adapter.
- `cannot be poisoned by an uncached command running concurrently` — `git mv` (writes dst + removes src through `ctx.fs`) runs alongside a cached `status`; the *next* command still sees `A  renamed-0.txt` / `D  tracked-0.txt`.
- `never shares a memo between two wrappers over one filesystem` at the unit level.

### 2. Negative `readFile` entries were unbounded

Also correct — `readFile` passed only `keep`, never `admit`, so an `ENOENT` was retained with no entry check and no byte accounting; a walk probing thousands of absent loose objects grew the `files` map without any ceiling.

`hasRoom()` now counts **all four maps** (`stats + lstats + dirs + files`) against one `maxEntries` cap, and `readFile` passes it as its `admit` gate, so misses are bounded by exactly the same budget as hits. The limits moved into an injectable `ReadCacheLimits` (`maxEntries` 50 000, `maxFileBytes` 1 MiB, `maxTotalBytes` 8 MiB) so the budgets are cheap to test rather than requiring 50 k calls.

New tests:
- `stops memoizing once the entry cap is reached, misses included` — fills a `maxEntries: 3` memo with unique ENOENT paths (each retained: one round trip for two calls), then asserts the next miss costs a round trip **every** time.
- `shares one entry budget across stat, readdir and readFile`.

### Verification

| gate | result |
| --- | --- |
| `npm run lint` | exit 0 |
| `npm run typecheck` | exit 0, all 9 projects |
| `npm run test` | 18,460 passed / 52 skipped; **2 failed** — `kernel/realm/slicc-fs-module.test.ts` (pyodide), pre-existing: confirmed identically failing with my changes stashed |
| `npm run test:coverage:webapp` | exit 0 with that file excluded — stmts 82.46 / branches 74.00 / funcs 82.38 / lines 84.47 vs floors 82 / 73 / 82 / 84. `fs-command-cache.ts` itself 94.78 % stmts, 96.7 % lines |
| `npm run build -w @slicc/webapp` | exit 0 |
| `check-first-load-size.mjs` | `First-load OK (allowance 5 kB per change; total 4965 kB)` |
| `check-touched-exemptions.mjs` | `no debt lists found`; pre-push ran all 7 checks, all passed |

Suite is 24 tests (was 21). `docs/webapp-details.md` updated with the "never make it instance state, a `GitCommands` instance serves concurrent `execute()` calls" invariant and the shared entry cap.

Head SHA: `719e4543a6b2d85f6e4a24a0c2140ef55fbb39de`

Issue #2713, opened 2026-09-01
git: `resolveRevision` tries `expandOid` before `resolveRef`, so resolving `HEAD` reads every pack index (3.8 MB, 34 requests)


trieloff:
Part of #2707.

## What happens

`git -C /mnt/slicc rev-parse HEAD`: 205 ms, **38 requests, 3.83 MB transferred** — 34 of the requests are reads of all 30 `.git/objects/pack/*.idx` files (plus `packed-refs`, `HEAD`…). Native: 31 ms, reads two small files. Warm runs still make all 38 requests (bodies come back `304`).

The same path is taken by every command that goes through `resolveRevision` (`show`, `diff <rev>`, `log <rev>`, `merge-base`, `~`/`^` suffixes…).

## Why

`packages/webapp/src/git/commands/revision.ts` `resolveBase()`:

```ts
try {
  return await git.expandOid({ fs, dir, oid: ref });   // scans every pack index for a prefix match
} catch {
  return await git.resolveRef({ fs, dir, ref });
}
```

`expandOid` iterates all `.idx` files (`readPackIndex` for each) looking for an abbreviated oid before the symbolic ref lookup is even attempted, and `HEAD`/`main`/`origin/foo` are never valid hex prefixes.

## Fix

Only try `expandOid` when the token matches `/^[0-9a-f]{4,40}$/i`; otherwise (or first) `resolveRef`. If both could apply (a branch literally named `deadbeef`), git resolves the ref first and warns, so `resolveRef` → `expandOid` is the correct order anyway.

PR #2749, opened 2026-09-02
perf(git): walk all branch tips in one pass, memoize the object store

### LINKED ISSUE #2712: git: `log --all` takes 150+ s — `objects/pack` listed once per object, loose-object miss before pack lookup, fresh pack cache per branch

trieloff:
## Summary

`git log --all` over a `--mount`ed host repo took **157 s / 72,767 hostfs requests** for
`--oneline --all -n 20` (29 branches; native git: 51 ms). This makes it one traversal and
stops isomorphic-git from re-listing the object store once per object.

## Root cause

1. **`logAllBranches()` was 29 walks.** It called `git.log({ ref: branch, depth: 50 })` once
   per branch, each with its own implicit isomorphic-git `cache`, then concatenated,
   deduped, re-sorted by author date and only then applied `-n`. All 30 pack indexes were
   parsed — and the 92 MB pack re-verified — once per branch (`GitPackIndex.fromIdx` = 7.2 %
   of the profile), and `-n 20` still paid for `29 × 50` commits.
2. **Every object read cost two extra round trips.** `_readObject` probes the loose path
   first (one read → 404 on a packed repo) and then `readObjectPacked` does
   `fs.readdir(gitdir/objects/pack)` on *every* call — no listing cache anywhere. That is
   the 25,005 `list` + a large share of the 46,696 `read` requests in the issue's numbers,
   and it is also what makes `show --stat HEAD` (1,997 lists) and `diff --stat HEAD~1 HEAD`
   (1,466 lists) slow.

## What changed

- **`packages/webapp/src/git/commands/log-walk.ts` (new)** — one date-ordered traversal:
  seed a binary heap with every `listBranches()` tip, pop the newest commit, emit it, push
  its parents. Each commit object is read at most once (dedup by oid before the read), and
  the walk breaks *before* expanding the last emitted commit, so `-n 20` never reads a 21st
  generation. One shared `cache` is threaded through every read.
  - Ordering key is **committer date**, matching real git and isomorphic-git's own
    single-ref walker (`compareAge`); the old code sorted by *author* date, so `--all` on a
    one-branch repo could disagree with plain `git log` on it.
  - Second-resolution timestamps tie constantly. Ancestry orders itself (a parent is only
    queued once its child has been emitted); the remaining ties break by branch enumeration
    order then push order, which reproduces what the per-branch loop got from
    `listBranches()` order plus a stable sort — the `--all` + pathspec ordering test from
    #2714 still passes unchanged.
  - `--all` without `-n` now returns the newest **50** commits across all branches. The old
    code read 50 per branch and returned all of them unbounded; 50 is the same window for a
    one-branch repo and a bounded one for 29. `--all -- <path>` scans a 500-commit candidate
    window (the old bound was `50 × branch count`).
- **`packages/webapp/src/git/vfs-fs-adapter.ts`** — `beginObjectScope()` returns a disposer
  and, while open, memoizes (a) the `objects/pack` listing and (b) the `objects/` fan-out
  listing, which is enough to answer "is there a loose object at `objects/ab/…`?" without a
  round trip. Both hold the *pending promise*, so a concurrent fan-out shares one request
  instead of racing N identical ones, and `readdir` hands out a copy because isomorphic-git
  sorts the array it is given in place. **Any write through the adapter clears both maps**,
  so an object written mid-command is visible to the next read.
  - #2709's command-scoped cache has not landed, so this is scoped narrowly and only to the
    two object-store lookups; it is the obvious thing for #2709 to extend rather than
    duplicate. Pack SHA-1 verification (#2710) is untouched.
- **`packages/webapp/src/git/git-commands.ts`** — `execute()` opens the scope and closes it
  in the existing `finally`, so the lifetime is exactly one `git` invocation. This benefits
  every command, not just `log`.
- Docs: `docs/shell-reference.md` (new "`git log --all` walks every branch in one pass"
  section, next to the #2713/#2708/#2719 ones) and `docs/webapp-details.md` § Git.

## Measured effect

A throwaway benchmark in the test harness (21 branches × 10 commits = 210 commits, fully
packed, all objects behind the VFS, `git log --oneline --all -n 20`):

| | before | after |
| --- | ---: | ---: |
| VFS operations | 1,077 | **315** |
| `readDir` calls | 411 | **3** |
| `objects/pack` listings | 410 | **1** |
| loose-object probes (all 404s) | 410 | **0** |
| elapsed (in-memory FS) | 667 ms | **26 ms** |

Each of those 410 avoided listings and 410 avoided probes is an HTTP round trip through the
hostfs bridge on a real mount.

## How verified

- `npm run lint` — exit 0 (43 pre-existing warnings, no errors).
- `npm run test` — `18454 passed | 1 failed`: the one failure is
  `kernel/realm/slicc-fs-module.test.ts` timing out loading pyodide under full-suite
  parallelism; it passes on its own (`16 tests passed`) and is untouched by this change.
- `npm run test:coverage:webapp` — **exit 0**: 84.49 % lines / 82.49 % stmts / 82.37 % funcs
  / 73.96 % branches against floors 84 / 82 / 82 / 73. `log-walk.ts` is at 100 % lines,
  98.68 % statements. No threshold was lowered.
- `npm run build -w @slicc/webapp` — exit 0; `check-first-load-size.mjs` → "First-load OK".
- `node packages/dev-tools/tools/check-touched-exemptions.mjs` — "no debt lists found"
  (both baselines are empty on main).
- **Mutation check**: with the fix reverted and the new tests kept, 4 of the 10 new
  `git-log-all-traversal` tests fail (`c1: 2` vs `c1: 1` commit reads, 17 vs ≤10 object
  reads, 5 vs 1 `objects/pack` listings, and a commit read past `-n`) — the guards are real.

## Tests

- `packages/webapp/tests/git/git-log-all-traversal.test.ts` (10) — correct newest-first set
  across interleaved branches, dedup across branches, `-n` stopping the walk, one read per
  commit object, first-parent order for same-second merge parents, unresolvable/dangling
  branch refs skipped, and a genuinely **packed** fixture (`packObjects` + `indexPack`, loose
  copies deleted) asserting exactly one `objects/pack` listing and zero loose probes per
  command — plus that a commit made after packing is still found.
- `packages/webapp/tests/git/vfs-fs-adapter-object-scope.test.ts` (10) — scope semantics:
  one listing inside a scope and a fresh one after it closes, concurrent readers share one
  request, the returned array is a copy, nothing cached outside a scope, nesting and
  double-close, the loose-probe short-circuit and its inverse, and cache invalidation on a
  write inside the scope.

Fixes #2712


trieloff:
All three addressed in e174afe09.

### 1. `--all -- <path>` candidate window (Copilot + Codex, P2) — valid, removed entirely

You are right, and the failure mode is exactly as described: with a fixed window the answer depends on how busy the *other* branches are, and the match disappears with no diagnostic.

`logAllBranches` is now built on `walkAllBranches`, an **async generator**, and `logAllBranchesPathspec` consumes it until `-n` matches are found or every branch is exhausted. There is no global cap — commits are pulled in batches (`ALL_BRANCHES_PATHSPEC_BATCH = 64`) purely so the filter can prime its `commit oid -> tree oid` memo from a whole batch, exactly as the old materialized list did. `a..b` combined with `--all -- <path>` goes through the same stream with an exclusion set, so it lost its cap too.

Making the walk lazy also kept the `-n` guarantee: parents are expanded only *after* their commit is yielded, so a consumer that `break`s never pays for the next generation of commit reads. That is what the existing "stops the walk there" test pins.

Regression test, as Codex specified — `finds a pathspec match behind hundreds of newer commits on another branch`: branch `feature` tip touches `src/target.ts`, then 520 newer commits land on `main`, then `log --all -n 1 -- src/target.ts` must find the feature tip. **Mutation-checked**: re-adding a 500-commit cap fails exactly that test with `expected [] to deeply equal [ 'feature touches target' ]`. A companion test asserts the uncapped walk still stops early when the match is near the tip (< 100 object reads with 120 commits behind it), so "no cap" did not turn into "always walk everything".

### 2. Invalidation only before the async write (Copilot) — valid, fixed

Mutations now run through a `mutate()` helper that drops the memo *and* re-drops it in a `finally`, so a listing a concurrent reader took while the write was in flight cannot outlive it.

Test `drops a memo a concurrent read repopulated while a write was in flight` gates `vfs.mkdir` mid-flight, has a concurrent reader repopulate the fan-out memo from the pre-write state, then releases the write and asserts the next read sees the new object. **Mutation-checked**: dropping the `finally` fails it with `promise rejected "FsError: ENOENT" instead of resolving`.

### 3. Shared depth-counted scope — mirrored #2748's `contextFor` seam

Thanks for the pointer to 719e4543a. I went further than depth-counting: the scope handle is gone entirely.

- `createIsomorphicGitFs(vfs, { objectCache: true })` — the memo is now a `const` created with the adapter, so **its lifetime IS the adapter object's**. No `beginObjectScope()`, no disposer, no depth counter, nothing to forget to close, and the memo is collectible the moment the command returns.
- `GitCommands.contextFor()` builds one adapter per `execute()` and the `ctx` field is gone, matching #2748's shape (`private contextFor(): GitCommandContext` + `const ctx = this.contextFor()` in `execute`). When #2748 lands, the rebase should be adding the `command` parameter and the read-cache line to a method that already exists.
- `objectCache` is **off by default**, so the adapter's other callers (tests, `git-preview-source`) are unchanged and nobody accidentally gets a long-lived memo.
- Unlike #2748's read cache, this one needs **no subcommand allowlist**: `.git/objects` is written only through the isomorphic-git adapter (nothing reaches past it to `ctx.fs` for the object store — I checked), and every write through it drops the memo. Noted in the `contextFor` docstring so the rebase does not have to re-derive it.

Tests: `does not share the object memo between two commands running at once` parks a `git log --all` inside its `objects/pack` listing, runs a second command beside it, and asserts two listings rather than one; plus an adapter-level `gives each adapter its own memo`. **Mutation-checked**: caching one adapter on the instance fails the first with `expected [ '/project/.git/objects/pack' ] to have a length of 2 but got 1`.

### Verification

- `npm run lint` — exit 0
- `npm run typecheck` — clean, all 9 projects
- `npm run test` — **18461 passed | 0 failed** (the pyodide flake from the first run did not recur)
- `npm run test:coverage:webapp` — exit 0: 84.49 lines / 82.49 stmts / 82.36 funcs / 73.95 branches vs floors 84 / 82 / 82 / 73. `log-walk.ts` 100 % lines, `log.ts` 96.83 %, `vfs-fs-adapter.ts` 94.66 %. Nothing lowered.
- `npm run build -w @slicc/webapp` — exit 0; `check-first-load-size.mjs` → "First-load OK"
- `check-touched-exemptions.mjs` — "no debt lists found" (both baselines empty on main)
- `git tests` — 456 passed

Docs updated in both places for the new shape: `docs/shell-reference.md` gains the "bound by what the caller asked for, never by a candidate window" rule and the `{ objectCache: true }` lifetime, `docs/webapp-details.md` § Git the same as invariants.

Expecting to rebase onto #2748 once it lands.

PR #2750, opened 2026-09-02
perf(webapp): carry readDir stat fields so consumers stop re-stating entries

### LINKED ISSUE #2716: VFS/hostfs: `readDir` drops size/mtime on the way up, so every consumer re-stats each entry (N+1 round trips per directory)

trieloff:
## Summary

`readDir` now carries the stat fields the backend already paid for, and the three consumers that were re-stat'ing every entry they had just listed use them.

## Root cause

`/api/hostfs/list` stats every dirent server-side, so `size` + `lastModified` (and, since #2708, `ctime`/`ino`/`uid`/`gid`/`mode`) arrive with the listing, and `HostFsMountBackend.readDir` parses them into `MountDirEntry`. `VirtualFS.readDirMounted` then kept only `{name, type}` and the isomorphic-git adapter reduced that to names — so every consumer that needed a type or a size issued one `/api/hostfs/stat` per entry, each a full HTTP request plus a CORS preflight (#2715):

- `git branch`: 125 requests for 29 refs — 22 `list` + **100 `stat`** (`FileSystem.readdirDeep` stats every name to decide whether to recurse)
- Files panel (`dirChildren`): one stat per file, purely for its size
- `ls -l` / `du`: one stat per entry (just-bash's `DirentEntry` carries no size)

## What changed

**`fs/types.ts`** — `DirEntry` extends a new `DirEntryStats` (all optional: `size`, `mtime`, `ctime`, `ino`, `uid`, `gid`, `mode`), plus `statsFromDirEntry()`: the single rule for when a listing may stand in for a stat. It refuses a **symlink** (the listing describes the link, a `stat` describes its target) and anything missing `size`/`mtime`, because zeroed placeholders are exactly what make `compareStats` call a file stale and rewrite `.git/index` per file (#2708). `ctime` falls back to `mtime`, matching `VirtualFS.stat`.

**`fs/virtual-fs.ts`** — `readDirMounted` projects `MountDirEntry` onto `DirEntry` with the fields the backend reported; `statDirEntry` and `readDirSync` report what the `lstat` they already do returned (free), leaving symlink entries bare.

**`git/vfs-fs-adapter.ts`** — `readdir` primes a stat cache that `stat`/`lstat` answer from. A file is cached only when the listing carried the full set; a directory is cached type-only (all `readdirDeep`/`GitWalkerFs` ask of one, and directories never appear in `.git/index`). Any write through the adapter drops the path, `rmdir` drops the whole table (subtree), keys are normalized, and `GitCommands.execute` clears the cache in a `finally` so a listing never answers the *next* command.

**`shell/vfs-adapter.ts`** — `readdir`/`readdirWithFileTypes` prime the same kind of table for the mount slow path; `stat`/`lstat` answer from it for 1 s (`LISTING_STAT_TTL_MS`), and every mutation through the adapter clears it. That bounds the one case the adapter cannot see: a change made outside it between the listing and the stat.

**`ui/wc/wc-workbench.ts`** — reads `entry.size` from the listing, stats only the entries a listing left unknown.

**`kernel/messages.ts`** — `VfsDirEntryEnvelope` mirrors the optional fields. Additive: a peer that sends none is answered exactly as before.

**`docs/mounts.md`** — new "Listings carry stats (no re-stat per entry)" section under the wire shape, including the two invariants a backend has to keep.

Not in scope: both bridges still send a bare `{name, kind}` for directories, so a consumer that needs a directory's size or mtime still stats it (noted in the doc). `MountIndex`'s fast path (FSA mounts only) stores paths, not sizes, and is unchanged.

## How verified

- `npm run lint` — **exit 0** (43 pre-existing warnings, no errors)
- `npm run typecheck` — **pass** (all 9 projects)
- `npm run test` — **18463 passed, 3 failed**; the 3 are pre-existing and unrelated (2× `slicc-fs-module.test.ts` pyodide `loadPyodide` 5 s timeout — reproduced identically on a stashed, clean tree; 1× `service-worker-mount-sign-and-forward-port.test.ts`, passes on re-run). Touched suites (`tests/fs`, `tests/git`, `tests/shell`, `tests/kernel`, `tests/ui/wc/wc-workbench`) are green.
- `npm run test:coverage:webapp` — **exit 0** (821 files / 14810 tests; statements 82.45%, branches 74.04%, functions 82.33%, lines 84.47%)
- `npm run build -w @slicc/webapp` — **exit 0**
- `node packages/dev-tools/tools/check-first-load-size.mjs` — **OK**, worker +2.1 kB vs base, 82 kB under ceiling (allowance 5 kB)
- `node packages/dev-tools/tools/check-touched-exemptions.mjs` — no debt lists (both baselines empty); commit hook re-ran the layer-back-edge and `Record<string, unknown>` gates: ok
- `npm run deadcode -- --no-progress --reporter compact` and `npm run deadcode:production-files …` — **exit 0**

New tests (counting backends assert **requests**, not just values):

- `tests/fs/dir-entry-stats.test.ts` — `statsFromDirEntry` rules; mount/local/sync `readDir` carry the fields; a listing agrees with `stat` for the same path.
- `tests/git/vfs-fs-adapter-readdir-cache.test.ts` — a `readdirDeep`-shaped walk of a `refs/heads` tree costs **2 lists, 0 stats**; a cached answer is field-for-field the stat it replaced; a half-described entry, a `clearStatCache()`, and a write each cost a real stat; symlinks are never served from the cache.
- `tests/git/git-commands-stat-cache.test.ts` — `GitCommands.execute` clears the cache on both the success and the failure path.
- `tests/shell/vfs-adapter-listing-stats.test.ts` — `ls -l` over a mount = **1 list, 0 stats**, identical values; `du`'s `lstat` too; TTL expiry, write invalidation and an unlisted path all fall back.
- `tests/ui/wc/wc-workbench.test.ts` — the panel stats **0** files when the listing carries sizes, and still stats when it does not.
- `tests/kernel/vfs-rpc-host.test.ts` — the fields survive the `vfs-read-dir` round trip.

Fixes #2716

### DIFF
--- a/docs/mounts.md
+++ b/docs/mounts.md
@@ -204,6 +204,28 @@ The stable dispatcher owns its body end to end: it is excluded from node-server'
 
 A bridge that predates the stable endpoint answers `POST /api/hostfs` with a framework 404 carrying no errno `code`; `HostFsMountBackend` treats exactly that as "no stable endpoint" and falls back to the per-op routes for the rest of the mount's life, at a cost of one wasted request. Every error the real route emits carries a `code`, so a genuine `ENOENT` never triggers the downgrade.
 
+### Listings carry stats (no re-stat per entry)
+
+`list` stats every dirent server-side, so a listing already knows each file's `size`, `lastModified` and — since [#2708](https://github.com/ai-ecoverse/slicc/issues/2708) — its `ctime`, `ino`, `uid`, `gid` and `mode`. Those fields ride all the way up: `HostFsMountBackend.readDir` → `MountDirEntry` → `VirtualFS.readDir` → `DirEntry` (all optional, all best-effort), and the panel-side `vfs-read-dir` RPC mirrors them.
+
+They exist so a consumer never stats what it just listed ([#2716](https://github.com/ai-ecoverse/slicc/issues/2716) — `git branch` spent 100 of its 125 requests on stats of 29 refs):
+
+- **isomorphic-git** (`git/vfs-fs-adapter.ts`): `readdir` primes a stat cache that `stat`/`lstat` answer from, cleared by `GitCommands.execute` at the end of every command and dropped for any path the adapter writes.
+- **Shell** (`shell/vfs-adapter.ts`): `ls -l` and `du` are a `readdir` plus a stat per name (just-bash's `DirentEntry` has no size), so a listing answers stats of its own entries for 1 s and any write through the adapter drops the table.
+- **Files panel** (`ui/wc/wc-workbench.ts`): reads `entry.size` and stats only the entries a listing left unknown.
+
+Three rules keep this honest:
+
+- **The backend opts in.** `MountBackend.listingStatsMatchStat` means "my listing reports the same numbers my `stat()` would" — consumers use them _instead of_ a stat and cannot tell the difference. Only `hostfs` (both bridges stat each dirent inside `list`) and `local` (both paths read `FileSystemFileHandle.getFile()`) set it. **S3 and AEM deliberately do not**: both answer `stat` from the body cache when they have one, where `mtime` is `cachedAt` — when the body was cached, not the object's mtime — and AEM reports the _decoded_ size there while a Source Bus listing carries the stored (compressed) one. Aligning those two sources is a separate change; until then their listings stay `{name, type}` and consumers pay the stat rather than getting a different number. `VirtualFS.readDir` drops the fields for any backend that has not opted in.
+- **A partial entry is never promoted.** A raced file the bridge could not stat comes back as a bare `{name, kind}` and costs a real stat, because answering with zeroed placeholders is exactly what makes isomorphic-git call a file stale and rewrite `.git/index` per file (#2708). Both bridges omit `size`/`lastModified` rather than sending zeros — including for a dangling symlink, whose `stat` op answers ENOENT.
+- **Symlinks are never promoted.** The listing describes the link, a `stat` describes its target.
+
+The last two are enforced in one place: `statsFromDirEntry` in `fs/types.ts`.
+
+Both consumer caches are bounded, because a TTL limits how long an entry may be _reused_, not how long it is _retained_: the shell's table sweeps what has expired whenever it primes and holds at most 20,000 entries, the git adapter's at most 100,000, and a single listing larger than the cap primes nothing at all rather than blowing past it.
+
+Directories over hostfs carry no stat fields today (both bridges send `{name, kind}` for them), so a consumer that needs a directory's size or mtime still stats it.
+
 ## Common error patterns
 
 | Error                                                                                         | What it means                                                                                 | Fix                                                                                              |
--- a/packages/swift-server/Sources/Server/HostFSRoutes.swift
+++ b/packages/swift-server/Sources/Server/HostFSRoutes.swift
@@ -453,13 +453,35 @@ enum HostFSRoutes {
         let entries: [LickSystem.JSONValue] = names.map { name in
             let full = dir + "/" + name
             var isDirectory: ObjCBool = false
-            FileManager.default.fileExists(atPath: full, isDirectory: &isDirectory)
-            if isDirectory.boolValue {
+            // Follows symlinks, like node's `stat(resolve(target, d.name))`:
+            // a link to a directory classifies as a directory, and a DANGLING
+            // link (or an entry deleted since `contentsOfDirectory`) does not
+            // resolve at all.
+            let resolves = FileManager.default.fileExists(atPath: full, isDirectory: &isDirectory)
+            if resolves && isDirectory.boolValue {
                 return .object(["name": .string(name), "kind": .string("directory")])
             }
-            let attrs = try? FileManager.default.attributesOfItem(atPath: full)
-            let size = (attrs?[.size] as? NSNumber)?.doubleValue ?? 0
-            let mtime = (attrs?[.modificationDate] as? Date).map { $0.timeIntervalSince1970 * 1000 } ?? 0
+            // An entry we could not measure reports the NAME ONLY — never
+            // `size: 0, lastModified: 0`, and never the dangling LINK's own
+            // attributes (`attributesOfItem` does not follow links, so it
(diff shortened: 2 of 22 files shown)


trieloff:
All four addressed in `a4ec75b` (branch rebased onto `main`, so the first commit is now `fba9067`).

**(1) P1 — shell listing cache retained without bound.** Correct, and the reason it was invisible: the TTL bounds how long an entry may be *reused*, not how long it is *retained* — expiry was only noticed on a same-path lookup, so an `ls -R` / `du -sh` walk piled every directory it ever touched into a map that lives as long as the shell. Priming now does both things you asked for: it sweeps what has expired, and it enforces a hard cap (`MAX_LISTING_STATS`, 20,000). The sweep is cheap because insertion order is age order — priming `delete`s a key before re-`set`ting it (a plain `set` on an existing key keeps its old position), so the scan stops at the first live entry and costs one step per entry actually removed. Tests: a 20-directory walk with the cap pinned to 10 asserts `listingStatsSize <= 10` after every listing, and a separate test proves the sweep evicts d0's expired entries when an unrelated directory is primed, while a fourth proves it does **not** evict entries still inside the TTL.

**(2) P2 — oversized listing blew the cap in both adapters.** Right, and worse than "clears then inserts": clearing to make room and inserting anyway leaves `entries.length` entries behind, i.e. no cap at all for exactly the directory that needs one. Both adapters now clear and prime **nothing** when a single listing exceeds the cap; the subsequent stats simply go to the backend. Tested on both sides (`statCacheSize()` / `listingStatsSize` is 0 and the next stat costs a request).

**(3) P2 — Swift/node listing parity.** Fixed, and the divergence was a bit wider than the zeros: `attributesOfItem` does not follow symlinks, so for a **dangling** link Swift reported the link's own size/mtime for a path whose `stat` op answers ENOENT — a listing that disagrees with `stat`, which is the exact hazard behind (4). `listResponse` now uses the following existence check (mirroring node's `stat(resolve(target, d.name))`) and reports the name only when the entry does not resolve or cannot be measured. New `testListOmitsMetadataForAnEntryItCannotStat` asserts `size`/`lastModified`/`ctime`/`ino`/`mode` are all absent for a dangling link, that a measurable sibling is unaffected, and that the `stat` op 404s for the same path. `swift build` + `swift test --filter HostFSRoutesTests` — 12 tests, 0 failures.

**(4) Divergent stat sources on S3/AEM.** Took the **narrower option: gated the passthrough**, did not align the backends. `MountBackend.listingStatsMatchStat` is an opt-in meaning "my listing reports the same numbers my `stat()` would"; only `hostfs` (both bridges stat each dirent inside `list`) and `local` (both paths read `FileSystemFileHandle.getFile()`) set it, and `VirtualFS.readDir` drops the fields for everything else. S3 (`backend-s3.ts:369`, `mtime: cached.cachedAt`) and AEM (`backend-aem.ts:305`, decoded body size vs the listing's stored size) therefore keep the historical `{name, type}` listing and keep paying the stat. Aligning their two sources is a real change to the body-cache/ETag semantics and belongs in its own PR — noted in `docs/mounts.md` with the reason. Test: a backend with the flag off has its listing fields dropped even though it reported them.

Each new guard was mutation-tested rather than just observed green — reverting the sweep, the oversized-listing guard (TS, both adapters) or the Swift guard fails exactly one test and no others; the Swift mutation reproduces the old `size: 108` from the dangling link.

**Verification** (full pass, then typecheck + the touched suites again after the rebase):

- `npm run lint` — exit 0 · `npm run typecheck` — exit 0 (9 projects, re-run post-rebase)
- `npm run test` — 18472 passed, 1 failed: `chrome-extension/tests/fetch-proxy-shared.test.ts`, unrelated to this diff and green on re-run (33/33). Post-rebase re-run of `tests/fs`, `tests/git`, `tests/shell`, `tests/kernel`, `tests/ui/wc/wc-workbench`: **7430 passed, 0 failed**.
- `npm run test:coverage:webapp` — exit 0 (statements 82.49%, branches 73.99%, functions 82.42%, lines 84.49%)
- `npm run build -w @slicc/webapp` — exit 0 · `check-first-load-size.mjs` — OK, page +0.0 kB, worker +2.9 kB vs base (81 kB under ceiling)
- `check-touched-exemptions.mjs` — no debt lists; commit hook's layer-back-edge + `Record<string, unknown>` gates ok · `deadcode` + `deadcode:production-files` — exit 0
- `swift build` — Build complete · `swift test --filter HostFSRoutesTests` — 12 tests, 0 failures

PR #2751, opened 2026-09-02
perf(git): share one object/pack cache across git commands

### LINKED ISSUE #2710: git: every command re-downloads and SHA-1-verifies the whole packfile; no cross-command object/pack cache

### LINKED ISSUE #2735: git: isomorphic-git's once-per-packfile SHA-1 verification runs once per *concurrent reader* — the slower the pack read, the more times the 92 MB pack gets hashed

trieloff:
## Summary

Every `git` subcommand handed isomorphic-git a **fresh** `cache` object (only `clone` passed one at all), so each command re-read the whole `.pack` off the VFS, re-parsed every `.idx`, and re-ran the deep SHA-1 integrity check over the pack payload. On the slicc checkout (92 MB largest pack, 30 packs) that was **47 % of a warm `ls-files`'s worker CPU (5.2 s) in native `digest`**, and 14.1 s of `digest` + 11.3 s of `GitPackIndex.fromIdx` for `log --all`.

`GitCommands` now owns **one cache per instance** and threads it into every isomorphic-git call that accepts one. Fixes #2735 in the same code path.

## Root cause

- `packages/webapp/src/git/commands/*` never passed `cache:`, and isomorphic-git parks its parsed `.idx` files (`PackfileCache`) and read `.pack` buffers **on the caller's cache object** — so their lifetime was one API call.
- `readObjectPacked` guards its packfile verification with `p._checksumVerified`, which is only set **after** the `await`. `_walk` fans object reads out with `Promise.all`, so every reader queued behind the pack read saw `false` and started its own SHA-1 over the whole pack (5,195 ms on hostfs vs 603 ms for one hash → the 92 MB pack was hashed 8–9 times; the slower the transport, the more duplicates). — #2735

## What changed

**`packages/webapp/src/git/git-cache.ts` (new)** — `GitCacheManager` owns the instance-wide cache and the bookkeeping that makes it safe to keep:

- **Invalidation**: the `objects/pack` listing + `packed-refs` mtime are re-sampled before a command, but **only for a repository the cache already holds packs for** — so a command that never reads a packfile still never goes near `objects/pack` (#2713's property is preserved). `fetch` / `pull` / `clone` drop the repo's packs outright. isomorphic-git's index cache self-invalidates on `lstat` and is deliberately left alone.
- **Memory bound**: at most `DEFAULT_MAX_RESIDENT_PACKS` (4, configurable via `GitCommandsOptions.maxResidentPacks`) pack **buffers** stay resident. Least-recently-used are unloaded; their parsed index — the expensive half to rebuild — stays cached. Eviction only runs with no command in flight, because `GitPackIndex.readSlice` awaits `this.pack` once per read.
- **Verification switch**: `skipDeepPackVerification` on the cache object, read by the patch below.

**`commands/types.ts` + all command modules** — `ctx.cache` added to `GitCommandContext` and threaded into all 114 isomorphic-git call sites that accept a cache. The per-command caches in `diff.ts` / `log.ts` / `clone.ts` now point at it.

**`patches/isomorphic-git+1.41.9.patch` (new)** — 1.41.9 exposes no option for either half, so this goes through the repo's patch-package flow (manifest entry + `renovate.json` routing added; `npm run lint:patches` green):

- Extracts the verification into `verifyPackfileIntegrity()` and **memoizes it as a promise on the cached pack index** (`p._verification`) → N concurrent readers verify exactly once. **#2735**
- Skips the full-payload SHA-1 when the caller's cache carries `skipDeepPackVerification`. The **O(1) trailer check always runs**, which is already more than canonical git does on an object read (it verifies packs on `fsck` / `index-pack`). SLICC sets it by default; `SLICC_GIT_VERIFY_PACKS=1` (or `GitCommandsOptions.verifyPackfiles`) opts back in.
- Stamps a monotonic `_lastUsedAt` on each pack index it serves a read from — the LRU signal the resident-buffer bound needs.

**Upstream**: both hunks are small upstream fixes. The memoization is a straight bug fix (`p._verification ??= verify(pack)`); the skip wants an isomorphic-git option (`verifyPackfiles: false`, or verification only on `indexPack`). `removeWhen` in `patches/patches.json` records both conditions; an upstream issue/PR still needs to be filed against isomorphic-git.

Only `index.js` (the ESM entry) is patched: the browser build and vitest both alias `isomorphic-git` to it on purpose, because the CJS entry imports Node crypto (`tests/git/isomorphic-git-esm.test.ts` pins that).

**Docs** — `docs/webapp-details.md` (Git section) and `docs/shell-reference.md` (new "Packfiles are read once per shell" section, incl. `SLICC_GIT_VERIFY_PACKS`).

Not touched: the loose-before-packed lookup and the `log --all` structure (#2712 is a separate PR).

## Tests

`packages/webapp/tests/git/git-pack-cache.test.ts` (new, 5 tests). The fixtures pack their objects for real (`packObjects` + `indexPack`, loose copies deleted) so the reads under test go through `readObjectPacked`; assertions count VFS reads and `crypto.subtle.digest` calls sized to the pack payload, because none of this cost is visible in stdout.

1. two consecutive commands read the `.idx` and the `.pack` exactly once each;
2. a changed pack directory forces a re-read of the cached index — and the refreshed cache then holds again;
3. the deep verification never runs by default, runs **once** with `verifyPackfiles: true`, and once via `SLICC_GIT_VERIFY_PACKS=1`;
4. **8 concurrent `readCommit` calls on a fresh cache hash the pack exactly once** (#2735);
5. with `maxResidentPacks: 1`, the least-recently-**used** pack buffer is unloaded (and its index kept) — the pack used first *and* last stays resident, which insertion order would get wrong.

Each was mutation-checked: reverting the isomorphic-git patch fails 3, 4 and 5; always-invalidating fails 1 and 5; never-invalidating fails 2.

## How verified

| Command | Result |
| --- | --- |
| `npm run lint` / `npm run verify` | pass — "All 7 checks passed" |
| `node packages/dev-tools/tools/check-touched-exemptions.mjs` | pass (0 new back-edges, 0 new `Record<string, unknown>`) |
| `npm run lint:patches` | pass — 5 patches OK, `isomorphic-git@1.41.9` matches installed |
| `npm run test` | 18,436 passed / 2 failed — both pre-existing load flakes (`wc-attach` `beforeAll` import, `slicc-fs-module` pyodide), green when re-run in isolation |
| `npm run test:coverage:webapp` | pass — 82.47 stmts / 73.99 branch / 82.36 funcs / 84.48 lines vs floors 82/73/82/84; `git-cache.ts` 96.25 % stmts |
| `node packages/dev-tools/tools/check-first-load-size.mjs` | "First-load OK (allowance 5 kB per change; total 4966 kB)" |

Not verified against a live 92 MB hostfs repo — the profile numbers above are from the issue; the tests prove the read counts, not the wall-clock win.

Fixes #2710
Fixes #2735


trieloff:
Both Codex findings fixed in e55486d70 — thanks, they were real.

**P1 — a rejected `.idx` read was memoized forever** (`git-cache.ts`). isomorphic-git stores the in-flight *promise* in `cache[PackfileCache]`, so one transient bridge `EIO` (the #2720 class) was replayed to every later command, and `beforeCommand` saw an unchanged pack-dir signature with nothing to invalidate — a shared cache loses the retry the old per-API-call caches got for free. `GitCacheManager` now runs `evictFailedPacks()` when a command settles (nothing in flight, so every promise inspected has settled): rejected entries are deleted, an `.idx` that resolved `undefined` (not IDX v2 — a half-written pack dir, not a verdict) is deleted, and a failed or empty `.pack` buffer is reset to `null` so the next read re-fetches it.

**P2 — the verification memo ignored which mode created it** (patch, hand-regenerated). `p._verification` was reused even when it had skipped the payload hash, so `SLICC_GIT_VERIFY_PACKS=1` later in the same shell silently verified nothing. The memo now stamps `p._deepVerified` and re-verifies when the caller wants the deep check and the memo did not include it. `deep` is passed into `verifyPackfileIntegrity()` rather than re-read from the cache, so the promise and the flag it was created under cannot disagree, and the #2735 guarantee is intact — the flag is set *before* the await, so concurrent readers still hash once (test 4 unchanged and still green).

**Tests** (now 6 in `git-pack-cache.test.ts`):
- an `.idx` read that rejects once then succeeds → the first command errors, the **second succeeds**;
- a default (skipped) read followed by an opt-in read **on the same cache** → the payload digest actually runs, exactly once, and a third verified command does not re-hash.

Both mutation-checked: commenting out `evictFailedPacks()` fails the retry test; restoring the old `if (!p._verification)` fails the opt-in test.

**Re-verified**

| Command | Result |
| --- | --- |
| `npm run verify` (incl. `lint:patches`) | pass — "All 7 checks passed"; 5 patches OK |
| `check-touched-exemptions.mjs` | pass (0 new back-edges, 0 new `Record<string, unknown>`) |
| `npm run test` | **18,444 passed, 0 failed** (the two flakes from the first run did not recur) |
| `npm run test:coverage:webapp` | pass — 82.50 stmts / 74.04 branch / 82.45 funcs / 84.52 lines vs floors 82/73/82/84 |
| `check-first-load-size.mjs` | "First-load OK (total 4967 kB)" |

Docs updated in the same commit (`docs/webapp-details.md`, `docs/shell-reference.md`) and the `patches.json` reason now records the mode-aware memo. Head is e55486d70.

PR #2752, opened 2026-09-02
feat(hostfs): serve ranged reads so a big packfile is reachable by git

### LINKED ISSUE #2711: hostfs: the 100 MB body cap makes any repo with a packfile > 100 MB unreadable by git (and every pack is read into worker memory)
Part of #2707.

## What happens

`HOSTFS_MAX_BODY_BYTES = 100 * 1024 * 1024` is enforced on both sides (`packages/node-server/src/hostfs.ts` returns `413 EFBIG`; `packages/webapp/src/fs/mount/backend-hostfs.ts` throws `EFBIG`). isomorphic-git reads packfiles whole (`readObjectPacked`: `fs.read(packFile)` → on failure `InternalError("Could not read packfile … too large to read into memory")`).

The slicc checkout's largest pack is 92 MB today, so `git log` still works — after `git gc` or a couple more months of history it will not, and the failure mode is "every git command errors out" (any object lookup iterates every pack index and the first pack containing the oid is read whole).

Even below the cap, every command that touches the big pack allocates ≥92 MB in the kernel worker (plus the `Buffer` polyfill copies seen in the profile: `Buffer.slice`/`Buffer.from` ~5 s in `log --all`), and the read goes through the browser HTTP cache as a single 92 MB entry.

## Fix ideas

- Support `Range` requests on `/api/hostfs/read` and give the isomorphic-git fs adapter a `read(path, {start,end})`-style path (isomorphic-git's own node build uses a chunked implementation for large packs — see the "The Node package build swaps in a chunked implementation" comment next to `shasumRange`).
- At minimum: raise the cap for `.git/objects/pack/*.pack` or stream the body instead of buffering it in Express (`res.send(body)` after `readFile`) and in the worker (`response.arrayBuffer()`).
- Surface a readable error in the git command layer instead of isomorphic-git's `InternalError`.

trieloff:
Fixes #2711 (part of #2707).

## Summary

`GET /api/hostfs/read` now speaks HTTP `Range` and streams every body off disk, and the ranged read is plumbed through the mount backend, `VirtualFS`, `RestrictedFS`/sudo-fs and the isomorphic-git fs adapter. Before: a `--mount`ed checkout whose largest packfile crossed the 100 MB body cap made **every** git command fail with `InternalError: Could not read packfile …`, and a pack under the cap still cost its full size in kernel-worker memory on each object lookup. Now: the bridge will serve any window of any file, the cap applies only to an unranged whole-file read, and the packfile error names the pack and the limit instead of `InternalError`.

## Why

`HOSTFS_MAX_BODY_BYTES = 100 MiB` was enforced on both sides (node-server `413 EFBIG`, `backend-hostfs.ts` `EFBIG`). isomorphic-git reads a pack whole (`readObjectPacked` → `fs.read(packFile)`) and turns a failed read into `InternalError`, so the cap was a hard wall. The slicc checkout's largest pack is 92 MB today — one `git gc` away from unusable.

Even below the cap, the #2707 benchmark shows the cost: express buffered 92 MB per read (`readFile` + `res.send`), the kernel worker materialized it again via `response.arrayBuffer()` (plus `Buffer` polyfill copies — `Buffer.slice`/`Buffer.from` ≈ 5 s in `log --all`), and it sat in the browser HTTP cache as one 92 MB entry.

**Repro** (before this PR): `npm run dev -- --mount=<repo-with-a->100MB-pack>:/mnt/x`, then `git -C /mnt/x log`.
Expected: the log. Actual: `fatal: Could not read packfile at /mnt/x/.git/objects/pack/pack-….pack. The file may be missing, corrupted, or too large to read into memory.`

**Wire** (`node-server/src/hostfs.ts`, mirrored in `swift-server/HostFSRoutes.swift`)
- `parseByteRange(header, size)` — one `bytes=` window, all three shapes (`a-b`, `a-`, `-N`), clamped to `[0, size-1]`. Malformed → ignored and the whole file is served (RFC 9110 §14.2); well-formed but outside the file → `416` + `Content-Range: bytes */<size>` + `EINVAL`, because answering *that* with the whole file would hand a pack reader bytes from the wrong offsets.
- `streamFileBody` replaces `readFile` + `res.send`. Headers are committed on the stream's `open`, so a failed open still rejects before anything is on the wire and maps to the usual errno JSON; after that the status line is spent, so a mid-body failure destroys the connection (which the client's existing retry path from #2720 already handles). `sendFsError` now no-ops into a `destroy()` once `headersSent`.
- **No size cap on a ranged read** — the window is its own bound and the body is streamed, so neither side ever holds more than the bytes asked for. That exemption is the actual fix for #2711. The cap survives for the unranged read, whose caller must hold the whole body.
- The read route moved out of `registerHostFsRoutes` into a module-level `readOp` to stay under the 150-line function cap.

**Client** (`backend-hostfs.ts` → `backend.ts` → `virtual-fs.ts` → `vfs-fs-adapter.ts`)
- `readFileRange(path, start, end)` is **half-open** (`subarray` semantics) everywhere in JS and converts to the wire's inclusive `bytes=start-(end-1)` at the boundary — chosen to match isomorphic-git's own `shasumRange({ start, end })`.
- A bridge that ignores `Range` answers `200` with everything; the backend detects the missing `206` and slices client-side, so an older launcher behind a fresh hosted UI stays *correct*, just not fast.
- `MountBackend.readFileRange` is optional; `VirtualFS.readFileRange` falls back to read-and-slice so callers never branch on backend capability.
- The adapter honors `readFile(path, { start, end })` because `FileSystem.read(filepath, options)` forwards `options` verbatim — a window wins over an encoding (a pack slice is bytes).

**Error surface** — `packfileReadHint` in `git/commands/shared.ts` rewrites both isomorphic-git packfile `InternalError`s into a message that names the pack and points at the mount/limit, instead of the generic "missing, corrupted, or too large to read into memory".

### What is deliberately NOT done: the upstream half

isomorphic-git 1.41.9 **cannot** consume a ranged reader, and I confirmed this in the installed build rather than assuming:
- `readObjectPacked` calls `fs.read(packFile)` with no options at all.
- `GitPackIndex.readSlice({ start })` does `pack.slice(start)` — the *entire remainder* of the pack — because it has no object end to bound a window with; it discovers the length only after inflating.
- The "deep integrity check" SHA-1s the whole payload. The comment near `shasumRange` ("the Node package build swaps in a chunked implementation for large packs") is about chunked **hashing of a buffer that is already resident** — `index.cjs` differs from `index.js` only in that it feeds `crypto.createHash('sha1')` in `SHA1_CHUNK_SIZE` slices. It does not read in chunks.

Teaching `readSlice` to work from an async windowed reader is a real upstream change (growable windows, a chunked trailer/payload verify), not a patch-package one-liner — and it belongs upstream anyway, since the browser build hits the same wall against any HTTP-backed fs. I did **not** add a `patches/` entry for it. Everything below isomorphic-git is now ready for that change, and until it lands an oversized pack still fails — with a readable error rather than `InternalError`. Written up in `docs/mounts.md`.

## Cross-cutting impact

- **Floats exercised**: CLI (node-server bridge) and Sliccstart (swift-server) — both bridges changed in lockstep; the swift `parseByteRange` test asserts the same table as the node one. Extension / hosted have no local bridge and are unaffected.
- **Other callers of the changed contract**: `sendFsError` gained a `headersSent` guard — reached by every hostfs route, but `headersSent` is only ever true on the newly-streamed read path. `MountBackend` gained an *optional* member, so `backend-local`/`s3`/`da`/`aem`/`proc` are untouched and take the `VirtualFS` read-and-slice fallback.
- **Security**: `readFileRange` is added to sudo-fs's `READ_ASYNC` list and to `RestrictedFS`. The sudo Proxy only intercepts the method names it lists, so a ranged read left off that list would have been a free pass around a `Read` policy — there is a test for exactly that. `RestrictedFS.readFileRange` gates identically to `readFile` (ENOENT for out-of-sandbox and traversal); it deliberately does **not** serve virtual devices or ephemeral fds, which have no stable byte offset.
- **Behavior change for existing clients**: an unranged read now carries `Accept-Ranges: bytes` and an explicit `Content-Length`, and the 413 message changed. No caller parses either.
- **Persisted state / bundle size**: none. First-load gate unchanged (`total 4964 kB`).

## Test plan

All run in this worktree at `82ecdc2`:

- [x] `npm run lint` — 0 errors (43 warnings, all pre-existing)
- [x] `npm run typecheck` — clean (all 9 projects)
- [x] `npm run test` — **18475 passed**, 52 skipped, 0 failed
- [x] `npm run test:coverage:node-server` — lines 83.97 / statements 81.94 / functions 85.81 / branches 72.66 (floors 83 / 81 / 84 / 71)
- [x] `npm run test:coverage:webapp` — lines 84.46 / statements 82.45 / functions 82.31 / branches 73.93 (floors 84 / 82 / 82 / 73)
- [x] `npm run build -w @slicc/webapp` and `npm run build -w @slicc/node-server` — both OK
- [x] `node packages/dev-tools/tools/check-first-load-size.mjs` — `First-load OK (allowance 5 kB per change; total 4964 kB)`
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — passes; none of the touched files are on a biome debt list
- [x] `npm run deadcode -- --reporter compact` and `npm run deadcode:production-files` — both clean
- [x] Swift: `swift build` OK; `swift test --filter HostFSRoutesTests` — **13 passed, 0 failures**; `swiftlint` and `swift format lint --strict` clean for the files touched (one pre-existing warning at `HostFSRoutesTests.swift:163` left alone)

New / changed coverage:

- `packages/node-server/tests/hostfs.test.ts` — route: 206 + `Content-Range`/`Content-Length`/`Accept-Ranges`, open-ended + clamped + suffix ranges, `416` + `Content-Range: bytes */<size>`, unparseable Range ignored, errno answers preserved under a Range. Unit: `parseByteRange` across all shapes, including a >2 GiB offset.
- `packages/webapp/tests/git/vfs-fs-adapter-range.test.ts` (new) — proves a ranged read **never requests the whole body** (the fake backend records every whole-file read and asserts it stayed empty), plus the read-and-slice fallback, clamping, zero-length windows, EINVAL, and error rebranding.
- `packages/webapp/tests/fs/backend-hostfs.test.ts` — the exact `Range` header sent, the 200-without-`206` client-side slice, EINVAL/EFBIG/EBADF, and that the cap bounds the *window* and not the file (a 512 MB offset is accepted).
- `packages/webapp/tests/fs/sudo-fs.test.ts`, `restricted-fs.test.ts` — the two policy boundaries above.
- `packages/webapp/tests/git/git-error-format.test.ts` — the rewritten packfile messages, incl. inside a `MultipleGitError`, and that an unrelated packfile error is left alone.
- `packages/swift-server/Tests/HostFSRoutesTests.swift` — the same range table and route behavior as node-server.

**Not done**: no manual smoke against a live instance with a >100 MB pack — the fix's remaining blocker is the isomorphic-git limitation described above, so a live run would still fail at the same point, just with the better message.

**Pre-existing failures**: none introduced. One flake seen once under `test:coverage:webapp` load — `kernel/realm/slicc-fs-module.test.ts > write_bytes: payload reaching _bridge.writeBytes is a real JS Uint8Array` (a pyodide test) — passed on re-run and in every `npm run test`; unrelated to this change.

## Documentation

- [x] `docs/mounts.md` — new "Ranged reads" subsection under the wire shape (semantics, why the cap is exempt, the upstream gap), plus an error-table row for the new packfile message
- [x] `packages/node-server/CLAUDE.md` and `packages/swift-server/CLAUDE.md` — Range support in the mount-table paragraphs
- [x] No `README.md` change — no user-facing command or flag changed

## Risk & rollback

Blast radius is the hostfs bridge's `read` route on both servers plus one new optional method on the mount/VFS surface. No flags, no persisted state, no migrations — a clean revert. The riskiest edge is the streamed-body error path (a failure after headers are committed becomes a truncated body rather than an errno JSON); that is what the client's `attemptRequest` retry from #2720 already exists for, and it can only trigger on an I/O failure between a successful `stat` and the end of the read.


trieloff:
Both findings were real. Fixed in `15ff86d8e`.

## P1 — streaming dropped the ETag (`hostfs.ts`)

Correct, and it was the worst possible regression to introduce here: `res.send(body)` had been handing express a buffered body to hash into an ETag for free, and `createReadStream` removed the buffer and the validator with it. A pack URL that cannot be revalidated is re-transferred on every object lookup — the opposite of the fix — and your 220,310/385,033 number is exactly the traffic that was at stake.

`cacheValidator(s)` now derives the validators from the `stat` the route already does, and they ride on **every** answer (200, 206, 416, 413 and 304), so a client can refresh its stored metadata off any of them:

- `ETag: "<size>-<mtimeMs>-<ino>"`, hex, and **strong** — no `W/`. That was deliberate: `If-Range` is only defined for a strong validator (RFC 9110 §13.1.5), and the ranged pack read is the case that matters most here. What justifies calling it strong is what goes into it — inode, size, and the unrounded high-resolution mtime, the same basis nginx and `serve-static` use. The honest limit is in the doc comment: a write landing on the same inode with an identical size *and* a deliberately restored mtime is invisible to it. Nothing git does looks like that (packs are written under a new name), and the alternative — hashing the body — is precisely the buffering this PR removed.
- `Last-Modified`, floored to the second (all an HTTP-date can carry).
- `If-None-Match` → 304, using the **weak** comparison the spec mandates for it, so a cached `W/"…"` form of our tag still matches; `*` and comma-separated lists handled. `If-Modified-Since` is consulted only in its absence (§13.2.1 precedence), and an unparseable date is treated as no claim at all rather than a guess.
- The 304 goes out via `.end()`, not `.json()` — no body, and no Content-Type for content that isn't there.
- `If-Range` gates whether the `Range` is honored at all, with the **strong** comparison; a mismatch quietly downgrades to a full 200. That ordering matters beyond the obvious: it means a stale `If-Range` carrying an out-of-file window yields a 200, not a bogus 416, because the Range is dropped before `parseByteRange` ever sees it. There's a test pinning that specific interaction.

Tests in `packages/node-server/tests/hostfs.test.ts` (9 new): validators present on both the whole-file and ranged branches and the tag is strong; repeat GET with the ETag → 304 with an empty body; weak/`*`/comma-list matching and a stale tag → 200; `If-Modified-Since` → 304, then `utimes` forward → 200 and the *new* tag revalidates; unparseable `If-Modified-Since` ignored; `If-Range` match → 206; mismatch → 200 with no `Content-Range`; the HTTP-date form of `If-Range` both fresh and stale; and the stale-`If-Range`-plus-bad-window → 200 case above.

## P2 — the Swift window was materialized whole (`HostFSRoutes.swift`)

Also correct — `Data(contentsOf:)` before the `ByteBuffer` meant the Swift side had removed the 100 MiB guard without replacing it, so `bytes=0-` on a multi-GB file would have allocated all of it. I went with streaming rather than a per-window cap, so the two bridges keep the same contract.

Windows now go out through Hummingbird's closure-backed `ResponseBody(contentLength:)`, writing `streamChunkBytes` (1 MiB) at a time; the unranged read is streamed the same way, so it no longer buffers up to 100 MiB either. The `FileHandle` is opened **eagerly**, before the body closure runs, so an ENOENT/EACCES still becomes errno JSON before any header is on the wire — matching how node-server commits headers on the stream's `open`.

New Swift test `testLargeAndOpenEndedRangesAreStreamed` writes a 2 MiB+ patterned file and checks an open-ended `bytes=0-`, a window straddling several chunk boundaries, and the unranged read. I mutation-tested it rather than trusting a green run: adding a `break` after the first `writer.write` fails all three assertions (`1048576 bytes` vs `2109497 bytes`), so it genuinely exercises the multi-chunk loop.

**Parity note (slightly beyond the finding):** swift-server never had an ETag either — Hummingbird doesn't synthesize one — so that half was a pre-existing gap rather than something this PR broke. But `HostFSRoutes.swift` is documented as mirroring `hostfs.ts` 1:1, and leaving Sliccstart unable to revalidate a pack while the CLI can is the cross-runtime-parity trap in the review checklist. So `cacheValidator`/`isNotModified`/`ifRangeAllowsRange` are mirrored on the Swift side too, with `testConditionalRequests` asserting the same matrix and `testChangedMtimeInvalidatesTheValidator` covering invalidation on both mtime and size.

The `RestrictedFS`/sudo-fs gating is untouched, as asked.

## Re-verified at `15ff86d8e`

| Gate | Result |
|---|---|
| `npm run lint` | 0 errors |
| `npm run typecheck` | clean |
| `npm run test` | **18484 passed**, 52 skipped, 0 failed |
| `test:coverage:node-server` | lines 84.07 / stmts 82.05 / funcs 85.90 / branches 72.86 (floors 83 / 81 / 84 / 71) — `hostfs.ts` 91.10 % lines |
| `test:coverage:webapp` | lines 84.48 / stmts 82.46 / funcs 82.42 / branches 74.02 (floors 84 / 82 / 82 / 73) |
| builds | `@slicc/webapp` + `@slicc/node-server` both OK |
| first-load / boy-scout | OK (total 4964 kB, unchanged); no touched file on a debt list |
| `deadcode` + `deadcode:production-files` | clean |
| Swift | `swift build` OK; `HostFSRoutesTests` **16 passed, 0 failures**; swiftlint + `swift format lint --strict` clean (one pre-existing `optional_data_string_conversion` warning at `HostFSRoutesTests.swift:292`, in code this PR didn't add) |

Docs updated in the same commit: `docs/mounts.md` gains the conditional-request paragraph, and both `CLAUDE.md`s note the validator + Swift chunked streaming.

Still unchanged from the PR description: the isomorphic-git side. 1.41.9 cannot consume a ranged reader (`readObjectPacked` passes no options, `readSlice` does `pack.slice(start)` over the whole remainder, and the "chunked" `shasumRange` only chunks *hashing* of an already-resident buffer), so a >100 MB pack still fails — with the readable message rather than `InternalError`. That remains an upstream change.

PR #2755, opened 2026-09-02
perf(git): persist pack caches across commands

### LINKED ISSUE #2710: git: every command re-downloads and SHA-1-verifies the whole packfile; no cross-command object/pack cache

trieloff:
Closes #2710

## Solution
Retain isomorphic-git's pack cache per repository, invalidating it when pack or packed-ref metadata changes; concurrent readers still complete one shared SHA-1 integrity check.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01M1GP0W8YF6WWYQQ380DCGABK&panel=chat)

### DIFF
--- a/packages/webapp/src/git/cached-isomorphic-git.ts
+++ b/packages/webapp/src/git/cached-isomorphic-git.ts
@@ -1,0 +1,149 @@
+/**
+ * isomorphic-git facade with a persistent object/pack cache per repository.
+ *
+ * The upstream API creates a new cache object whenever a caller omits one.
+ * GitCommands binds its per-repository cache to the filesystem view for each
+ * invocation; cache-aware calls below then receive it automatically.
+ */
+
+import * as upstream from 'isomorphic-git';
+import { normalizePath } from '../fs/path-utils.js';
+import type { IsoGitFsPromises } from './vfs-fs-adapter.js';
+
+export * from 'isomorphic-git';
+
+const cacheForFs = new WeakMap<object, () => Promise<object>>();
+
+interface CacheEntry {
+  fingerprint: string;
+  cache: object;
+}
+
+/** Stable missing-file errors; any other failure disables reuse for safety. */
+const MISSING_CODES = new Set(['ENOENT', 'ENOTDIR']);
+
+function errorCodeOf(error: unknown): string | undefined {
+  const code = (error as { code?: unknown } | null)?.code;
+  return typeof code === 'string' ? code : undefined;
+}
+
+async function statFingerprint(fs: IsoGitFsPromises, path: string): Promise<string | null> {
+  try {
+    const stat = await fs.stat(path);
+    return `${stat.mtimeMs}:${stat.size}`;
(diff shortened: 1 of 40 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01M1GP0W8YF6WWYQQ380DCGABK)⚡</sup>

👋 I've got this PR.

trieloff:
Closing as superseded: #2710 was fixed and merged in #2751 (cross-command object/pack cache with pack-dir/`packed-refs` invalidation and a bounded LRU of resident pack buffers), which also carries the #2735 verification-race fix. Verified on merged main — warm `git log --all` went from 638 MB transferred to 6.5 MB; numbers on #2707. Thanks — reopen if you think something here is still missing.

PR #2759, opened 2026-09-02
perf(fs): index mount directory children

### LINKED ISSUE #2734: MountIndex: `getDirectoryEntries` scans every indexed path on each `readDir` (O(N) per listing; 761k entries on a real picker mount)
Part of #2707.

## What happens

`packages/webapp/src/fs/mount-index.ts` `getDirectoryEntries(mountPath, dirPath)` — the "fast path" `VirtualFS.readDirMounted` takes whenever an index is `ready` — answers a directory listing by iterating **all** of `data.files` and `data.directories` and prefix-matching each path:

```ts
for (const path of data.files) { if (!path.startsWith(prefix)) continue; … }
for (const path of data.directories) { … }
```

Picking a real project folder through the mount picker produced an index of **761,394 entries** (the folder contained `node_modules`/build output; the default cap is 2,000,000). From then on every `readDir` under that mount — `ls`, the Files panel, every isomorphic-git `readdir` (thousands per `git status`, ~25k per `git log --all`, #2712) — walks 761k strings and allocates a `Map`, so the "fast path" is slower than the FSA call it replaces, and it burns kernel-worker CPU that the cone's own commands need.

## Fix ideas

- Store the index as a tree (`Map<dirPath, { files: string[]; dirs: string[] }>`) or at least a per-directory bucket map built once at index time, so a listing is O(children).
- Keep the flat `files` array only for the name-search use case (or replace it with a trie / sorted array with binary search on the prefix).
- Consider skipping `node_modules`/VCS/build directories in the index by default (the same skip list #2717 introduced for sprinkle discovery) — most of those 761k entries are never useful for file discovery and make the index cost dominate.

trieloff:
Closes #2734

**Solution**
- Build and incrementally maintain per-directory child buckets so mounted `readDir` work is proportional to immediate children.

**Testing**
- Added a 10,000-path regression proving a small listing does not iterate the flat indexes.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01M1GP0SPZ84S7RHW9BM2R88DZ&panel=chat)

### DIFF
--- a/packages/webapp/src/fs/mount-index.ts
+++ b/packages/webapp/src/fs/mount-index.ts
@@ -207,6 +207,8 @@ export interface MountIndexState {
   abortCause?: MountIndexAbortCause;
 }
 
+type DirectoryChildren = Map<string, 'file' | 'directory'>;
+
 interface MountData {
   handle: FileSystemDirectoryHandle;
   state: MountIndexState;
@@ -214,6 +216,8 @@ interface MountData {
   files: Set<string>;
   /** Set of all directory paths under this mount (absolute VFS paths) */
   directories: Set<string>;
+  /** Immediate children keyed by their absolute parent directory path. */
+  childrenByDirectory: Map<string, DirectoryChildren>;
   /** Abort controller for cancelling in-progress indexing */
   abortController: AbortController | null;
   /**
@@ -250,6 +254,7 @@ export class MountIndex {
       state: { status: 'pending', indexed: 0 },
       files: new Set(),
       directories: new Set(),
+      childrenByDirectory: new Map(),
       abortController,
       limits,
     };
@@ -293,6 +298,7 @@ export class MountIndex {
     data.state = { status: 'pending', indexed: 0 };
     data.files.clear();
     data.directories.clear();
+    data.childrenByDirectory.clear();
     if (limits) data.limits = limits;
     this.notifyListeners();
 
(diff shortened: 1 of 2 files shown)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01M1GP0SPZ84S7RHW9BM2R88DZ)⚡</sup>

👋 I've got this PR.

trieloff:
Picking this up rather than duplicating it — thanks. Context: this is the last open piece of the #2707 benchmark work (waves 1 and 2 are merged; see the verification tables on #2707).

Reviewed the diff: the per-directory child buckets are the right shape, the flat `files`/`directories` sets are correctly kept for the search consumers, and the incremental maintenance covers add/remove/rename including subtree moves. Marking it ready so CI and the review bots run.

One gap worth naming, since the PR says "Closes #2734": the issue has two halves and this covers the first. The second — **not indexing `node_modules`, `.git`, build output and dot-directories by default** (mirroring the skip list #2717 added in `fs/bounded-walk.ts` for sprinkle discovery) — isn't here. Without it a picker mount of a real project still ingests ~761k entries; listings are now O(children) so the pathology is much reduced, but the index build itself still walks and retains all of it. I'll file that as a follow-up rather than hold this PR, unless you'd prefer to add it here.

trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01M1GP0SPZ84S7RHW9BM2R88DZ)⚡</sup>

Agreed — I'll keep this PR scoped to eliminating the per-`readDir` full scan. Default directory filtering changes index semantics and is better handled in the follow-up you described.

PR #2766, opened 2026-09-02
fix(ffmpeg): recycle the wasm core after a trap instead of poisoning the realm

trieloff:
## Problem

`getFfmpeg` memoizes one `FFmpeg` instance per realm, and the `.catch()` that clears the cache covers only a **load** failure — a path that never produced an instance in the first place. So a trap raised from inside an already-loaded core is permanent: WebAssembly linear memory is left inconsistent, every later entry re-traps at once, and the cached promise keeps handing the dead instance to every subsequent `ffmpeg`.

Found on a live leader. After a 10 MB `.ts` remux blew the heap at 15:13, every later `ffmpeg` failed instantly (4–130 ms, versus 11–126 s for real work) with the same message:

```
ffmpeg: RuntimeError: memory access out of bounds
```

Including a 64x64 synthetic encode with **no input file at all**:

```
$ ffmpeg -f lavfi -i testsrc=duration=1:size=64x64:rate=5 /tmp/syn.mp4
ffmpeg: RuntimeError: memory access out of bounds
```

Six commands over the next 12 minutes produced zero output and the session never recovered — only a tab reload would have cleared it. `ffmpeg -version` kept answering the whole time, because that gate resolves the core package without entering wasm, so the command looked healthy.

The heap pressure is cumulative rather than size-based: a 10.17 MB `.ts` remuxed fine at 15:12, and the near-identical 10.28 MB `.ts` trapped a minute later. `cleanupMemfs` deletes staged files but cannot reclaim heap the encoder grew.

## Fix

`recycleFfmpeg()` drops the cached promise and `terminate()`s the worker — terminating is what actually releases the dead core's heap. `runWasmFfmpeg` calls it when the wasm path throws.

A throw is already the right trigger: bad flags and unsupported codecs come back as a non-zero **exit code**, which is left alone. MEMFS cleanup is skipped after a fault, since every `deleteFile` would re-enter the trapped module. The caller gets told the core was recycled so a retry is worth trying.

This does not stop a large job from exhausting the heap — it stops one exhausted job from taking the rest of the session with it.

## Tests

`ffmpeg-wasm.test.ts` is new (mocks `@ffmpeg/ffmpeg`, so no core boots): instance sharing until recycle, `terminate()` on the stale worker, no-op when nothing is loaded, no unhandled rejection when recycling a failed load, and tolerance of an already-dead worker.

`ffmpeg-command.test.ts` covers the policy: recycle on a trap and on a staging throw, the retry hint, cleanup skipped after a fault, and **no** recycle for an ordinary non-zero exit or a missing/empty output. Reverting the source makes 3 of them fail.

## Verification

`verify` · `typecheck` · `test` (18,622) · `test:coverage` · both builds · `bundle-size` + first-load · `check-touched-exemptions` — all green. `ffmpeg-wasm.ts` statement coverage 91%.

Docs: new `docs/pitfalls.md` section — `magick-wasm.ts` and `v86-wasm.ts` have the same memoized shape and are listed as not yet recycled.

---

## Live verification (added after the first push)

Ran on an isolated local harness (bridge `:5715`, own wrangler `:8797`, ephemeral Chromium profile; production `:5710`/`:9222` PIDs verified unchanged before and after) against a build of this branch composed with #2767.

A real heap exhaustion was induced rather than simulated — a ~3 GB rawvideo encode:

```
$ ffmpeg -f lavfi -i testsrc=duration=40:size=1920x1080:rate=25 \
         -c:v rawvideo -pix_fmt yuv420p -f rawvideo big.yuv
frame=  623 fps= 51 q=-0.0 size= 1892352kB time=00:00:24.92 ...
ffmpeg: RangeError: Array buffer allocation failed
ffmpeg: the wasm core faulted and was recycled; retry the command (a large input may need to be split into smaller passes)
```

The core died at 1.89 GB — the same class of fault as the production incident, reached by a different route (`RangeError` on the JS side rather than a `RuntimeError` trap inside wasm; the recycle policy keys on the throw, not the message, so both are covered).

**The next commands succeed.** On the unpatched production leader this is exactly where everything stopped working for the rest of the session:

| Command run immediately after the fault | Result |
| --- | --- |
| `ffmpeg -f lavfi -i testsrc=duration=1:size=64x64:rate=5 syn.mp4` | 2835 B ✅ (this is the invocation that failed forever in the field) |
| `ffmpeg -f concat -safe 0 -i list.txt -c copy j3.mp4` | 16503 B ✅ |

Console watcher clean across the run.

### DIFF
--- a/docs/pitfalls.md
+++ b/docs/pitfalls.md
@@ -205,10 +205,30 @@ package without entering wasm — so the command looked healthy.
 promise _and_ `terminate()` the worker (terminating is what actually releases
 the dead heap), then let the next call boot a clean core.
 
+Three things the recycle path has to get right:
+
+- **Retire a generation, not "the cache".** Callers share the instance, so a
+  slow run can unwind from a core that a faster retry has already replaced.
+  Pass the faulting instance in and no-op unless it is still the current one,
+  or the late unwind terminates the healthy core the retry just booted.
+- **Revoke the `blob:` URLs.** `terminate()` does not. The core JS + wasm are
+  ~31 MB per generation, so a recyclable loader that never revokes feeds the
+  memory pressure that caused the fault.
+- **Cover every entry point into the shared core.** In `ffmpeg-command.ts` that
+  is the main command path _and_ `transcodeCapturedBytes` (the webcam
+  post-capture pass), which uses the same cached instance.
+
 Distinguish a fault from an ordinary failure: these commands report bad flags
 and unsupported inputs as a **non-zero exit code**, so a _throw_ out of the
 WASM call path is already exceptional and is the right trigger. Skip MEMFS
 cleanup afterwards — every `deleteFile` would re-enter the trapped module.
+
+Two traps for that heuristic. The core can return a stale exit 0 after an
+internal `Aborted()`, so the real trap surfaces on the **readback** — catching
+it locally as "produced no output" leaves the poisoned instance cached. And
+keep non-WASM work (writing the artifact back to the VFS) outside the fault
+`catch`, or a read-only mount gets reported as a core fault and costs a
+needless reboot.
 
 | Site                                         | Status                                                                           |
 | -------------------------------------------- | -------------------------------------------------------------------------------- |
--- a/docs/pitfalls.md
+++ b/docs/pitfalls.md
@@ -184,6 +184,38 @@ the closure returns.
 | `core/image-processor.ts`                        | Snapshots via `new Uint8Array(data)`                                               |
 | `cdp/browser-api.ts`                             | Consumes the view inside the callback to build base64 (no escape)                  |
 
+## A WASM Trap Poisons the Cached Instance, Not Just the Call
+
+A WebAssembly trap — `RuntimeError: memory access out of bounds`,
+`unreachable`, an emscripten `Aborted(…)` — is terminal for the **instance**.
+Linear memory is left inconsistent, so every later entry into the module
+re-traps immediately. Any loader that memoizes one instance per realm
+therefore turns a single bad call into a dead command for the rest of the
+session.
+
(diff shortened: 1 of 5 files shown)


trieloff:
All four addressed in `00e78034f`.

**P1 — guard recycling to the instance that faulted.** Real race, fixed. `recycleFfmpeg(faulted?)` compares against a new module-level `currentFfmpeg` and no-ops unless that instance is still current; a `null` `currentFfmpeg` (load in flight) is by definition not the faulting one. Test: retire gen 1, boot gen 2, replay `recycleFfmpeg(gen1)` — gen 2 survives and is never terminated.

**P1 — let readback traps reach the recycling handler.** Also real, and it defeated the whole patch on that path. Readback moved into `readEncodedOutput`, which rethrows when `isCoreFault(err)` (WASM `RuntimeError`, `RangeError`, `Aborted`, allocation failure, …) and otherwise keeps the "produced no output" result. Two tests: a `RuntimeError` on `readFile` now recycles and skips cleanup; a plain `FS error: no such file` still reports a missing output and does **not** recycle.

**P2 — keep VFS write errors out of the fault handler.** The write now happens after the try/finally and reports `cannot write <path>: …`, with no recycle and with MEMFS cleanup still running. Test asserts an `EROFS` write does not mention the core.

**P2 — revoke core blob URLs.** Fixed, and it mattered more than it looks: the URLs are now tracked per generation and revoked on recycle *and* on a failed `load()`. Tests assert the revoked set matches the created set in both cases.

One thing the review didn't flag that came out of it: `transcodeCapturedBytes` (the webcam post-capture pass) shares the same cached core and had the identical exposure, so it recycles too — gated on `isCoreFault`, since its own non-zero-exit throw leaves the core healthy.

Lessons recorded in `docs/pitfalls.md`.

PR #2767, opened 2026-09-02
fix(ffmpeg): stage concat-list members so -f concat works at all

trieloff:
## Problem

`loadResolvedInputs` stages into MEMFS only the paths that appear as `-i`. The concat demuxer opens the files named **inside** the list file itself, so those members never reached MEMFS:

```
$ printf "file '01.mp4'\nfile '02.mp4'\n" > list.txt
$ ffmpeg -f concat -safe 0 -i list.txt -c copy joined.mp4
```

`list.txt` was staged; `01.mp4` and `02.mp4` were not. The demuxer then opened files that did not exist. This was not a size or argument problem — `-f concat` could not work under any arguments.

Found on a live leader while joining six clips into one cut. The agent worked around it by remuxing every clip to mpegts, `cat`-ing them together and remuxing back — which is exactly what you do when the concat demuxer is unavailable.

## Fix

Expand a `-f concat` input before staging: parse the list, read each `file` directive from the VFS, and rewrite the list to point at the MEMFS names the members get staged under.

Details that matter:

- **Member paths resolve against the list file's directory**, matching ffmpeg, not the shell cwd. A list at `/clips/list.txt` naming `01.mp4` reads `/clips/01.mp4` even when cwd is `/home`.
- **Members are written before the list**, since the demuxer opens them as soon as it reads it.
- **Non-`file` directives pass through untouched** — `duration`, `inpoint`, `outpoint`, `ffconcat version 1.0`, comments, blanks — so list-level options still reach the demuxer.
- **Quoting follows ffmpeg**: bare, `'single'`, `"double"`, and an embedded apostrophe written `'\''`.
- Members are removed again in `cleanupMemfs`, so a concat run does not leave megabytes of stale media in the wasm heap.
- A member missing from the VFS is **named in the error** rather than failing somewhere inside wasm.

Out of scope: the `concat:` **protocol** form (`-i "concat:a.ts|b.ts"`). It already fails with a clear `input file not found`, and the help text now points at `-f concat` instead.

The staging loop moves into `stageInputsIntoMemfs` — inlining the members loop pushed `runWasmFfmpeg` to cognitive complexity 26, one over the cap.

## Tests

`parseConcatList` is exported and covered directly: quoting variants, `'\''` unquoting, directive/comment pass-through, and that a filename beginning with `file` is not mistaken for a directive.

End-to-end through the command: members staged and the list rewritten to match, write ordering, list-relative path resolution, the named missing-member error, MEMFS cleanup of members, and that an ordinary single input is untouched by the new path. Reverting the source branch makes 5 of them fail.

## Verification

`verify` · `typecheck` · `test` (18,622) · `test:coverage` · both builds · `bundle-size` + first-load (allowance 5 kB, unchanged) · `check-touched-exemptions` — all green.

Docs: `docs/shell-reference.md` ffmpeg row, plus a concat example in `ffmpeg --help`.

Independent of #2766 (recycling a trapped core); both branch off `main` and touch different regions.

---

## Live verification (added after the first push)

Ran on an isolated local harness (bridge `:5715`, own wrangler `:8797`, ephemeral Chromium profile; production `:5710`/`:9222` PIDs verified unchanged before and after) against a build of **both** ffmpeg branches composed together.

The staging fix alone was not enough. With members staged, plain `-f concat -i list.txt` joined correctly — but the **canonical** form still failed:

```
$ ffmpeg -f concat -safe 0 -i list.txt -c copy joined.mp4
__in0_list.txt: Invalid data found when processing input
```

`-safe` was missing from `VALUE_TAKING_FLAGS`, so it parsed as a boolean toggle and its `0` fell through to `handlePositionalToken`. That made `0` a phantom output path **and** flushed the pending options into that phantom output — dropping the `-f concat` bound to the very next `-i`. The rebuilt argv had no `-f concat`, so the core probed the rewritten list as media.

Fixed in `dc0f1b7f5` with a regression test that pins the whole parse (`inputs[0].raw`, `outputPath`, `outputOpts`). Removing the `-safe` entry makes it fail.

After the fix, live:

| Check | Result |
| --- | --- |
| `ffmpeg -f concat -i list.txt -c copy` (2 s + 3 s clips) | 16503 B, `time=00:00:04.70` |
| `ffmpeg -f concat -safe 0 -i list.txt -c copy` | 16503 B — identical |
| Console watcher during the run | clean |

### Note for reviewers: the same class of bug is still live elsewhere

Any value-taking flag missing from `VALUE_TAKING_FLAGS` corrupts the invocation silently. `-bsf:v` / `-bsf:a` are also absent, so

```
ffmpeg -i in.mp4 -c copy -bsf:v h264_mp4toannexb -f mpegts out.ts
```

loses **both** `-c copy` and the bitstream filter, and re-encodes instead of stream-copying. That is a separate fix from this PR; flagging it rather than widening the diff.


trieloff:
Addressed in `d2b502c46`.

**P1 — parse `-safe` as value-taking.** Correct, and it was the difference between "the docs example works" and "it doesn't". Already fixed in `dc0f1b7f5`, pushed after the reviewed commit: `-safe` joined `VALUE_TAKING_FLAGS`, with a regression test pinning the whole parse (`inputs[0].raw`, `outputPath`, `outputOpts`). Verified live afterwards — `-f concat -safe 0 -i list.txt -c copy` now produces the same 16503 B / 4.70 s join as the form without `-safe 0`.

**P2 — decode backslash escapes in bare paths.** Fixed. A bare `file` directive now runs through `rest.replace(/\\(.)/g, '$1')`, so `file clip\ one.mp4` resolves `clip one.mp4` instead of looking for a member with a literal backslash and reporting a valid list as missing.

**P2 — preserve the demuxer's safe validation.** Fixed, and the reasoning holds: rewriting members to flat `__cat…` names makes ffmpeg's own check pass vacuously, so the check has to run on the path the user wrote. `concatSafeMode()` reads the effective `-safe` off the input's own options (defaulting to enabled, as ffmpeg does) and `isSafeConcatPath()` rejects absolute, drive-letter, and `..`-traversing members with `unsafe file name: <path> (pass -safe 0 to allow it)`. Three tests: `../secret.mp4` and `/abs.mp4` refused at the default level, `/abs.mp4` accepted once `-safe 0` is given.

Reverting each fix fails its guard (3 of 3).

PR #2776, opened 2026-09-02
refactor(webapp): resolve @ffmpeg/core through an explicit layout list and add a live canary

trieloff:
Follow-up to #2744 (magick 0.0.43 layout break + `magick-wasm-live.test.ts`), applying the same pattern to `ffmpeg-wasm.ts`.

## Summary

`ffmpeg-wasm.ts` hardcoded where `@ffmpeg/core` keeps `ffmpeg-core.{js,wasm}` inside the package. The location is now an explicit, commented layout list (`FFMPEG_CORE_LAYOUTS`, one entry today, exclusions spelled out), and a live canary test boots the REAL installed package so a dependency that changes shape, exports, or ABI fails a test instead of shipping a silent "not installed".

## Why

Same exposure as #2744: every existing ffmpeg test drives the loader over a synthetic file map, so a reorganised `dist/` would keep the whole suite green while `ffmpeg` reports `@ffmpeg/core is not installed`.

**Facts first**, checked with `npm pack` against the registry rather than assumed:

| Versions | Layout | Usable with the bundled `@ffmpeg/ffmpeg` 0.12.15? |
| --- | --- | --- |
| 0.12.0 … 0.12.10 (pinned) | `dist/esm/` + `dist/umd/`; `@ffmpeg/core-mt` identical plus `ffmpeg-core.worker.js` | `dist/esm/` yes. `dist/umd/` no: the wrapper always spawns a module worker, `importScripts` throws, it falls back to `(await import(coreURL)).default`, and the UMD build has no default export |
| ≤ 0.11.x | flat `dist/ffmpeg-core.{js,wasm,worker.js}` | No: pre-0.12 `createFFmpeg` protocol. A version mismatch, not a layout to paper over |

So the layout moved exactly once (0.11 → 0.12) and has been identical across every 0.12.x release. There is one supported layout; the list exists so the next move is a one-line edit with the canary saying where. No evidence-free candidates were added.

- `FFMPEG_CORE_LAYOUTS` + a private `findFfmpegCoreFiles(pkgDir, pkg, reader)`; `tryLoadFfmpegCoreFromNodeModules` now resolves through it. Loader caching, `selectFfmpegCore`, the command surface and `ffmpeg-command.ts` are untouched.
- The canary cannot go through `getFfmpeg`: the `@ffmpeg/ffmpeg` wrapper needs a real `Worker`, which Node lacks. It goes one level down instead: real `selectFfmpegCore` over a file-for-file mirror of the installed package (the file list comes from disk, nothing in the test hardcodes `dist/esm`), then boots the returned glue + wasm exactly the way the wrapper worker does (`import()` → `.default` → `createFFmpegCore({ wasmBinary })`) and runs `-version` plus a real `-f lavfi` → WAV encode. The glue is compiled worker-only, so the test stubs `self` with a `location`; the source is imported via a `data:` URL, Node's analogue of the loader's `blob:` URL.
- Stated in the file header: not covered are the wrapper's blob-URL/postMessage plumbing (unit-tested with fakes) and `@ffmpeg/core-mt` (not a devDependency; pthreads need real workers).

## Cross-cutting impact

- **Floats exercised**: none at runtime. Behavior for a correctly installed 0.12.x core is byte-identical; only a partial install or unknown layout still returns `null` as before.
- **Other callers**: `ffmpeg-command.ts`'s `-version` gate and `selectFfmpegCore` call the same exported function with the same contract.
- **Bundle size**: no new imports on the kernel-worker eager path.

## Test plan

- [x] `npm run verify` — 7/7 checks passed
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs origin/main`
- [x] `npm run typecheck`
- [x] `npm run test` — full suite 18691 tests; every failure in the saturated run was a load timeout in an unrelated file (see below)
- [x] `npm run test:coverage:webapp -- --maxWorkers=8` — 14944 passed, 0 failed; statements 82.64 (floor 82), lines 84.65 (floor 84)
- [x] `npm run build` and `npm run build -w @slicc/chrome-extension`, then `npm run bundle-size`
- [x] New behavior covered: `tests/shell/supplemental-commands/ffmpeg-wasm-layouts.test.ts` (7 resolution cases: `dist/esm` hit for both cores, -mt worker requirement, `null` for `dist/umd`-only, flat 0.11 layout, glue-without-wasm and wasm-without-glue, failing `readBytes`) and `ffmpeg-wasm-live.test.ts` (3 tests, ~1 s).
- [x] **Proof the guards bite** (each mutation reverted afterwards):
  1. Source pointed at a wrong layout (`dist/es`): 2 resolution tests + the canary fail.
  2. Real package moved on disk (`dist/esm` → `dist/esm-next`), source untouched: all 64 existing mocked ffmpeg tests stay green, only the canary fails, listing the files it actually found and pointing at `FFMPEG_CORE_LAYOUTS`. That is the #2744 scenario, caught.
- [x] N/A — no UI change, no manual smoke needed (no runtime behavior change for a healthy install)

**Pre-existing failures** (load-induced, unrelated to this PR): the dev box was shared with several other sessions' vitest/build jobs (load average 200–430). Under that load 14 unrelated files timed out (`Test timed out in 5000ms` / `Hook timed out`, e.g. the Pyodide load in `slicc-fs-module.test.ts`, `git init` in `check-swift-resolved-drift.test.mjs`, the 1 s race in `wc-sprinkles.test.ts`). Every one of them passed when re-run once load fell; the failing set shrank monotonically with load (16 → 5 → 3 → 2 → 0). The uncapped coverage run alone pushes this many-core box past load 150, hence the `--maxWorkers=8` run above, which is faster and clean.

## Documentation

- [x] No documentation changes needed — no doc outside `ffmpeg-wasm.ts` records the layout; the source comments carry the evidence and exclusions.

## Risk & rollback

Blast radius is the `ffmpeg` command's not-installed path only. Clean revert. The canary adds a ~32 MB wasm read to the webapp unit run (~1 s), no network.

## Checklist

- [x] Public API / tool / shell-command / lick-channel changes are intentional and reflected in docs (none)

### DIFF
--- a/packages/webapp/src/shell/supplemental-commands/ffmpeg-wasm.ts
+++ b/packages/webapp/src/shell/supplemental-commands/ffmpeg-wasm.ts
@@ -49,7 +49,8 @@ export const FFMPEG_CORE_MT_PACKAGE = '@ffmpeg/core-mt';
 
 /**
  * Read-only VFS context the loader needs to read an ipk-installed
- * `@ffmpeg/core/dist/esm/{ffmpeg-core.js,ffmpeg-core.wasm}` pair.
+ * `@ffmpeg/core` glue + wasm pair (see {@link FFMPEG_CORE_LAYOUTS} for
+ * where inside the package they live).
  * Mirrors the {@link IpkResolutionContext} shape used by
  * `esbuild-wasm.ts` and `biome-command.ts` so every float
  * (standalone/hosted/extension/Node) wires the loader the same way.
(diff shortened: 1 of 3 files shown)


trieloff:
Addressed the first review note in de629ccf6: the canary header now states the real cost (~1 s under vitest, ~100 ms in bare Node). The second note (raw `readdirSync` ENOENT if `@ffmpeg/core` ever stops hoisting to the root `node_modules`) is left as is, per the review: still a loud canary failure, just a less friendly message.

Ready for a human look; not merging from here.

trieloff:
Rebased onto `main` (67e9f496a). The only conflict was an add/add on `tests/shell/supplemental-commands/ffmpeg-wasm.test.ts`: #2767 landed a lifecycle suite for the new `recycleFfmpeg` under that name, so this PR's layout suite now lives at `ffmpeg-wasm-layouts.test.ts` and `main`'s file is kept verbatim. `ffmpeg-wasm.ts` itself merged cleanly (the recycle changes and the layout list touch different regions). On the rebased tree: all 4 ffmpeg test files pass (101 tests, including the new lifecycle suite), typecheck and the 8-check lint gate are green.

PR #2788, opened 2026-09-02
fix(ffmpeg): support -f null analysis sinks without requiring output

trieloff:
## Summary

- Recognize ffmpeg **analysis sinks** so Remotion-style `silencedetect` / `loudnorm` invocations work in the WASM wrapper: `-f null` with `-` or `/dev/null` (and bare `/dev/null`) no longer require a MEMFS/VFS artifact.
- Treat a lone `-` as a positional output filename (native ffmpeg stdin/stdout convention), not a boolean flag — that was why `… -f null -` failed with `at least one output file must be specified`.
- On a successful sink run, return the captured core log on stderr with exit 0 and skip VFS writeback; normal encodes still fail on missing/empty output. Sink success does **not** call `recycleFfmpeg` (#2766).

### Detection rule (defended)

An invocation is an analysis sink iff:

1. `outputPath` is `-` or `/dev/null`, **and**
2. either `-f null` appears in the output options, **or** the path is bare `/dev/null`.

Why this shape:

- Remotion teaches exactly `-f null -` and `-f null /dev/null`; both are covered.
- Bare `/dev/null` is also a sink because MEMFS has no `/dev/null` device — writing there can never succeed usefully.
- Bare `-` **without** `-f null` is **not** a sink (stdout piping we do not emulate) so it still hits the missing-output failure.
- `-f null dump.bin` stays a normal encode (artifact required) — keeps failed encodes from reporting success.

### Interaction with #2766

Built on the recycled `runWasmFfmpeg` shape after #2766 landed on `main`. Sink success sets `early = { exitCode: 0, stderr }` **before** `readEncodedOutput`, so missing MEMFS output is never confused with a core fault and `recycleFfmpeg` is not called. `cleanupMemfs` still runs for staged inputs + `__null_sink`.

### Pre-fix failures (verified)

| Form | Message |
|------|---------|
| `-f null -` | `ffmpeg: at least one output file must be specified` |
| `-f null /dev/null` | `ffmpeg: produced no output file for /dev/null` |

### Tests

- Unit: both sink forms succeed and return log text; normal missing/empty still fails; sink path does not call `recycleFfmpeg`.
- Guard proof: reverted the command body and the new sink tests failed (exit 1 instead of 0).
- Live harness (ports 5755 / 8845; Playwright Chromium):

```text
ffmpeg -f lavfi -i anullsrc=r=44100:cl=mono -t 1 -af silencedetect=noise=-50dB:d=0.3 -f null -
EXIT:0
[silencedetect @ 0xdf3380] silence_start: 0
[silencedetect @ 0xdf3380] silence_end: 1.0449 | silence_duration: 1.0449
```

(`Aborted()` appears as a core log line after success; exit code remains 0.)

Production listeners: session preflight `:5710=4132` `:9222=5106` — unchanged after harness. (Earlier in the session those PIDs were `17234`/`17558`; they changed while other work was running, not by this harness.)

## Test plan

- [x] Unit tests in `ffmpeg-command.test.ts` (sinks + regression + no-recycle)
- [x] `npm run verify`, debt gate, typecheck, test, coverage, builds, bundle-size
- [x] Live silencedetect on lavfi `anullsrc` via CDP harness
- [ ] Reviewer: confirm rebase stays clean vs any remaining ffmpeg topic PRs (#2767 / #2769) if they land after


trieloff:
Reviewed. The implementation is careful and I'm happy with the core of it:

- `hasNullMuxer` scans **`outputOpts`**, not `parsed.inputs[].format`. That was the trap I expected someone to fall into — `-f null -i in.mp4 out.mp4` binds `-f null` to the *input*, and a naive `input.format === 'null'` check would have misfired. This gets it right.
- The detection is conservative in the right direction: bare `-` without `-f null` is **not** a sink, so a genuinely failed encode still fails the missing-output check rather than reporting success.
- `__null_sink` still flows through `cleanupMemfs` via `outputName`, so nothing is stranded.
- Returning through `early` correctly gates the post-try VFS write, so no literal `/dev/null` path gets written back, and no recycle fires.

I verified the pre-fix messages independently and they match what's claimed.

## One integration defect to fix — after #2769 lands

#2769 (`fix/ffmpeg-value-taking-flags`) adds `hasLaterPositional()`, which decides unknown-option arity by scanning ahead for a positional:

```ts
if (!tok.startsWith('-')) return true;
```

A lone `-` starts with `-`, so once this PR makes `-` a **positional**, the two disagree. I test-merged the branches and confirmed:

```
parseFfmpegArgs(['-i','in.mp4','-unknown_opt','val','-f','null','-'])
  → outputPath = "-"   outputOpts = ["-f","null"]
```

`-unknown_opt val` is silently dropped — the exact class of bug #2769 exists to fix. The sink itself still resolves correctly, so this is not a functional break for the feature, but it re-opens the hole one PR over.

Fix is one line in `hasLaterPositional`:

```ts
if (!tok.startsWith('-') || tok === '-') return true;
```

You can't apply it here yet — this branch is based on `main` without #2769, so `hasLaterPositional` doesn't exist on it. Sequencing: #2767 → #2769 → rebase this PR → add the fix plus a regression test → enqueue. I'll carry that out; no action needed from you.

Nice work on the live proof — the real `silencedetect` output is exactly the evidence I wanted.

trieloff:
Addressed the three Codex findings on this PR:

1. **P1 poisoned core** — analysis sinks now call `ensureCoreHealthy` (MEMFS write+delete probe) before accepting exit 0; `cleanupMemfs` rethrows `isCoreFault` and recycles if cleanup is the first place poison surfaces.
2. **P2 bare `/dev/null`** — `ensureNullMuxerOpts` injects `-f null` before exec. Live-verified: `ffmpeg … /dev/null` → EXIT:0 with `Output #0, null, to '__null_sink'` and silencedetect lines.
3. **P2 lone `-`** — `-` is only accepted as an output positional when `-f null` is already pending; `ffmpeg -i in.wav -f mp3 -` now fails with “at least one output file must be specified”.

Each fix has a unit test that fails when the change is reverted. Full verify pass green locally. Not enqueueing.

PR #2789, opened 2026-09-02
feat(webapp): add emulated ffprobe shell command via ffmpeg-wasm

trieloff:
## Summary

- Adds a shell `ffprobe` command that **emulates** media probing on the same `@ffmpeg/core` wasm instance as `ffmpeg`. There is no separate ffprobe wasm/ipk for the 0.12.x pin — we stage the input, run `ffmpeg -hide_banner -i <file>` (no output), parse the Input #N banner, and format a useful subset of fields.
- Supports machine-readable callers: `-show_format` / `-show_streams` / `-show_entries` / `-select_streams` / `-of json|csv|default=nw=1:nk=1`. Unsupported options are **rejected by name** (never silently dropped).
- Wired into supplemental registration, `help`, `docs/shell-reference.md`, and `builtin-shadow-map` (`ffprobe` + `@ffprobe-installer/ffprobe` → `@ffmpeg/core`). Lazy-loads `ffprobe/run.ts` on first use so the worker first-load budget stays green.
- Agent skill: `packages/vfs-root/workspace/skills/ffmpeg/SKILL.md` (ffmpeg + emulated ffprobe — install, supported flags, honesty about the emulation).
- Shares `isCoreFault` from `ffmpeg-wasm.ts` with `ffmpeg-command.ts` (single home next to `recycleFfmpeg`) and gates MEMFS `deleteFile` on a `faulted` flag after recycle — same #2766 invariant as `runWasmFfmpeg`.

**Honesty:** this is not real ffprobe. Output shape is a practical subset for Remotion-style channel/duration/container sniffing — not a drop-in for every ffprobe flag or exact CLI banner fidelity.

### Codex review follow-ups (this push)

1. **Concurrent probes** — serialize access to the shared ffmpeg instance via `withProbeLock` so log listeners and MEMFS staging cannot interleave across scoops/jobs.
2. **Runtime skill** — agent-facing `skills/ffmpeg/SKILL.md` (developer `docs/shell-reference.md` alone is not enough).
3. **`5.1(side)` layouts** — `channelsFromLayout` strips parenthetical qualifiers before `CHANNEL_LAYOUTS` lookup so Remotion-shaped `channels` queries return `6`.
4. **CSV writer** — emit section prefix (`format`/`stream`) per upstream `p=` semantics; CSV-escape fields that contain commas/quotes.
5. **Apostrophes in filenames** — stage under opaque MEMFS names (`__probe_<seq>_<ts>.ext`) so banners with `from '…'` still parse.

### Shared infrastructure note (not ffprobe-only)

**ipk tarball fetch default timeout: 30s → 120s** in `packages/webapp/src/shell/ipk/registry.ts` (`DEFAULT_TARBALL_TIMEOUT_MS`). Packument timeout stays 30s. Host pinning via `EXPECTED_TARBALL_HOST` is unchanged. Needed so `@ffmpeg/core` (~20 MB) can install through the fetch proxy without timing out; this default applies to every `ipk add` tarball download, not only ffmpeg.

## Live proof (CDP harness)

Isolated lane: `PORT=5722` / `WRANGLER_PORT=8803` / Playwright Chromium. Production listeners unchanged (`:5710`/`4132`, `:9222`/`5106`).

```text
$ which ffprobe
/usr/bin/ffprobe

$ cat /workspace/skills/ffmpeg/SKILL.md | head -1
# ffmpeg / ffprobe

$ ipk add -g @ffmpeg/core@0.12.10
ipk: installed @ffmpeg/core@0.12.10 -> /shared/lib/node_modules/@ffmpeg/core

$ ffmpeg -y -f lavfi -i testsrc=duration=1:size=320x240:rate=30 \
    -f lavfi -i sine=frequency=440:duration=1 \
    -c:v libx264 -pix_fmt yuv420p -c:a aac /tmp/probe-clip.mp4

$ ffprobe -v error -show_format -show_streams -of json /tmp/probe-clip.mp4
# (format + video/audio streams; duration ~1s; channels=1)

$ ffprobe -v error -select_streams a:0 -show_entries stream=channels -of default=nw=1:nk=1 /tmp/probe-clip.mp4
CH:1

$ ffmpeg -y -f lavfi -i "sine=frequency=440:duration=1" -ac 6 /tmp/surround.wav
$ ffprobe -v error -show_entries stream=channel_layout,channels -of default=nw=1:nk=1 /tmp/surround.wav
channel_layout=5.1(side)
channels=6
CH:6

$ ffprobe -v error -select_streams a:0 -show_entries stream=channels -of csv /tmp/surround.wav
CSV:stream,6

$ cp /tmp/surround.wav "/tmp/O'Brien.wav"
$ ffprobe -v error -show_entries stream=channels -of default=nw=1:nk=1 "/tmp/O'Brien.wav"
APOSTROPHE_CH:6

$ ffprobe -v error -count_frames /tmp/probe-clip.mp4
BAD_EC:1
ffprobe: unsupported option '-count_frames' (this is an emulated probe — see --help)
```

## Guard proofs (unit)

- Unsupported flags exit non-zero with a clear message.
- `-of default=nw=1:nk=1` is parsed via `indexOf('=')` (not `split(/=/, 2)`, which drops the remainder).
- Core faults call `recycleFfmpeg(instance)` and **do not** call `deleteFile` afterward (asserted; fails if the `faulted` gate is removed).
- `isCoreFault` lives in `ffmpeg-wasm.ts` and is covered there (including the live `Array buffer allocation failed` RangeError shape).
- Concurrent probes: mutex keeps `maxActive === 1` (fails if `withProbeLock` is removed).
- `5.1(side)` → channels `6` (fails if qualifier strip is removed).
- CSV emits `stream,6` / quotes comma-bearing fields (fails if section/`csvEscape` is reverted).
- Opaque MEMFS names refuse apostrophes in the staged basename (fails if `inferMemfsName` stages the raw basename).

## Test plan

- [x] `npm run verify`
- [x] `check-touched-exemptions.mjs`
- [x] `npm run typecheck`
- [x] `npm run test` / `test:coverage` (ffprobe suite green; repo-wide run hit a pre-existing `v86-wasm-live` pin mismatch unrelated to this PR)
- [x] `npm run build -w @slicc/webapp` + chrome-extension + `bundle-size`
- [x] Live CDP harness above (channel layout + apostrophe + CSV + skill path)
- [ ] Reviewer: spot-check `--help` honesty and Remotion-shaped `channels` probe

## Queue note

Not enqueued. Parent drives order: #2767 → #2769 → #2788 → #2789. Based on current `main` (includes #2766 / #2776).


trieloff:
Reviewed. The approach is the right call and honestly framed — `@ffmpeg/core` 0.12.x genuinely has no separate ffprobe artifact, and emulating via the `Input #N` banner while **rejecting** unsupported flags rather than silently dropping them is exactly the right instinct (silently dropping options is the bug class #2769 exists to kill). Good catch registering `@ffprobe-installer/ffprobe` in the shadow map, and lazy-loading for the worker budget.

Three things.

## 1. `isCoreFault` is duplicated byte-for-byte — please deduplicate

`ffprobe/run.ts:452` reproduces `isCoreFault` exactly as it exists at `ffmpeg-command.ts:1179`, regex and all.

That regex is not decorative. Its contents were derived empirically from a live production incident — `RangeError: Array buffer allocation failed` is in there because I hit it at 1.89 GB on a real core, not because it was guessed. It **will** be widened as we learn more fault signatures, and when that happens one copy will silently diverge and `ffprobe` will start leaving poisoned cores cached again.

Please give it one home. `ffmpeg-wasm.ts` is the right one — it already owns `recycleFfmpeg` and the instance lifecycle, and `run.ts` already imports from it, so nothing new is coupled. Export it there, import it in both call sites.

## 2. MEMFS cleanup runs after the core was recycled

In `runProbe`'s `finally`:

```ts
try { await ffmpeg.deleteFile(memfsName); } catch { /* noop */ }
```

This runs even when the `catch` above just called `recycleFfmpeg(ffmpeg)`, which `terminate()`s the worker. #2766 deliberately gates this (`if (!faulted) await cleanupMemfs(…)`) precisely because every `deleteFile` re-enters a dead module.

I checked and it is harmless *today*: `classes.js:66-68` — after `terminate()` sets `#worker = null`, `#send` returns `Promise.reject(ERROR_NOT_LOADED)`, which your `catch {}` swallows. So this is not a hang and not a bug you can currently observe. But it contradicts an invariant we just spent a PR establishing, and it depends on wrapper internals rather than on our own guarantee. Add a `faulted` flag and gate the cleanup the same way `runWasmFfmpeg` does.

## 3. The ipk tarball timeout change — accepted, but flagging it

`registry.ts` changes the **default** tarball timeout for every package from 30s to 120s. That is shared infrastructure and unrelated to ffprobe.

I'm accepting it: the need is real (a ~20 MB core over a 30s budget is genuinely tight), the host is already pinned via `EXPECTED_TARBALL_HOST` so the widened window is not much of an exposure, and the comment explains itself. Worth knowing it's in here rather than discovering it later — please make sure the PR body says so prominently, since a reviewer looking for "ffprobe" will not expect to find a global ipk default in the diff.

## Merge sequencing

Four PRs now touch this area. Order is forced by overlaps: **#2767 → #2769 → #2788 → this one**. This PR is based on current `main` and adds mostly new files, so I expect its rebase to be the easiest of the set. I'll drive the queue; just push 1 and 2.

PR #2800, opened 2026-09-03
feat(worker): redirect webhook deliveries from a superseded tray

trieloff:


trieloff:
#2799 has merged, so this is retargeted to `main` and rebased onto it — the diff is now just this PR's own two commits.

Note the follow-up that landed in #2799 after review: a browser cannot suppress redirect-following, so the hub gained a `?redirect=manual` opt-out that answers the pre-#1957 terminal shape instead of a 308. That opt-out deliberately does **not** apply to the webhook surface here, and `703e1c47a` documents and pins why: a webhook sender has no channel to be told about a hop through — it reads no `Link` header and parses no SLICC body — so the redirect is the only thing that saves the delivery.

Re-verified on the rebased tip: verify 8/8, typecheck, 19,053 tests, coverage, both builds.

Issue #2276, opened 2026-08-21
feat(kernel): CapabilityBroker with per-float adapters


trieloff:
Part of #1666 — RFC "Phase 6: normalize privileged capabilities".

## Goal
A work unit never asks "am I running in the extension?". It asks an injected `CapabilityBroker` for browser / network / secrets / devices / mounts / approvals and receives a typed result or `CapabilityUnavailable`.

## Scope
- Define the `CapabilityBroker` protocol with explicit per-operation allowlists (no generic shared handler maps).
- Adapters: Node (`node-server`), Swift (`swift-server`), extension service worker, browser/page, restricted hosted/cloud.
- Move runtime detection (`isChromeExtensionRealm`, float-topology checks) out of `scoops/` and `tools/` business logic to composition time in the kernel host.
- One conformance suite run against every adapter.

## Exit criteria (from the RFC)
- [x] Privileged surfaces expose explicit operation allowlists.
- [x] Node, Swift, extension and hosted adapters pass one conformance suite.
- [x] The kernel receives one capability descriptor at composition time.
- [x] Direct webapp ↔ extension source imports reduced or moved to shared protocol packages (review-patterns category 10). (#2891, pending merge)

trieloff:
## #2276 complete

All nine PRs across five slices merged:

| Slice | PR | Summary |
| --- | --- | --- |
| A | #2786 | `CapabilityBroker` protocol + composition-time injection: per-operation allowlists, `CapabilityUnavailable`, one host-injected broker |
| B | #2805 | Real adapters keyed by float topology (`node-rest`, `extension-direct`, `extension-delegate`, `connect`) + shared conformance suite |
| C (network) | #2829 | `scoops/` loses its only float/topology read — `createTrayFetch` moved to `shell/tray-fetch.ts` |
| C (secrets) | #2834 | Secrets domain call-site migration to `broker.secrets.listMaskedEnv()` |
| C (mounts) | #2836 | Mounts domain call-site migration to `broker.mounts.signRequest()` |
| C (approvals) | #2837 | Approvals domain call-site migration — `createSudoBroker` takes the composed broker as a required parameter |
| C (browser + leftovers) | #2841 | Final domain PR — `kernel/telemetry.ts`, `kernel/host.ts`, `crontask-command.ts` |
| D | #2843 | `check-no-float-probes.mjs` lint gate — bans the ten float/topology identifiers under `scoops/`, `tools/`, `kernel/` |
[...]

PR #2805, opened 2026-09-03
feat(webapp): real CapabilityBroker adapters keyed by float topology (#2276 slice B)

trieloff:
Slice B of #2276 (Phase 6 of #1666). Slice A (#2786) shipped the protocol and two stubs; this replaces the stubs with real adapters.

## Why

`CapabilityBroker` existed but only `network.localNodeServer` was live, so nothing below the kernel host could actually stop probing its float. Slice C cannot migrate a single call site until the transports exist.

## Approach

**Adapters are keyed by `FloatTopology`, not by the issue's five product-float names.** `shell/float-topology.ts` already has the axis the transports genuinely differ on. `kernel/host.ts` resolves it once and hands the answer to `createCapabilityBrokerForTopology`; the annotation there (`const topology: CapabilityAdapterId = resolveFloatTopology()`) is the compile-time check that the two unions stay in sync.

| Adapter | Transport | Live operations |
| --- | --- | --- |
| `node-rest` | `/api/*` on the local privileged server | `network.localNodeServer`, `network.crossOriginFetch`, `secrets.*`, `mounts.signRequest`, `approvals.request` |
| `extension-direct` / `extension-delegate` | `chrome.runtime` messages / the named `secrets.crud`, `mount.sign-and-forward`, `fetch-proxy.fetch` Ports | `network.crossOriginFetch`, `secrets.*`, `mounts.signRequest`, `approvals.request` |
| `connect` | none | none — a hosted `?connect=1` cone genuinely has no privileged surface |

**There is no Swift adapter.** node-server and swift-server expose the same privileged routes, so `node-rest` serves both. `packages/shared-ts/fixtures/capability-rest-contract.json` is the one wire definition, replayed three ways: the requests the browser adapter emits, the Express routes, and the Hummingbird routes. Both server guards were mutation-tested (a bogus route in the fixture fails each of them).

Three protocol changes fell out of making the adapters real:

- **`CapabilityFailure` joins `CapabilityUnavailable`.** "This float has no transport for that" is a permanent shape fact a caller can branch on once at composition; "that attempt failed" (HTTP 5xx, a disconnected Port, a malformed reply) is retryable and worth showing the user. Both are `ok: false`, so a caller that only checks `ok` still fails closed. Without this, a REST adapter would have had to report a 500 as "this float cannot do secrets".
- **Page-gesture ops are a channel, not an adapter.** A directory picker and the WebUSB/WebSerial/WebHID choosers need a page-realm gesture in *every* topology, so the host injects one `PageGestureChannel` all four adapters layer on. Nothing injects one yet, so `mounts.pickDirectory` and `devices.*` stay unavailable exactly as before; slice C's mounts PR wires the real channel.
- **Allowlists are derived, not declared.** `composeCapabilityBroker` builds each domain from the operations an adapter supplied *by name*, so "listed but unimplemented" and "implemented but unlisted" are both unrepresentable. The conformance suite now enforces both directions: an unlisted op must answer `CapabilityUnavailable`, a listed one never may.

`MountSignRequest` / `MountSignResult` and `ApprovalRequest` / `ApprovalDecision` grew into the shapes the endpoints actually take (sign-and-forward envelopes rather than `{url}`; `always` + pattern + authenticated requester). The placeholder shapes could not have carried a slice-C migration without losing behaviour — an approval downgraded from `always` to `allow` would silently drop the persisted NOPASSWD pattern.

`browser.*` stays unavailable on every adapter: it rides the `/cdp` bridge, which is not one of these transports and is not on the worker's boot graph. Called out in `docs/work-unit.md` rather than faked.

## Cross-cutting impact

- **Bundle**: the adapters' wire code sits behind first-use dynamic imports (`rest-ops.ts`, `extension-ops.ts`), so only the broker's shape joins the kernel worker's eager closure. The first cut was **+8.1 kB and failed the gate**; after the split it is **+2.2 kB page/worker combined**, inside the 5 kB allowance. No ceiling was raised. The split also keeps `secrets-bridge-client`, `mount-bridge-client` and the sudo brokers off the boot graph.
- **Floats**: no behaviour change. The only live call site (`shell-and-skills.ts` webhook topology) asks the same question and gets the same answer; every other operation was unavailable before and is now either implemented or still unavailable.
- **Security**: the REST adapter reads secrets only from `/api/secrets/masked` — a real value cannot be requested through the broker. An approval fails closed on any unrecognised body, and a `suggestedPattern` is forwarded, never suggested (suggesting costs an LLM call and is policy above the broker). Both server contract suites use a sudo backend that throws if reached, so no case can raise a real OS prompt in CI.
- **Persisted state**: none.

## Known gap, recorded rather than hidden — tracked as #2806

`POST /api/secrets` (persisted secret creation) exists on node-server and **not** on swift-server, so a `secrets.set` with the default `persisted` scope is unavailable on a Swift float. Session-scoped sets work on both. The fixture carries a `servers` field marking it node-only and explaining why, and the Swift replayer skips it — so this is a *recorded* gap, not a silent one: any other drift in either server's route table still fails the contract suites.

Closing it is **#2806** (a Keychain-backed `POST /api/secrets` on swift-server), referenced from the fixture's `//servers` note and the adapter header. When it lands, both `servers` markers and the note come out and the existing Swift replayer starts enforcing the two entries with no test changes. A Keychain write route deserves its own review rather than a drive-by inside a broker-adapter PR.

## Test plan

Everything below was run on this branch:

- `npm run lint` → exit 0; `npm run lint:ci` → exit 0 (43 pre-existing warnings, 0 errors)
- `npm run typecheck` → clean (all 9 projects)
- `npm run test` → **19049 passed**, 52 skipped, 0 failed
- `npm run test:coverage:webapp` → 84.91 lines / 82.92 stmts / 82.92 funcs / 74.55 branches, all above floors; `coverage-gate.mjs webapp` and `node-server` both exit 0. The new `work-unit/capability` dir measures ~94% statements.
- `swift test` in `packages/swift-server` → **602 passed, 0 failures** (sandbox off, `AI_ALIGNED_XCODE_DISABLE=true`)
- `npm run build -w @slicc/webapp` and `-w @slicc/chrome-extension` → both clean
- `check-first-load-size.mjs` → page +0.1 kB, worker +2.2 kB, PASS
- `check-touched-exemptions.mjs` → exit 0

New tests: the conformance suite runs each adapter twice — once over a transport that answers, once over one that fails every call — which is what caught the extension adapter letting a rejected Port escape as a throw. `capability-extension-adapter.test.ts` drives the **real** default transports (a scripted `chrome.runtime.sendMessage` and a scripted `chrome.runtime.connect` port), so a Port name or message-shape change fails here rather than in a float.

## Docs

- `docs/work-unit.md`: new phase-table row 6b, an adapter table, and the reasoning for the three protocol changes.
- `packages/webapp/CLAUDE.md`: unchanged — its existing `CapabilityBroker` rule is still exactly right, and the file is at its size cap.
- `docs/review-patterns.md`: unchanged — category 10's wording already covers this.

## Open questions — both answered

- **`CapabilityResult`'s third variant**: keep it. The `CapabilityFailure` / `CapabilityUnavailable` split is the right one and "never throws" stays load-bearing.
- **The Swift persisted-secret gap**: does not block this slice; filed as #2806 and referenced from the fixture and the adapter header.

Part of #2276


trieloff:
Heads-up on the recorded gap in `packages/shared-ts/fixtures/capability-rest-contract.json`: #2823 implements `POST /api/secrets` on swift-server (Keychain-backed, byte-mirroring `packages/node-server/src/routes/secrets.ts`), which closes #2806.

I based it on `main` rather than stacking on this branch — the fixture only exists here, but stacking would have cost it verification CI, and the Swift route is independent of the adapter work. So the fixture cleanup is **not** in #2823 and whichever of us lands second should do it. It is a pure deletion:

- `"servers": ["node"]` on the `secrets.set/persisted` operation entry
- `"servers": ["node"]` on the `persisted set rejects a body without domains` server case
- the `//servers` note

`Tests/CapabilityRestContractTests.swift` then starts enforcing both entries with no test changes, and it will pass — #2823 implements both, including the 400-on-a-body-without-domains case that fixture pins (`testPostSecretRejectsBodyWithoutDomains`).

Nothing breaks if this is missed in either order: the replayer skips entries its server is not listed for, so a stale marker only under-enforces on Swift. The `//servers` note's "closing it is issue #2806" line just goes stale.

One thing worth a look from this side: swift-server refuses an explicitly **empty** `domains` array fail-closed, and I mirrored node-server's **500** (from `EnvSecretStore.set`) rather than "improving" it to a 400, since the premise here is one wire for two servers. If the adapter maps a 500 differently from a 400, that's the one place the two servers' rejection semantics are worth double-checking against `rest-ops.ts`.

Issue #2040, opened 2026-08-09
DIP unlock: ffmpeg multi-threaded core (@ffmpeg/core-mt) on isolated leaders


trieloff:
With the leader cross-origin isolated via Document-Isolation-Policy (#2039), SharedArrayBuffer is available in the kernel worker — so ffmpeg.wasm's multi-threaded core works.

Today `ffmpeg-wasm.ts` loads the single-threaded `@ffmpeg/core` from the ipk-installed VFS `node_modules` (shadow-map bootstrap `ipk add @ffmpeg/core@<pinned>`). Plan:

- Loader resolves `@ffmpeg/core-mt` first when `globalThis.crossOriginIsolated === true` and the package is installed; falls back to `@ffmpeg/core` otherwise (embedded floats — Cherry/spoon/Electron — can never isolate, so the single-thread path stays the universal baseline).
- Shadow-map/help hint mentions the `-mt` option when isolated.
- Validate live in the Tier-1 harness: `ipk add @ffmpeg/core-mt`, transcode something, compare wall-clock vs single-thread.

Part of the DIP unlock series (#2036, PR #2039).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01G8Pt2nX6koz4K868SWBMN9

PR #2809, opened 2026-09-03
feat(webapp): ffmpeg engines: core-mt guidance, WORKERFS inputs, mediabunny fast path

trieloff:
## Summary

Three improvements to the `ffmpeg` / `ffprobe` shell commands (one commit each), plus a first-use split of the command body:

1. **Multi-threaded core: opt-in, with guard rails.** The loader has been able to boot `@ffmpeg/core-mt` on a cross-origin-isolated leader since #2040, but no install hint ever named it, so it never ran — and when I ran it live on the standalone harness it **deadlocks on every multi-input job** (ffmpeg's per-input demux threads call `pthread_create` from inside a pthread; emscripten proxies that to a main thread blocked in `exec`; details and the bisection table in #2810). Single-input encodes work and are several times faster. So: `@ffmpeg/core` stays the default everywhere; `FFMPEG_CORE=mt` opts into the mt core, which refuses multi-input jobs with the reason and gets `-threads 8 -filter_threads 2` injected unless the caller set them (ffmpeg's defaults exhaust the 32-worker pool on a 16-core machine). Install guidance names the core the loader will boot; `-version` prints a `core:` line with the package, thread count, the opt-in, and the limit. `@ffmpeg/core-mt` is a devDependency for a live layout canary.
2. **Inputs never enter the wasm heap.** Each run mounts its inputs read-only via emscripten WORKERFS from lazily read `Blob`s. New `VirtualFS.getNativeFile()` hands out the native OPFS / FSA `File` so nothing is read up front; other backends read once into a `Blob` that still stays out of the 2 GiB heap. Input size is now bounded by disk instead of "fits in the heap twice"; only the output is buffered.
3. **mediabunny fast path behind the same argv.** One-input → one-output jobs whose every option maps exactly run on mediabunny (WebCodecs hardware encoders, streaming reads, no 31 MB core). Everything else stays on the wasm core. `ffprobe` answers from mediabunny's container index with typed fields and no wasm boot. `FFMPEG_ENGINE=auto|wasm|mediabunny` forces an engine; stderr names the one that ran.
4. **`ffmpeg` body loads on first use.** `ffmpeg-command.ts` becomes the same registration stub `ffprobe` and `pdftotext` already use; the body moves to `ffmpeg/run.ts` unchanged. The kernel worker's eager payload goes from +4.1 kB to **-18 kB** vs main.

## Why

The recorded `RangeError: Array buffer allocation failed` at ~1.89 GB was an input copied three times (kernel worker → transfer → MEMFS) plus its output, all inside a 2 GiB heap. The mt core was wired but unreachable. And WebCodecs is 10-50× faster than wasm x264 for the common transcode, without a download or an isolation requirement.

- **Translator is conservative on purpose.** `ffmpeg/bunny-translate.ts` is pure argv → `BunnyPlan`. Any option without a WebCodecs equivalent REJECTS with the option named, never a silent drop, so the failure mode is "the wasm core ran it", not "a subtly different file". Effort-only knobs (`-preset`, `-tune`, `-threads`, `-maxrate`, …) are accepted as no-ops; `-c copy` and `-bsf:v h264_mp4toannexb` / `aac_adtstoasc` are honoured because mediabunny handles them itself.
- **`declined` vs `failed`.** Pre-flight declines (unreadable container, a track mediabunny would drop on its own such as no encoder in this browser) fall back to wasm, which then either does the job or reports ffmpeg's own error. A mid-run failure is the exit code; redoing minutes of work on a broken input would be worse.
- **Default codec divergence, documented.** Without an explicit `-c`, ffmpeg re-encodes to the muxer default; mediabunny copies streams the container can hold. Faster and lossless, and `FFMPEG_ENGINE=wasm` restores ffmpeg's behaviour when byte-identical output matters.
- **Staging names are relative** (`__in<id>/<name>`) so the concat demuxer's `safe` check keeps passing; the stage id also namespaces `__out<id>_<basename>`, closing a latent collision between concurrent runs on the realm-shared core.
- **`getNativeFile` joins the VFS read surface**: `RestrictedFS` gates it like `readFile` (answering `null`; the caller's fallback read raises the sandbox ENOENT), `sudo-fs` lists it in `READ_ASYNC`, `VfsAdapter` exposes it beyond just-bash's `IFileSystem` for duck-typing.
- **New dependency**: `mediabunny@1.55.6` (MPL-2.0, zero runtime deps), reached only through `import()` from `ffmpeg/bunny-run.ts` / `bunny-probe.ts`, so it never rides the kernel worker's eager graph. Vite dev pre-bundles it like magick-wasm. `@ffmpeg/core-mt@0.12.10` is a devDependency for the canary only.
- **Not in this PR**: streaming the *output* to OPFS. Both engines still buffer it (MEMFS / `BufferTarget`, the latter in the JS heap rather than the wasm heap). That needs a VFS write stream plus an "adopt handle into the ZenFS index" API; noted in `docs/webapp-details.md`.

## Changes since the first review round

- **Codex findings** (all three fixed, with tests): explicit `-c:v libx264` now always re-encodes and `-c copy` never does (a copy-mode track whose codec the container cannot hold is *declined* to wasm, which then fails like ffmpeg); `-vf` chains are accepted only in the order mediabunny applies them (`crop,scale`, `transpose` alone); `-shortest` is rejected instead of ignored.
- **Multi-threaded core demoted to opt-in** after the live run showed it deadlocks on every multi-input job (#2810). Default guidance is `@ffmpeg/core`; `FFMPEG_CORE=mt` opts in with a multi-input refusal and an injected thread budget.
- **Two pre-existing bugs found only by running it live**: (1) a failing wasm run reported `exited with code 1` with none of ffmpeg's output, and `-f null` analysis sinks returned the load log instead of the filter measurements, because the result was built from a stderr snapshot taken before `exec` — fixed with a live reader plus fakes that log during exec; (2) a VFS-native `File` was, in some sessions, not structured-cloneable and read as garbage once wrapped — now every lazy `File` is clone-probed and checked (size + first 64 bytes) against the VFS's own view before it is mounted, falling back to a VFS read with a stderr note. Neither reproduced in later sessions; the guards stay.
- The `ffmpeg` body moved behind a first-use stub (`ffmpeg/run.ts`), and the duplicated command table the rebase re-introduced in `docs/shell-reference.md` is collapsed again.

## Cross-cutting impact

- **Floats**: all of them run the same code in the hosted leader's kernel worker. Only the hosted origin is cross-origin isolated (Document-Isolation-Policy, Chrome ≥ 137), so only there does `-version` recommend `core-mt`; Cherry / Electron keep the single-threaded guidance. mediabunny needs no isolation and works everywhere WebCodecs does; where a codec is missing it declines and wasm takes over.
- **Other callers**: `ffprobe/run.ts` shares the staging module and the engine switch. The camera-capture transcode path (`transcodeCapturedBytes`) moved to the same WORKERFS staging. `FFMPEG_CORE_NOT_INSTALLED` is unchanged for non-isolated realms, so existing tests and copy stay valid.
- **Persisted state**: none. `FFMPEG_ENGINE` is a plain shell env var.
- **Async lifecycle**: the `faulted` guard still keeps every teardown call (`unmount`, `deleteDir`, `deleteFile`) away from a recycled core. mediabunny `Input`s are disposed in `finally`.
- **Bundle size**: mediabunny is a lazy chunk (~660 kB minified) reached only via `import()`; the eager graphs shrank (see first-load below). The webapp total-JS size-limit ratchet now sits at 24.83 of 25 MB — headroom is thin, and the next big lazy dependency will need that budget revisited.
- **Extension smoke test** (`continue-on-error` in CI) runs `ffmpeg -i in.wav -c copy out.wav`, which now takes the mediabunny path and no longer loads `ffmpeg-core.js`; its assertion on that was already stale since the core moved to ipk/VFS.

## Test plan

- [x] `npm run verify` — biome, prettier, custom lints, boy-scout debt, manifest, deadcode, deadcode-prod, autofix-drift: all 8 pass
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — nothing to gate
- [x] `npm run typecheck` — all nine projects clean
- [x] `npm run test` — 1077 files / 19109 tests passing, 0 new failures (rerun after the stub split: identical)
- [x] `npm run test:coverage:webapp` — gate passes (webapp statements 83%, lines 85%)
- [x] `npm run build -w @slicc/webapp` + `npm run build -w @slicc/chrome-extension` — both clean
- [x] `npm run bundle-size` (incl. first-load delta gate) — first-load page +0.0 kB, worker **-18.0 kB** vs merge-base (allowance 5 kB); size-limit webapp total 24.83 MB of 25 MB, extension 100.71 kB of 105 kB
- [x] New behaviour covered: `tests/shell/supplemental-commands/ffmpeg/{staging,input-blob,bunny-translate,bunny-run,bunny-probe,engine}.test.ts`, `tests/fs/{native-file,get-native-file}.test.ts`, plus new cases in `ffmpeg-command.test.ts` (WAV converts through mediabunny with `getFfmpeg` never called; `FFMPEG_ENGINE` forcing; fallback note), `ffprobe-command.test.ts` (WAV probe with no wasm boot), `sudo-fs.test.ts`, `ffmpeg-wasm.test.ts`, `ffmpeg-wasm-live.test.ts` (real `@ffmpeg/core-mt` layout).
- [x] Live on the standalone harness (`dev:standalone:fresh`, bridge :5715, own wrangler :8797, isolated Chrome for Testing): all on the isolated leader (`crossOriginIsolated === true`, 16 cores), read back through the kernel terminal:
  - `ffmpeg -version` before any install → the ST guidance; after `ipk add -g @ffmpeg/core@0.12.10` → `core: @ffmpeg/core 0.12.10 (single-threaded; the multi-threaded core is opt-in … FFMPEG_CORE=mt, single-input jobs only)`; with `FFMPEG_CORE=mt` → `core: @ffmpeg/core-mt 0.12.10 (multi-threaded, 16 threads; single-input jobs only)`.
  - wasm engine, two lavfi inputs → `/tmp/clip.mp4` (h264+aac, 3 s); wasm engine on a real input via WORKERFS (`FFMPEG_ENGINE=wasm … -vf scale=160:-2`) → written; `-f null` + `silencedetect` → the filter log comes back (49 lines incl. `Input #0`); a file written by one command and mounted by the next → fine.
  - mediabunny: `ffprobe -of json` → `mov,mp4,…`, `h264` 320×240 `30/1`, `aac`, duration 3.000000; `-vf scale=160:-2 -c:v libx264 -c:a aac` → `engine mediabunny (WebCodecs)`, wrote 88 kB; `-c copy` remux to `.mkv` → wrote; `ffprobe -of csv` on an `.m4a` → `stream,aac,44100`.
  - mt opt-in: `FFMPEG_CORE=mt FFMPEG_ENGINE=wasm … -c:v libx264 -c:a aac` single input → ok, `speed=19.4x`; the same with two inputs → refused with the #2810 explanation instead of hanging.

## Documentation

- [x] `docs/shell-reference.md` (ffmpeg / ffprobe rows; also collapsed four duplicated ffprobe rows)
- [x] `docs/webapp-details.md` — new "Media commands" section
- [x] `packages/vfs-root/workspace/skills/ffmpeg/SKILL.md` (engines, install per core, size guidance)
- [x] `ffmpeg --help` / `ffprobe --help`

## Risk & rollback

- Blast radius: the two media commands, all floats. A wrong mediabunny "accept" would mean a file that differs from ffmpeg's; the translator's reject-by-default and the `FFMPEG_ENGINE=wasm` kill switch bound that.
- Rollback: revert cleanly; no migrations.
- What CI cannot catch: real WebCodecs behaviour (hardware encoder availability per platform, AAC on Linux) — mediabunny declines those and wasm runs, but that path is only unit-tested with a fake core. Worth one manual pass on a live leader before relying on it.

## Checklist

- [x] Public API / tool / shell-command changes are intentional and reflected in docs
- [x] Contract used by other floats checked (shared kernel-worker code path)


trieloff:
Addressed the review round and finished the live pass on the standalone harness (`dev:standalone:fresh`, isolated leader, 16 cores). Summary of what changed since the reviewed commit:

**Codex findings — all three fixed with tests**
- Explicit codec intent: `-c:v libx264` always re-encodes (forceTranscode), `-c copy` never does — a copy-mode track the container cannot hold is *declined* so the wasm core reports ffmpeg's own error (`copyViolations()`), and `-c copy` + filter/bitrate is rejected as ffmpeg does.
- Filter order: only chains mediabunny applies in the written order are accepted (`crop,scale`, `fps` anywhere, `transpose` alone); `scale,crop` and any `transpose` mixed with a size change are rejected.
- `-shortest` is rejected instead of silently ignored.

**Multi-threaded core is now opt-in.** Running it live, `@ffmpeg/core-mt` deadlocks on every job with more than one input (per-input demux threads → `pthread_create` proxied to a main thread blocked in `exec`; bisection table in #2810). Single-input jobs work (x264 at 2/4/8 threads). So `@ffmpeg/core` stays the default; `FFMPEG_CORE=mt` opts in, multi-input jobs are refused with the reason, and `-threads 8 -filter_threads 2` are injected unless set.

**Two pre-existing bugs only the live run exposed**, both fixed here: a failing wasm run lost ffmpeg's own log (stderr snapshot taken before `exec` — `-f null` analysis sinks were returning the load log instead of the filter measurements), and a VFS-native `File` was in some sessions not cloneable / read as garbage once mounted — lazy Files are now clone-probed and verified against the VFS (size + 64-byte head) before mounting.

Live results (all read back through the kernel terminal): wasm lavfi encode, WORKERFS-mounted real input, `-f null` sink with its filter log, mediabunny probe JSON / WebCodecs transcode / `-c copy` remux, `FFMPEG_CORE=mt` single-input encode at 19× and the multi-input refusal. Details in the updated description.

Issue #2382, opened 2026-08-24
feat(ui): mount the WC shell on WorkUnitClient — transcript, queue, composer, model, one mount path


trieloff:
Follow-up to #2274 (part of #1666). The client protocol landed with both adapters, one conformance suite, and the **tab strip** rendering from `toTabDescriptors` on leader and follower. The rest of the shell still has two wirings.

## What is left

| Surface | Leader today | Follower today | After |
| --- | --- | --- | --- |
| transcript | `requestScoopMessages` → `onScoopMessagesReplaced` → `controller.loadMessages` | `onSnapshot` → `controller.loadMessages` | `client.snapshot(id)` / `subscribe(id)` |
| selection | `setSelectedScoopJid` + replay | `sync.selectScoop` + `onSnapshot` | `client.snapshot(id)` (selection IS the call) |
| queued pile | `queuedIds` on the replay envelope (#2362) | not sent | `snapshot.queuedIds`, absent ≠ empty |
| composer send | `AgentHandle.sendMessage` (implicit selection) | `sync.sendMessage` (implicit selection) | `client.send(id, input)` |
| stop | `stopScoop(jid)` | `sync.stop()` | `client.signal(id, 'stop')` |
| model pill | `modelFor(record)` + `set-scoop-model` | `wc-follower-model-surface` + `model.select` | `summary.model` as the single read |
| mount | `mountWcUiLive` / `mountWcUiExtension` | `mountWcUiFollower` | one mount path taking a `WorkUnitClient` |

## Why it was not in #2274

`attachWcClient` binds `OffscreenClient` to the VFS, the terminal, the monitor, sprinkles and the permissions surface as well as the chat — none of which the protocol covers or should. Collapsing the mounts means deciding what those get instead, which is a design question of its own (and overlaps #2276's `CapabilityBroker`). Shipping the protocol with the strip cut over keeps the risk of each step legible.

## Done when

- [ ] The transcript, the queued pile, the composer and the model pill read from `WorkUnitClient` on both sides.
- [ ] A tab click goes through `client.snapshot(id)` rather than two transport-specific selection calls.
- [ ] `mountWcUiLive` / `mountWcUiFollower` / `mountWcUiExtension` share one mount that takes a client.
- [ ] The conformance suite grows the cases the new surfaces need (queue reconcile parity, send/steer, stop mid-turn).
- [ ] No follower feature needs a parallel UI state machine (review-patterns category 8).

Legacy path deletion stays with #2365.

[...]

Issue #2354, opened 2026-08-24
Reconcile a held queue against the backend, ahead of the processing transition (#2312 follow-up)


trieloff:
Follow-up to #2312 / PR #2327 (Codex P2 on `wc-chat-controller.ts`).

#2327 lets a cone's queued pile survive a read-only detour into one of its own scoops (`stashQueued` / `restoreQueued`), and reconciles the held snapshot against the canonical replay on the way back, so a prompt the cone consumed while the user was away is no longer re-added as a phantom queue card.

What is **not** covered yet is the ordering around the processing transition:

- `selectScoop` calls `setProcessing(client.isProcessing(jid))` synchronously, while the queue restore lands later, inside the async `loadMessages` replay.
- The queued-flush is **edge-triggered** on the processing rising edge. Returning to a cone whose turn is already running therefore sees the rise while the pile is still empty; the restored prompts are installed afterwards and wait for the *next* rise.

In practice the replay dedupe covers the case that actually loses data (a consumed prompt showing twice / forever). The residue is a display-ordering nuance: a prompt that is genuinely still pending stays rendered as queued until the following turn starts, rather than flushing with the turn that consumes it.

Proper fix is to reconcile against the **backend** queue rather than only the replay, and to sequence the restore ahead of the processing transition — both of which want a way to ask the orchestrator for a scoop's pending queue, which does not exist on the panel protocol today.

Scope:
- add a queue-read to the panel protocol (or piggyback the pending ids on the messages replay);
- restore + reconcile before `setProcessing`;
- test: return to a cone mid-turn with a held prompt that the running turn then consumes.

PR #2820, opened 2026-09-03
feat(webapp): render the transcript and selection through WorkUnitClient

trieloff:
## Summary

PR **B** of four for #2382. The transcript and the selection now go through
`WorkUnitClient` on the leader mount and the dedicated follower mount:
selection IS `snapshot(id)`, and the thread is rendered from `subscribe(id)`.

Scope, precisely: `onScoopMessagesReplaced` is gone from the leader's callback
bag; the follower's `onSnapshot` handler survives but only for the selection
bookkeeping the frame also carries (the leader names the unit it mirrors, and
on a fresh join that is how the follower learns which one it has) — it no
longer reads the transcript. The leader-capable float's own follower path
(`wc-tray.ts` `buildFollowerOptions`) is the third follower wiring and is
STILL on the raw frames; it collapses in PR D, as it did for the composer in
PR A.

Stacked on PR A (#2803, merged). Rendering is unchanged; first-load is +0.0 kB.

## Why

The transcript was the largest remaining row of the two-implementations table
in `docs/work-unit-client.md`: a kernel replay reached the thread through the
leader's callback bag, a tray frame through the follower's options bag, and
nothing said the two had to agree. Selection was `setSelectedScoopJid` +
`requestScoopMessages` on one side and `sync.selectScoop` on the other — which
is exactly what the protocol's `snapshot(id)` already performs.

## Approach

**Rendered from `subscribe`, not from the awaited `snapshot`.** Awaiting
`snapshot(id)` and calling `loadMessages` with the result would paint the same
replay the subscription already delivers, and — worse — the local adapter's
no-answer fallback resolves with an *empty* transcript after 5 s, which must
never be allowed to wipe a live thread. The subscription only ever sees
snapshots a transport really published, so it is the safe painter.
`snapshot(id)` is still the call that performs the selection.

**Every replay now arrives on one path.** The kernel pushes replays the shell
never asked for (a new session, a compaction, a freezer restore); those used to
come through `onScoopMessagesReplaced`. A live per-unit subscription covers both
that and the one a selection asks for.

**One seed rule, in the adapters.** A subscriber that joins an in-flight
`snapshot(id)` neither re-asks nor takes the cached seed — both would produce a
second wholesale render of a transcript about to be replaced (dips disposed and
rehydrated, a flash of stale history, and the one-shot held-queue restore
consumed against a stale `queuedIds`, #2354). A tab click asks before
subscribing and so sees only the fresh replay; a fresh join and a guest seat
have nothing in flight and still get their seed, which for them is often the
only snapshot that unit will ever have. The mounts decide nothing.

**An unanswered request is recovered, once.** That dedup makes the single
request load-bearing: if it is dropped, nothing else would re-issue it, and the
PREVIOUS unit's transcript would stay on screen while the composer, thread
context and navbar attention already belong to the new one — the user reads B
and sends into A. `LocalWorkUnitClient` re-asks once on timeout; if that is
dropped too it publishes the unit's own last-known transcript (or an empty one)
with `queuedIds` absent, so the held pile stands. The recovery answer is
emitted, never cached — it is the client's, not the transport's.

**`resetSelection` drops the dead channel's transcripts.** `addFollower`
re-sends a snapshot and the roster back to back on reconnect and the roster can
win that race, so a seed from the old channel could paint a transcript the
leader has since frozen or cleared.

**`WorkUnitSnapshot.summary` is now optional.** This is the one protocol change
and it fixes a real bug the cutover would otherwise have shipped: a **biscotto
seat is pinned to one thread and is deliberately never sent `scoops.list`**
(`sendScoopsListToFollower` refuses it — the inventory's labels would leak what
else the owner is working on). The remote adapter used to *hold* a snapshot
until the roster named its unit, so for a guest that hold was permanent and the
thread would have rendered blank. It now publishes un-rostered snapshots with
no `summary`, and the `earlySnapshots` machinery is gone. No `src/` consumer
reads `snapshot.summary`.

**`LocalWorkUnitClient.subscribe` no longer re-asks** for a transcript that a
`snapshot(id)` is already fetching — two asks meant two replays and therefore
two `loadMessages`, and each one consumes the held-queue restore.

Two extractions were needed to stay under the function-size cap and are pure
moves: `createUnitWatcher` and `reconcileQueueForSwitch` in `wc-live.ts`.

## Cross-cutting impact

- **Floats**: leader (CLI + extension) and follower (standalone, cherry, hosted,
  extension side panel) both render from the protocol. The leader-capable float
  in `wc-tray.ts` still has its own follower wiring; it collapses in PR D.
- **Followers / wire**: no tray wire field changes. iOS untouched.
- **Guest seats**: the un-rostered publish is a fix, not a regression — see
  above. The follower conformance case pins it.
- **Persisted state**: none. **Bundle**: page and worker both +0.0 kB.

## Test plan

Run locally on this branch:

- `npm run lint` then `npm run lint:ci` — clean. Two function-size caps were
  paid down by extraction rather than exempted.
- `npm run test` — **19244 passed**, 0 failed.
- `npm run test:coverage:webapp` — lines 85.00 / statements 83.01 / functions
  83.03 / branches 74.65, all above the floors (84 / 82 / 82 / 73).
- `npm run build -w @slicc/webapp`, `npm run build -w @slicc/chrome-extension`,
  `check-first-load-size.mjs` — page 759 kB, worker 4217 kB, **+0.0 kB** each.
- `check-touched-exemptions.mjs` — clean.
- **Fake-LLM e2e** on isolated ports (`SLICC_E2E_WRANGLER_PORT=8817
  SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8818 SLICC_E2E_BRIDGE_PORT=5735
  SLICC_E2E_CDP_PORT=9252 SLICC_E2E_FAKE_LLM_PORT=5795`):
  `multiple-cones`, `multiple-cones-follower` and `multiple-cones-licks` —
  **5 passed (1.1m)**, against a real leader + real follower pair. Re-run after
  the review fixes.

New conformance cases (over **both** adapters): queue reconcile parity (`[]`
locally vs `undefined` remotely, from one envelope), a second snapshot
superseding the first, a snapshot delivered before the roster names its unit,
and a subscribe joining an in-flight snapshot without a second transport ask.
The mount tests that used to drive the callback bag now drive the real protocol
path.

## Docs

- `docs/work-unit-client.md` — the optional-`summary` rule, the two adapters'
  different answers to a pre-roster snapshot, the seed policies, and a
  "Mounting the shell on the client (#2382)" section.
- `docs/work-unit.md` — phase row 5c.

## Review rounds 1 and 2 (applied in `5997d03cd`)

Both adversarial reviewers plus the parent's read. Eight items, all applied:

- **Dropped replay never recovered (P1)** and **wrong unit under the new chrome
  (P1)** — folded into one rule for the local adapter's timeout path: retry
  once, then publish this unit's own transcript with no `queuedIds`. Four
  fake-timer tests, and a leader A→B→A mount test asserting the last
  `loadMessages` carries the destination's ids. Mutation-tested: removing the
  publish fails three of them.
- **Follower double render on re-selection (P2)** — fixed with the in-flight
  gate above rather than by call site, per the round-2 refinement.
  Mutation-tested against the mounted follower.
- **`resetSelection` leaked the old session's caches (P1)** — now clears
  `lastSnapshots` and `pendingSnapshots`; conformance case added.
- **Weak conformance assertion (P2)** — "at most one request" became "exactly
  `['cone_1']`", which no longer passes a client that never asks.
- **Wording and docs drift (P2/P3)** — the "both mounts" claim is now precise
  about `wc-tray.ts`; the `WorkUnitClientEvent` snippet no longer lists the
  `error` variant the prose denies; the follower's surviving `onSnapshot` is
  described as bookkeeping only; the stale "held until the roster names it"
  comment is gone; `work-unit.md` row 5b reads "done (#2803)".

Part of #2382


trieloff:
## Screenshots — rendering is unchanged

Both pairs are the same e2e spec (`multiple-cones-follower.test.ts`, "follower
mirrors the cone strip and drives one cone's model") run against a real leader
+ real follower pair over the fake LLM: once on `main`, once on this branch.
Frames come from the Playwright traces, so they are the rendered UI. Timestamps
differ; nothing else does.

### Follower — the transcript now comes entirely from `subscribe`

The turn on the right ("review the docs" → "Reviewer cone online.") used to be
painted by the mount's `onSnapshot` handler. It is now painted by
`WorkUnitClient.subscribe`, and the mount's handler is gone.

| before (`main`) | after (this PR) |
| --- | --- |
|  |  |

### Leader — selection is `snapshot(id)`

`setSelectedScoopJid` + `requestScoopMessages` became one `snapshot(id)`, and
the replay it asks for arrives on the same subscription as every replay the
kernel pushes on its own.

| before (`main`) | after (this PR) |
| --- | --- |
|  |  |

### What the screenshots cannot show

The guest-seat fix has no e2e: a biscotto seat never receives `scoops.list`, so
the remote adapter's old "hold until the roster names the unit" would have held
its transcript forever once the mount stopped reading `onSnapshot` directly.
That is covered by the conformance case "delivers a snapshot that arrives
before the roster names its unit", which asserts the un-rostered publish and
its absent `summary`.

PR #2825, opened 2026-09-03
fix(shell): make advertised bash builtins runnable instead of 127

### LINKED ISSUE #2816: Shell builtins jobs/bg/fg/trap/select are in `help` but not runnable (127)
## Summary

Five shell builtins are advertised by the shell's own `help` table but are **not runnable** — every invocation exits 127 `command not found` (or a syntax error for `select`). This looks like a gap between the documented builtin table and what `just-bash` actually implements.

Found while auditing `/man/*` pages on www.sliccy.com against real command behaviour.

## Reproduction

Each was run in a clean, non-interactive `bash -c` to rule out job-control noise from the interactive session:

```console
$ bash -c 'sleep 5 & jobs'
bash: jobs: command not found

$ bash -c 'sleep 5 & bg'
bash: bg: command not found

$ bash -c 'sleep 5 & fg'
bash: fg: command not found

$ bash -c "trap 'echo x' INT; echo trap-ok"
trap-ok
bash: trap: command not found

$ bash -c 'select x in a b; do echo $x; break; done <<< 1'
bash: syntax error near unexpected token 'select'
```

| Builtin | Symptom | Exit |
|---|---|---|
| `jobs` | `command not found` | 127 |
| `bg` | `command not found` | 127 |
| `fg` | `command not found` | 127 |
| `trap` | `command not found` — but see note | 127 |
| `select` | `syntax error near unexpected token 'select'` | 2 |

### `trap` is the interesting one

`trap` is **partially** wired up: the command after it still runs (`trap-ok` is printed) and *then* the `trap: command not found` error is emitted. So the parser accepts the statement and continues, but no builtin backs it — signal handlers are silently not installed. Code that relies on `trap` for cleanup will appear to work while never actually running its handler, which is worse than a hard failure.

### `select` differs in kind

`select` fails at **parse** time, not lookup time, so it is a missing shell *keyword* rather than a missing builtin. It probably belongs to a different fix than the other four.

## Expected

Either:
1. Implement these (at minimum `trap`, since silent no-op handlers are a correctness trap), **or**
2. Remove them from the `help` builtin table so they are not advertised as available.

Job control (`jobs`/`bg`/`fg`) may be genuinely out of scope for a browser shell — in which case dropping them from `help` is the honest fix, and `trap` failing loudly instead of half-working would still be an improvement.

## Builtins that DO work

Verified working in the same session, for contrast: `cd`, `export`, `set`, `eval`, `exec`, `source`, `read`, `wait`, `shift`, `unset`, `getopts`, plus control flow `if/fi/for/do/while/until/case/then`.

## Environment

- SLICC browser shell (`just-bash` WASM)
- Observed 2026-09-03

trieloff:
Fixes #2816

## What was wrong

just-bash ships bash's complete `help` topic table while implementing only part of it. Thirteen advertised names reached command lookup and answered `command not found` (127) — five more than the issue found:

`bg` · `caller` · `disown` · `enable` · `fc` · `fg` · `jobs` · `logout` · `suspend` · `times` · `trap` · `ulimit` · `umask`

(The issue also listed `kill` and `exec` as working — they are, because SLICC registers `kill` and just-bash implements `exec`. A bare-just-bash probe reports both as missing; the numbers above come from probing the real `AlmostBashShellHeadless`.)

`trap` was the dangerous one, exactly as the issue says: the parser accepted `trap 'cleanup' EXIT`, the script kept running, and a handler that was never installed looked like it had worked.

## What this does

Registers all thirteen in `bash-builtins-command.ts`. Custom commands are consulted only *after* builtins, so these names reach dispatch precisely because upstream has no builtin for them — the list can never shadow one just-bash later implements.

Two behaviours, no third.

**Faithful** — where real bash without job control already answers with a diagnostic, this shell answers with bash's own text and exit code (verified against `/bin/bash -c`):

| command | behaviour |
|---|---|
| `jobs` | empty table, exit 0; a jobspec gets `%1: no such job`, exit 1 |
| `fg` / `bg` | `bash: fg: no job control`, exit 1 |
| `suspend` | `bash: suspend: cannot suspend: no job control`, exit 1 |
| `disown` | `bash: disown: current: no such job`, exit 1 |
| `logout` | ``bash: logout: not login shell: use `exit'``, exit 1 |
| `trap -l` | the five signals the kernel can actually deliver (`kill --help`) |
| `trap` / `trap -p` | empty trap table, exit 0 |
| `trap - SPEC`, `trap '' SPEC` | exit 0 — nothing is trapped, nothing raises these |

`jobs` is empty rather than wrong: `&` runs its command synchronously here, so a backgrounded command has already finished by the time `jobs` runs. Long-lived kernel processes are `ps` / `kill`'s job.

**Loud** — anything the shell genuinely cannot do exits **2** with a one-line reason and, where one exists, the working alternative. Never a silent no-op, so `set -e` scripts stop instead of carrying on:

| command | why, and what to use instead |
|---|---|
| `trap 'cmd' SPEC` | no signal-delivery path into a running script — the handler could never fire |
| `caller` | the interpreter exposes no caller frames |
| `enable` | builtins cannot be enabled/disabled/loaded; list with `help` or `commands` |
| `fc` | no history editing; use `history` |
| `times` | no per-process CPU accounting; use `time <command>` |
| `ulimit` | interpreter limits are fixed at boot; see `df` and `meminfo` |
| `umask` | the VFS has no file-creation mask; use `chmod` |

Every one of the thirteen answers `--help` with real usage.

This is option 2 from the issue plus the part of option 1 that is honestly available. Implementing `trap` for real is not reachable from a custom command — there is no signal delivery into a running script and no interpreter hook to install one — and a partial `trap` would reintroduce the exact "looks like it worked" failure. So the handler form fails loudly instead, which the issue names as an improvement in its own right.

## `select`

Out of reach and left alone. It is a missing shell **keyword**, not a builtin, so it fails in just-bash's parser (`syntax error near unexpected token 'select'`, exit 2) before command lookup happens — no registration can reach it. It is also *not* in the `help` table (`help select` → `no help topics match`), so nothing advertises it. Documented in `docs/shell-reference.md`; a real fix belongs upstream in the just-bash parser.

## Tests

`bash-builtins-command.test.ts` (42 tests). The load-bearing one walks the **live `help` listing** through a real `AlmostBashShellHeadless` and asserts no advertised name answers 127 — that is what keeps the table and dispatch from drifting apart again across a just-bash upgrade, rather than a hardcoded list that goes stale.

## First-load budget

Behaviour and usage text live in `bash-builtins/run.ts`, imported on first use. `supplemental-commands/index.ts` is boot-critical; a single-file version put 5.5 kB into the eager worker closure and failed the first-load gate. Split, it passes (`First-load OK`).

## Verification

`npm run verify` (8/8) · `typecheck` · `test` (19382 passed) · `test:coverage:webapp` (exit 0) · `build` · `build -w @slicc/chrome-extension` · `bundle-size` (First-load OK) · `check-touched-exemptions`.

### DIFF
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -40,16 +40,16 @@ just-bash ships bash's complete `help` topic table while implementing only part 
 
 **Faithful** — where real bash without job control already answers with a diagnostic, this shell answers with bash's own text and exit code:
 
-| command                       | behaviour                                                     |
-| ----------------------------- | ------------------------------------------------------------- |
-| `jobs`                        | empty table, exit 0; a jobspec gets `%1: no such job`, exit 1 |
-| `fg` / `bg`                   | `bash: fg: no job control`, exit 1                            |
-| `suspend`                     | `bash: suspend: cannot suspend: no job control`, exit 1       |
-| `disown`                      | `bash: disown: current: no such job`, exit 1                  |
-| `logout`                      | ``bash: logout: not login shell: use `exit'``, exit 1         |
-| `trap -l`                     | the five signals the kernel can deliver (see `kill --help`)   |
-| `trap` / `trap -p`            | empty trap table, exit 0                                      |
-| `trap - SPEC`, `trap '' SPEC` | exit 0 — nothing is trapped, and nothing raises these         |
+| command            | behaviour                                                                                        |
+| ------------------ | ------------------------------------------------------------------------------------------------ |
+| `jobs`             | empty table, exit 0; a jobspec gets `%1: no such job`, exit 1                                    |
+| `fg` / `bg`        | `bash: fg: no job control`, exit 1                                                               |
+| `suspend`          | `bash: suspend: cannot suspend: no job control`, exit 1                                          |
+| `disown`           | `-a`/`-r` sweep the empty table and exit 0; every other form gets `current: no such job`, exit 1 |
+| `logout`           | ``bash: logout: not login shell: use `exit'``, exit 1                                            |
+| `trap -l`          | the five signals the kernel can deliver (see `kill --help`)                                      |
+| `trap` / `trap -p` | empty trap table, exit 0                                                                         |
+| `trap - SPEC`      | exit 0 — restores the default disposition, which is what an untrapped signal already gets        |
 
 `jobs` is empty rather than wrong: `&` runs its command synchronously here, so a backgrounded command has already finished by the time `jobs` runs. For long-lived kernel processes use `ps` and `kill`.
 
--- a/docs/shell-reference.md
+++ b/docs/shell-reference.md
@@ -32,6 +32,43 @@ This is a **convention, not a sandbox** — one unit can still read and write an
 
 Its cleanup boundary is the explicit **New session** control: **Save & start new**, **New chat — skip memory**, and **Erase & start new** each remove the entries under the selected cone's own `$TMPDIR` before its chat is cleared — a sibling cone's scratch is never in the blast radius. Active mount roots below it and the directories containing them stay attached and are never traversed, so mounted Local, S3, and DA contents are not treated as scratch data. Page reload, app restart, and scoop creation do not clear `/tmp`.
 
+### Bash builtins: `help` lists more than just-bash implements
+
+just-bash ships bash's complete `help` topic table while implementing only part of it. Thirteen advertised names — `bg`, `caller`, `disown`, `enable`, `fc`, `fg`, `jobs`, `logout`, `suspend`, `times`, `trap`, `ulimit`, `umask` — reached command lookup and answered `command not found` (127). `trap` was the dangerous one: the parser accepted `trap 'cleanup' EXIT`, the script kept running, and a handler that was never installed looked like it had worked.
+
+`packages/webapp/src/shell/supplemental-commands/bash-builtins-command.ts` registers all thirteen; the behaviour and usage text live in `bash-builtins/run.ts`, imported on first use because `index.ts` is boot-critical. Custom commands are consulted only after builtins, so these names reach dispatch precisely because upstream has no builtin for them — they can never shadow one just-bash later implements. Two behaviours, no third:
+
+**Faithful** — where real bash without job control already answers with a diagnostic, this shell answers with bash's own text and exit code:
+
+| command                       | behaviour                                                     |
+| ----------------------------- | ------------------------------------------------------------- |
+| `jobs`                        | empty table, exit 0; a jobspec gets `%1: no such job`, exit 1 |
+| `fg` / `bg`                   | `bash: fg: no job control`, exit 1                            |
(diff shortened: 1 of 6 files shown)


trieloff:
Both review findings addressed in e07b513.

**P1 — `trap '' SPEC` (Codex): confirmed and fixed.** I traced the path before acting: Ctrl+C sends `terminal-signal SIGINT` → `terminal-session-host.ts` `handleSignal` → `pm.signal(pid, 'SIGINT')` → the process `AbortController` aborts the running exec. Nothing on that path consults trap state, so acknowledging a mask would let a "protected" critical section be interrupted anyway — the same silent lie as accepting a handler, which is the whole point of this PR. `trap '' SPEC` now refuses with exit 2 and names the reason.

`trap - SPEC` still succeeds, and I think correctly: it asks for the *default* disposition, and the default for a delivered signal is exactly what already happens. Nothing is claimed that isn't true.

**`disown -a` / `-r`: confirmed against `/bin/bash` and fixed.** Both exit 0 on an empty table; they sweep the whole table, so an empty one satisfies them. `-h` and an explicit jobspec still report `no such job`.

That second one also surfaced a real collision worth flagging: `subcommand-help.ts`'s `isHelpRequest` treats `-h` as a help request, but `-h` is a genuine `disown` flag ("do not send SIGHUP"), so `disown -h` was printing usage instead of erroring. Bash builtins take `--help` only, so this family now matches `--help` alone rather than borrowing the verb-dispatcher helper.

Re-verified: `verify` 8/8 · `typecheck` · `test` (19387 passed) · `test:coverage:webapp` · both builds · `bundle-size` (First-load OK) · `check-touched-exemptions`.

---

Separately, a note for anyone reading the `no job control` wording as final — it isn't. I probed whether real job control is reachable and it is, without patching just-bash:

- `Bash.exec()` applies plugins registered via `registerTransformPlugin`, and the AST carries `Statement.background`. Rewriting background statements into a job-spawning command works, compound statements included.
- Concurrent `exec()` on one `Bash` instance is already isolated — cwd/env don't cross-talk or leak back — which is exactly bash subshell semantics for `&`.
- The substrate exists: `bash-tool.ts` already spawns killable, tee'd, lick-completing detached jobs via `jobHost.spawn`.

The trap there is `wait`: it's an *implemented* builtin, so a custom command can't shadow it, and async `&` without a `wait` rewrite would silently break `cmd & wait` — this exact bug class. Doable in the same transform pass. That's a follow-up PR, not this one.

trieloff:
Filed the job-control findings as #2846 so they don't only live in a PR comment: the `registerTransformPlugin` hook, the concurrency-isolation proof, the existing `jobHost.spawn` substrate, and the `wait` trap that any async-`&` change has to handle in the same pass.

Nothing in this PR blocks on it — #2846 replaces the `jobs`/`fg`/`bg` behaviour here wholesale when it lands.

Issue #2821, opened 2026-09-03
bug: Node fetch-proxy does not unmask application/x-www-form-urlencoded bodies (Swift does)


trieloff:
## Summary

Node and Swift `/api/fetch-proxy` disagree on whether `application/x-www-form-urlencoded` request bodies are text for secret unmask. Swift unmasks form POSTs; Node leaves the masked token in the body. A form POST that works on Sliccstart can send the masked secret to the upstream API on the Node CLI.

Spotted while auditing fetch-proxy parity for #2818. Not the JPEG-body bug that PR fixes.

## Float (runtime)

- Standalone CLI / `npm run dev`
- Electron / Sliccstart (macOS app)

## Surface

- Shell / AlmostBashShell / supplemental commands
- Provider / model / API key handling

## Area

- `area/node-server`
- `area/swift`

## Reproduction

1. Register a domain-scoped secret, then POST a form body that contains the masked token:

```javascript
await fetch(url, {
  method: 'POST',
  headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
  body: 'token=$masked$API_KEY&grant_type=client_credentials',
});
```

2. On Sliccstart (Swift): upstream receives the real secret.
3. On Node CLI: upstream receives `$masked$API_KEY` unchanged.

## Suspected root cause / location

Request-side classifiers:

- Node `unmaskRequestBody` in [`packages/node-server/src/routes/fetch-proxy.ts`](packages/node-server/src/routes/fetch-proxy.ts) uses shared-ts [`isTextContentType`](packages/shared-ts/src/content-type.ts), which does **not** match `urlencoded`.
- Swift [`isTextRequestContentType`](packages/swift-server/Sources/Server/APIRoutes.swift) explicitly includes `urlencoded` because "form bodies carry secrets".

Unify them. The Swift comment is the right request-side policy: form POSTs (OAuth token exchange, many provider APIs) carry secrets and must unmask. Shared-ts `isTextContentType` is also used for **response** scrub, so either:

- add `urlencoded` to the shared predicate (response-side form bodies are rare, but then both hops match), or
- keep a request-only variant and use it on both Node and Swift.

Related: #1996 (predicate triplication), #2818 (JPEG empty-CT path; left this drift out of that PR).

## SLICC version / commit

`b33887d1a` / PR #2818

## Pre-flight

- [x] I have stripped secrets, tokens, and PATs from logs and screenshots.

PR #2827, opened 2026-09-03
fix(swift-server): scrub fetch-proxy response bodies on the shared text predicate

### LINKED ISSUE #2822: bug: fetch-proxy response-scrub classifiers still differ between Node and Swift
## Summary

Node and Swift `/api/fetch-proxy` still use different predicates to decide whether an **upstream response body** is text (and therefore secret-scrubbed). #1996 unified Node onto shared-ts `isTextContentType`; Swift's response path never followed. XML / JS / SVG / HTML-as-`application/*` replies can leak real secrets on Sliccstart while Node scrubs them — and Swift will UTF-8-scrub some `charset=` binary types that Node leaves alone.

Spotted while auditing fetch-proxy parity for #2818. Not the JPEG request-body bug that PR fixes.

## Float (runtime)

- Standalone CLI / `npm run dev`
- Electron / Sliccstart (macOS app)

## Surface

- Shell / AlmostBashShell / supplemental commands
- Provider / model / API key handling

## Area

- `area/node-server`
- `area/swift`

## Reproduction

With a secret whose real value appears in the upstream body:

| Upstream `Content-Type` | Node (`isTextContentType`) | Swift (`makeStreamingProxyResponse`) |
| --- | --- | --- |
| `text/html` | scrub | scrub (`text/`) |
| `application/json` | scrub | scrub (`application/json` prefix) |
| `application/xml` | scrub (`xml`) | **no scrub** |
| `application/javascript` | scrub | **no scrub** |
| `image/svg+xml` | scrub (`svg` / `xml`) | **no scrub** |
| `text/event-stream` | scrub (`text/`) | scrub (`event-stream`) |
| `application/octet-stream; charset=utf-8` | **no scrub** | **scrub** (`charset=`) |

```javascript
// Node scrubs; Swift forwards the real secret.
await fetch(url); // upstream: Content-Type: application/xml
```

## Suspected root cause / location

- Node: [`streamUpstreamBody`](packages/node-server/src/routes/fetch-proxy.ts) → `isTextContentType(upstream.headers.get('content-type') ?? '')` from [`packages/shared-ts/src/content-type.ts`](packages/shared-ts/src/content-type.ts) (`text/`, json, xml, javascript, ecmascript, html, css, svg).
- Swift: [`makeStreamingProxyResponse`](packages/swift-server/Sources/Server/APIRoutes.swift) uses a local classifier:

```swift
let isText =
    contentType.hasPrefix("text/")
    || contentType.hasPrefix("application/json")
    || contentType.contains("charset=")
    || contentType.contains("event-stream")
```

That is the old Node response-side shape from #1996, plus `charset=` / `event-stream`. Swift request-side already has a closer copy (`isTextRequestContentType`); response-side does not use it.

Fix: drive both response paths from one shared predicate (or a documented Swift twin of `isTextContentType`). Drop the `charset=` catch-all — it turns some binary types into a UTF-8 String round-trip.

Related: #1996 (closed after Node unification), sibling request-side form drift, #2818 (left this out).

## SLICC version / commit

`b33887d1a` / PR #2818

## Pre-flight

trieloff:
Closes #2822.

## Problem

`/api/fetch-proxy` scrubs real secret values out of an **upstream response body** only when that body is UTF-8 text. #1996 unified node-server onto shared-ts `isTextContentType`; swift-server's response path kept a local classifier, so the two floats disagreed:

| Upstream `Content-Type` | node-server | swift-server (before) |
| --- | --- | --- |
| `text/html` | scrub | scrub |
| `application/json` | scrub | scrub |
| `application/xml` | scrub | **forwarded verbatim** |
| `application/javascript` | scrub | **forwarded verbatim** |
| `image/svg+xml` | scrub | **forwarded verbatim** |
| `text/event-stream` | scrub | scrub |
| `application/octet-stream; charset=utf-8` | no scrub | **scrubbed** (lossy `String` round-trip) |

So on Sliccstart an XML / JS / SVG reply handed the agent a real secret that the Node CLI masked, and a binary body that merely carried a `charset=` parameter was decoded and re-encoded for no reason.

## Fix

- New `packages/swift-server/Sources/Server/ContentType.swift` — the Swift twin of shared-ts `isTextContentType`, with the drift history documented in the doc comment.
- `makeStreamingProxyResponse` now calls it instead of re-deriving the list inline.
- Dropped the `charset=` catch-all (a parameter is not a type) and the `event-stream` special case (`text/event-stream` already matches the `text/` prefix, which is how Node matches it).

## Tests

The classification table is pinned on **both** sides, so the next drift fails a test instead of leaking a secret:

- `packages/swift-server/Tests/CrossImplementationTests.swift` → `testIsTextContentTypeMatchesPinnedTable`
- `packages/shared-ts/tests/content-type-parity.test.ts` (new)

Both tables include the `application/octet-stream; charset=utf-8` row as the regression guard for the dropped catch-all.

## Docs

`packages/swift-server/CLAUDE.md` and `packages/shared-ts/CLAUDE.md` now name `ContentType.swift` as the twin and say not to re-derive the list inline.

## Coordination with the parallel track (#2821)

#2821 fixes the **request**-side half of the same drift (Node not unmasking `application/x-www-form-urlencoded` bodies) and is in flight on `bb/close-node-swift-parity-gap-thr_uzkt2n9bgz`. The two changes were deliberately kept disjoint:

- This PR touches only the response hop (`makeStreamingProxyResponse`, ~line 1030 of `APIRoutes.swift`); #2821 touches the request hop (~line 500).
- This PR does **not** touch `shared-ts/src/content-type.ts`, `shared-ts/tests/content-type.test.ts`, `shared-ts/tests/cross-impl-vectors.test.ts`, or `node-server/src/routes/fetch-proxy.ts` — all of those belong to #2821.
- The Swift parity test is appended at the end of `CrossImplementationTests.swift`; #2821's goes near the top of the class.
- The TS parity table is a new file rather than an append to `cross-impl-vectors.test.ts`, which #2821 extends.

The one shared surface is `ContentType.swift`. This PR lands the **base** predicate there; #2821 has been asked (thread `thr_uzkt2n9bgz`) to layer its request variant on top as `isTextContentType(ct) || urlencoded` rather than spelling the list out a second time, which is also what its shared-ts half already does. That keeps one predicate per hop instead of re-triplicating what #1996 consolidated.

## Verification

`npm run verify` (8/8) · `check-touched-exemptions` · `npm run typecheck` · `npm run test` (19,255 passed) · `npm run test:coverage` · `npm run build` + chrome-extension build · `npm run lint:swift` (0 serious) · `npm run lint:swift:format` · `swift build` · `swift test` (609 passed, 0 failures) · `swift-coverage-check.sh packages/swift-server` (lines 66.47% / floor 65%).


trieloff:
Thanks for the review. Both non-blocking observations, addressed:

**1. Chunk-boundary caveat.** Confirmed out of scope and unchanged by this PR. The per-chunk transform in `ScrubbingAsyncStream` handles UTF-8 codepoint boundaries (it carries incomplete trailing bytes into the next chunk) but not *secret-value* boundaries, so a masked value split across two chunks still slips through. That is deliberate and matches node-server's `createScrubStream`, which has the same shape and the same documented limitation. Fixing it properly means buffering a sliding window of `max(secretLength) - 1` bytes on both floats — a separate change, and one that has to land on both at once or it just recreates the drift this PR closes.

**2. Confirming #2821 layers on the base rather than re-spelling the list.** Done — it now does. Its branch (`bb/close-node-swift-parity-gap-thr_uzkt2n9bgz`) carries a dedicated commit for exactly that, `refactor(swift-server): layer the request predicate on a base isTextContentType`, and its `ContentType.swift` now reads:

```swift
/// Whether a content type is safe to decode as UTF-8 text.
func isTextContentType(_ contentType: String) -> Bool { /* base list, byte-identical to this PR */ }

/// Request-side variant of `isTextContentType`, additionally matching
/// `application/x-www-form-urlencoded`.
func isTextRequestContentType(_ contentType: String) -> Bool { … }
```

So the predicate stays single-sourced per hop on both floats and #1996's consolidation is not undone. The base body is byte-identical to the one here, so whichever of the two PRs lands second rebases with a trivial resolve on that one file; everything else in the two changes is disjoint by construction (this PR is the response hop only, and touches none of #2821's TS or node-server files).

Issue #2271, opened 2026-08-21
feat(work-unit): per-cone workspace and memory for extra cones


trieloff:
## Addendum 2026-08-22 (PR #2286 — rebase + this before merge)

With the `agentic-memory` flag on, compaction-time extraction is off (`shouldExtractMemories: !flag`) and the curator archives only `session-cone`, so extra cones get **no memory at all**. Decision: **run the curator per cone** — archive `session-<folder>` and write to `workspaceFor(unit).memoryPath`. Don't fall back to extraction and don't document it as primary-only.

Also: use `WorkUnitManager.rootOf(id)` in `agent-bridge.ts` instead of re-walking ownership.

---
Part of #1666 (follow-up to #2262).

## Problem
Extra cones share `/workspace` and `/workspace/CLAUDE.md` with the primary cone (`workspaceFor` returns the same coordinates for every root). Two cones therefore see each other's files and append to one memory file.

## Scope
- Decide the layout for non-primary roots (proposal: `/cones/<folder>/workspace` + `/cones/<folder>/CLAUDE.md`, `/tmp` stays shared; the primary keeps `/workspace` for compatibility).
- `workspaceFor` in `packages/webapp/src/work-unit/descriptor.ts` becomes the single source; `ScoopContext.ensureDirectoryStructure`, skills seeding (`createDefaultSkills`) and `ConeMemoryStore` / `appendConeMemory` honour it.
- Sudo policy and `RestrictedFS` defaults for scoops spawned by an extra cone (`scoop_scoop` defaults `visiblePaths: ['/workspace/']`) follow the owning cone's root.
- Migration: existing extra cones created under #2262 keep working (they already live in `/workspace`; either migrate on first boot or treat "folder === cone" specially — document the choice).

## Acceptance
- [ ] Two cones cannot see each other's workspace files by default; `/shared` remains shared.
- [ ] Memory auto-extraction for cone B lands in cone B's `CLAUDE.md`.
- [ ] `docs/work-unit.md` limits section updated; `docs/mounts.md` if mount roots are affected.

Issue #2329, opened 2026-08-23
Follower can latch an empty model catalog and never show a model picker (#2310 wire path)


trieloff:
Part of #1666 (multiple cones). Fallout from #2310 / #2318, observed while building the two-instance e2e in #2313 (PR #2328). Lower priority: **observed on CI, not yet reproduced locally.**

## Symptom

A follower joins a leader's tray, mirrors the cone strip correctly, can select a cone — and never shows a model picker. `<slicc-composer-meta>` stays `display:none` carrying its first-paint placeholder `model="Preview"` for the rest of the session.

Observed in the `e2e` job of PR #2328 (run [32607141424](https://github.com/ai-ecoverse/slicc/actions/runs/32607141424)): the follower spec timed out waiting for the pill, with the locator resolving **26 × as hidden**.

## What the follower demonstrably did receive

Inferred from assertions that passed *before* the failing step, not from frame logs (there is no tray-sync frame-level debug log to enable — follower-side logging in `tray-follower-sync.ts` covers `hello`, user messages, aborts and close, not per-frame receipt):

| Frame | Arrived? | Evidence |
| --- | --- | --- |
| `hello` | yes | the sync came up; nothing logged a legacy peer |
| snapshot | yes | the thread rendered |
| `scoops.list` | **yes** | the follower's tab strip matched the leader's exactly (`[sliccy, reviewer, helper]`), and selecting cone B on the follower moved its active tab |
| `models.list` | unknown — **empty or non-matching** | the pill stayed hidden, which `createFollowerModelSurface.apply()` only does when there is no `state` or no catalog entry matching `state.activeModelId` |
| `model.state` | unknown | same gate; cannot be distinguished from the outside |

Timing relative to the join: the follower attached, the strip mirrored within seconds, cone B was selected, and the pill was still hidden 30 s later. The leader had already resolved its own model twice by then (`[model] Cone using model: fake-cone-primary (provider: local-llm)` appears twice in the leader console before the failure).

## Why this looks like a product bug, not test timing

The catalog is pushed to a follower **exactly once**, when it attaches:

- `tray-leader-sync.ts` `addFollower()` → `broadcast.sendModelCatalogToFollower(bootstrapId)`
- `tray-leader/broadcast.ts` `sendModelCatalogToFollower()` → `buildModelCatalog()` → `follower.sync.send({ type: 'models.list', models })`, then `sendModelStateToFollower()`

[...]

PR #2838, opened 2026-09-04
refactor(webapp): type the shell's selection surface on WorkUnitSummary (#2382 PR D2a)

trieloff:
## What

Re-types the WC shell's **selection surface** from `RegisteredScoop` (a
leader-only record) to `WorkUnitSummary` (the client protocol's unit).
`WcShellBoot.selectScoop` / `getSelected` and every helper downstream of them
now speak summaries.

No mount changes, no rendering changes. This is the blocker D1's PR body
called out: `attachWcChat` cannot hand a follower's unit to a shell whose
selection type only a leader can produce.

Part of #2382.

## Why this is a PR of its own

D2b (`attachWcChat` / `attachWcWorkbench` / `mountWcShell` and deleting the
three mounts) is a mount change. Doing the re-type in the same PR would mix a
mechanical field rename across 11 files with the one change a reviewer
actually has to think about. Split so each is readable.

## What moved

- `unit.jid` -> `unit.id`, `parentJid === null` -> `isRootSummary(unit)`, and
  the switcher label reads the carried `role` instead of re-deriving it from
  the ownership edge.
- `wc-unit-context.ts` is typed on summaries throughout (`orderForSwitcher`,
  `rootForSelection`, `rootForConeFolder`, `chatSessionIdFor`,
  `isPrimaryRootSummary`), and its callers read the roster from the client
  (`getUnits()`), not `OffscreenClient.getScoops()`.
- `wc-live-freezer.ts`, `wc-nav.ts`, `wc-cone-actions.ts`,
  `wc-sprinkles.ts`, `wc-live-controller.ts`, `wc-live-composer.ts` and
  `wc-live-thinking-hydration.ts` follow the same surface.

## Record-only fields are read at the LEAF

Two leader-only fields have no summary equivalent, and neither re-widens the
selection type:

1. **The reasoning level** behind the thinking pill. `applyThreadContext`
   takes an optional `getRecord(id)` and the leader passes its own roster
   lookup. A follower passes nothing and renders correctly — it has no
   thinking pill.
2. **The session archive's folder work** in the freezer rail. The archive is a
   record operation (`folder`, plus `jid` so a curator pass is parented to the
   cone it curates, #2271), so `recordFor(client, unit)` resolves the cone
   summary back to a record with one lookup at the point the VFS work needs
   it. A follower has no freezer.

The label those archives carry goes through `archiveConeTarget(root)`, which
resolves the record's role once instead of at each archive call site.

## Tests

Every mounted wc test now selects summaries. One new test pins the seam this
PR introduces:

- `wc-live.test.ts` — "reads the selected unit's thinking level from its
  RECORD at the leaf": a summary carries no reasoning level, so the mount has
  to hand `applyThreadContext` a record lookup. **Mutation-tested**: dropping
  the `getRecord` argument makes it fail (the pill goes blank on every
  selection).
- `wc-thread-model-pill.test.ts` gained a `byId` record source, so the
  owning-cone model rule and the leaf thinking read are asserted together.
- `wc-queue-reconcile.test.ts`'s mounted fixture selects summaries — it was
  passing records through `as never`, which would have silently selected
  `undefined`.

## Verification

Run locally on this branch (rebased onto `origin/main` right before push):

- `npm run lint` then `npm run lint:ci` — clean. Two errors surfaced and were
  paid down rather than exempted: `wireFreezerRail` went over the 150-line cap
  (extracted `archiveConeTarget` + `recordFor`), and one import went unused.
- `npm run test` — **19461 passed**, 0 failed, **0 unhandled errors**.
- `npm run test:coverage:webapp` — lines 85.07 / statements 83.09 / functions
  83.19 / branches 74.76, all above the floors (84 / 82 / 82 / 74).
- Both builds; `check-first-load-size.mjs` **+0.0 kB** on both graphs;
  `check-touched-exemptions.mjs` clean; `npm run deadcode` clean.
- **Fake-LLM e2e** on isolated ports (`SLICC_E2E_WRANGLER_PORT=8827
  SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8828 SLICC_E2E_BRIDGE_PORT=5745
  SLICC_E2E_CDP_PORT=9262 SLICC_E2E_FAKE_LLM_PORT=5805`): the five cone specs,
  **5 passed**.

## Rendering is unchanged

The evidence is the five cone e2e specs above: they drive a real leader + a
real follower over the fake LLM and assert the strip, the selection, the
read-only scoop transcript and the model pill on both sides. All five pass
unchanged.

Frames were also captured from the Playwright traces of
`multiple-cones-follower.test.ts` ("follower mirrors the cone strip and drives
one cone's model") run once on `main` and once on this branch, at the same
step: leader and follower render identically, timestamps aside. They are not
attached — `gh image` has been retired and `--attach` is behind an interactive
device flow — so they stay on the machine that produced them at
`/tmp/claude/thr_sqvjwe2vn2/shots-d2a/{before,after}-{leader,follower}.jpeg`.

While comparing frames I found a **pre-existing** transient worth recording:
after "New cone", the leader's thread can show only the assistant's greeting
for ~5 s (the user's initial prompt and the date divider appear late). It
reproduces on `main` as well as on this branch — the first replay for a
brand-new cone goes unanswered and the adapter's 5-second recovery paints the
full transcript. Not caused by this PR; not fixed here.

## What is NOT in this PR

D2b: `attachWcChat(boot, client, host: WcChatHost)` /
`attachWcWorkbench(boot, offscreenClient)`, `mountWcShell`, and deleting
`mountWcUiLive` / `mountWcUiFollower` / `mountWcUiExtension`.


trieloff:
Round 1 applied in `7921bd8ba`. All six, each with a test that fails when the fix is reverted.

**1 (P1, regression) — `defaultRootOf` took the first root in roster order.** Now `orderRoots(units.filter(isRootSummary))` then primary-or-first, so it reads roots exactly as the strip does. `orderRoots` is exported from `presentation.ts` rather than re-implemented, so the two cannot drift again.

Two cases in `wc-unit-context.test.ts`: two extra cones with the newer one FIRST in the array and the older `addedAt` (the existing "oldest root" case has a single extra root and passed either way), asserting `defaultRootOf(roster)?.id === orderForSwitcher(roster)[0]?.id`; and a half-timestamped roster, where the all-or-nothing rule keeps transport order and the two reads still agree.

**2 (P2) — `onScoopCreated` looked the payload up in the roster.** Back to projecting the payload (`recordToWorkUnitSummary(scoop, {fill, phase, status})`), and the false comment is gone — replaced with why the lookup cannot work: the base handler runs before `emitList`, and the adapter has no roster of its own, it reads `getScoops()`. Test drives it with an EMPTY `getScoops()`.

**3 (P2) — `pendingSelect` was cleared before the lookup.** One helper now: `unitFor(record)` prefers the client's roster and falls back to the leader's own projection of the record it is already holding, so neither the landed cone nor the drop survivor can be lost to a roster that is a tick behind. Two tests, via a `holdUnits()` harness hook that pins the client roster to an earlier instant: the new cone still gets selected, and the drop moves onto a survivor the client roster has not seen yet (the "make a second cone, drop the first" ordering — with the survivor already in the held roster the old code passes, so the test had to be built around the young cone).

**4 (P2, D2b contract) —**
- `rootForSelection` returns `undefined` for a child whose `parentId` is `undefined`. Deliberately narrow: a child whose NAMED parent is missing from the roster keeps the historical default-root fallback, because that is a render race and not an absent edge — narrowing it further would have changed leader behaviour. Callers already handle `undefined` (`if (!root) return false` in the model write, `recordFor(...)` in the freezer, `?? unit.id` in the queue reconcile).
- The thinking pill now writes only when a record answered or a locked effort level is set; otherwise it keeps its value. `metaThinkingForScoop(undefined)` is `off`, and writing that was #2329's latch in a second place.
- `docs/work-unit-client.md` has the contract table (`parentId`, `addedAt`, the record behind `getRecord?`) with what absent means and what the shell does, so D2b inherits it.
- `wc-live.test.ts` comment fixed: it wrote `off`, it did not go blank.

**5 (P2) — the workbench workspace.** Extracted to `workspaceForSelection({client, clearSelection, selectedId})` so it is testable at all, and a selection the roster no longer knows now clears itself instead of resolving to the first root — the next roster event re-selects via `ensureSelection`, and the caller gets the same default workspace that selection lands on.

**6 (P3) — comments.** `orderForSwitcher`'s "project, order, project back" is gone (both sides order the same type now); the doc's `chatSessionIdFor` claim is corrected — it stays on `work-unit/record.ts` and keys off `folder`, which both shapes carry — and the roster is named as `currentUnits()` via the shell's `getUnits()`, not `subscribeList`.

**Verification** (rebased onto `origin/main` right before push; the previous head `8d1bc60d9` was all-green on CI):

- `npm run lint` + `lint:ci` clean; `typecheck` clean (9 projects).
- `npm run test` — **19471 passed**, 0 failed, 0 unhandled errors.
- Coverage: lines 85.07 / statements 83.09 / functions 83.17 / branches 74.77 (floors 84 / 82 / 82 / 74).
- Both builds; first-load **+0.0 kB** on both graphs; exemptions + deadcode clean.
- The five cone e2e specs on isolated ports: **5 passed**.

Issue #2840, opened 2026-09-04
New cone: the leader thread shows only the greeting for ~5 s before the initial prompt paints


trieloff:
## What happens

Create a cone from the freezer rail's action row ("New cone", with an initial
prompt). The new cone is selected and its thread paints — but for about five
seconds it shows only the assistant's greeting. The user's own initial prompt
and the date divider above it are missing, then appear all at once.

## Repro

The leader half of `packages/webapp/tests/e2e/multiple-cones-follower.test.ts`
("follower mirrors the cone strip and drives one cone's model") walks straight
through it. Run it with `--trace on` and step the screencast frames:

```bash
SLICC_E2E_WRANGLER_PORT=8827 SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8828 \
SLICC_E2E_BRIDGE_PORT=5745 SLICC_E2E_CDP_PORT=9262 SLICC_E2E_FAKE_LLM_PORT=5805 \
npx playwright test --config packages/webapp/tests/e2e/playwright.config.ts \
  multiple-cones-follower --trace on --output /tmp/cone-trace
```

Observed on one run: the Create click lands at t=6.69 s, the thread shows the
greeting alone from t=6.9 s through t=11.3 s, and the full transcript (date
divider + "review the docs" + greeting) is there by t=11.84 s. 6.75 + 5.0 =
11.75.

**This reproduces on `main`.** It was found while diffing before/after frames
for #2838 (#2382 PR D2a) and is not caused by that PR — a first `main` run
happened to take the fast path, a second `main` run on an idle machine showed
the same five-second window. It is timing-sensitive: when the kernel answers
the first replay promptly the transcript is correct within ~300 ms.

## Hypothesis

The five seconds is the local adapter's unanswered-snapshot recovery
(`LocalWorkUnitClient`, #2382 PR B): `subscribe` deliberately does not re-ask
while a `snapshot(id)` is in flight, so the single request is load-bearing and
an unanswered one is re-asked once on a 5 s timeout. The retry is what paints
the transcript here.

The request is very likely made before the kernel has the brand-new unit —
`selectScoop` fires as soon as the cone's real record lands in the roster —
so the kernel has nothing to replay and drops it. The transcript that
eventually arrives is correct, which is consistent with the retry succeeding.

**Suggested fix**: the recovery should not have to wait out the full timeout
when there is a better signal. When the roster later NAMES a unit whose
[...]

Issue #2839, opened 2026-09-04
docs: demo-recording skill still documents the retired `gh image`


trieloff:
`.agents/skills/demo-recording/SKILL.md` § *Embedding in GitHub PRs* tells an
agent to upload a recording with `gh image` from the `ai-aligned-gh` wrapper.
That subcommand has been retired:

```
$ gh image --help
Error: 'gh image' has been retired.

  gh 2.99.0 or newer attaches images and videos natively
  via the repeatable --attach flag. You have gh 2.97.0.

    gh issue create --attach ./repro.png
    gh issue comment 123 --attach ./repro.png
    gh pr create --attach ./before.png
    gh pr comment 456 --attach ./result.mp4
```

So the documented path fails for every agent that follows the skill, and the
`gh image --markdown` example in it produces nothing embeddable.

**What to change**

- Replace the `gh image` examples with `--attach` (repeatable), and state the
  gh >= 2.99 requirement.
- Note that `--attach` goes through the wrapper's write path, so it can
  prompt the as-a-bot device flow — an agent that cannot complete an
  interactive authorization has no upload path and should say so rather than
  block. (Hit on #2838: the frames were captured but could not be attached.)
- Drop the claim that "plain `gh` cannot attach media" — it can now.

Anything else in the skill that names `gh image` (the front-matter
description, § *Full Example*) needs the same pass.

Review pattern 12 (agent skill freshness).

PR #2844, opened 2026-09-04
refactor(webapp): one mount path for the WC shell (#2382 PR D2b)

trieloff:
## What

**One mount path for the WC shell.** Every float — standalone, Electron
overlay, hosted leader, extension side panel and popout, browser follower,
Cherry embed — now builds the same frame, wires the same chat surface onto
it, and differs only in the transport it hands over and whether it has a
kernel to open a workbench on.

```text
mountWcShell(app, log, { floatKind, connect })
├── buildWcShellFrame        the DOM, unchanged (was `mountWcShell`, renamed)
├── connect(boot) ─────────► { client, host, workbench? }   ← the float's half
├── attachWcChat             strip · transcript · queue · composer · stop ·
│                            selection · model pill    (every float)
└── workbench?(boot, chat)   VFS · terminal · monitor · sprinkles ·
                             permissions · export · sudo · stats  (leader only)
```

Closes out #2382: PR A put send/stop/model on the protocol, B the transcript
and selection, C the model pill, D1 the leader-capable float's follower role,
D2a the selection surface — and this one removes the parallel UI state
machine the RFC set out to delete.

## The three mounts are gone

`mountWcUiLive`, `mountWcUiExtension` and `mountWcUiFollower` are deleted.
What is left are three CONNECTORS — `bootLeaderFloat`, `bootExtensionFloat`,
`bootFollowerFloat` — each a prelude, a transport and a `mountWcShell` call.
None of them wires a controller, a strip, a composer or a selection any more;
the follower alone shed ~300 lines of parallel chat wiring, including its own
`WcChatController`, its own strip publisher, its own `watchUnit`, its own
selection chrome and its own submit/stop listeners.

## `WcChatHost` — the seam

Four verbs a client protocol cannot carry: `sendSprinkleLick`,
`sendToolUiAction`, `deleteQueuedMessage`, `emitAgentError`. Plus the
transport facts a float answers differently — its agent event stream, whether
it has records (`getRecord`), whether it narrows what a send may address
(`addressableUnitId`), whether it speaks replies, owns the model pill, or
renders tool-UI cards read-only.

A follower's host states its refusals at each member instead of scattering
null checks through code both floats run. Its `deleteQueuedMessage`
**rejects** rather than resolving: the queue is the leader's, the tray has no
verb for it, and a silent success would leave the pile out of step with the
backend that still holds the entry.

## Judgement calls worth a reviewer's eye

1. **`connect` is a callback, not a value.** The requested shape was
   `mountWcShell(app, log, { client, workbench?, floatKind })`. A leader's
   `OffscreenClient` is constructed FROM this shell's callback bag
   (`createWcLiveCallbacks(boot.wiring)`), so the frame must exist before the
   client does — a client-as-value would have forced the leader to keep its
   own mount. `connect(boot)` is the smallest change that keeps ONE path.
2. **`mountWcShell` was taken.** It was the DOM frame builder in
   `wc-shell.ts`; it is now `buildWcShellFrame`, freeing the name for the
   mount path. Mechanical, compiler-checked, 55 call sites.
3. **`currentUnits()` joined the protocol.** The strip repaints on paths that
   cannot await — a tab click re-orders the same units around the new
   selection, and a leader's `awaitingInput` moves with no transport event at
   all. Both adapters already answered it; this writes it down instead of
   duck-typing it. Conformance case added on both.
4. **The guest carve-out stays explicit.** A biscotto seat is pinned to one
   thread and never sent `scoops.list`, so there is no summary to select
   (the D2a contract table predicted exactly this). `boot.watchUnit(id)`
   renders that transcript with no chrome change and the composer stays
   writable — the control on a guest's message is the leader-side review
   gate.
5. **Page theming moved out of the shared controller.** Custom-theme
   overrides and the sprinkle theme broadcast are the leader's now: applying
   them under a Cherry host would fight the theme that host pushed. Same for
   dictation cues and spoken replies, gated by `speaksReplies` — a follower
   arms PTT but has no reply voice, and turning one on would be a behaviour
   change, not a mount change.
6. **The transcript watcher now dedups.** Re-pointing at the unit already
   being watched would tear down a live subscription and take a fresh SEED in
   its place. A follower re-asserts its shown unit on every 5 s roster push,
   so this is load-bearing, not an optimisation — mutation-tested.

## Tests

- `wc-chat.test.ts` (new) drives `attachWcChat` with a fake client and a
  hand-written host, because the point of the module is that it does not know
  which float it is on: strip publish, tab-click selection (and re-click
  no-op), submit, stop-only-while-processing, staged attachments, and a send
  addressed at what the HOST allows rather than at the selection.
- The follower gained two: a repeated roster push does not re-render the
  transcript, and a guest thread the roster never describes still renders
  with a writable composer.
- `conformance.ts` pins `currentUnits()` on both adapters.
- `leader-chat-host.ts` (test helper) attaches a leader shell in the same
  order the mount does, so a test cannot assert against a seam the shell does
  not use.
- **Mutation-tested**: dropping the watcher dedup fails the roster-push test;
  ignoring `addressableUnitId` fails the send test; skipping `watchUnit` for
  a guest fails the guest test.

## Verification

Run locally on this branch (rebased onto `origin/main` = `7c17a7c5d` right
before push; `git merge-tree` prints a bare tree):

- `npm run lint` then `npm run lint:ci` — clean.
- `npm run test` — **19489 passed**, 0 failed, 0 unhandled errors.
- `npm run test:coverage:webapp` — lines 85.03 / statements 83.06 / functions
  83.13 / branches 74.77, all above the floors (84 / 82 / 82 / 74).
- Both builds; `check-first-load-size.mjs` **−0.0 kB page / +0.0 kB worker** —
  a mount collapse should not move the eager graph, and it does not;
  `check-touched-exemptions.mjs` clean; `npm run deadcode` clean.
- **Fake-LLM e2e** on isolated ports (`SLICC_E2E_WRANGLER_PORT=8827
  SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8828 SLICC_E2E_BRIDGE_PORT=5745
  SLICC_E2E_CDP_PORT=9262 SLICC_E2E_FAKE_LLM_PORT=5805`): the five cone
  specs, **5 passed** — leader and follower, including the read-only scoop
  and the follower-driven model pick.
- The **whole** e2e suite also ran (26 specs): 25 passed, 1 skipped, and
  `transcript-export.test.ts` failed — which **also fails on `main`**
  at `0be5ada8b` with the same `TranscriptExportError: schema-invalid`.
  Pre-existing, not from this PR.

## Rendering is unchanged

Frames were captured from the Playwright traces of
`multiple-cones-follower.test.ts` on `main` and on this branch and compared at
matched steps: the follower pair is identical, and the leader pair is
identical at the settled end state. They are not attached — `gh image` is
retired and `--attach` is behind an interactive device flow (#2839) — so they
stay at `/tmp/claude/thr_sqvjwe2vn2/shots-d2b/{before,after}-{leader,follower}.jpeg`
on the machine that produced them.

While comparing I hit the transient from #2840 again, and one more worth
naming: during the phase where a follower clicks through tabs, the leader's
chrome and its transcript are momentarily out of phase (the tab flips, the
replay lands a beat later). It reproduces on `main` — sampled there as
"sliccy tab + reviewer transcript", and here as "reviewer tab + empty
transcript" — so it is the same pre-existing behaviour caught at a different
instant, not a regression.


trieloff:
Round 1 applied in `a4d09b2e2` (rebased onto `origin/main` = `17d5b3bec`; `git merge-tree` prints a bare tree). All five, each with a test that fails when the fix is reverted.

**1 (P1, regression) — every roster push re-selected the shown unit.** You are right about the mechanism and it is worse than a wasted round trip: `handleScoopSelection` has no same-jid short-circuit, so each 5 s `scoops.list` re-assert asked for a snapshot the leader answered with a full replay, and the watcher re-rendered it — dips disposed and rehydrated, in-thread form state wiped, every five seconds. `showUnit` now selects only when `boot.getSelected()?.id !== unit.id`, the same guard the tab-click path in `attachWcChat` already had. A re-assert of the SAME unit still runs `setComposerState`, because the connection can move even when the unit does not (that is what `applyFollowerSelectionChrome` used to cover on every frame).

The test was a false green exactly as you describe — no `currentSync`, so `snapshot()` rejected into the adapter's `.catch`. It now installs `currentSync.selectScoop`, asserts it fires **once** on the first roster of a session (a snapshot-only join still gets its unit, because `getSelected()` is null then) and **never** on a re-assert. One correction to my own first attempt at it: I initially also asserted no `loadMessages` after a leader-PUSHED snapshot, which is wrong — a pushed snapshot is a real render. The assertion is scoped to roster frames.

**2 (P2) — the conformance case compared ids.** Now `expect(currentUnits()).toEqual(await list())` plus field-level `toMatchObject` on a cone and a scoop (`role`, `parentId`, `model`, `state`), and two synchronous-reflection checks with no await in between: `emitStatus` → `state === 'working'`, and a roster change → new id list. `{ id }` stubs and a remembered id list both fail it now. Runs on both adapters (verified in the verbose reporter, not assumed).

**3 (P2) — the biscotto carve-out.** Added: the strip stays empty; `scoops.select` never reaches the wire; Stop is a no-op before `unitConfirmedThisSession`; and the model pill / thread context keep the FRAME's own seeds (`model="Preview"`, `context="cone"`, `accent="var(--waffle)"` from `buildWcShellFrame`) rather than anything this page's catalog or `applyThreadContext` would have written. I had to correct my first phrasing — asserting `null` was wrong, because the frame seeds those attributes; "unchanged from the frame's seed" is the honest claim and the stronger one.

**4 (P3) — double floatbar subscription.** The connector's `installFloatbarStatus` is gone; only `mountWcShell`'s remains.

**5 (P3) — stale names.** `attachWcClient` → `attachWcWorkbench` and `AttachWcClientOptions` → `AttachWcWorkbenchOptions` everywhere (comments in `wc-sprinkles.ts`, `wc-onboarding.ts`, `layout-apply-registry.ts`, `wc-permissions.ts`, `wc-shell.ts`, `wc-live-callbacks.ts` included). The `wc-live.ts` JSDoc you flagged turned out to be dangling — it documented `prepareWcShell` before the collapse moved code around, and was sitting above `reconcileQueueForSwitch` documenting nothing; it is now `prepareWcShell`'s own doc and says what actually happens (frame first, transport second, on every float). `docs/work-unit.md` row 5f is marked `done (this PR)`.

**Verification** (all after the rebase):

- `npm run lint` + `lint:ci` clean; `typecheck` clean (9 projects).
- `npm run test` — **19516 passed**, 0 failed, 0 unhandled errors.
- Coverage: lines 85.03 / statements 83.05 / functions 83.09 / branches 74.74 (floors 84 / 82 / 82 / 74).
- Both builds; first-load **−0.0 kB page / +0.0 kB worker**; exemptions + deadcode clean.
- The five cone e2e specs on isolated ports: **5 passed**.

Mutation checks: forcing `showUnit` to select unconditionally fails the roster-push test; the guest and conformance additions were written against the current code and re-run after each edit.

PR #2849, opened 2026-09-04
fix(transcript): stop export redaction from corrupting schema-controlled fields

### LINKED ISSUE #2845: transcript-export e2e fails locally with TranscriptExportError: schema-invalid (green on CI)
## What fails

```
packages/webapp/tests/e2e/transcript-export.test.ts:170:3
› transcript export — local ZIP download
› exports ZIP: cone + scoop conversations, binary unchanged, credential redacted
```

The "Export transcript" avatar-menu action never produces a download, so the
test times out at step 4:

```
TimeoutError: page.waitForEvent: Timeout 60000ms exceeded while waiting for event "download"
  226 |
  227 |     // ── 4. Trigger the "Export transcript" UI action ──────────────
> 228 |     const downloadPromise = page.waitForEvent('download', { timeout: 60_000 });
```

The browser console says why the download never came:

```
error [main] Transcript export failed TranscriptExportError: schema-invalid
    at ar (…/assets/wc-live-*.js)
    at lr.export (…)
```

So `DefaultTranscriptExportService.export()` built a document that
`validateTranscriptDocumentV1` rejected, and `wc-live`'s `onExportTranscript`
logged it and swallowed it — which is the intended UI behaviour for a failed
export, and why the test can only observe a missing download.

## It is not a regression

Found while verifying #2382 PR D2b. It reproduces on **`main` at `0be5ada8b`**
with the same error, from a clean `git switch --detach 0be5ada8b` +
`npm run build -w @slicc/webapp`. Both my branch and `main` fail identically.

Meanwhile the CI `e2e` job is green on `0be5ada8b` **and** on `7c17a7c5d`
(current `main` at the time of writing), so this is local-harness-specific,
not "e2e is broken".

## Repro

```bash
SLICC_E2E_WRANGLER_PORT=8827 SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8828 \
SLICC_E2E_BRIDGE_PORT=5745 SLICC_E2E_CDP_PORT=9262 SLICC_E2E_FAKE_LLM_PORT=5805 \
npx playwright test --config packages/webapp/tests/e2e/playwright.config.ts \
  transcript-export
```

Reproduced 3/3 times (twice on the branch, once on `main`).

## What differs from CI, and what is already ruled out

- **Isolated ports.** The run above overrides all five `SLICC_E2E_*` ports so
  it cannot collide with a live local stack; CI uses the defaults. Nothing in
  the export path reads a port, so this is unlikely to be the cause — recorded
  because it is the most visible difference.
- **Build freshness / scope.** Locally only `npm run build -w @slicc/webapp`
  ran; the CI `e2e` job additionally builds `@slicc/shared-ts` and
  `@slicc/node-server`. `validateTranscriptDocumentV1` lives in
  **`packages/shared-ts/src/transcript-export.ts`**, so a stale or
  differently-built `shared-ts` against a freshly built webapp is the most
  promising lead: the validator and the document builder could be out of step
  locally in a way CI never sees.
- **Browser profile: ruled out.** `cdp-browser.ts` uses `chromium.launch()`,
  a fresh temp profile per launch, and nothing was left listening on the CDP
  port between runs — so accumulated IndexedDB/OPFS state from earlier runs is
  NOT the explanation, even though `reuseExistingServer: !process.env['CI']`
  would allow reuse in principle.
- **Machine speed.** The spec's own comments call out CI-runner slowness for
  its earlier steps; a local machine is faster, so a race that only shows up
  when the turn completes QUICKLY (the scoop's conversation not yet flushed
  when the export walks sessions) would fail locally and pass on CI. This is
  the other plausible lead.

## The first thing to fix, either way

`assertValid` throws away the reason:

```ts
function assertValid(document: TranscriptDocumentV1): void {
  const validation = validateTranscriptDocumentV1(document);
  if (!validation.ok) {
    throw new TranscriptExportError('schema-invalid');   // `validation.error` dropped
  }
}
```

`TranscriptValidationResult` carries a specific message (`session.id must be a
string`, `privacy.reasoningExcluded must be true`, …) and none of it reaches
the log, so a failed export is undiagnosable from the console alone. Carrying
`validation.error` into the thrown error (or at least logging it) would turn
this issue from "something in the document is wrong" into a one-line answer,
and would pay for itself the first time a user reports a failed export.

## Why it matters

Export failing is silent by design in the UI — the user clicks
"Export transcript", nothing downloads, and only the console says why. If the
cause turns out to be a real document-shape bug rather than a harness
artefact, it is the kind that only shows up for the sessions people most want
to export.

Labels: `bug`, `area/webapp`.

trieloff:
Closes #2845

## Summary

Transcript export corrupted its own document and then failed validation. Before: a user with a
short secret in their secret store clicked "Export transcript" and nothing happened — no file, no
message, only a bare `TranscriptExportError: schema-invalid` in the console. Now: the export
succeeds, and if one ever does fail the console names the offending field and the UI says so.

**Which of the two outcomes in the issue this is: outcome 2 — a real validator/builder mismatch
that CI is masking.** Not a build-freshness artefact; the issue's strongest lead is disproven
under "What was ruled out".

## Why

The issue's first ask was to stop guessing and capture the reason. Carrying `validation.error`
into the thrown error turns the console line into the whole answer:

```
error [main] Transcript export failed
  TranscriptExportError: schema-invalid: session.state must be "active" or "frozen"
```

**Root cause.** Export known-secret redaction (`SecretsPipeline.redactForExport`) is a plain
`replaceAll` of each stored secret's value across every string leaf of the transcript document.
Nothing kept it away from the fields the schema *constrains*.

Secrets shorter than `MIN_MASKABLE_SECRET_LENGTH` are deliberately still substituted on the
export path — three tests in `secrets-pipeline.test.ts` pin that, since export must not leak even
a short secret. So a one-character secret rewrites unrelated text. Probed against the real local
secret store on the reporting machine, through the same `/api/secrets/redact-export` endpoint the
webapp uses:

```
POST /api/secrets/redact-export  {"texts":["active","frozen","complete", ...]}
-> {"texts":["acti⟦REDACTED:known-secret:k4⟧e","frozen","complete", ...]}
```

`session.state` becomes `"acti⟦REDACTED:known-secret:k4⟧e"`, `validateTranscriptDocumentV1`
rejects it, `assertValid` throws, and `onExportTranscript` logs and swallows — so the only
user-visible symptom is a download that never arrives.

This is not confined to enums. The same walk covers ISO timestamps, SHA-256 hexes, bundle paths
and the ids the relational checks join on, so any secret value occurring inside one of them fails
the entire export.

**Repro** (needs a secret in the local store whose value occurs in a schema field — a
one-character secret guarantees it):

```
SLICC_E2E_WRANGLER_PORT=8827 SLICC_E2E_WRANGLER_SUPERVISOR_PORT=8828 \
SLICC_E2E_BRIDGE_PORT=5745 SLICC_E2E_CDP_PORT=9262 SLICC_E2E_FAKE_LLM_PORT=5805 \
npx playwright test --config packages/webapp/tests/e2e/playwright.config.ts transcript-export
```

Expected: ZIP downloads. Actual (before this PR): times out waiting for `download`, console shows
`schema-invalid`.

- `isStructuralTranscriptPointer()` lives in `packages/shared-ts/src/transcript-export.ts`, next
  to the validator that constrains the same fields — deliberately co-located so the coupling is
  visible: a field that gains a constraint in the validator needs an entry there too. The
  redactor filters those leaves out of its walk.
- **Redaction stays fail-closed.** Every content-bearing field is deliberately absent from the
  set and is still fully redacted: `session.title`, conversation `name`/`folder`, message text,
  message `error`, tool-call `input`, `attachments[].originalName`, and the text attachments
  themselves. A test asserts exactly that.
- **Alternative rejected:** dropping short secrets from `redactForExport` (aligning it with the
  masking path, which already refuses them for this exact collision reason). That reverses a
  deliberate, tested security decision — three tests pin that export redacts short secrets — so
  it is not mine to make unilaterally. Skipping fields that cannot contain a secret achieves the
  fix without weakening the guarantee.
- `TranscriptExportError` gains an optional `detail` folded into `.message` only. The wire code is
  unchanged.

## Cross-cutting impact

- **Other callers of the changed contract:** every consumer of `TranscriptExportError` reads
  `.code`, never `.message` — checked `cherry-host-transport.ts` (`exportErrorCodeFromRejection`),
  `tray-follower-sync.ts` (`handleExportError`), `scoops/tray-leader/transcript-export.ts`, and
  `session-command.ts` (which prints `err.code`). Nothing parses the message, so widening it is
  safe. The `transcript.export.error` wire envelope still carries only the canonical code.
- **Floats:** the redaction change is in `packages/webapp/src/transcript/`, shared by every float
  that exports (CLI, extension, Electron, cloud leader, and the tray follower path, which goes
  through the same service). `redactForExport` itself is untouched, so the extension service
  worker and node-server REST paths behave identically.
- **Security:** narrows *what* is redacted, never *whether*. Only machine-generated fields are
  skipped; the change cannot cause a secret to be emitted from a content field. No secret value
  appears in the new `detail` — it carries the validator's own field-name message.
- **Bundle size:** `npm run bundle-size` → First-load OK. The UI notice rides the already-lazy
  `wc-transcript-export.js` chunk, so nothing is added to the boot bundle.
- **Persisted state:** none. No schema, IndexedDB, or storage-key change.

## What was ruled out

- **Build freshness (the issue's strongest lead): disproven.** `packages/webapp/vite.config.ts`
  aliases `@slicc/shared-ts` to `packages/shared-ts/src/index.ts`, so the browser bundle always
  compiles the validator from source. A stale `shared-ts/dist/` cannot desynchronise the builder
  from the validator in the webapp — only node-server consumes the built `dist/`.
- **CI hiding a retry: disproven.** The `e2e` job log for `0be5ada8b` shows the spec passing on
  its first attempt (`✓ 27 … (28.6s)`, `26 passed`), no flaky/retry marker.
- **Machine speed / scoop-flush race: not the cause.** The failure is deterministic and the
  validator names a field, not a missing conversation.

## Test plan

- [x] `npm run verify` — clean (the `suppressions/unused` warnings on `packages/cherry/**`,
      `chrome-extension/**`, `cloudflare-worker/**` reproduce on untouched files locally: the
      GritQL `lint/plugin` is not loaded in this local invocation. Unrelated to this PR.)
- [x] `npm run typecheck`
- [x] `npm run test` — 19560 passing, 0 failures
- [x] `npm run test:coverage` — no floor lowered
- [x] `npm run build`
- [x] `npm run bundle-size` — First-load OK
- [x] **New behavior covered by unit tests** —
      `npx vitest run packages/webapp/tests/transcript/redact.test.ts packages/shared-ts/tests/transcript-export.test.ts packages/webapp/tests/ui/wc/wc-transcript-export.test.ts`
- [x] E2E A/B on the reporting machine, both arms on a freshly started harness:

| arm | result |
|---|---|
| fix reverted | `✘ … TranscriptExportError: schema-invalid: session.state must be "active" or "frozen"` |
| fix applied | `✓ transcript-export.test.ts:170 … (4.9s)` |

The three new tests in `packages/webapp/tests/transcript/redact.test.ts` pin redactor output
against `validateTranscriptDocumentV1` using a redactor that reproduces `redactForExport`'s
substring behaviour for a degenerate secret. All three were confirmed to **fail** with the fix
reverted, so this regression is now caught in the fast suite rather than only in e2e.

**Pre-existing failures:** none.

**No screencast:** the second commit adds a dialog that only appears on a failed export, which no
longer reproduces once the first commit lands. Happy to record one against an artificially failed
export if a reviewer wants it.

## Documentation

- [x] `docs/` — `docs/transcript-export.md` gains a "Schema-controlled fields are never redacted"
      subsection and an updated `schema-invalid` row in the error table.
- [ ] `README.md` — no user-facing behavior change beyond the bug going away.

## Risk & rollback

- **Blast radius:** transcript export only, all floats. If the pointer set is wrong in the
  *conservative* direction (a content field listed by mistake) a secret could survive into that
  field — the set was written field-by-field against the schema and is asserted by a test that
  every listed pointer is one the validator constrains.
- **Rollback:** revert cleanly, both commits. No persisted-state migration.
- **CI cannot catch this class:** CI's secret store is empty, so no CI job can exercise the
  interaction between real stored secrets and the export document. The unit tests substitute a
  fake degenerate redactor precisely to close that gap.

## Note for whoever debugs the e2e next

Unrelated to this bug, but it cost a debugging cycle: the Playwright `webServer` entries use
`reuseExistingServer: !CI`, and a reused `wrangler dev` keeps serving the **asset snapshot from
its own start-up**. Rebuilding `dist/ui` between local runs does not reach the browser — it loads
the previous build's hashed chunks, so a fix appears not to work. Kill the harness ports between
builds when iterating on e2e locally.

## Checklist

- [x] Commits are focused and package-local where possible (fix + UI notice are separate commits)
- [x] Public API changes are intentional and reflected in docs (`TranscriptExportError.detail` is
      additive; the wire code is unchanged)
- [x] Contract used by other floats checked — see "Cross-cutting impact"

### DIFF
--- a/docs/transcript-export.md
+++ b/docs/transcript-export.md
@@ -170,6 +170,25 @@ Redaction is **fail-closed**: if the redactor cannot initialize, the export abor
 Each redaction entry in `privacy.redactions[]` records the detector, category, and JSON
 Pointer to the redacted location. `privacy.redactionCounts` gives per-category totals.
 
+#### Schema-controlled fields are never redacted
+
+Both detectors run over every _content-bearing_ string in the document — `session.title`,
+conversation names and folders, message text, tool-call `input`, message `error`, and
+`attachments[].originalName` — plus the text attachments themselves.
+
+They deliberately skip the **schema-controlled** fields: enum discriminators
+(`session.state`, `conversations[].kind`, message `role`, content-block `type`,
+`attachments[].handling`), ISO timestamps, SHA-256 hexes, bundle paths, and the ids the
+relational checks join on. Those are machine-generated and cannot carry a secret, while
+known-secret redaction is plain substring replacement — so a stored secret value that merely
+_occurs_ inside one of them would rewrite it into something `validateTranscriptDocumentV1`
+rejects, failing the whole export instead of the one field. A one-character secret makes that
+certain: it turns `"active"` into `"acti⟦REDACTED:…⟧e"`.
+
+The list lives in `isStructuralTranscriptPointer()`
+(`packages/shared-ts/src/transcript-export.ts`), next to the validator that constrains the same
+fields. A field that gains a constraint in the validator needs an entry there too.
+
 ### Binary attachments — unchanged and potentially sensitive
 
 Text attachments (`text/*` MIME types) are redacted inline before export. Binary attachments
@@ -358,15 +377,15 @@ Any export can be cancelled via `AbortSignal`. On cancellation:
 
 ### Stable error codes
 
-| Code                    | Cause                                                                                                                                                            | Retry?                                                                                                                           |
-| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-| `permission-denied`     | User denied the approval dialog. In the cloud float this also covers a delegated prompt that timed out (120 s), was never shown, or whose follower disconnected. | Only after a new approval.                                                                                                       |
-| `redaction-unavailable` | Redactor failed to initialize.                                                                                                                                   | Not useful — log and report.                                                                                                     |
-| `session-not-found`     | The export service is not registered, or the frozen session ID does not exist.                                                                                   | Wait for boot to complete, or check the session ID.                                                                              |
-| `transfer-aborted`      | Cancelled or disconnected mid-stream.                                                                                                                            | Retry from start — no partial resume.                                                                                            |
-| `transfer-corrupt`      | Byte length or SHA-256 mismatch.                                                                                                                                 | Retry from start — the entire transfer must be re-run.                                                                           |
-| `schema-invalid`        | The assembled transcript failed v1 validation.                                                                                                                   | Report as a bug.                                                                                                                 |
-| `attachment-unreadable` | A text attachment could not be decoded or redacted (fail-closed safety guard).                                                                                   | Do not retry — report as a bug. Binary or missing files do NOT throw this error; they complete as partial with `present: false`. |
+| Code                    | Cause                                                                                                                                                                                                     | Retry?                                                                                                                           |
+| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| `permission-denied`     | User denied the approval dialog. In the cloud float this also covers a delegated prompt that timed out (120 s), was never shown, or whose follower disconnected.                                          | Only after a new approval.                                                                                                       |
+| `redaction-unavailable` | Redactor failed to initialize.                                                                                                                                                                            | Not useful — log and report.                                                                                                     |
+| `session-not-found`     | The export service is not registered, or the frozen session ID does not exist.                                                                                                                            | Wait for boot to complete, or check the session ID.                                                                              |
+| `transfer-aborted`      | Cancelled or disconnected mid-stream.                                                                                                                                                                     | Retry from start — no partial resume.                                                                                            |
+| `transfer-corrupt`      | Byte length or SHA-256 mismatch.                                                                                                                                                                          | Retry from start — the entire transfer must be re-run.                                                                           |
+| `schema-invalid`        | The assembled transcript failed v1 validation. The thrown error's `detail` carries the validator's own message (e.g. `session.state must be "active" or "frozen"`); the wire code stays `schema-invalid`. | Report as a bug, with the `detail`.                                                                                              |
+| `attachment-unreadable` | A text attachment could not be decoded or redacted (fail-closed safety guard).                                                                                                                            | Do not retry — report as a bug. Binary or missing files do NOT throw this error; they complete as partial with `present: false`. |
 
 ### Retry semantics
 
--- a/docs/transcript-export.md
+++ b/docs/transcript-export.md
@@ -172,22 +172,33 @@ Pointer to the redacted location. `privacy.redactionCounts` gives per-category t
 
 #### Schema-controlled fields are never redacted
 
-Both detectors run over every _content-bearing_ string in the document — `session.title`,
-conversation names and folders, message text, tool-call `input`, message `error`, and
-`attachments[].originalName` — plus the text attachments themselves.
+Both detectors run over every string in the document **except** two narrow categories, so
+redaction is fail-closed by default — `session.title`, conversation names and folders, message
+text, message `error`, `source`, `channel`, `stopReason`, `model.*`, tool-call `input`, and
+`attachments[].originalName` / `.mimeType` are all still redacted, as are the text attachments.
 
-They deliberately skip the **schema-controlled** fields: enum discriminators
-(`session.state`, `conversations[].kind`, message `role`, content-block `type`,
-`attachments[].handling`), ISO timestamps, SHA-256 hexes, bundle paths, and the ids the
-relational checks join on. Those are machine-generated and cannot carry a secret, while
-known-secret redaction is plain substring replacement — so a stored secret value that merely
-_occurs_ inside one of them would rewrite it into something `validateTranscriptDocumentV1`
-rejects, failing the whole export instead of the one field. A one-character secret makes that
-certain: it turns `"active"` into `"acti⟦REDACTED:…⟧e"`.
+The two exempt categories:
+
+1. **Validator-constrained** — corrupting them fails `validateTranscriptDocumentV1`: the enum
+   discriminators (`session.state`, `session.completeness.status`, `export.format`,
+   `export.producer.application`, `conversations[].kind`, message `role`, content-block `type`,
+   `attachments[].handling` / `.missingReason`), the ISO message `timestamp`, the SHA-256 hex,
+   the bundle `path`, the non-empty tool-call `id` / `name`, and the ids the relational checks
+   join on.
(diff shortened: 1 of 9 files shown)


trieloff:
Good catch — fixed in d71aa49, and you were right on both halves.

Verified each claim before trimming:

- `source`, `channel`, `stopReason`, `model.*` — `validateMessage` checks only `role`, `id`,
  `sequence`, `timestamp`, `content`, `usage` and (for `tool-result`) `toolCallId`. Unconstrained,
  and `source` can hold a scoop name, so exempting them was exactly the fail-open asymmetry you
  describe.
- `attachments[].mimeType` — string-ness only, and it is *meant* to be redacted: the handling and
  opaque-path decisions read the pre-redaction `PendingAttachment`, and `attachments.ts:603` lists
  `mimeType` as redacted metadata by design.
- `session.createdAt` / `updatedAt` / `frozenAt` / `snapshotAt`, `conversations[].createdAt` /
  `updatedAt`, `export.id` / `generatedAt` / `producer.version`, `delegations[].timestamp` — none
  parsed by the validator. All dropped back to being redacted.

Since exempting a field is fail-open, the set now holds exactly two categories and nothing else:

1. **Validator-constrained** — corrupting them fails `validateTranscriptDocumentV1`.
2. **Join mirrors** — `session.id` and `conversations[].parentConversationId`. Neither is
   constrained beyond string-ness, but each holds a copy of a `conversations[].id` from category 1
   (`session.id` is the cone's own id on the active path — `wc-live.ts` builds it from
   `cone?.id`). Redacting one side and not the other silently desyncs the link, so they have to
   follow the id they mirror. Called out explicitly rather than left looking arbitrary.

One entry I kept that your list implied might go: **`delegations[].toolCallId` is** relationally
checked — `checkEachDelegation` joins it against the source conversation's tool-call ids
(`transcript-export.ts:1037`).

Three tests pin the new boundary so it can't drift back:

- unconstrained pointers are asserted **not** in the set;
- `source` / `channel` are asserted to still be redacted end-to-end by `redactTranscript`;
- every category-1 pointer is corrupted in turn and `validateTranscriptDocumentV1` is asserted to
  reject it — which pins the JSDoc claim itself, so a field that loses its constraint fails this
  test instead of silently staying exempt.

`docs/transcript-export.md` rewritten to match, including the "adding an entry is fail-open"
warning.

Re-verified: `npm run verify`, `npm run typecheck`, `npm run test` (19563 passing), both builds,
`npm run bundle-size` (First-load OK).

PR #2854, opened 2026-09-04
feat(webapp): render images and video inline from markdown

trieloff:
## What

Markdown `alt` in a chat message had no bridge to the VFS and no video path at all. The only way to show the user a picture or a clip was `open`, which spends a browser tab per file — the CDP target list on my dev instance had **six** tabs of the same `interview-cut-v4.mp4`.

One markdown syntax now carries both media types; the file decides which element.

## The three defects, all silent

**1. A VFS path in `` failed with a 200.** `image()` emitted the href verbatim, so `/shared/x.png` resolved against the app origin. Measured live:

| request | status | content-type |
|---|---|---|
| `/shared/video-cut/interview-cut-v4.mp4` | **200** | **`text/html`** |
| `/preview/shared/video-cut/interview-cut-v4.mp4` | 200 | `video/mp4` |

The SPA fallback answers, so the element decodes HTML as an image and fails with nothing logged anywhere. `toPreviewUrl()` already existed and was wired into `open`, the file tree, sprinkles and the follower controller — everything except the renderer.

**2. Video could not survive the sanitizer.** `video` was absent from `PURIFY_CONFIG.ALLOWED_TAGS`, so any clip the agent emitted was deleted outright by DOMPurify.

**3. Video could not seek.** The preview SW returned whole files with no `Accept-Ranges` — 23.2 MB in one `Response` for that mp4 — so a `<video>` had to buffer everything and could never seek.

## What changed

- **`base/message-media.ts`** — the single decision point: which element, and whether the href is rewritten. Extension-based typing is correct here (a path, no bytes) and uses `base/mime-types.ts`, the *serving*-MIME table.
- **`base/preview-url.ts`** — `toPreviewUrl`/`isPreviewUrl` moved out of `shell/supplemental-commands/shared.ts`, which pulls in `just-bash` and the ipk resolver — far too heavy for the render path. `shared.ts` re-exports, so every existing caller is unchanged. First-load gate stays green (allowance 5 kB, unused).
- **`ui/message-renderer.ts`** — `image()` emits `` or `<video controls preload="metadata" playsinline>`; alt text becomes `aria-label` on video, where `alt` is inert. Raw-HTML `src="/…"` resolves the same way, so the two spellings cannot disagree. `groupMediaGalleries()` grids runs of 2+.
- **`ui/preview-sw-handler.ts`** — `Range` → 206 + `Content-Range`, 416 for unsatisfiable, `Accept-Ranges: bytes` on binary reads. Text reads stay whole (they are strings; a byte range would corrupt non-ASCII, and nothing seeks a stylesheet). This fixes seeking; it still reads the whole file first, so peak memory is a follow-up.
- **`slicc-agent-message.ts`** — media capped at the column width, gallery grid, video floor so `preload="metadata"` cannot leave a 0px-tall element with unclickable controls.

`.shtml` references are left completely untouched — `hydrateDips()` keys off `img[src$=".shtml"]` and a rewritten src or a `<video>` would break that handshake.

## Screenshots

Eight new stories under **Chat/AgentMessage Media**. A few below; CI posts the full light+dark set.

| | |
|---|---|
| Single image | Single video — real controls, poster frame |
| 4-up gallery (2×2) | Narrow pane — tiles shrink, no overflow |

Media interleaved with headings, tables, inline code and a blockquote (dark) checks vertical rhythm.

## Note for review: a doc cap moved

`packages/vfs-root/shared/CLAUDE.md` is bundled into every agent context window and has a hard **3000-byte** cap. It sat at 2995 — five bytes of headroom. Documenting the syntax (which has no other discovery path; there is no `--help` for markdown) needs ~230 bytes, so `AGENT_CLAUDE_MAX_BYTES` moves 3000 → 3300 rather than thinning instructions that are all deliberate. Flagging it explicitly because it is a budget change, not just content.

The doc also warns that a ```` ```html ```` fence renders **source**, not markup — I verified this rather than assuming it, and the original plan for this PR had it wrong.

## Verification

`verify` (8/8) · `typecheck` (all 9 projects) · `test` 19,629 passed / 0 failed · `test:coverage` floors met · `build` + extension build · `bundle-size` within budget · `check-touched-exemptions`.

Two full-suite runs showed 4 and 1 failures respectively in `service-worker-secrets`, `slicc-fs-module` and one 5 s timeout — all pass in isolation and none are in files this PR touches; they are load flakes on a 1,103-file parallel run.

Baseline for the Range fix was measured on the live instance: a `bytes=0-99` request returned all 24,341,182 bytes with status 200 and no `Accept-Ranges`.


trieloff:
## Storybook: markdown media in chat messages

Captured from the eight new `Chat/AgentMessage Media` stories at the real 520px column width (260px for the narrow case).

Single image between two paragraphs

Single video with native controls and poster frame

Four images as a balanced 2x2 gallery

Four-up gallery in a 260px pane, tiles shrink without overflow

Images and a video in one gallery

Media interleaved with heading, table, inline code and blockquote in dark mode

trieloff:
Fixed the audio gap from the review — good catch, that was genuinely half-wired.

`resolveMessageMedia` now has an `audio` kind and the renderer emits `<audio controls>` through the same branch as video, so `vo` is a player rather than an `` that cannot decode. Galleries group audio too. Agent doc and package CLAUDE.md updated; 9 new tests (incl. the `.ogg` audio vs `.ogv` video split, the one pair the MIME table separates).

One non-obvious thing the story surfaced that a unit test would not have: **`height: auto` on the shared `.msg__media` rule collapses an `<audio>` to 0px.** It has no intrinsic aspect ratio to resolve against, so the control bar vanished completely while the element still reported `readyState: 4`. `height: auto` is now scoped to the elements that do have a ratio.

Re-verified: verify 8/8 · typecheck all 9 projects · **19,697 tests passed, 0 failed** · both builds · bundle budgets green (first-load allowance untouched).

Audio player inline in a message, 54px control bar

Images and video in one gallery, unchanged

PR #2856, opened 2026-09-04
fix(webapp): hold a replay racing a cone create until its brief is buffered (#2840)

### LINKED ISSUE #2840: New cone: the leader thread shows only the greeting for ~5 s before the initial prompt paints

trieloff:
Closes #2840

## Summary

Creating a cone with an initial prompt showed only the assistant's answer for ~5 s — the user's own prompt and the date divider above it were missing, then appeared all at once. Now the transcript is complete as soon as it paints.

## Why

**The kernel does not drop the replay request — it answers it, with an empty transcript.** That is not what #2840 hypothesised, so the mechanism is worth spelling out.

`orchestrator.registerScoop()` is what lists the new cone to the page, and the freezer rail selects a cone the instant its real record lands (`wc-cone-actions.ts` `render()`). So the page's `request-scoop-messages` routinely arrives while `handleConeCreate` is still inside that `await`, before the brief has been buffered. `handleRequestScoopMessages` then finds the scoop, finds nothing in any of its four sources — buffer, live agent messages, canonical record, legacy session store — and emits the empty replace it is *designed* to emit rather than leave a foreign transcript under the new cone's chrome.

The panel wholesale-replaces the thread with nothing, the assistant's reply streams in on top of that emptiness, and the user's own prompt reappears only when `LocalWorkUnitClient`'s recovery re-asks 5 s later.

Instrumented trace from a `main` run (iteration 2 of 8):

```
237368  kernel coneCreate registered      ← logged before `await registerScoop`
237389  snapshot() ask  cone_1788504237368
237389  kernel req      cone_1788504237368
237389  kernel EMPTY-REPLACE              ← answered, with nothing
237389  publishSnapshot ... 0 messages    ← thread wiped
242393  recoverUnanswered re-ask          ← +5.0 s
242394  publishSnapshot ... 2 messages
242821  prompt visible, 5506 ms after the Create click
```

**Repro** (the leader half of `multiple-cones-follower` walks straight through it; #2840 has the exact command):

1. Boot a leader, open the freezer rail → **New cone**, give it a name and a brief.
2. Click **Create** while the current cone is mid-turn.

Expected: the thread shows the brief and then the reply.
Actual (~20 % of the time): the thread shows the reply alone, then the brief pops in 5 s later.

Fixed at the source rather than in the client. **The client-side change #2840 suggests — re-ask immediately when the roster later NAMES a unit whose snapshot is still unanswered — would not fire here at all**, because the snapshot *was* answered. There is no unanswered request to re-ask.

- `Bridge.conesBeingCreated: Map<jid, Promise<void>>` — a per-jid gate held from *before* `registerScoop` (the call that lists the unit) until the brief is in the buffer. `handleRequestScoopMessages` awaits it before it resolves any source.
- **The gate opens on the prompt being BUFFERED, not on the turn finishing.** `handleUserMessage` pushes the user message synchronously before it awaits the turn, so the gate is released on the *call* rather than the answer. Waiting for the model would park the previous cone's transcript on screen for the length of the first turn — a worse bug than the one being fixed. There is a test for exactly this.
- A registration that throws releases the gate in a `finally`, so a failed create cannot wedge every later replay for that jid into an unresolvable await. Also tested.
- No new ordering requirement anywhere else: `await this.conesBeingCreated.get(jid)` on a jid with no in-flight create is `await undefined`.

## Cross-cutting impact

- **Floats exercised**: the change is in `kernel/facade.ts`, which every local float shares (CLI kernel worker and extension offscreen differ only in transport). Cone creation from the rail is leader-only — followers never mount `wc-cone-actions.ts` — so a follower reaches this code only via its own leader role.
- **Other callers of `handleRequestScoopMessages`**: it is reached only from the `request-scoop-messages` panel message. Every other replay path (`handleRequestScoopTranscript`, `handleRequestScoopChatMessages`) is side-effect-free and deliberately left ungated.
- **Persisted state**: none. The gate is in-memory and per-create; it holds no buffer, writes nothing, and is deleted when it opens.
- **Async lifecycle**: the one real hazard is a gate that never opens, which is why the release is idempotent and lives in a `finally` — covered by the third test.

## Test plan

Three regression tests in `packages/webapp/tests/kernel/bridge.test.ts`. The first constructs the exact interleaving (replay arrives while `registerScoop` is in flight) and **fails on `main`** — verified by reverting only `facade.ts`:

```
× holds a replay racing the create until the brief is buffered (#2840)
✓ opens the replay gate on the brief being buffered, not on the turn finishing (#2840)
✓ does not wedge later replays when the registration throws (#2840)
```

That is a deterministic on-demand reproduction of the race, which is why the timing-sensitive e2e is not leaned on for the guard.

### A/B under load

A green run proves nothing here. A leader-only Playwright probe created a cone with a brief and recorded, from the Create click, when the assistant's reply appeared and when the user's own prompt appeared. The metric is `prompt − reply`: *how long after the answer does the question show up*. Timing the prompt from t=0 instead would score a prompt that painted early and was then wiped by the empty replace as a pass.

Unloaded the race almost never fires (0 hits in 10). **Making it fail on demand needed the load the original repro had from its live follower + real tray**: each cone is created while the previous one is mid-stream, so the kernel worker is streaming and persisting when `registerScoop` runs. That widens the window between "the new cone is listed" and "its brief is buffered" from a couple of milliseconds to something the page reliably lands inside.

Same machine, same load, alternating arms, 12 rounds × 10 cone creations per arm:

| arm | n | hits (`prompt − reply` > 1.5 s) | fast-path delta (min / median / max) | hit delta (min / median / max) |
| **unfixed** (`main`) | 120 | **24 (20.0 %)** | 1 / 2 / 372 ms | 4996 / **5196** / 5624 ms |
| **fixed** (this PR) | 120 | **0 (0.0 %)** | 1 / 2 / 145 ms | — |

The distribution is cleanly bimodal — either ~2 ms or ~5.2 s, nothing in between — which is the `SNAPSHOT_TIMEOUT_MS` signature, not noise. Fisher exact p ≈ 1e-7. One unfixed round, showing the shape:

```
probe0 prompt=5559 reply=368  delta=5191
probe1 prompt=74   reply=72   delta=2
probe3 prompt=5215 reply=57   delta=5158
probe4 prompt=138  reply=134  delta=4
```

The probe spec and its fixtures were scaffolding and are not part of this PR.

### Verification pass

- [x] `npm run verify` — all 8 checks passed
- [x] `npm run typecheck`
- [x] `npm run test` — 19553 passing, 52 skipped, no new failures
- [x] `npm run test:coverage` — no floor breached
- [x] `npm run build`
- [x] `npm run bundle-size` — First-load OK, 4979 kB across both graphs
- [x] New behavior covered by unit tests in `packages/webapp/tests/kernel/bridge.test.ts`
- [x] No UI change to screencast — this is a timing fix in the kernel; the visible difference is the absence of a 5 s gap

**Pre-existing failures**: none.

## Documentation

- [x] `docs/` — `docs/webapp-details.md` (Kernel Host) documents the gate and why it opens on the buffer rather than the turn

## Risk & rollback

- **Blast radius**: cone creation with an initial brief, on every local float. A replay for a jid with no in-flight create is untouched.
- **Failure mode if wrong**: a gate that never opens would hang replays for that jid — hence the idempotent `finally` release and its test.
- **Rollback**: reverts cleanly; no flag, no persisted state, no migration.
- **Not caught by CI**: the race itself. The A/B above is the substitute; the unit test pins the interleaving deterministically.

## Noted, not changed

`snapshot()`'s fallback timer in `LocalWorkUnitClient` is never cancelled, so `recoverUnanswered` fires 5 s after **every** `snapshot()` call, including ones the transport answered promptly — an extra `request-scoop-messages` whenever the unit is still subscribed. That spurious retry is what accidentally repaired the thread here. It is a real defect (an extra wholesale `loadMessages`, which spends the one-shot held-queue restore per #2354), but it is a separate one, and removing it would take away a safety net for other "kernel answered wrongly" paths. Left alone deliberately; worth its own issue.

## Checklist

- [x] No public API / tool / shell-command / lick-channel change
- [x] Contract used by other floats: the `request-scoop-messages` → `scoop-messages-replaced` exchange is unchanged in shape; only its timing during a cone create moves. Follower/leader/offscreen paths checked.


trieloff:
Filed the follow-up the review asked for: #2859 tracks the uncancelled `snapshot()` fallback timer in `LocalWorkUnitClient`, so removing that safety net is a tracked change rather than a note in this PR's body.

No other changes — the review requested none.

Issue #2819, opened 2026-09-03
Advertised commands patch, uptime, id are not implemented (127)


trieloff:
## Summary

Three commands are advertised as available but are not implemented: `patch`, `uptime`, `id`. Each has a published man page on www.sliccy.com and each is listed in the "batteries included" command list on the sliccy.com homepage, but all three exit 127 `command not found`.

## Reproduction

```console
$ patch --version
bash: patch: command not found
$ uptime --version
bash: uptime: command not found
$ id --version
bash: id: command not found

$ which patch uptime id
(no output, rc=1)
```

| Command | `which` | Runs? | Man page | Listed on homepage |
|---|---|---|---|---|
| `patch` | not found | no (127) | [/man/patch](https://www.sliccy.com/man/patch) | yes |
| `uptime` | not found | no (127) | [/man/uptime](https://www.sliccy.com/man/uptime) | yes |
| `id` | not found | no (127) | [/man/id](https://www.sliccy.com/man/id) | yes |

## Why these three are filed here (and not upstream)

These look like SLICC-level rather than pure-bash concerns:

- **`id`** — user/group identity is a SLICC runtime question (there is no real uid/gid in a browser sandbox). It needs a decision about what identity even means here, not a coreutils port.
- **`uptime`** — likewise; the meaningful value would be session/runtime uptime, which only SLICC knows.
- **`patch`** — applying diffs to the VFS. Useful for agent workflows, but it has to be VFS-aware.

`realpath` and `yes` are being filed separately against `vercel-labs/just-bash` as plain shell-utility gaps.

## Requested

For each: either implement it, or drop it from the homepage battery list and unpublish its man page so it is not advertised. Happy to do the content side (this is the sliccy.com repo's own docs) once you decide which way each one goes.

## Related

- Non-functional builtins `jobs`/`bg`/`fg`/`trap`/`select`: #2816
- `nohup` is also advertised (homepage + /man/nohup) and also 127 — noted here as a fourth instance of the same pattern; tell me if you would prefer it split out.

## Environment

SLICC browser shell (`just-bash` WASM), observed 2026-09-03.

trieloff:
All three are implemented rather than unpublished — one PR each, independent, land in any order:

| Command | PR | What it does |
|---|---|---|
[...]

PR #2868, opened 2026-09-04
feat(shell): implement `patch` over the VFS

trieloff:
## Summary

`patch` is listed in the sliccy.com homepage battery list and has a published man page at [/man/patch](https://www.sliccy.com/man/patch), but it was never implemented: `patch --version` answered `bash: patch: command not found` (127). This adds a real, VFS-aware `patch` that parses unified diffs and applies them to files.

## Why

Closes the `patch` third of #2819 ("Advertised commands patch, uptime, id are not implemented"). `patch` is the VFS-aware one of the three: applying a diff means reading and writing files, not just reporting runtime identity — which is why the issue filed it here rather than upstream against `just-bash`.

**Repro (before):**

```console
$ patch --version
bash: patch: command not found
$ which patch
(no output, rc=1)
```

**After:**

```console
$ diff -u note.txt edited.txt > change.patch
$ patch < change.patch
patching file note.txt
```

**Where the reusable starting point was.** `git/diff.ts` already produces unified diffs (Myers) and `git/merge-file-core.ts` already merges three versions of a file — but nothing in the repo could take a diff back *apart*. `src/git/patch-core.ts` is the missing inverse, and it sits next to the producer so a future `git apply` can reuse it without a shell → git back-edge. `upgrade`'s three-way merge is left alone; it solves a different problem (merging bundled VFS files across releases).

- **`src/git/patch-core.ts`** — pure text in / text out. Unified-diff parser (multi-file, `diff --git` preambles, mail wrappers, `/dev/null` creations and deletions, `\ No newline at end of file` on either side, blank-line-as-context tolerance) plus a hunk applier with GNU-style outward offset search and context fuzz, reverse, and `.rej` rendering.
- **`shell/supplemental-commands/patch-command.ts` + `patch/{args,run}.ts`** — argv, VFS I/O, and the GNU-shaped report. `index.ts` gets a registration stub only; the grammar and applier are imported on first use, so the kernel worker's eager graph does not grow.

Advertised surface (`-p NUM`, `-R`, `--dry-run`) plus `-i FILE`, `-F NUM`, `-s`, `--help`, `--version`, `FILE PATCHFILE` operands, and stdin.

**Two non-obvious decisions:**

1. **Fuzz never ignores a `-` line.** Only genuine context lines at a hunk's ends may be dropped (`leadingContext` / `trailingContext` are tracked during the parse for exactly this). Fuzzing away a removed line would silently apply a hunk to text it does not describe — the worst failure mode a patch tool has. Pinned by a test.
2. **With no `-p`, the target is auto-detected.** The name as written, then one component stripped at a time; the first candidate that *exists* wins, else the first whose *parent directory* exists (that is what `+++ b/src/new.ts` creating `src/new.ts` looks like), else the basename — which is what POSIX `patch` does with no `-p` at all. Strict POSIX alone would mean `patch < git-format.diff` lands on a same-named file in the cwd; this keeps that as the fallback but lets the filesystem speak first. Documented in `--help` and `docs/shell-reference.md`.

No new dependencies.

## Cross-cutting impact

- **Floats exercised**: the command is registered in `createSupplementalCommands()`, so it is identical in CLI, extension, Electron and the kernel worker — there is no float-specific path. Nothing native, no follower wiring, no CDP.
- **Bundle size**: first-load gate reports **worker +0.1 kB** (the stub), with `patch-core` + `patch/run` in a lazily-imported chunk. `node packages/dev-tools/tools/check-first-load-size.mjs` → `First-load OK`.
- **Other callers**: `git/patch-core.ts` is a new module — nothing else imports it yet. `git/diff.ts` is untouched; the round-trip test consumes its output rather than changing it.
- **Persisted state**: none. `patch` writes only the files the diff names (plus `FILE.rej` on rejection), through the same gated `ctx.fs` every other command uses — so a patch aimed at `/etc/sudoers` still raises the normal approval.
- **Security**: no secrets, no network. `-p`/auto-detection strips leading components and never rejoins `..` segments back into a path; targets resolve through `ctx.fs.resolvePath(ctx.cwd, …)` like every other file command.

## Test plan

- [x] `npx vitest run packages/webapp/tests/shell/supplemental-commands/ packages/webapp/tests/git/` — **4329 passed, 1 file skipped**, no new failures
- [x] `npm run build -w @slicc/webapp` — clean; `check-first-load-size.mjs` → OK
- [x] `npm run lint:docs` (doc-refs, doc sizes) and `npm run lint:duplication` — clean
- [x] **New behavior is covered by unit tests** mirroring the changed `src/` paths:
      - `packages/webapp/tests/git/patch-core.test.ts` — 23 pure tests (parsing, fuzz, offsets, reverse, EOF-newline, creation/deletion, garbage input)
      - `packages/webapp/tests/shell/supplemental-commands/patch-command.test.ts` — 33 tests on a real `VirtualFS` through the real shell, including a round trip of this shell's own `diff -u` output back through `patch`
- [ ] Manual smoke — not run; the command has no UI and the shell-level tests drive the real `AlmostBashShellHeadless` over a real `VirtualFS`
- [x] UI change? No.

## Documentation

- [x] `docs/` — new `patch` row in `docs/shell-reference.md` (supplemental command table), including the auto-`-p` rule
- [x] `commands` listing — `patch` added to the "Text processing" category next to `diff`
- [ ] `README.md` — not needed; the homepage/man-page content side stays with the issue author, and nothing there needs unpublishing now that the command answers

## Risk & rollback

Blast radius is one new shell command plus one new `git/` module; nothing existing changes behavior. No feature flags, no migrations, no persisted state. Reverting this PR is clean — `patch` goes back to 127.

Nothing here needs a real browser to verify: no CDP, no OAuth, no native permissions.

## Checklist

- [x] Public shell-command surface change is intentional and reflected in docs
- [x] No cross-float contract change (one registration, all floats)

Refs #2819


trieloff:
Fixed in 521831461. Short flags are now consumed as a getopt-style cluster: booleans bundle in any order (`-Rs`, `-sR`), and the first value flag takes the rest of the cluster (`-p1`, `-Rp1`) or, failing that, the next argument (`-p 1`). An unknown letter names itself in the error — `unrecognized option '-z'` for `-Rz`, not the whole bundle. Four `parsePatchArgs` cases added.

PR #2869, opened 2026-09-04
feat(shell): implement `uptime` from the last page load

trieloff:
## Summary

`uptime` is listed in the sliccy.com homepage battery list and has a published man page at [/man/uptime](https://www.sliccy.com/man/uptime), but it was never implemented: `uptime` answered `bash: uptime: command not found` (127). This adds it, measured from the last load of the browser window.

## Why

Closes the `uptime` third of #2819. The issue is explicit that this one is a SLICC-level question rather than a coreutils port — "the meaningful value would be session/runtime uptime, which only SLICC knows".

**Repro (before):**

```console
$ uptime
bash: uptime: command not found
```

**After:**

```console
$ uptime
 11:42:07 up 2:07,  1 user,  load average: 0.15, 0.09, 0.05
$ uptime -p
up 2 hours, 7 minutes
$ uptime -s
2026-09-04 09:35:00
```

**The clock starts at the last page load.** In a browser the "system" is the page: a reload is a reboot. Every realm can read `performance.timeOrigin`, but the shell runs in the **kernel worker**, where `timeOrigin` is when the *worker* was constructed — some way into the page's boot, and materially wrong on a slow cold boot. So the page hands its own origin over in `kernel-worker-init`, right beside `pageBuildId`, which already travels for exactly this kind of reason. `base/page-load-time.ts` records it (same shape as `base/slicc-version.ts`'s registered reader) and falls back to this realm's own `performance.timeOrigin` when nothing was registered — which is *correct* in the page realm and the closest available answer anywhere else.

`spawn.ts` has one production caller (`ui/wc/wc-live.ts`, at boot), so the value is set exactly once per page load.

**The load average is real, not simulated.** The man page promised "Load average is simulated"; making up three numbers seemed worse than reporting nothing, and it turns out we do not have to. `ProcessManager` already records `startedAt` / `finishedAt` for every process — the same table `ps` reads — so the average number of processes alive over the last 1/5/15 minutes is computed **exactly**, by integrating lifetimes over each window, rather than sampled the way the kernel does it. No timers, no background sampler, no new state; `loadAverage()` is a pure function over the process list.

Two deliberate details:

- The window is **clamped to the elapsed uptime**, so a runtime that has been up ten seconds and busy the whole time reports `1.00`, not `0.17`.
- The known limitation is upstream and documented: `ProcessManager` retains a bounded number of *terminated* records, so the 5- and 15-minute windows under-report after a long burst of short commands. Called out in `--help` and in `docs/shell-reference.md` rather than papered over.

**The user count is `1`.** Not a placeholder: a SLICC runtime has exactly one user identity — the same one `id` reports in its sibling PR.

No new dependencies.

## Cross-cutting impact

- **Floats exercised**: `KernelWorkerInitMsg.pageLoadedAt` is optional, so a page from an older build simply omits it and the worker falls back to its own origin (same tolerance `pageBuildId` has). All floats that spawn a kernel worker get the accurate value through the one `bootstrapKernelWorker` init site; realms that run the shell in the page realm read `performance.timeOrigin` directly and are already correct.
- **Other callers of the changed contract**: `KernelWorkerInitMsg` gained an optional field — no existing consumer reads it, and `configureWorkerEnvironment` is the only place it is consumed. `ProcessManager` is untouched; `uptime` only imports its *types* and calls `list()`, exactly as `ps` does.
- **Bundle size**: `check-first-load-size.mjs` → **OK**, worker graph unchanged within the allowance (the command has no heavy imports, so it registers eagerly like `uname` rather than needing a lazy stub).
- **Persisted state**: none. No localStorage keys, no IndexedDB, no migration.
- **Async lifecycle / security**: no timers, no listeners, no network, no secrets. `uptime` reads memory and formats a string.

## Test plan

- [x] `npx vitest run packages/webapp/tests/shell/supplemental-commands/ packages/webapp/tests/kernel/` — **5340 passed, 1 file skipped**, no new failures
- [x] `npm run build -w @slicc/webapp` — clean; `check-first-load-size.mjs` → `First-load OK`
- [x] `npm run lint:docs`, `npm run lint:layer-back-edges` — clean
- [x] **New behavior is covered by unit tests** in `packages/webapp/tests/shell/supplemental-commands/uptime-command.test.ts` (27 tests): registration and `--version` (the issue's own repro), the `page-load-time` registry and its fallbacks, `formatUptime` / `formatPretty` / `formatSince`, seven `loadAverage` cases including window clipping and the fresh-boot clamp, the rendered line against a fixed clock, a live `ProcessManager`, and an end-to-end run through the real `AlmostBashShellHeadless` checking `commands` and `which` too
- [ ] Manual smoke — not run; no UI surface, and the shell-level test drives the real shell
- [x] UI change? No.

## Documentation

- [x] `docs/` — new `uptime` row in `docs/shell-reference.md`, covering the page-load clock, the flags, and the load-average caveat
- [x] `commands` listing — `uptime` added to the "Process" category next to `ps` / `kill` / `meminfo`
- [ ] `README.md` — not needed; the homepage/man-page content side stays with the issue author. Worth noting for them: the man page's "Load average is simulated" line is now wrong in our favour

## Risk & rollback

Blast radius is one new shell command plus one optional field on the kernel-worker init message. Nothing existing changes behavior; the field is additive and tolerated when absent. No flags, no migrations. Reverting this PR is clean — `uptime` goes back to 127.

One thing CI cannot check: that the reported uptime matches wall-clock time since a real page reload. The plumbing is pinned by tests, but a reviewer with a running instance can confirm with `uptime -s` against the time they last reloaded.

## Checklist

- [x] Public shell-command surface change is intentional and reflected in docs
- [x] Cross-float contract (`KernelWorkerInitMsg`) change is additive and optional; the leader/worker paths were checked

Refs #2819


trieloff:
Thanks — both optional notes read right to me, and I have deliberately kept both as they are:

- **`-p` under a minute.** `up N seconds` beats procps' `up 0 minutes` for a browser runtime, where "just reloaded" is a state you are actually in often enough to want the number. It is tested, so it will not drift by accident.
- **`-p -s` last-wins.** Matches procps, and neither ordering is more obviously right than the other.

No changes pushed for this PR.

PR #2871, opened 2026-09-04
feat(shell): implement `id` from the home directory

trieloff:
## Summary

`id` is listed in the sliccy.com homepage battery list and has a published man page at [/man/id](https://www.sliccy.com/man/id), but it was never implemented: `id` answered `bash: id: command not found` (127). This adds it, taking identity from the home directory. It also shadows `whoami`, which just-bash hardcodes to `user`, so the two agree.

## Why

Closes the `id` third of #2819. The issue is explicit that this one needed a decision before it needed code: *"user/group identity is a SLICC runtime question (there is no real uid/gid in a browser sandbox). It needs a decision about what identity even means here, not a coreutils port."*

**Repro (before):**

```console
$ id
bash: id: command not found
```

**After:**

```console
$ id
uid=1000(lars) gid=1000(lars) groups=1000(lars),10(cone)
$ id -un
lars
$ id researcher                      # a scoop
uid=41277(researcher) gid=41277(researcher) groups=41277(researcher),20(scoop)
```

**The decision: identity is the home directory.** The shell already answers this question — it just never exposed the answer. `shell/home-dir.ts` resolves `$HOME` from the `/home` tree (onboarding's `/home/<slug>`) and `$USER` follows as its basename; `scoops/scoop-context/shell-env.ts` pins `HOME=/scoops/<folder>/home` and `USER=<folder>` for a scoop. `id` reports what the shell already knows, in id(1)'s shape, rather than inventing a parallel identity scheme.

**The numbers follow the cone/scoop split** — the one identity distinction this codebase already treats as fundamental ("cone and scoop are roles over one `WorkUnit`"):

| | uid | groups |
|---|---|---|
| **cone** — the human user | `1000`, like the first account on any Linux install. One per runtime: every cone shares the one `/home/<slug>`. | user-private group + `cone` (10) |
| **scoop** — a service identity | FNV-1a of the folder name into `2000–60000`. Stable across reloads, distinct per scoop, no passwd file to keep in sync. | user-private group + `scoop` (20) |

Collisions in the scoop range are possible and harmless — the name in parentheses is the identity that matters, exactly as on a real system with a duplicated uid. Scoop uids can never collide with the cone's, which is pinned by a test.

An unknown `USER` operand is `id: 'nobody': no such user` (exit 1), not an invented identity. `-r` is accepted and ignored: there is no setuid here, so real and effective never differ. `-n` alone is refused, as coreutils does.

**`whoami` is shadowed too.** just-bash's bundled `whoami` returns the literal `user` regardless of the environment. Shipping `id` alone would leave two commands in the same shell disagreeing about who is running it — `id -un` reading the home onboarding created, `whoami` not. A bundled command can be shadowed the way `rm` / `stat` already are in `symlink-aware-file-commands.ts`. This is a deliberate, small scope addition; flagging it because it changes an existing command's output. **On a profile that never onboarded both still answer `user`** — they only diverge where the old answer was wrong. Pinned by a test either way.

No new dependencies.

## Cross-cutting impact

- **Floats exercised**: one registration in `createSupplementalCommands()`, so CLI, extension, Electron and the kernel worker all get the same command. No native, follower, or CDP paths.
- **Other callers of the changed contract**: `whoami` is the one behavior change. Grepped for every `whoami` reference in `src/` and `tests/` — the hits are the `commands` category list, an agent-command fixture, an upskill fixture, and two realm sync-exec/sync-fs acceptance tests that use it as a trivial command whose *output* they do not assert. All of `tests/kernel/`, `tests/net/` and `tests/shell/` run green (6790 tests).
- **`home-dir.ts` is read, not changed**: `currentIdentity` uses `$USER`/`$HOME` (always set before the first command by `initHomeAndProfile`); the `/home` scan is a defensive fallback behind a small `IFileSystem` → `HomeDirFS` adapter.
- **Bundle size**: `check-first-load-size.mjs` → **OK**. No heavy imports, so it registers eagerly like `uname` rather than needing a lazy stub.
- **Persisted state**: none — `id` reads env and probes two directory paths. No writes, no localStorage, no migration.
- **Security**: no secrets, no network. The uid is derived from a folder name that is already public within the runtime; it grants nothing (SLICC's real authorization is `/etc/sudoers` + the capability broker, untouched here).

## Test plan

- [x] `npx vitest run packages/webapp/tests/kernel/ packages/webapp/tests/net/ packages/webapp/tests/shell/` — **6790 passed, 1 file skipped**, no new failures
- [x] `npm run build -w @slicc/webapp` — clean; `check-first-load-size.mjs` → `First-load OK`
- [x] `npm run lint:docs`, `npm run lint:layer-back-edges` — clean
- [x] **New behavior is covered by unit tests** in `packages/webapp/tests/shell/supplemental-commands/id-command.test.ts` (25 tests): registration and `--version` (the issue's own repro), the cone identity, the scoop identity and its uid stability / distinctness / non-collision, every selector including `-un` / `-gn` / `-Gn` and the `-n`-alone refusal, `USER` lookup against `/home` and `/scoops`, the unknown-user error, and — through the real `AlmostBashShellHeadless` over a real `VirtualFS` — `commands`, `which`, `whoami` agreeing, and `whoami` still answering `user` on a fresh profile
- [ ] Manual smoke — not run; no UI surface, and the shell-level tests drive the real shell
- [x] UI change? No.

## Documentation

- [x] `docs/` — new `id` row in `docs/shell-reference.md`, covering the identity model, the uid scheme, the groups, and the `whoami` shadow
- [x] `commands` listing — `id` added to the "Environment & shell" category next to `whoami` / `hostname`
- [ ] `README.md` — not needed; the homepage/man-page content side stays with the issue author. Worth noting for them: the man page says "Returns simulated values", which is no longer accurate — the values are derived from the runtime's real home directories

## Risk & rollback

Blast radius is one new command plus one shadowed one. No flags, no migrations, no persisted state. Reverting is clean: `id` goes back to 127 and `whoami` back to upstream's hardcoded `user`.

The one thing a reviewer may want to weigh in on rather than CI: the identity model itself (cone = uid 1000, scoop = hashed). It is a decision, not a derivation, and it is the part of the issue that was left open. If a different model is preferred, the change is contained to `identityFor` / `scoopUid` / `groupsOf`.

## Checklist

- [x] Public shell-command surface change is intentional and reflected in docs (`id` added, `whoami` behavior changed)
- [x] No cross-float contract change (one registration, all floats)

Refs #2819


trieloff:
Fixed in f7eb44486. You are right that it invented an identity — the fallback now derives the name per role: basename for a cone, parent folder for a scoop (`nameFromHome`). Still unreachable in practice, since `buildScoopShellEnv` pins `$USER` for every scoop, but the fallback should be correct rather than merely present. Test added asserting a scoop with `HOME=/scoops/researcher/home` and no `$USER` reports `(researcher)` and never `(home)`.

PR #2873, opened 2026-09-04
fix(webapp): resolve sliccy:skill refs/assets from the skill root

### LINKED ISSUE #2866: sliccy:skill refs/assets resolve to scripts/ but all 89 skills ship them at the skill root
## Summary

`sliccy:skill` exposes `refs` and `assets` as **script-relative**
(`<dir>/references`, `<dir>/assets` where `dir` is the running script's
directory). Every skill in `ai-ecoverse/skills` ships those directories at the
**skill root**, one level above `scripts/`. So both properties point at paths
that do not exist, and no skill can use them to read its own bundled files.

## The numbers, from a fresh clone of ai-ecoverse/skills

```
root-level references/ : 89
scripts/references/    :  0
root-level assets/     :  5
scripts/assets/        :  0
```

Not a majority — unanimous. The API resolves to a location zero skills use.

## Root cause

`packages/webapp/src/kernel/realm/skill-global.ts:79`

```ts
const scriptPath = deps.argv[1] ?? '';
const dir = dirname(scriptPath);
const refs = joinChild(dir, 'references');
const assets = joinChild(dir, 'assets');
```

For the standard layout `<skill>/scripts/foo.jsh`, `dir` is `<skill>/scripts`,
so `skill.assets` is `<skill>/scripts/assets`.

This is consistent with the documented contract in
`skills/skill-authoring/jsh-runtime-extensions.md:192` (`skill.assets` —
`` `<dir>/assets` ``), so the implementation matches its own docs. The mismatch is
between that contract and the repo's packaging convention.

Notably, the module's own header cites `oryx` as one of the skills whose
`argv[1]` dirname math it replaces — and `oryx` ships **root-level** `assets/`
and `references/`. The motivating example has the layout the API does not
support.

## How it surfaces

Nothing fails at load time; the paths simply do not exist. It surfaces later, in
whatever way the consumer happens to read them:

- `remotion-clipper` — `remotion render`, the skill's headline command, failed
  outright on a fresh install with "could not locate the render harness".
- `interview-me` — one asset read sat inside a deliberately non-fatal
  `try/catch`, so `install` **silently** skipped a bundled README with no error
  at all. That is the dangerous shape: an unusable path plus a tolerant caller
  equals a feature that quietly does nothing.

Both were only ever exercised from a checkout whose layout happened to match, so
neither showed up until installation via `upskill`.

## Suggested fix

Any one of these, in rough order of preference:

1. Expose the skill root (e.g. `skill.root`, stripping a trailing `/scripts`),
   and resolve `refs` / `assets` from it — matching all 89 skills.
2. Resolve `refs` / `assets` by probing root-level first, then script-relative.
3. If script-relative is intended, say so unambiguously in the skill-authoring
   docs and fix the 89 skills — but that contradicts the existing convention and
   every SKILL.md reference to `references/…`.

Whichever is chosen, `skill.refs` / `skill.assets` currently cannot be used as
documented, and callers are silently working around it (`ai-ecoverse/skills#339`
derives the root by stripping `/scripts`).

## Environment

SLICC 6.128.7

trieloff:
Closes #2866

## Summary

`require('sliccy:skill')` now points `refs` and `assets` at the **skill root**, not the running script's directory. A script at `<skill>/scripts/foo.jsh` reads `<skill>/references` and `<skill>/assets` — the layout every published skill ships.

Before this, those properties resolved to `<skill>/scripts/references` and `<skill>/scripts/assets`, which do not exist. Callers failed later (or silently skipped bundled files): `remotion-clipper`'s render harness, `interview-me`'s install README.

## Why

From a fresh clone of `ai-ecoverse/skills`:

```
root-level references/ : 89
scripts/references/    :  0
root-level assets/     :  5
scripts/assets/        :  0
```

The API matched its own docs (`<dir>/references` where `dir` is the script folder) and mismatched the packaging convention. `oryx`, the motivating example in `skill-global.ts`, ships root-level `assets/` and `references/`.

**Repro (before):**
1. Install a skill whose `.jsh` lives under `scripts/` (`upskill` / `ai-ecoverse/skills`).
2. In that script, `require('sliccy:skill')` and `fs.readFile(`${skill.refs}/…`)` or `${skill.assets}/…`.
3. Expected: the bundled file.
4. Actual: ENOENT, or a `try/catch` that skips the file with no error.

Preferred option from the issue: expose the skill root and resolve `refs`/`assets` from it. No probing, no changes to the 89 skills in the other repo.

- `skill.dir` stays the directory containing the running script.
- `skill.root` is new: if `basename(dir) === 'scripts'`, the parent; otherwise `dir`.
- `skill.refs` / `skill.assets` are `<root>/references` and `<root>/assets`.
- `skill.config()` still reads/writes `<dir>/.config` (typically `scripts/.config`) because `upskill` preserves dotfiles there.

Rejected probing (root-level first, then script-relative): it would make boot async, and zero skills use the script-relative layout.

## Cross-cutting impact

- **Floats exercised:** CLI and Chrome extension share the JS realm (`createSkillGlobal`); both pick this up. Electron / Sliccstart / worker use the same kernel path.
- **Other callers:** `skill.dir` is unchanged. Existing scripts next to `SKILL.md` (not under `scripts/`) keep the same `refs`/`assets` paths. Additive `skill.root` only.
- **Persisted state:** none. Path math at realm boot.
- **Bundle size:** first-load worker +0.1 kB vs merge-base (under the 5 kB allowance).

## Test plan

- [x] `npm run verify` — all 8 lint-gate checks passed
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs` — no debt lists
- [x] `npm run typecheck`
- [x] `npm run test` — 19750 passed, 46 skipped
- [x] `npm run test:coverage` — floors held
- [x] `npm run build`
- [x] `npm run build -w @slicc/chrome-extension` (covered by the workspace build)
- [x] `npm run bundle-size` — first-load OK; extension size-limit OK
- [x] New behavior covered in `packages/webapp/tests/kernel/realm/skill-global.test.ts` and `sliccy-modules.test.ts`:
  - `<root>/scripts/foo.jsh` → refs/assets at `<root>/`
  - script at skill root still resolves refs/assets there
  - directory named `my-scripts` is not treated as `scripts/`
  - `.config` stays at `scripts/.config`
- [x] N/A for UI screencast — kernel path math only

## Documentation

- [x] `packages/vfs-root/workspace/skills/skill-authoring/jsh-runtime-extensions.md` (agent-facing contract)
- [x] `packages/vfs-root/workspace/skills/skill-authoring/SKILL.md`
- [x] `docs/shell-reference.md`

## Risk & rollback

- Blast radius: `.jsh` scripts that call `sliccy:skill` for `refs`/`assets`. Scripts that already derived the root by stripping `/scripts` keep working; hypothetical `scripts/references/` layouts would miss files (none exist in `ai-ecoverse/skills`).
- Rollback: revert this PR. No persisted-state migration.
- Additive `skill.root`; `skill.dir` / `config()` / `token()` unchanged.


trieloff:
Addressed the nested-`scripts/` observation from the gated review: `skill.root` is now the parent of the first `scripts/` path segment, so a helper at `<root>/scripts/lib/foo.jsh` still resolves `refs`/`assets` at `<root>/`. Follow-up: `d5d8ddfee`.

PR #2879, opened 2026-09-04
docs(vfs-root): cut /shared/CLAUDE.md to constitutional invariants

### LINKED ISSUE #2860: Cut /shared/CLAUDE.md to ~2 kB: constitutional invariants, not a duplicated manual
## Proposal

Cut `packages/vfs-root/shared/CLAUDE.md` from **3,253 bytes to ~2,000** by removing content the runtime already injects, and lower `AGENT_CLAUDE_MAX_BYTES` to match.

> The ontology is right; the abstraction level is wrong. The file petrifies a command catalogue, lick dispatch table, workflow API, and sudo grammar inside permanent context. SLICC already builds each unit's prompt from its actual role, policy, workspace, memory, and loaded skills. The shared prompt should contain constitutional invariants, not duplicate manuals.

## This is verifiable in the code, and the case is stronger than "redundant"

`scoops/scoop-context/system-prompt.ts` assembles the prompt "from the unit's policy + workspace, its memories, and the loaded skills". Two lines matter:

- **:27** — *"advertises come from the policy so the text can never promise a tool"* the unit lacks. A hand-maintained command list in `/shared/CLAUDE.md` has no such guard, so it can and will contradict the injected capability set.
- **:87–88** — `/shared/CLAUDE.md` is injected as **global memory read by every unit**, while the role is injected separately as `` isRoot ? 'Cone' : 'Scoop' ``.

That second point is a live defect, not just bloat: the shared file currently opens with

```
- **Cone**: You. Orchestrate scoops, talk to the human, full FS access.
```

which is **false for every scoop that reads it** — and scoops all read it. The runtime already told each unit which role it is; the shared file then tells all of them they are the cone.

Related invariants the file should defer to rather than restate: `work-unit/policy.ts:213` (`child capabilities ⊆ parent capabilities`) and `work-unit/descriptor.ts` (`isRootUnit` / `parentJid: null`) — the WorkUnit migration (#1666) permits coexisting root cones, so both "Cone: You" and any generic `cone` target are unsafe.

## Proposed body

```markdown
# sliccy

SLICC is a browser-native runtime for coding, browsing, automation, and parallel agents.

## Runtime

Injected role, policy, workspace and tools are authoritative. Cone: root work unit. Scoop: delegated child. Root cones may coexist. Do not infer authority or assume runtime parity.

Sprinkles are persistent `.shtml` panels owned by long-lived scoops; dips are ephemeral `shtml`; licks are addressed events; trays are remote runtimes.

## Discover

Before claiming absence, inspect `commands`, `<cmd> --help`, `man <topic>`, and `skill list`; search with `upskill search "<query>"` or `upskill tabs`. Verify locally: manuals and skills define syntax and capability.

Reusable capability belongs in a skill. Read `/workspace/skills/skill-authoring/SKILL.md` before authoring.

## Delegate

Cones orchestrate and synthesize; scoops do independent work. Give each scoop a self-contained prompt; assume no chat context. Read `man delegation`.

Parallelize non-browser work as useful. Browser drivers share a serialized bridge: keep at most 3-4 active, use waves and back off on contention. If work goes quiet, inspect pending approvals.

## UI and events

One identically named scoop owns each sprinkle. A cone must not edit its `.shtml`, run `sprinkle`, or handle its licks; route them to the owner. A cone may author a dip directly. Read its skill first.

Route licks to the addressed work unit; never hardcode a singleton cone. Discovery events are optional signals. Handoffs and privileged actions are human-gated. Automation targets the intended unit; never assume `cone`.

Render local media as `![label](/absolute/path)`.

## Operate

Respect policy and sudo; request least privilege. After an ambiguous failure, verify state before repeating a mutation. Preserve evidence, inspect outputs and try a different path. Verify results and artifacts before claiming completion.

Keep only durable facts in unit or shared memory; prune stale entries. Be direct, professional, and emoji-free.
```

## Two implementation notes

**1. The gate counts bytes, not characters.** The draft is exactly 1,999 chars (2,000 with the trailing newline) — but `check-doc-sizes.mjs` uses `measureBytes` for this file, and the original `3–4` used an **en-dash** (3 bytes in UTF-8), landing at 2,002 bytes. The block above substitutes an ASCII hyphen, which lands it at exactly **2,000 bytes**. Suggest setting `AGENT_CLAUDE_MAX_BYTES = 2048` for a little headroom rather than pinning to 2,000 with zero — the current 3,000-byte cap ran at five bytes of headroom, which is what forced it up to 3,300 in #2854.

**2. The one operational number is not sourced from this repo.** "3-4 concurrent browser drivers" comes from an operator observation — eight drivers contending for one serialized bridge roughly doubled wall time — not from anything currently measured or asserted in the codebase. I could not find a documented limit in `cdp/` or `docs/`. It is proposed for permanence *because* it is the kind of fact a unit cannot discover from `--help`; worth confirming before it is enshrined, and worth a comment near the bridge if it holds.

## What must not be lost

Everything below is currently only in this file. If it moves, it needs a destination:

- Sprinkle ownership (`man sprinkle` covers the mechanics, not the cone prohibition)
- Lick routing / actionable verbs — the dispatch table is a manual, but "route to the addressed unit" is an invariant
- Workflow API surface → belongs in `man workflow` or a skill
- Sudo grammar → `docs/approvals.md` and `man sudo`
- Markdown media syntax (#2854) — retained above as one line

Fixes the underlying tension noted in #2854, where documenting one new capability required raising the cap because every remaining byte was a deliberate instruction.

trieloff:
Fixes #2860

## Summary

Cuts `packages/vfs-root/shared/CLAUDE.md` from 3,253 to **1,826 bytes** and lowers `AGENT_CLAUDE_MAX_BYTES` from 3300 to **2048**. The file keeps only what the prompt assembler does not already say; the duplicated command catalogue, lick dispatch table, workflow API, and sudo grammar are gone.

Before: every scoop read `- **Cone**: You. Orchestrate scoops, talk to the human, full FS access.` in its own system prompt. Now the file defines Cone and Scoop as structural terms and asserts no identity, deferring to the injected role.

## Why

`/shared/CLAUDE.md` is injected verbatim into **every** work unit's system prompt as the `GLOBAL MEMORY` block, so every byte is paid for on every turn of every cone and every scoop. It had drifted from constitutional invariants into a duplicated manual.

It also carried a live defect. `scoop-context/memories.ts::loadMemories` returns the shared file for every unit unconditionally — the `canWriteSharedMemory` check only gates the write-back mirror — so every scoop read "Cone: **You**" one screen after `buildScoopSystemPrompt` (`system-prompt.ts:34`) had already told it it is "a scoop assistant". Roles are derived from the parent edge (`work-unit/types.ts:39` — "never stored") and root cones may coexist (#1666), so the singleton framing was wrong twice over.

What remains is only what the assembler does *not* say: the explore-before-refusing imperative, the dip / lick / tray vocabulary, the markdown media syntax (no `--help` exists for markdown — the reason #2854 raised the cap), lick routing to the addressed unit, and tone.

Dropped content and where it stays reachable:

| Dropped | Destination |
| --- | --- |
| "Cone: You" + scoop tool names | `system-prompt.ts:34`, `:51-62` (policy-gated) |
| SLICC-native command list | `commands` (`help-command.ts`) |
| Delegation principle | `system-prompt.ts:64-74`; `delegation` skill; `man delegation` |
| Sprinkle ownership rules | `sprinkles/SKILL.md:219-229`; `man sprinkle` |
| Dip authoring pointer | injected `AVAILABLE SKILLS` entry for `dips` |
| Lick dispatch table | `automation/SKILL.md`; `man lick`; `lick_confirm`/`lick_dismiss` tool schemas |
| Workflow API | `workflows/SKILL.md`; `man workflow`; `workflow --help` |
| Sudo grammar | `man sudo`; `docs/approvals.md`; readable `/etc/sudoers` |
| Memory hierarchy | `system-prompt.ts:84-92` |

Two fixes ride along:

1. **`system-prompt.ts:66` advertised a `delegate_to_scoop` tool that is not registered anywhere.** The real tool is `feed_scoop` (`scoop-management-tools.ts:894`); the only other hits for the old name are comments. The policy guard cited in the issue governs the capability *gate*, not the tool *name* — and this mattered the moment the shared file stopped naming the real one.
2. **The cap was documented in three places, two already stale at 3,000** — `docs/CLAUDE.md`, `claude-md-compactor/lib.mjs`, `docs/dev-tools-details.md`. All three now say 2,048.

`packages/vfs-root/CLAUDE.md` also cited a "trigger line in `shared/CLAUDE.md`" for handoffs that this change deletes; it now points at the skill.

Deliberate choices:

- **2,048, not a hard 2,000.** The issue's draft lands at exactly 2,000 bytes. Zero headroom is what forced 3,000 → 3,300 in #2854; this ships at 1,826 with 222 bytes to spare.
- **`man` topics are singular.** `man` is a network fetch to `www.sliccy.com/man/<topic>.plain.html` (`man-command.ts:48`). Verified live: `commands`, `delegation`, `sprinkle`, `handoff`, `sudo`, `workflow`, `lick` → 200; the plurals `sprinkles`, `workflows`, `licks` → 404. The file does not *depend* on `man` — "Explore first" lists local paths too, since egress-blocked floats cannot reach it.
- **Left out the issue's "keep at most 3-4 active browser drivers" line.** That number is an operator observation with no source in `cdp/` or `docs/`. Enshrining an unmeasured limit in a constitutional file contradicts the issue's own thesis. Worth raising separately, as a comment near the bridge, if it holds.

## Cross-cutting impact

- **Floats**: the file is bundled into the VFS, so all floats (CLI, extension, Electron, Sliccstart, worker) get the new text identically. No float-specific path touched.
- **Every work unit's prompt changes** — that is the point. Cones lose the false-but-harmless "you are the cone" restatement; scoops lose a line that actively contradicted their injected role.
- **`system-prompt.ts` tool-name fix** affects the delegation instructions in every unit with `canManageChildren`. It replaces a name that resolves to nothing with the name that does.
- **No persisted state, no data shape, no migration.** `session-freezer` already excludes this file; the orchestrator's legacy `## Auto-extracted` migration reads the bundled default dynamically rather than by content.
- **Bundle size**: vfs-root docs inline into the kernel-worker bundle verbatim, so this is a ~1.4 kB reduction. `npm run size` reports worker −0.0 kB, 68 kB under ceiling.

## Test plan

- [x] `npx prettier --check` on every changed markdown file — clean
- [x] `npm run typecheck` — clean (all 9 projects)
- [x] `npm run test` — **19,740 passed, 52 skipped, 0 failed**
- [x] `npm run build`
- [x] `npm run lint` and `npm run lint:docs` — green; gate reports `shared/CLAUDE.md is 1826/2048 bytes`
- [x] `npm run size -w @slicc/webapp` — First-load OK
- [x] 8/8 pre-push checks
- [x] **Guard inverted rather than assumed**: padded the file to 2,126 bytes → `check-doc-sizes` exits 1 with the expected error; restored to 1,826. A green gate you have not seen fail proves nothing.

No new unit test: this PR changes prose in a bundled asset and one string in a prompt template, and no test asserts either literally. `orchestrator.test.ts:2577` reads the bundled default dynamically, `upgrade-command.test.ts` writes its own content, `session-freezer.test.ts` asserts absence. The `# sliccy` H1 is preserved because `.agents/skills/cdp-smoke-test/tier1-infrastructure.md:43` documents `head -1 /shared/CLAUDE.md` as a smoke-test assertion.

**Not verified locally** — worth a reviewer's eye on a live instance: that a scoop's `GLOBAL MEMORY` block no longer claims the reader is the cone.

**Pre-existing failures**: none observed; the suite was fully green.

## Documentation

- [x] Relevant `CLAUDE.md` (developer / package architecture) — `docs/CLAUDE.md` size-budget table, `packages/vfs-root/CLAUDE.md` handoff section
- [x] `docs/` (agent reference) — `docs/dev-tools-details.md` compactor exclusion note
- [x] `packages/vfs-root/shared/CLAUDE.md` (agent-facing) — this is the change

## Risk & rollback

- **Blast radius**: all floats, but only the agent's instruction text. No runtime behavior, no persisted state, no migration. Reverts cleanly as a single commit.
- **The real risk is discovery**, not breakage: an agent that would have skimmed the dropped command list must now run `commands`, `--help`, or `skill list`. That is why the explore-before-refusing paragraph stayed and got *stronger* wording rather than being trimmed with the rest.
- **Skill descriptions are only injected when the skill is loaded** (`loadSkills` over `SKILLS_LIBRARY_DIR`). A unit with a restricted `skillsFs` loses that discovery path; the local-first `commands` / `--help` guidance is the fallback.
- **CI cannot catch a prompt-quality regression.** If delegation, sprinkle handling, or lick routing gets noticeably worse in practice, the fix is to add the specific missing invariant back — there are 222 bytes of headroom — not to restore the manual.

## Checklist

- [x] If this changes a contract used by other floats, I checked the followers/leader/offscreen paths — the file is bundled identically for every float; no float-specific handling exists

### DIFF
--- a/docs/CLAUDE.md
+++ b/docs/CLAUDE.md
@@ -23,7 +23,7 @@ This file covers the documentation surface in `docs/`.
 | ------------------------------------------------------- | ------ | ----- | ------------------------ |
 | `CLAUDE.md` (root)                                      | 15,000 | chars | enforced                 |
 | `packages/*/CLAUDE.md`                                  | 20,000 | chars | enforced                 |
-| `packages/vfs-root/shared/CLAUDE.md`                    | 3,000  | bytes | bundled into the VFS     |
+| `packages/vfs-root/shared/CLAUDE.md`                    | 2,048  | bytes | bundled into the VFS     |
 | `.github/copilot-instructions.md` + `*.instructions.md` | 4,000  | chars | Copilot truncation limit |
 
 All package `CLAUDE.md` files are now within the 20,000-char cap. The grandfathered
--- a/docs/dev-tools-details.md
+++ b/docs/dev-tools-details.md
@@ -73,7 +73,7 @@ Two details that are easy to get wrong and are therefore pinned by tests:
   (`PACKAGE_CLAUDE_MAX_CHARS`, via `lint:docs`); the compactor's policy is
   10,000 → 9,500 across _every_ tracked `CLAUDE.md`. A 12,000-char guide passes
   `lint:docs` and is still compaction work. `packages/vfs-root/shared/CLAUDE.md`
-  (3,000 **bytes**, bundled into the VFS) is excluded by construction. Sizes are
+  (2,048 **bytes**, bundled into the VFS) is excluded by construction. Sizes are
   measured with `String.length`, never bytes. Every oversized guide fans out
   as its own Claude job (largest first; `max_guides` caps N, default all); a
   `--check` miss still publishes a **partial** shard when that guide actually
--- a/packages/dev-tools/claude-md-compactor/lib.mjs
+++ b/packages/dev-tools/claude-md-compactor/lib.mjs
@@ -89,7 +89,7 @@ export function computeMaxTurns(worklist, { targetChars = COMPACTOR_TARGET_CHARS
  * Guides the compactor must never select, whatever their length.
  *
  * `packages/vfs-root/shared/CLAUDE.md` is the agent-facing runtime guide. It is
- * bundled into the VFS and budgeted at AGENT_CLAUDE_MAX_BYTES = 3,000 BYTES by
+ * bundled into the VFS and budgeted at AGENT_CLAUDE_MAX_BYTES = 2,048 BYTES by
  * `packages/dev-tools/tools/check-doc-sizes.mjs` — an order of magnitude
  * stricter than this policy, measured in bytes rather than characters, and it
  * already sits close to its cap. It can therefore never be oversized at a
--- a/packages/dev-tools/tools/check-doc-sizes.mjs
+++ b/packages/dev-tools/tools/check-doc-sizes.mjs
@@ -33,11 +33,12 @@ const Filename = fileURLToPath(import.meta.url);
 const repoRoot = resolve(dirname(Filename), '..', '..', '..');
 
 const ROOT_CLAUDE_MAX_CHARS = 15000;
-// Raised from 3000 in the markdown-media change: the file had 5 bytes of
-// headroom, and inline image/video syntax has no other discovery path (there
-// is no `--help` for markdown). Everything still in the file is a deliberate
-// instruction, so the budget moved rather than the guidance being thinned.
-const AGENT_CLAUDE_MAX_BYTES = 3300;
+// Lowered from 3300: the file is injected verbatim into every work unit's
+// system prompt, so anything the prompt assembler, the injected skill index, or
+// `man` already states is paid for on every turn. It now holds only what those
+// do not say. Headroom is deliberate — the previous cap ran at 5 bytes and had
+// to be raised to document one new capability.
+const AGENT_CLAUDE_MAX_BYTES = 2048;
 const COPILOT_INSTRUCTIONS_MAX_CHARS = 4000;
 
 const measureChars = (text) => text.length;
--- a/packages/vfs-root/CLAUDE.md
+++ b/packages/vfs-root/CLAUDE.md
@@ -117,9 +117,9 @@ apply` merge) would prompt the owner to approve the default already in force (#2
 ## External Handoffs
 
 - Mechanism: RFC 8288 `Link` response header carrying `https://www.sliccy.ai/rel/handoff` or `https://www.sliccy.ai/rel/upskill` on a main-frame document response → `navigate` lick → cone approval card.
-- Agent-facing flow: `packages/vfs-root/workspace/skills/handoff/SKILL.md` (bundled to `/workspace/skills/handoff/SKILL.md`) and the trigger line in `shared/CLAUDE.md` (bundled to `/shared/CLAUDE.md`).
+- Agent-facing flow: `packages/vfs-root/workspace/skills/handoff/SKILL.md` (bundled to `/workspace/skills/handoff/SKILL.md`); the agent reaches it through the injected skill index, not through `shared/CLAUDE.md`, which only states that handoffs are human-gated.
 - Protocol reference: `docs/slicc-handoff.md`.
-- When handoff behavior changes, keep the skill, `shared/CLAUDE.md`, and `docs/slicc-handoff.md` aligned — do not duplicate their content here.
+- When handoff behavior changes, keep the skill and `docs/slicc-handoff.md` aligned — do not duplicate their content here or in `shared/CLAUDE.md`.
 
 ## Important Distinction
 
--- a/packages/vfs-root/shared/CLAUDE.md
+++ b/packages/vfs-root/shared/CLAUDE.md
@@ -1,62 +1,33 @@
 # sliccy
 
-AI agent in SLICC, a browser-native runtime: code, automate, browse, orchestrate parallel agents.
+SLICC is a browser-native runtime for coding, browsing, automation, and parallel agents.
 
-## Vocabulary
+## Roles
 
-- **Cone**: You. Orchestrate scoops, talk to the human, full FS access.
-- **Scoops**: Isolated sub-agents (`scoop_scoop`, `feed_scoop`, `drop_scoop`, `agent` one-shot).
-- **Sprinkles**: Persistent UI panels (`.shtml`), owned by a long-lived scoop.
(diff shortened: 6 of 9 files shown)


trieloff:
Thanks — addressed the one actionable observation in 12276a872 rather than deferring it.

**Stale `delegate_to_scoop` comments**: swept all three (`scoop-message-router.ts:165,170`, `orchestrator.ts:1422`). `grep -rn delegate_to_scoop packages/webapp/src` now returns nothing, so the name is fully retired rather than surviving only in the comments someone would read it back out of. Comments only; no logic change. Verified with `npm run typecheck` (clean), `npm run lint:ci` (clean), and `npx vitest run --project webapp packages/webapp/tests/scoops/` — 2,010 passing.

**Discoverability**: no change requested, and I agree with the posture you described. Noting for the record that the 222 bytes of headroom are the intended lever if delegation or lick routing regresses in practice.

For completeness on this PR's review coverage: Copilot and Codex both declined with usage-quota messages, so this review is the only one that actually ran.

Issue #2887, opened 2026-09-04
playwright-cli response-body --filename ignores CDP base64Encoded and can UTF-8-expand saved bytes


trieloff:
## Summary

`playwright-cli response-body <index> --filename=<path>` is the download sibling of https://github.com/ai-ecoverse/slicc/issues/2878. Capture stores `Network.getResponseBody.body` and **drops `base64Encoded`**. Save then guesses from MIME: `atob` if it looks binary, else `VirtualFS.writeFile(string)` which UTF-8-encodes. A captured JPEG/zip/mp4 can be written expanded (`EF BF BD` / `C2 80`) with exit 0.

The HAR recorder next door already keeps the flag (`packages/webapp/src/cdp/har-recorder.ts` stores `responseBodyBase64 = bodyResult.base64Encoded === true`). This command does not.

## The hop

Capture (`packages/webapp/src/shell/supplemental-commands/playwright/handlers/network-requests.ts`):

```ts
interface NetworkGetResponseBodyResult {
  body?: string;
  base64Encoded?: boolean; // declared…
}
// …
entry.responseBody = r.body ?? null; // …never stored
```

Save:

```ts
const isBinary =
  entry.mimeType !== null &&
  !entry.mimeType.startsWith('text/') &&
  !entry.mimeType.includes('json') &&
  !entry.mimeType.includes('javascript') &&
  !entry.mimeType.includes('xml');

if (filename) {
  if (isBinary) {
    try {
      const binary = atob(entry.responseBody);
      // …write Uint8Array — OK when Chrome actually sent base64
    } catch {
      await fs.writeFile(filename, entry.responseBody); // STRING → TextEncoder
    }
  } else {
    await fs.writeFile(filename, entry.responseBody); // STRING → TextEncoder
  }
}
```

`VirtualFS.writeFile` of a string is `new TextEncoder().encode(content)` (`virtual-fs.ts`). Playwright-cli is wired to `VirtualFS` directly, not the latin1-preserving `VfsAdapter`.

Failure modes, all silent:

| CDP `base64Encoded` | MIME heuristic | Result |
|---|---|---|
| `true` (Chrome's usual binary) | `isBinary` | `atob` + `Uint8Array` write — OK |
| `false` (latin1 code units in `body`) | `isBinary` | `atob` throws or garbage; catch writes the string as UTF-8 → expansion |
| `true` | missing / `text/*` / `json` | writes the **base64 alphabet** as the file (wrong, not FFFD) |
[...]

Issue #2886, opened 2026-09-04
overlay CSP Fetch proxy UTF-8-encodes HTML-document POST bodies (Electron + Sliccstart)


trieloff:
## Summary

The overlay CSP-strip Fetch proxy (Electron `electron-controller.ts` and Sliccstart `OverlayTargetSession.swift`) re-issues intercepted **HTML document** requests, including POSTs, and encodes `request.postData` as UTF-8. A top-level form POST (`Accept: text/html`) whose body is not valid UTF-8 — multipart with a file, or any high byte — is forwarded expanded (`80` → `C2 80`) or, on Swift, sometimes mis-read as base64. The fulfill succeeds. The origin sees a corrupt body.

This is **not** the jsh `fetch` / `/api/fetch-proxy` path (fixed in https://github.com/ai-ecoverse/slicc/pull/2818 and https://github.com/ai-ecoverse/slicc/pull/2826). Those skip non-text Content-Types. This is the overlay iframe's CDP `Fetch.requestPaused` → Node/`URLSession` hop, and it has no Content-Type gate.

## The hop

Both implementations only intercept when `Accept` contains `text/html` (document navigations). XHR/fetch of `image/jpeg` is `Fetch.continueRequest` and is fine. A **form that navigates** (`<form method=POST enctype=multipart/form-data>`) typically *does* send `Accept: text/html`.

**Electron** (`packages/node-server/src/electron-controller.ts`):

```ts
const postData = request.postData;
// …
if (postData) {
  proxyReq.write(postData); // JS string → Node writes UTF-8
}
```

**Sliccstart** (`packages/swift-server/Sources/Browser/OverlayTargetSession.swift`):

```swift
if let postData {
    request.httpBody = Data(base64Encoded: postData) ?? postData.data(using: .utf8)
}
```

Two extra Swift failure modes: (1) latin1 bytes that happen to be valid base64 are decoded as base64, not as the POST body; (2) `data(using: .utf8)` is lossy for unpaired high bytes (nil → empty body, or replacement depending on Foundation overlay — either way not the original bytes).

Neither side calls `Fetch.getRequestPostData` (the CDP API that returns base64 for binary POST). They trust the JSON `postData` string.

## Why it must stay binary

[...]

Issue #2885, opened 2026-09-04
git stash round-trips worktree files through UTF-8, corrupting binaries


trieloff:
## Summary

`git stash` (push and pop/apply) reads every dirty worktree file with `readTextFile` (`VirtualFS.readFile({ encoding: 'utf-8' })`, non-fatal `TextDecoder`) and writes it back with `TextEncoder`. A dirty JPEG, zip, wasm, or packfile is stashed as replacement characters and restored as `EF BF BD`. The command succeeds.

This is not `git merge-file` (explicitly a text tool). Stash claims to save the worktree, including binaries git itself would treat as binary.

## The hop

Push, `packages/webapp/src/git/commands/stash.ts`:

```ts
workdirContent = await ctx.fs.readTextFile(`${cwd}/${filepath}`);
// …
const oid = await git.writeBlob({
  fs: ctx.lfs,
  dir: cwd,
  blob: new TextEncoder().encode(workdirContent),
});
```

`readTextFile` is:

```ts
async readTextFile(path: string): Promise<string> {
  const content = await this.readFile(path, { encoding: 'utf-8' });
  return content as string;
}
```

and `readFile` default/`utf-8` is `new TextDecoder('utf-8').decode(body)` with `fatal: false` → U+FFFD per invalid sequence.

Pop/apply does the same in reverse: `TextDecoder().decode(blob)` then `threeWayMerge` then `encoder.encode(mergedText)` into the worktree. A binary that survived push by accident still dies on pop.

Unmodified files take the raw-blob path (`git.writeBlob({ blob })` of the HEAD bytes) and are fine. **Only dirty binaries** are corrupted.

## Codec probe (same JPEG SOI as PR #2818)

```
in  ff d8 ff 98 00 41 7f 80 fe     (9 bytes)
readTextFile + TextEncoder.encode →
out ef bf bd × 4, 00 41 7f, ef bf bd × 2   (21 bytes)
```

## Why it matters

Silent. `git stash` / `git stash pop` exit 0. A dirty `*.png` in a repo (screenshots, fixtures, wasm, vendor zips) comes back unopenable. The user blames the image tool, not stash. Same family as composer zip-as-text (PR #1031) and fetch JPEG expansion (PR #2818): the call that should have been a byte copy went through UTF-8 and reported success.

## Suggested reproduction

```bash
# in a SLICC git repo with a dirty binary
[...]

Issue #2884, opened 2026-09-04
composer VFS image pick UTF-8-decodes bytes before the vision inline copy


trieloff:
## Summary

Picking an image from the VFS add-menu (not a drag/drop `File`, not a camera snap) reads it with `VirtualFS.readFile` **default utf-8**, then `TextEncoder().encode`s the replacement-character string into the inline vision `data` field. The original path is still attached, so the file on disk is fine. The model is handed a JPEG/PNG whose high bytes have become `EF BF BD`. Nothing errors.

This survived https://github.com/ai-ecoverse/slicc/pull/1031. That PR stopped inlining *non-image* files as `kind:'text'`. Image picks still take this hop for the vision copy.

## The hop

`packages/webapp/src/ui/wc/wc-attach.ts` `stageVfsFile`:

```ts
if (IMAGE_EXT.test(name)) {
  const reader = await openReader();
  const raw = await reader.readFile(id); // LocalVfsClient → VirtualFS, default encoding utf-8
  const bytes = typeof raw === 'string' ? new TextEncoder().encode(raw) : raw;
  const attachment = attachmentFromBytes(name, bytes); // kind:'image', data: base64 of `bytes`
  stage.add({ ...attachment, error: undefined, path: id });
  return;
}
```

`VirtualFS.readFile` with no options:

```ts
const encoding = options?.encoding ?? 'utf-8';
if (encoding === 'utf-8') return new TextDecoder('utf-8').decode(body); // fatal: false
```

Dropped `File`s are safe (`file.arrayBuffer()`). Camera/screen captures are safe (canvas data URL). Non-image VFS picks are safe (path only, no read). **Only VFS-picked `png|jpe?g|gif|webp|avif` (and svg, which is usually valid UTF-8 so it happens to survive).**

## Codec probe (same JPEG SOI as PR #2818)

```
in  ff d8 ff 98 00 41 7f 80 fe     (9 bytes)
TextDecoder('utf-8') + TextEncoder →
out ef bf bd × 4, 00 41 7f, ef bf bd × 2   (21 bytes)
```

The existing composer test (`references a VFS file pick by its existing path without inlining`) uses `notes.md`, which never enters this branch. Image drop tests use `new File([Uint8Array])`, which never hits `readFile`.

## Why it matters

[...]

PR #2901, opened 2026-09-05
feat(dev-tools): sweep for regression clusters when a release lands

trieloff:
## Summary

Adds **Regression Cluster Hunter** — a release-triggered workflow that takes one bug fix that just shipped and sweeps the tree for the *same defect shape* still living somewhere else, filing an issue per confirmed sibling.

New: `.github/workflows/regression-cluster-hunter.yml` + `packages/dev-tools/regression-cluster-hunter/` (77 unit tests). Read-only on code — `Edit`/`Write`/`gh pr create` are in `--disallowedTools`; the only writes are `gh issue create`.

## Why

Fixing one instance of a bug and leaving its siblings in the tree is this repo's most repeated failure mode. Four clusters, all real:

| Seed fix | Siblings | How they were found |
| --- | --- | --- |
| #2818 binary bodies UTF-8-corrupted | #2878, #2883, #2884, #2885, #2886, #2887 | by hand, the next day |
| #2071 `readFile` swallow clobbers persisted state | #2154, #2400, #2703 | **one per week** |
| #1996 text/binary predicate triplicated | #2821, #2822 | while auditing a different PR |
| #2166 unknown flag ignored, exit 0 | #2255, #2404, #2405, #2816, #2819, #2863, #2864, #2865, #2880 | ten separate discoveries |

None was found by a sweep. `agentic-debt-triage.yml` rotates seven *static* debt taxonomies and files **one** issue per night — which is exactly why the `readFile` family surfaced one per week for four weeks. This is the complement: it starts from a fix that actually shipped.

Same two-part house shape as the other agentic workflows: a deterministic Node selector (unit-tested in the `dev-tools` project) writes a composed prompt to `$GITHUB_OUTPUT`, then `claude-code-action` does only the judgement a grep cannot — *is this lookalike actually the same bug?*

**The evidence bar is a surviving construct** — the analogue of the flake hunter's "same commit, two outcomes". The fix deleted some code, and code of the same kind is still present in files the fix did not touch. No survivors, no dispatch.

**Two searches, because one is not enough.** Diff-derived tokens find files sharing *vocabulary*, which is not the same as sharing a *bug*. Replaying the selector over #2818 proves the gap: its tokens were all about `latin1` — the convention the fix **removed** — while its five real siblings were `TextEncoder` / `base64Encoded` sites sharing no words with it. So `shapes.mjs` is a catalog of shapes that have provably clustered here, each with its own `detect(diff)` and `probe(text)`. `probe` is a function rather than a regex list because the strongest signal is often an **absence**: `har-recorder.ts` (#2887) reads CDP's `base64Encoded` and never decodes it, so there is no wrong call to grep for, only a missing right one. Every shape carries an `evidence` field naming its real cluster, enforced by a test.

**Token extraction** reads removed lines only, in product-source hunks, skipping comments. Tokens that also appear on added lines survived the fix and are ranked *down*, not dropped.

## Cross-cutting impact

- **No runtime code changes.** Everything here is CI tooling under `packages/dev-tools/` plus one new workflow; no float, shell, VFS, or UI path is touched. `npm run typecheck` and the full suite are unaffected.
- **Trigger — the trap worth reviewing.** `on: release: [published]` **would never fire**: `release.yml` runs semantic-release with `secrets.GITHUB_TOKEN`, and GitHub suppresses workflow runs for `GITHUB_TOKEN`-created events (same anti-recursion constraint that makes `coverage-ratchet.yml` push with `BOT_PAT`). It listens for the **Release workflow** completing via `workflow_run` instead — but `Release` runs on every push to main and usually publishes nothing, so the selector confirms a `chore(release):` commit separately.
- **Cost.** Releases land ~8x/day here (~27 `fix` commits/day), so three gates bound model spend: a `MIN_INTERVAL_HOURS` cooldown (default 12), the release-actually-landed check, and the surviving-construct bar.
- **Secrets/vars**: none new. Reuses `AWS_BEARER_TOKEN_BEDROCK` like `agentic-debt-triage.yml`. Optional vars `CLUSTER_BEDROCK_MODEL`, `CLUSTER_MIN_INTERVAL_HOURS`, `CLUSTER_MAX_ISSUES` all have defaults.
- **State** stays GitHub-native, like the rest of the family: previous successful runs are the cooldown clock, and the `swept-fix:N` HTML marker on filed issues is the dedup key. No state file, branch, or Actions cache.
- **Third-party actions** pinned to immutable commit SHAs, matching `agentic-debt-triage.yml`.

## Test plan

- [x] `npm run lint` — exit 0
- [x] `npx vitest run --project dev-tools` — **1424 passed / 0 failed** (77 of them new)
- [x] `node packages/dev-tools/tools/check-touched-exemptions.mjs origin/main` — boy-scout gate clean
- [x] `check-doc-refs` / `check-doc-sizes` — clean
- [x] **New behavior is covered by unit tests**: `npx vitest run --project dev-tools packages/dev-tools/regression-cluster-hunter/`
- [x] N/A for UI / screencast — no UI surface
- [x] Builds not run: no `src/` in any bundled package is touched

**Live end-to-end run** of the selector step, with the workflow's exact env (real `GITHUB_OUTPUT` / `GITHUB_STEP_SUMMARY`, non-dry-run):

```
Release 6.131.0: sweeping #2888 — fix(webapp): keep playwright-cli upload binary files byte-exact.
Its construct survives in 57 untouched file(s). Matched known shape(s): binary-encoding-hop.
has_candidate=true  pr=2888  version=6.131.0
```

It picked up the live UTF-8 cluster (#2893/#2895/#2896/#2897 landed during development) and rejected five other fixes for having too few survivors.

**What that run caught, and the second commit fixes:** the selector proposed this hunter's *own* `shapes.test.mjs` as a suspect. `NON_PRODUCT_PATH` matched a `tests/` directory but `packages/dev-tools` colocates suites as `lib.test.mjs`. The second-order damage mattered more — tests in the corpus inflated the per-token file counts used to reject over-common tokens, so good signatures were silently discarded. Filtering at load took the corpus 3324 → 1861 files and moved #2888 from 23 → 57 surviving sites. `isProductSource` is now exported and shared by all three call sites, so they cannot drift — which is, with some irony, the #1996/#2821/#2822 shape this workflow exists to hunt.

**Cannot be tested before merge:** `workflow_dispatch` returns HTTP 404 from a topic branch — GitHub only registers dispatch triggers from the default branch. The `claude-code-action` step therefore has not executed. Suggested first run after merge: `workflow_dispatch` with `dry_run: true`.

**Pre-existing failures:** none. (`check-touched-exemptions.test.mjs` failed against the older `main` before I rebased; it passes on this branch and on current `main`.)

## Documentation

- [x] `packages/dev-tools/CLAUDE.md` — one-line entry in the tooling map
- [x] `docs/dev-tools-details.md` — new `regression-cluster-hunter` section
- [x] `packages/dev-tools/regression-cluster-hunter/README.md` — full notes, incl. **measured recall** and the tuning knobs
- [x] No `README.md` / agent-skill changes needed — no user-facing command or agent surface changed

## Risk & rollback

- **Blast radius**: CI only. Worst realistic case is noisy issues, capped at `CLUSTER_MAX_ISSUES` (default 5) per run behind a 12h cooldown. It cannot push, branch, or open PRs.
- **Rollback**: revert cleanly, or disable the workflow in the Actions UI. No persisted state, no migration.
- **Measured recall is documented, not oversold**: over #2818 the selector recovers two of five known siblings within a 25-file cap (ranking #2887's `har-recorder.ts` first); over #2888 it ranks #2883's `mouse.ts` first. The brief explicitly tells Claude the table is a lead rather than a finding and to grep for the *concept*. Raising the caps trades a reviewable shortlist for a grep dump.
- **What CI cannot catch**: whether Claude's filed issues are worth a human's time. That needs one real run to judge — please review the first batch before letting it run unattended.

## Checklist

- [x] Public API / shell-command changes: none
- [x] Cross-float contracts: none touched

Generated with [Claude Code](https://claude.com/claude-code)


trieloff:
## Review findings addressed — `43273ec5c`

All five, plus a structural test for the class of bug behind one of them.

**P1 / Minor — cooldown clock reset on quiet runs.** Correct, and worse than "cuts into recall": it was fatal. `workflow_run` fires on every push to main and a quiet run succeeds too, so with pushes far more frequent than 12h the clock reset several times a day and the hunter would **never have dispatched again after its first run**. The recommended `dry_run: true` smoke test would have blocked the next real release too.

The clock is now the `regression-cluster-dispatch` artifact, uploaded only by a run that actually dispatched — so the ledger can't drift from what happened, and it stays GitHub-native (no state file/branch/cache). The sweep window now also starts from the last *dispatch*, capped at `MAX_WINDOW_HOURS` (72) so a long quiet spell can't widen it without bound.

**Major — `min_interval_hours: 0` didn't disable the cooldown.** Confirmed; the repro was right. Fixed with a `min` argument on `envInt` rather than a blanket `>= 0`, since the reviewer's own caveat applies: `MAX_ISSUES=0` and `WINDOW_HOURS=0` are meaningless rather than useful overrides, so only `MIN_INTERVAL_HOURS` passes `0`. Blank still falls back (`parseInt('')` is `NaN`).

**P2 — `cross-runtime-predicate` could never clear the floor.** Correct, and it was live: my own end-to-end run printed `cross-runtime-predicate: 0 files` and I didn't chase it. The shape's probe tops out at 2 against a default floor of 3, so it rendered an empty list every run — silently disabling the parity check for the very shape that produced #2821 and #2822. Shapes can now declare their own floor (`minHits: 2` here), and a new structural test asserts **every** shape can reach its own. I mutation-tested that guard: reverting the fix makes it fail with `cross-runtime-predicate can never reach its floor of 3`.

**P2 — PR files read one page deep.** Correct. A fix touching >100 files had its later-page repairs missing from `fixedFiles`, so the sweep could hand Claude a file that very fix had already repaired. Now paginated, capped at 10 pages.

**P2 — comparator contradicted the documented invariant.** Correct that code and docs disagreed. Before changing it I measured, because the ordering had been deliberately flipped earlier for recall. Against #2818's four known siblings, with the corpus as it now stands:

| Ordering | mouse | stash | har-recorder | network-requests | recall @25 |
| --- | --- | --- | --- | --- | --- |
| `precise → hits → incidental` (was) | 21 | 52 | 1 | 71 | **2/4** |
| `precise → incidental → hits` (docs) | 64 | 21 | 1 | 40 | **2/4** |
| 3 weighted blends | — | — | 1 | — | 2/4 and 1/4 |

Every variant recovers exactly 2 of 4; they only swap *which* two. Given the tie I took your ordering — it follows the recorded evidence that every real sibling was a file whose job was something else that happened to touch bytes. The A/B is now recorded in the comparator comment and the README so it isn't re-litigated: **2/4 is the static-probe ceiling, not a tuning gap**, which is exactly why the brief tells the model the table is a lead rather than a finding.

### Verification

`npm run lint` exit 0 · `npm run typecheck` exit 0 · `npx vitest run --project dev-tools` **1429 passed / 0 failed** (82 in this directory) · doc-refs clean.

Re-ran the selector live against the real API after the changes: artifact-based clock resolves with no prior dispatch, `MIN_INTERVAL_HOURS=0` now yields `0` (was silently `12`), `MAX_ISSUES=0` still correctly falls back to `5`.

The `claude-code-action` half still cannot execute until this is on the default branch (`workflow_dispatch` 404s from a topic branch), so the first post-merge run should be `dry_run: true`.

Issue #2883, opened 2026-09-04
playwright-cli drop --path corrupts binary files the same way as upload (UTF-8 → EF BF BD)


trieloff:
## Summary

`playwright-cli drop --path=<vfs-file>` reads the VFS file as **text** and re-encodes it with `TextEncoder`, so every byte ≥ `0x80` becomes U+FFFD (`EF BF BD`). The drop "succeeds". The `DataTransfer` `File` that lands on the page is a corrupt JPEG / zip / mp4.

This is the same hop as https://github.com/ai-ecoverse/slicc/issues/2878 (`playwright-cli upload`), in a different command. **Do not fold this into #2878** — upload is already dispatched; `drop` will still be broken after that lands unless this path is fixed too.

Runtime: current `main` (`packages/webapp/src/shell/supplemental-commands/playwright/handlers/mouse.ts`).

## The hop

```ts
const content = await fs.readFile(vfsPath); // VirtualFS default encoding is utf-8
const bytes = typeof content === 'string' ? new TextEncoder().encode(content) : content;
files.push({ name, type, base64: uint8ToBase64(bytes) });
```

`VirtualFS.readFile` with no options is:

```ts
const encoding = options?.encoding ?? 'utf-8';
if (encoding === 'utf-8') return new TextDecoder('utf-8').decode(body); // fatal: false → U+FFFD
```

Then the page-side injector does `new File([bytes], …)` from that base64. Call reports success.

This is **not** the just-bash `VfsAdapter` (latin1-preserving). Playwright handlers talk to `VirtualFS` directly, which is why `curl` / `ffmpeg` / `zip` are fine and this is not.

## Codec probe (same 9-byte JPEG SOI as PR #2818)

```
in  ff d8 ff 98 00 41 7f 80 fe     (9 bytes)
out ef bf bd × 4, then 00 41 7f, then ef bf bd × 2   (21 bytes)
```

Six high bytes → six U+FFFD. ASCII `00 41 7f` survives. That is exactly the #2878 arithmetic.

The existing test (`drop with --path reads file from VFS`) stubs `readFile` to return the string `'file-bytes'` and never checks byte identity.

## Why it matters

Silent. `drop` is the path for hidden file inputs and drag-and-drop upload widgets that `upload` (focused `<input type=file>`) cannot hit. A 0-exit drop of an `.mp4` / `.zip` that the receiving page then probes as garbage is the same failure mode as #2878.

## Suggested reproduction

[...]

Issue #2878, opened 2026-09-04
playwright-cli upload appears to corrupt binary files (high bytes become EF BF BD) — hunch, needs investigation


trieloff:
## Summary

`playwright-cli upload` appears to mangle binary files: the bytes that arrive at the far end are not the bytes on disk, and the corruption pattern looks like a UTF-8 text decode. **Filing this as a hunch — I have the symptom measured, but I have NOT confirmed the mechanism and have not built a controlled reproduction. Please treat the cause below as a hypothesis and investigate.**

Runtime: `6.129.0-mtmq2fae`.

## What was measured

While driving a real TikTok Studio upload through a file input:

| | bytes |
|---|---|
| source file on VFS (`.mp4`, h264, 1080x1920, 29.93s) | 24,349,008 |
| what the server received | 45,152,034 |

Ratio ~1.854. Sampling the received payload showed runs of `EF BF BD` (the UTF-8 replacement character, U+FFFD) where the original had bytes >= 0x80.

## Why it matters: the failure is silent

Nothing errored. The upload "succeeded", and then:

- the receiving service probed the video as **480x360** instead of 1080x1920
- transcoding never completed
- the UI's Post button stayed permanently disabled, with no diagnostic

Two attempts were burned before the byte counts were compared. A corrupting upload that reports success is worse than one that throws, because the symptom surfaces far downstream (or, in this case, as a vendor-side transcode that just never finishes).

## Hypothesis (UNVERIFIED)

The VFS file is read as **text** (UTF-8) rather than as bytes somewhere in the upload path, so every byte sequence that is not valid UTF-8 is replaced with U+FFFD = `EF BF BD`, expanding 1 byte to 3.

A pure 1->3 expansion of every high byte would predict a ratio near 2.0 for roughly half-high-byte data; the observed 1.854 is consistent with *some* incidental byte pairs forming valid UTF-8 and passing through unchanged. That is suggestive rather than conclusive.

**What I did not do:** no minimal reproduction, and no source confirmation — my local checkout was at 6.112.4 against a 6.129.0 runtime, so its lack of a matching code path proves nothing. I did not want to assert a mechanism I had only read out of stale source.

## Suggested reproduction (cheap, deterministic)

1. Serve a trivial page containing `<input type="file">` plus a handler that reports `file.size` and a hash.
2. Create a 256-byte file containing every byte value `0x00`..`0xFF` exactly once.
3. `playwright-cli upload` it and compare the reported size against 256.

[...]

PR #2906, opened 2026-09-05
fix(dev-tools): keep the cluster sweep in one session

trieloff:
## Summary

Follow-up to #2901. The first live run of the merged workflow ([33960038570](https://github.com/ai-ecoverse/slicc/actions/runs/33960038570)) went **green and filed nothing**. This makes the sweep run in one session.

## Why

Claude fanned the sweep out to one sub-agent per package, then ended its turn with *"All three remaining agents are still actively reading… I'll wait for their final reports rather than poll further."* Nothing waits — the session terminates when Claude stops. Result: 34 turns, **$7.27** spent almost entirely on fan-out cache reads (3.7M cache-read tokens), **zero issues filed**, `conclusion: success`, and nothing in the digest saying anything went wrong. Its own final message shows it had already *cleared* webapp and dev-tools and was mid-verdict on three more packages — findings that died with the session.

Worth being precise about the failure mode: the run is indistinguishable from the legitimate "filing nothing is a good outcome" path unless you read the log. That is the part that makes it worth fixing rather than tolerating.

## Approach

- **`Task` added to `--disallowedTools`.** The previous `--allowedTools` list didn't include it, but that only restricts *some* surfaces; the fan-out happened anyway. Explicit denial is the fix.
- **The brief carries the contract twice**, because the model has to internalise it before it starts planning:
  - *One session, no delegation* — with the failure narrated, so the instruction has a reason attached rather than being a bare prohibition.
  - *File as you confirm, never at the end* — the moment a candidate passes the rule, `gh issue create` and move on. A session that ends mid-batch loses everything it found.
  - *Budget your turns* — stop investigating, file what's confirmed, print what you didn't reach. A partial sweep that files two real issues beats a thorough one that files none.

## Cross-cutting impact

- CI tooling only; no runtime code, same as #2901.
- The `$7.27` was on `us.anthropic.claude-opus-4-8` (`vars.RUM_BEDROCK_MODEL`), not the `claude-sonnet-4-6` default in the workflow's fallback chain. Removing the fan-out should cut the dominant cost (cache reads), but **the per-run cost is still worth watching** — see below.

## Test plan

- [x] `npm run lint` — exit 0
- [x] `npx vitest run --project dev-tools` — **1433 passed / 0 failed**
- [x] New test asserts the brief carries both halves of the contract
- [x] N/A for UI

Still unproven: that a non-delegating run actually files good issues. That needs another live dispatch after this merges, which is the next step.

## Documentation

- [x] `packages/dev-tools/regression-cluster-hunter/README.md` — new "The sweep runs in one session" section with the run ID and cost
- [x] Workflow header comment explains why `Task` is denied

## Risk & rollback

- Blast radius: CI only. Revert cleanly.
- **Open question for the reviewer**: at ~$7/run on opus-4-8 with a 12h cooldown, this is roughly $14/day if it dispatches every window. If the non-delegating run doesn't come down substantially, the honest options are pinning `CLUSTER_BEDROCK_MODEL` to a cheaper model, raising `CLUSTER_MIN_INTERVAL_HOURS`, or tightening the candidate caps. I'd rather surface that now than let it run up quietly.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/regression-cluster-hunter.yml
+++ b/.github/workflows/regression-cluster-hunter.yml
@@ -31,6 +31,12 @@ name: Regression Cluster Hunter
 #   2. claude-code-action reads that fix's diff, states the defect as a rule,
 #      tests each candidate against it, and files an issue per confirmed sibling.
 #
+# `Task` is in --disallowedTools deliberately. On the first live run Claude fanned
+# the sweep out to one sub-agent per package, then ended its turn with "I'll wait
+# for their final reports" — but nothing waits, the session simply terminates.
+# The run went green, filed zero issues, and cost $7.27 in fan-out cache reads.
+# One session, working sequentially and filing as it confirms, is the contract.
+#
 # TRIGGER — why `workflow_run` and not `on: release`. semantic-release publishes
 # the GitHub Release with `secrets.GITHUB_TOKEN` (see release.yml), and GitHub
 # does not start workflow runs from events created by GITHUB_TOKEN. An
@@ -199,6 +205,6 @@ jobs:
           claude_args: |
             --model ${{ vars.CLUSTER_BEDROCK_MODEL || vars.RUM_BEDROCK_MODEL || 'global.anthropic.claude-sonnet-4-6' }}
             --allowedTools "Read,Grep,Glob,Bash(gh issue:*),Bash(gh pr:*),Bash(gh search:*),Bash(git:*)"
-            --disallowedTools "Bash(gh pr create:*),Bash(gh pr merge:*),Bash(gh issue close:*),Edit,Write"
+            --disallowedTools "Task,Bash(gh pr create:*),Bash(gh pr merge:*),Bash(gh issue close:*),Edit,Write"
             --max-turns 200
           prompt: ${{ steps.scan.outputs.prompt }}
--- a/packages/dev-tools/regression-cluster-hunter/README.md
+++ b/packages/dev-tools/regression-cluster-hunter/README.md
@@ -108,6 +108,17 @@ This is exactly why the brief tells Claude the table is a lead rather than a
 finding, and instructs it to grep for the _concept_. Do not raise the caps to
 chase the last siblings — that trades a reviewable shortlist for a grep dump.
 
+## The sweep runs in one session
+
+`Task` is in `--disallowedTools`, and the brief says so twice. On the first
+live run (`33960038570`) Claude fanned the sweep out to one sub-agent per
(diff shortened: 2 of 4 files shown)


trieloff:
## Verified live after merge

Run [33961543009](https://github.com/ai-ecoverse/slicc/actions/runs/33961543009), a forced sweep of #2888 (`pr: 2888`, `min_interval_hours: 0`).

**The fix holds.** `subagent_type` occurrences in the log: **0** (was 293). The sweep ran sequentially in one session for 68 turns — twice the 34 it managed before dying mid-fan-out — and produced a complete verdict instead of stopping to wait for reports that never come.

| | before (33960038570) | after (33961543009) |
| --- | --- | --- |
| delegation | 293 `subagent_type` | **0** |
| turns | 34, ended mid-work | 68, completed |
| cache-read tokens | 3.69M | 3.18M |
| cost | $7.27 | **$3.88** |
| outcome | nothing, no explanation | full verdict + audit trail |

**It filed nothing, and that is the correct answer here.** It stated the rule crisply (precondition / wrong behavior / observable damage), then tested ~20 candidates and rejected each with a one-line reason — `rsync-command.ts` reads `{ encoding: 'binary' }`; `open-command.ts` uses byte-preserving latin1 `String.fromCharCode`→`btoa`; `FsClient.swift` has a correct `readBinaryFile` twin; and so on. Its own summary explains why: the webapp shell "has been swept 6+ times (#2818, #2878, #2883–2887)" and is now uniformly defended. The binary-encoding cluster really is exhausted — we just finished fixing it.

So this is the documented *"filing nothing is a good outcome"* path working, now distinguishable from the silent failure it used to look identical to.

### Also verified in production

- The **`workflow_run` trigger fires on its own** — run `33960961800` was triggered by a real Release completion, not a manual dispatch.
- The **dispatch-only cooldown clock works**: a later run's window started at `10:14`, the dispatch artifact's timestamp, not at the several quiet runs in between. Dry runs correctly skip the marker step.
- The **release gate works**: two runs exited quietly with *"No `chore(release):` commit in the window"*, which is the `workflow_run`-is-not-proof case from #2901.

### Open, for a human call

At **$3.88/run on `claude-opus-4-8`** (via `vars.RUM_BEDROCK_MODEL`) with a 12h cooldown, this is ~$8/day if it dispatches every window. Halving it was worth doing, but if that's more than the sweep is worth, the levers are `CLUSTER_BEDROCK_MODEL` (pin something cheaper), `CLUSTER_MIN_INTERVAL_HOURS` (raise), or the candidate caps. Flagging rather than deciding.

PR #2912, opened 2026-09-06
fix(webapp): generate MCP aliases against the child_process bridge

trieloff:
## Summary

`mcp add <url> <name>` auto-writes a `.jsh` alias shim at
`/workspace/.mcp/aliases/<name>.jsh`. That shim called a bare `await exec(cmd)`,
but the JS realm has no top-level `exec` — so **every generated alias died with
`ReferenceError: exec is not defined`**. Before: running the alias printed a
realm stack trace. After: the alias forwards to `mcp invoke` and returns the
tool's output.

`mcp invoke <name> <tool>` was never affected — only the alias ergonomic that
`mcp add` advertises on success.

## Why

Found while wiring a local adb MCP server into SLICC to drive an Android phone
from the shell.

**Repro**

1. Start any MCP server, e.g. on `http://localhost:5799/mcp`
2. `mcp add http://localhost:5799/mcp adb` → reports `alias: /workspace/.mcp/aliases/adb.jsh`
3. `adb`

Expected: forwards to `mcp invoke adb …` and prints the tool result.
Actual:

```
ReferenceError: exec is not defined
    at eval (eval at Vo (…/js-realm-worker-DopctPNd.js:38:25466), <anonymous>:13:11)
```

Confirmed directly in the realm: `node -e "console.log(typeof exec, typeof require)"`
prints `undefined function`.

`docs/…/jsh-runtime-extensions.md` already documents the supported path —
`require('child_process')` resolving to a shim over `exec.start`, with
`util.promisify` support. The shim now uses that.

One non-obvious consequence, and the reason this isn't a one-line swap: the
`child_process` bridge follows Node semantics and **rejects** on a non-zero
exit, with the captured output and status attached to the error
(`node-child-process.ts` → `reject(Object.assign(error, { stdout, stderr }))`,
`code` set from the child's exit). The old `await exec(cmd)` shape resolved a
`{ stdout, stderr, exitCode }` record instead. A naive port would have turned
every failing MCP tool call into an unhandled rejection, so the shim wraps the
call and rebuilds `stdout` / `stderr` / `exitCode` from the rejection.

## Cross-cutting impact

- **Floats exercised**: standalone (wrangler + node-server thin-bridge). The
  generated shim is plain `.jsh` text and runs through the same realm in every
  float, so extension / Electron behavior is unchanged by construction.
- **Other callers**: `aliasContent` has exactly two callers — `writeAliasShim`
  (on `mcp add`) and the test suite. `removeAliasShim` matches on path, not
  content, so `mcp delete` is unaffected.
- **Persisted state**: none. `servers.json` is untouched; only the generated
  `.jsh` body changes. Aliases written by an older build keep the broken
  content until the server is re-added — re-running `mcp add <url> <name>`
  overwrites it.
- **Security**: no change to argument escaping — the `escape()` helper is
  byte-for-byte the same.

## Test plan

- [x] `npx prettier --check` on both changed files — clean
- [x] `npx biome check` on both changed files — clean
- [x] `npx vitest run packages/webapp/tests/shell/supplemental-commands/mcp-command.test.ts packages/webapp/tests/shell/jsh-discovery.test.ts` — **106 passing**
- [x] New behavior covered by unit tests in `mcp-command.test.ts`: the existing
      `aliasContent` assertions were updated (they pinned `'await exec(cmd)'`,
      so they had encoded the bug), plus two new guards — one rejecting a bare
      `exec(` global, one pinning the failure-path forwarding.
- [x] `node packages/dev-tools/tools/check-layer-back-edges.mjs` — ok
- [x] `node packages/dev-tools/tools/check-record-string-unknown.mjs` — ok
- [x] **Manual smoke (standalone harness)**: `PORT=5715 WRANGLER_PORT=8788 npm run dev:standalone:fresh`,
      registered a real adb MCP server, wrote this exact generated shim into the
      VFS, and exercised both paths against a physically attached device:

  ```
  / $ adb adb_devices; echo "exit=$?"
  List of devices attached
  ZY22MP4VVM   device  product:naples_g_syse  model:moto_g67  device:naples
  exit=0

  / $ adb nope_tool > /dev/null 2>&1; echo "exit=$?"
  exit=1
  ```

  The failure case forwards `mcp invoke`'s stderr and propagates exit 1 rather
  than throwing — that's the rejection handling doing its job.
- [ ] `npm run test` (full) / `npm run build` — not run; change is confined to a
      generated string and its tests.
- [ ] UI change — N/A, no UI surface touched.

`npx tsc --noEmit -p tsconfig.json` reports **232 errors on a clean
`origin/main` at 9d1f3b941, and the identical 232 with this branch applied** —
none in `mcp-command.ts` or its test. Mostly `work-unit` test fixtures
(`TraySudoKind`, `ScoopSummary.isCone`).

## Documentation

- [x] No documentation changes needed — `docs/shell-reference.md` documents
      *that* `mcp add` writes the shim, not its contents.

## Risk & rollback

Blast radius is one generated string. Reverts cleanly; no migration. Worst case
is a return to today's behavior, where the alias is broken anyway.

Reviewer note: the thing CI can't catch is the realm's actual global surface.
The regression guard asserts on generated text, not on execution — I verified
execution by hand in the harness (above), since there's no existing test rig
that runs a generated `.jsh` inside a live realm.


trieloff:
Correction to the **Pre-existing failures** note in the PR description — that claim was wrong, and I'd rather flag it than let a reviewer act on it.

I wrote that `npx tsc --noEmit -p tsconfig.json` reports 232 errors on a clean `origin/main`. It does not. `main` typechecks **clean**. Those 232 errors were an artifact of my throwaway worktree, whose `node_modules` was symlinked from another checkout with only some per-package `node_modules` present — module resolution failed and produced spurious errors. Run in a normal checkout, the same command returns **0 errors**.

Re-verified properly: applying this branch's `mcp-command.ts` in a full checkout also typechecks clean. (One error initially appeared and was likewise my fault — I'd copied the file onto a branch where `resolveMcpRedirectUri` still has the zero-arg signature, so main's `resolveMcpRedirectUri(resolveFloatTopology())` call didn't match. Cross-branch mixing, not a defect in this change.)

Nothing else in the description changes: the 106 passing tests, the lint/ratchet gates, and the live harness verification all stand as written.

PR #2924, opened 2026-09-07
fix(webapp): retry invalidated ZenFS File snapshots

trieloff:
## Summary

A concurrent OPFS overwrite can invalidate the File snapshot ZenFS is about to read and reject startup with `NotReadableError`. Patch the pinned `@zenfs/dom` to reacquire a fresh File and retry that byte read up to three total attempts.

## Why

Upstream: https://github.com/zen-fs/dom/issues/46 (full inline reproduction).

One six-byte file changed from `before` to `after!` reproduces the failure on dom 1.2.13 / core 2.7.2 as well as our pinned versions. The browser fixture schedules a real native overwrite after `getFile()` and before `arrayBuffer()`; it supplies no fake File or exception. Limiting concurrency to one does not prevent this interleaving.

Run `packages/webapp/tests/e2e/zenfs-opfs-read-race/README.md`. The once-overwrite case should mount successfully and read `after!` from the synchronous cache. Set `always: true` to confirm that repeatedly invalidated snapshots still fail after three attempts.

Related pull request: #2923 (preload concurrency). The two patches can merge independently.

Only native `DOMException` byte-read failures named `NotReadableError` are retried. Handle lookup and snapshot acquisition stay outside the retry catch. Permission, not-found, abort, and non-native errors propagate immediately; exhaustion preserves the final native error.

This shared backend patch applies wherever the hosted CLI, extension, and other consumers use WebAccess. It does not provide an atomic multi-file snapshot or repair concurrent file-size changes. The patch manifest includes the upstream reference and removal criteria, plus the existing empty-file issue's reference.

## Verification

- `npm run verify`, touched-file debt gate, and `npm run typecheck` pass.
- `npm test`: 20,325 passing tests; `npm run test:coverage` passes.
- `npm run build`, `npm run build -w @slicc/chrome-extension`, and `npm run bundle-size` pass.
- Nine unit cases cover stable reads, two invalidations followed by recovery, bounded exhaustion, and errors that must not be retried. The recovery and exhaustion guards fail without the patch.
- Chrome for Testing 151.0.7922.34, macOS 26.5.2 arm64: the patch independently recovers the once-overwrite fixture to `after!`, and the always-overwrite control fails after exactly three attempts. Earlier repeated runs gave 20/20 expected outcomes for each case.

## Risk & rollback

An unreadable snapshot can incur two extra reads before its error surfaces. The patch does not reset local data or change its format. Revert this PR to remove the snapshot retry hunk.


trieloff:
Checked both non-blocking notes:

- The instruction file is **3,997 characters** by the JavaScript string-length measure used in `check-doc-sizes.mjs`, and **4,035 UTF-8 bytes**. It passes the enforced 4,000-character gate.
- A merge simulation against `origin/main` at `42b206ac1` completes without conflicts and retains release version **6.134.1**. This PR changes neither `package.json` nor `package-lock.json`.

No code changes are needed for these notes. The snapshot retry's existing unit, browser, and full local verification results still apply to the reviewed commit.

PR #2929, opened 2026-09-07
fix(webapp): drop keyboard mode while a frame holds the focus

trieloff:


trieloff:
Both Codex P2s addressed in `2cbb885`.

**Restore a manually entered mode after leaving a frame** — right, and the same defect was already there for window blur: under `trigger: "esc"` the mode never auto-enters, so `applyTriggerSettle()` does nothing for a focus that is merely not typing, and the clear was permanent. A suspension now RECORDS that the mode was up (`settler.suspend()`, used by both the frame gate and `onWindowBlur`), `settle` puts it back when the keyboard returns, and a live `setTrigger` drops the record (`dropSuspension()`) — a switch decides the mode and must not be undone by a stale suspension. The inline `focusin` drop for a frame also routes through `suspend()`, otherwise the mode was already off by the time `settle` ran and there was nothing left to remember.

**Re-settle when a focused frame is detached** — also right. A `MutationObserver` on `document.documentElement` (childList + subtree), armed only while a frame holds the focus and disconnected the moment it is gone; the callback is a single `frame.isConnected` test, then a `schedule()`. Also released in `dispose()`.

Tests: frame removal takes the mode back, an Escape-entered mode survives both a frame and a window round trip, and a switch to `esc` is not undone by a pending suspension. Verification re-run: `npm run verify` (8/8), `check-touched-exemptions`, `typecheck`, `test:coverage:webapp` (no floor breach), `build -w @slicc/webapp`; all 73 `tests/ui/wc` files pass.

PR #2948, opened 2026-09-07
feat: unread indicator for scoop tabs

trieloff:
## Unread indicator for scoop tabs

With multi-cone standard, the strip is a place you leave things running — so a tab has to be able to say it has news. A scoop that spoke while you were looking elsewhere now wears a dot in its own hue, notched onto the top-right of its status ring.

Three treatments were drafted in Storybook and reviewed (segment-corner dot, count pill, bold-label emphasis); the glyph-anchored dot won and the other two are gone, along with the temporary `unread-style` / `unread-dot` attributes that held the comparison.

### Why on the glyph rather than in the segment corner

- Costs the segment no padding, so it pushes no tab into the overflow.
- Cannot drift as a label changes length, and cannot be clipped by the track's `overflow: hidden` — the corner dot was clipped on the last tab at narrow widths.
- Binds "unread" to the mark that already carries this agent's state: one place to look per tab.

Hue is the agent's own even on a broken tab, whose ring stays red — state is the ring's job, identity is the dot's. The halo takes `--canvas` on the selected segment and `--ghost` elsewhere, so it matches whichever surface the tab sits on in both themes. Nothing animates; the rotating fullness arc already owns motion here.

### Rules

- **The selected tab never shows it.** The tab you are reading is read.
- **A collapsed tab is not silent.** Its count reaches `<slicc-scoop-overflow>`, which repeats the dot per row and marks its trigger once in ink (that trigger stands for several agents, so it carries no single hue). Both aria labels carry the count.
- **Nothing persists.** Unread is per-page-session, the way an unseen streamed reply is.

### Where the count comes from

`work-unit/client/unread.ts`. `UnreadLedger.sync(units, selectedId)` folds a roster plus the selection into per-unit counts, and `toTabDescriptors` puts them on the descriptor.

The signal is the **roster**, not a message stream — which is what lets both floats share one implementation. A leader could count `turn_end` per unit, but a follower subscribes only to the transcript of the unit it is showing and would have nothing to count for the rest. Both sides see every unit's presentation state, so one increment is one `working` unit that stops working. **A follower therefore needs no new tray-wire field**, and a leader too old to send `state` simply presents no transition to count.

It counts transitions, not states: a repeated roster push adds nothing, and a first roster opens with every tab clean.

### One fix that came out of the wiring

The leader's 15 s stats poll was a second strip publisher, projecting descriptors straight from the records. It dropped everything the real publisher adds on top of the roster and wiped every unread dot on each tick, so it now repaints through `refreshScoops` like every other path.

## Testing

- 13 new component tests (`slicc-agent-tabs`, `slicc-scoop-overflow`): suppression on the selected tab, non-finite/negative counts, hue-over-broken-ring, the surface-matched halo, the aria wording, and the hand-off to the overflow.
- 7 ledger tests plus leader (`wc-live.test.ts`) and follower (`wc-tray-unread.test.ts`) integration tests, so the parity is pinned on both floats.
- Full pass green: `verify`, `check-touched-exemptions`, `typecheck`, `test` (20,370), `test:coverage`, both builds, `bundle-size`, and the 2,290 webcomponents browser tests.

## Stories

`Switcher/AgentTabs` → **Unread**, **UnreadOnSelected**, **UnreadNarrow360**; `Switcher/ScoopOverflow` → **Unread**, **UnreadClosed**.


trieloff:
Storybook, light theme — the shipped treatment: hue dots on the strip, the overflow trigger's ink dot at 360px, and the per-scoop dots repeated on the overflow rows. Stories: `Switcher/AgentTabs` → Unread / UnreadOnSelected / UnreadNarrow360.

trieloff:
Fixed in `cee5ec8` — you were right that sampled state cannot answer this.

`ScoopSummary.turns` is now a per-unit count of completed turns, monotonic since the leader's page loaded. The leader increments it in `onStatusChange` (at the event, because everything downstream samples), projects it through `recordToWorkUnitSummary` → `toScoopSummaries`, and the follower reads it back in `summaryToWorkUnit`. `UnreadLedger` treats it as a **version**: it keeps the value it last saw per unit and counts the increase, so a frame the 50 ms coalescing window swallowed still leaves its increment in the next one. A decrease means the leader reloaded and restarted at zero, which re-baselines instead of counting backwards. Where the field is absent (an older leader) the ledger keeps watching `state` leave `working`, exactly as before.

It deliberately carries the leader's **turns**, not the leader's unread count: read state is per-device, so each side counts from its own baseline and clears on its own selection.

New tests: a follower roster that only ever reports `idle` but bumps the counter still dots the tab (`wc-tray-unread.test.ts`), a burst of turns collapsed into one frame counts all of them, a first sighting at `turns: 7` opens clean, a reloaded leader re-baselines, an aborted turn (transition without an increment) counts nothing, and the wire round-trip is pinned in `wc-tray-scoops.test.ts`. iOS shows no unread indicator and has no field for it, so the corpus classifies `turns` as `dropped` — the `addedAt` arrangement.

PR #2949, opened 2026-09-07
fix(secrets): bound the persisted-store read so `secret list` cannot hang

trieloff:
## Summary

`secret list` could hang forever on Sliccstart. `GET /api/secrets` read the Keychain on every request, and that read is synchronous and uncancellable — when the item's ACL does not trust the running binary it sits on an "allow access" dialog that a request from a browser tab has nobody to answer. This bounds the read, and reports what went wrong instead of hanging or answering with an empty list.

## Why + Repro

The trigger is the recurring-prompt condition already documented in `docs/swift-server-details.md`: the `ai.sliccy.slicc / __envfile__` item trusts one cdhash, and every rebuild produces a new one.

Nothing bounded the wait at any layer — `SecItemCopyMatching` cannot be cancelled, Hummingbird sets no `idleTimeout`, and the webapp's secret backend called `fetch` with no signal.

Reproduced with a store that stalls on demand (no dialog needed), which is now the regression test:

```
PROBE masked route answered in 0.0013s   ← while list was stalled
PROBE list finished within 6s: false     ← nothing bounded the request path
```

`secret get` / `printenv` / `secret test` were unaffected because they read the in-memory masked snapshot — which matches the reports. Two further honest notes:

- **node-server is not affected.** Its `GET /api/secrets` does `readFileSync`: no dialog, no hang. This is Sliccstart-only.
- **#2944 does not fix this.** That suppression is conditional on `SLICC_KEYCHAIN_NONINTERACTIVE=1`, which is off in a normal Sliccstart run.

## Approach

Server (`packages/swift-server`):

- New `BoundedStoreCall` runs a store call on a Dispatch queue — not a cooperative thread, which blocking would starve — and races a 5 s deadline.
- Every request-path store touch goes through it: list, set, delete (`get` + `remove`), peek, scope (`get` + `save`).
- A miss answers `503` with `errorCode: "persisted-store-unavailable"` and a message naming the likely cause and the fix. 503, not 500: the store is unreachable, not broken.
- **The abandoned `SecItem*` call keeps its thread** until the dialog resolves. Uncancellable by design and accepted: an idle background thread beats a wedged route. Documented at the helper so nobody "fixes" it by waiting.

Client (`packages/webapp`):

- Every `/api/secrets*` call carries `AbortSignal.timeout(10_000)` — above the server's 5 s so the server's specific diagnosis wins the race, and at the control-plane budget so a broker waiting on us never gives up first.
- A transport failure is shaped like a failed response rather than rethrown, so callers report it through the path they already use for server errors instead of surfacing a raw `DOMException`.
- `SecretBackend.list()` now returns `{ entries, warnings }`. It reads two independent stores and could lose one; carrying the reason alongside the entries is the honest shape. Both backends previously dropped a failed half silently, including the extension one, which ignored the handlers' `error` field.
- `secret list` prints whatever was readable, puts the reason on stderr, and exits 1. `No secrets stored` now means both stores answered and both were empty, so a `0` exit is trustworthy. `secret test` blames the unreadable store rather than claiming the secret does not exist — a name it could not look up is not a scope decision.

## Cross-cutting impact

- **Cross-runtime parity**: the client change covers all five topologies; the CLI/Electron/Sliccstart path shares one REST client. node-server needed no change (no dialog, no hang), so the 503 is a Swift-only response, reachable only on a deadline miss — no contract case changes.
- **Follower wiring parity**: `createMessageSecretBackend` (extension-direct + bridge) gained the same warnings, so a failed SW store read is no longer invisible there either.
- **Error-path coverage**: the new paths are the error paths, and each is tested.
- Session secrets are untouched throughout — they live in memory, so `secret list` still renders its SESSION rows while the saved half is unavailable.

## Test plan

New, and each fails without its fix (verified by reverting the wiring):

- `packages/swift-server/Tests/BoundedSecretStoreRoutesTests.swift` (4) — stalled store answers 503 with an actionable message in ~5 s instead of hanging; `/api/secrets/masked` and the session route stay sub-2 s throughout; prompt results and thrown store errors pass through the wrapper unchanged; a call past its deadline is not waited for.
- `secret-backends.test.ts` (+3, and 3 existing expectations updated) — the bridge's 503 diagnosis reaches the caller verbatim; every call carries an abort signal; an abort and an unreachable bridge each become a readable warning.
- `secret-command.test.ts` (+5) — partial list still prints SESSION rows with the reason on stderr and exit 1; never `No secrets stored` when a store was unreadable; a genuinely empty store still reports empty with exit 0; `secret test` blames the store, and still says `no secret named` when the stores read cleanly.

Full pass, all green: `swift build`, `swift test` (652), `swiftlint` (59 warnings, unchanged from baseline), `swift format lint --strict`, `npm run typecheck`, webapp shell suite (5180), `npm run lint`, `npm run lint:docs`, touched-file debt gate.

## Documentation

- `docs/secrets.md` — new "When a store cannot be read" section with the actual output shape and the exit-code contract; `secret list` table row.
- `docs/swift-server-details.md` — records that the startup read is not the only one, the bounded request path, the 5 s/10 s relationship, and the abandoned-thread trade-off.

## Risk & rollback

Low. Behaviour changes only where a store call misses its deadline or fails — previously a hang or a silent empty list. The `list()` signature change is internal to the webapp shell (two call sites, both updated).

`secret list` now exits 1 on a partial answer where it used to exit 0 with an incomplete list; that is the point, but it is the one change a script could notice.

Rollback is the single revert commit.

## History

Opened stacked on #2945 (both touch `secret-command.ts`). #2945 and #2944 have since merged, so this now targets `main` directly and has been rebased onto it. The rebase resolved one documentation conflict: #2944 corrected the `SLICC_KEYCHAIN_NONINTERACTIVE` paragraph in `docs/swift-server-details.md`, and that correction is kept intact with this PR's request-path paragraph added above it.

Verified again after the rebase, on top of #2944's Keychain changes: `swift test` 658, webapp shell suite 5190, typecheck, `npm run lint`, `lint:docs`.

## Checklist

- [x] Tests added and verified to fail without the fix
- [x] Docs updated
- [x] Full verification pass run locally
- [x] No secret values in code, logs, tests, or error messages


trieloff:
Both P1 findings from the Codex review are addressed in `a4033c26c`, along with a third exposure the first one pointed at.

**Timed-out writes committing later** — correct, and worse than the review says: the route also skipped `reload()`, so the pipeline kept injecting the superseded credential. A write timeout now answers `errorCode: "persisted-store-write-unknown"` and states the outcome is unknown and may still apply, rather than implying failure and inviting a retry that double-applies a rotation. `BoundedStoreCall` gained an `onLateCompletion` hook, and the three write sites (set, remove, scope-save) use it to run `SecretInjector.reload()` if the abandoned call lands after the response.

**Incomplete lookup read as absence** — correct, and it was a gate bypass, not just a wrong answer: `getInfo() -> null` told `handleSetSession` the name was new, skipping the `value` approval. `resolveInfo` now raises when the name is missing *and* a store went unread, in both backends; absence is only reported when every store answered. `secret set` refuses with the reason and never reaches `setSession`.

**`reload()` was still unbounded**, as your second finding implied — so session set/delete could hang on the same Keychain read even with the routes bounded. It is bounded now, and a miss keeps the previous snapshot instead of replacing it with a partial one: dropping live secrets would silently stop masking them, which is worse than briefly missing the newest entry.

New tests (7 swift / 8 webapp in total on the branch): a timed-out write reports an unknown outcome and never echoes the value; a late completion is reported so it can be reconciled; a session mutation survives a stalled reload with the earlier secret still masked; `getInfo` raises on a partial list, still reports genuine absence, and still resolves a found record when the *other* store failed; `secret set` refuses without calling the backend or the broker.

Re-verified: `swift test` 655, webapp shell suite 5184, `swiftlint` 59 warnings (unchanged from baseline), `swift format lint --strict`, typecheck, `npm run lint`, `lint:docs`, debt gate. Docs updated for the write contract and the fail-closed `secret set`.

Issue #2950, opened 2026-09-07
/usr/bin/diff defaults to unified output (not POSIX normal format); 'diff a b | grep ^<' silently matches nothing, and normal format is unreachable


trieloff:
## Summary

`/usr/bin/diff`, invoked with no flags, emits **unified** output (`---`/`+++`/`@@`/`+`/`-`) instead of POSIX **normal** format (`<`/`>`). POSIX and GNU `diff` both default to normal format; unified is opt-in via `-u`. The consequence is that the extremely common idiom `diff a b | grep -E '^[<>]'` silently matches nothing here, so a script or agent that counts changed lines that way concludes "no differences" on files that differ.

It also emits a non-standard 67-character `===...===` banner as the first line, which neither GNU normal nor GNU unified output produces, and normal format appears to be unreachable by any flag.

## Reproduce (deterministic)

```
$ printf 'a\nb\nc\n' > d1.txt
$ printf 'a\nX\nc\n' > d2.txt

$ diff d1.txt d2.txt          # no flags
===================================================================
--- d1.txt	
+++ d2.txt	
@@ -1,3 +1,3 @@
 a
-b
+X
 c

$ diff d1.txt d2.txt | grep -cE '^[<>]'      # normal-format markers
0
$ diff d1.txt d2.txt | grep -cE '^[-+@]'     # unified-format markers
5
```

`diff -u d1.txt d2.txt` produces byte-identical output to the no-flag invocation, confirming the default IS unified.

## What GNU/POSIX diff does instead

```
$ diff d1.txt d2.txt          # GNU default = normal format
2c2
< b
---
> X
```

## Reachability of normal format

```
$ diff --normal d1.txt d2.txt   -> diff: unrecognized option '--normal'   (exit 1)
$ diff -c d1.txt d2.txt         -> diff: invalid option -- 'c'
$ diff --version                -> diff: unrecognized option '--version'
$ diff -q d1.txt d2.txt         -> Files d1.txt and d2.txt differ         (works)
```

So there is no flag that yields normal `<`/`>` format, and `-c` (context) is also unavailable.

## Impact

`diff a b | grep '^<'` / `'^>'` is a standard idiom for "lines only in a / only in b", taught as THE normal-format contract. An agent or script relying on it here gets a false negative with no error — the same failure class as a filter flag being silently ignored. In my case it made a file that differs by 3,565 bytes read as "0 changed lines" until a `cmp` contradicted it.

## Suggested fix, in order of preference

[...]

PR #2954, opened 2026-09-07
feat(git): implement `git diff --no-index`

trieloff:
## What

Implements `git diff --no-index <path> <path>` in SLICC's git shim.

Before, the unknown flag was dropped and the first path fell through to revision resolution:

```
$ git diff --no-index d1.txt d2.txt
fatal: ambiguous argument 'd1.txt'
```

Now it does what real git does. The whole point of the mode is that **no repository is involved** — every test in the new suite runs from a cwd that was never `git init`ed.

New module `packages/webapp/src/git/commands/diff-no-index.ts` reads both operands straight off the VFS and hands them to the same `unifiedDiff` / `--stat` renderer every other diff mode uses. It never calls isomorphic-git.

## Behavior

- **Operands** — two paths, else the usage line and exit 129. File-vs-file, **directory-vs-directory** (recursive, union of both trees, sorted by relative path), and `<dir> <file>` in either order, which git resolves to `<dir>/<basename of file>`. `/dev/null` is an empty file whether or not the path exists, so `git diff --no-index /dev/null new.txt` yields an add patch.
- **Exit code** — `--no-index` implies `--exit-code`: 1 when the paths differ, 0 when identical (with no output). A path that cannot be read is `error: Could not access '<path>'` + exit 1, never an empty diff; the left operand is reported first.
- **Output flags** — `--name-only` (prints the b-side path, or `/dev/null` for a deletion), `--stat` (rows labeled with git's rename-compressed `{A => B}/f.txt` form; `Bin <n> -> <m> bytes` for binary pairs), and `-U<n>` / `--unified=<n>`.
- **Binary** — a pair with a NUL in its first 8000 bytes reports `Binary files a/X and b/Y differ` instead of being decoded and line-diffed.

### Directory support: implemented, not punted

The task left this open. It turned out to be cheap — a recursive `readDir` on each side and a sorted union — so it's in, and adds/deletes render with `/dev/null` on the absent side exactly as git does.

## Fidelity against the host oracle

Every expectation in the test suite was taken from real `git diff --no-index` (git 2.50.1) on the host. Output is byte-identical **except** for two deliberate omissions:

- **No `index` / `new file mode` / `deleted file mode` header lines.** Note the task brief said real git omits the `index` line for `--no-index`; it does not — it prints `index de98044..0c02ccc 100644`. We omit it because SLICC's diff output carries no object IDs or modes *anywhere* (`git diff`, `log -p`, `show`, `stash` all already omit them), and making `--no-index` the sole exception would be the inconsistent choice.
- **No rename or copy detection.**

Two *fixes* to the shared formatter fell out of getting the `@@` line exactly right, and they now apply to every diff producer, not just `--no-index`:

- A side that contributes no lines is numbered by what came **before** the hunk. An addition to an empty file was heading `@@ -1,0 +1,3 @@` — a hunk header naming a line the old file does not have. It is now `@@ -0,0 +1,3 @@`, and a `-U0` insertion after line 2 is `@@ -2,0 +3 @@`.
- A one-line side omits its count: `@@ -1 +0,0 @@`, not `@@ -1,1 +0,0 @@`.

Both forms are what `git apply` and `patch` expect to read back, and `patch-core.ts`'s hunk regex already accepts the optional count. Full-suite blast radius of the change was exactly zero — all 20,453 tests pass unmodified.

`-U<n>` is normalized to `--unified=<n>` before the flag parse, because `mri` mis-reads a short flag with an attached value (`-U3` becomes `{ '3': …, U: '' }` and eats the next positional). It applies to all diff modes, which previously dropped `-U<n>` silently.

## Not implemented

**Implicit `--no-index`** — real git *infers* the mode when a plain `git diff` outside a repository is given two paths. The task marked this optional; it is not inferred here, so pass the flag. Inferring it means "am I in a repo, and are both positionals existing paths rather than revisions?", which changes the failure mode of ordinary `git diff <rev> <rev>` typos, and did not seem worth that on this pass.

## Tests

`packages/webapp/tests/git/diff-no-index.test.ts` — 20 cases: differing files, identical files (exit 0, empty output), missing path (either side, and both), a cwd outside any repository, `--name-only`, `--stat`, `-U<n>`, absolute operands, `/dev/null`, binary, wrong operand count, the three directory dispositions, identical trees, `<dir> <file>` joining, and `pprintRename` units. New module is at 99% statement coverage.

## Docs

`docs/shell-reference.md` — new `git diff --no-index needs no repository` and `Hunk headers are byte-compatible with git` sections, plus the `git` row in the supplemental-command table.

## Verification

`npm run verify`, `check-touched-exemptions`, `typecheck`, `test` (20,453 passed), `test:coverage` (exit 0), both builds, `bundle-size` all pass. The kernel-worker graph grows +4.5 kB against a 5 kB per-change allowance (42 kB under ceiling) — the new module rides in statically, consistent with every other git subcommand, none of which is lazily imported.

## Issue

Refs #2950. This closes **only** the `git diff --no-index` breadcrumb at the bottom of that issue. The issue's actual subject — plain `diff` defaulting to unified format — is a just-bash builtin, not ours, and is filed upstream as vercel-labs/just-bash#412. Plain `diff` is untouched here and no supplemental `diff` shadow was added.


trieloff:
Both findings were real and reproduced. Fixed in e13c85c04 rather than documented as deviations — the trailing-newline one in particular re-introduced exactly the failure this PR set out to remove.

**Major — trailing-newline-only differences.** `unifiedDiff` popped the trailing empty element from both splits, so `a\nb\n` and `a\nb` compared equal and rendered `''` at exit 1.

Every line handed to the Myers diff now carries a one-character **prefix** recording whether it ended with a newline, so an incomplete final line can never compare equal to the same text followed by one. A prefix is collision-free where a sentinel suffix would not be: exactly one character is added and exactly one is stripped, whatever the line contains. The affected line is closed by `\ No newline at end of file`, placed and uncounted exactly as git places it — including the case where both sides share an incomplete last line, where git emits the marker once on the context line.

One thing that turned up while fixing it: `patch-core.ts` has always *parsed* that marker (`oldNoNewlineAtEof` / `newNoNewlineAtEof`, and its "one more line" lookahead). This is the first SLICC producer to emit it, so `git diff | patch` now round-trips a missing trailing newline. The fix reaches `diff`, `log -p`, `show`, and `stash` as well, all of which previously showed no diff at all for a newline-only change.

**Minor — binary-only `--stat` summary.** Matched git's actual rule from `print_stat_summary`: a zero clause is suppressed only when the *other* one is non-zero. So a binary-only change now reports `0 insertions(+), 0 deletions(-)`, while an insertions-only change still correctly omits the deletions clause.

Three new cases cover the newline behaviour (real hunk, shared incomplete last line, `--stat` counts), and the binary `--stat` assertion is now exact rather than `toContain`. Full suite: 20,456 passed, no regressions. `verify`, `typecheck`, `test:coverage`, both builds, and `bundle-size` all re-run green.

PR #2955, opened 2026-09-07
feat(webapp): give the exhausted-budget error a card that can fix it

trieloff:
## Summary

The Adobe proxy's new rate limit reached the chat as a raw
`429 {"error":{"type":"quota_exceeded","message":"Weekly budget has been fully used. Resets on 2026-09-14. You can also connect your own LLM provider.","resets_at":"…"}}`
line, under a generic **Something went wrong** header, behind a **Try again**
button that cannot work — the budget is gone until it resets.

Now the card parses that envelope and offers the two remediations that do work:
**Switch provider and try again** (only when another provider is actually
connected) and **Add a provider**.

## Why

The generic treatment fails the user three ways at once: the body is JSON, the
header says nothing, and the only affordance re-runs the exact request that just
got refused. The information needed to fix it — that a *different* provider would
work, or that connecting one is a click away — is nowhere on screen.

**Which CTA, and why it is conditional** (`errorCardEl`, `wc-message-view.ts`):

- **Another provider connected** → primary `change-model` labelled *"Switch
  provider and try again"*, with *"Add a provider"* (`settings`) as an outline
  secondary. `wireWcNav` already routes `change-model` to the composer model
  picker **and** stages a pending replay, so "and try again" is literal: picking
  a model re-runs the failed turn without a second click. No new nav wiring.
- **Exhausted account is the only one** → the single CTA is *"Add a provider"*.
  Opening a model picker that holds only the exhausted account is a dead end.

"Another provider connected" is the new `getAlternativeModelProviders`
(`account-store.ts`), derived from `getAllAvailableModels` — so hidden,
build-excluded and policy-denied providers are filtered out exactly as the picker
filters them. A CTA that opens a dropdown must not promise an entry the dropdown
lacks. It is read at render time inside a `try`/`catch`: the store touches
`localStorage` and a provider catalog, and neither may take an error card down.

**Detection** (`core/error-families.ts`) matches the machine-readable
`error.type`, not the prose, so re-worded proxy copy — or a
`Scoop "…" failed with unrecoverable error: ` wrapper — never drops the failure
out of the family. `parseQuotaExceededError` is detect-and-parse in one call; a
truncated or message-less envelope still yields a detail, because a parse miss
must not fall back to dumping JSON at the user. The trailing "You can also
connect your own LLM provider." sentence is stripped: the CTAs *are* that now.

**Component** (`slicc-error-card`): new optional `secondary-action` /
`secondary-button-label`, drawn from the same four variants and firing the same
events. Absent **or unknown** means no button — deliberately *not* normalized to
`retry` the way `action` is, so a host that never asked for a second button
cannot grow one.

**Two judgment calls a reviewer should weigh in on:**

1. The ask read as either "one CTA, conditionally chosen" or "two CTAs". I chose
   to show **both** when a switch target exists. Reverting to strictly
   one-or-the-other is a two-line change in `quotaCtaAttrs`.
2. I extended the *reword* (not the CTAs) to the iOS follower and to read-only
   scoop transcripts. See below.

## Cross-cutting impact

- **Floats exercised**: leader (CLI / hosted). The extension side panel needs no
  change — `wc-follower.ts` already forwards `slicc-error-open-settings` and
  `slicc-error-change-model` to the leader tab, and this PR reuses both events
  rather than inventing new ones.
- **iOS follower**: `QuotaExceededDetail` (`Models/ErrorFamilies.swift`) mirrors
  the *text* only. `ErrorCard` still renders **no** CTA — both act on
  leader-side state the follower→leader protocol carries no message for — but a
  reader who cannot act is still owed prose instead of a JSON envelope.
- **Read-only scoop transcripts** (#2312): same reasoning. The label/body rewrite
  now happens *before* the `no-action` early return, so a scoop transcript keeps
  `no-action` and loses every CTA, while still showing readable text.
- **Telemetry**: no new wiring. Both skip sites (`trackError`,
  `trackScoopLifecycle`) already call `isUserFixableError`, which the quota
  family now joins — so the exhausted-budget cascade stops beaconing as a
  regression. `#emitErrorCardBeacon` was switched from its hand-copied
  three-family list to that same predicate, which is now the single shared list.
- **Bundle size**: `wc-message-view.ts` gained a static import of
  `account-store.ts`, which was **already** an eager page chunk (65 kB). The
  first-load gate measures **page +1.3 kB / worker +0.2 kB** vs the merge base —
  well inside the 5 kB allowance, 20 kB / 46 kB under the ceilings.
- **Other callers of the changed contract**: `isUserFixableError` gained a
  fourth family; its three call sites (two telemetry, one error-card beacon) all
  want exactly that widening. The `slicc-error-card` change is additive —
  `#bindAction`'s primary lookup became `.retry:not(.ghost)` so the shared
  `.retry` class hook keeps matching both buttons for existing CSS and
  shadow-piercing tests.
- **Persisted state**: none. No new storage keys, no migrations.

## Test plan

- [x] `npm run verify` — 8/8 (biome, prettier, custom-lints, boy-scout-debt,
      manifest, deadcode, deadcode-prod, autofix-drift)
- [x] `npm run typecheck`
- [x] `npm run test` — **20,456 passing**, 54 skipped, no new failures
- [x] `npm run test:coverage:webapp` / `:webcomponents` — both gates pass
- [x] iOS: 737 unit tests pass; `swift-coverage-check.sh` passes
      (`ErrorFamilies.swift` at 100% lines / 95% regions)
- [x] `swift format lint --strict` + `swiftlint` clean
- [x] `npm run bundle-size` — first-load gate green (numbers above)
- [x] **New behavior covered by unit tests** in four suites:
  - `packages/webapp/tests/core/error-families.test.ts` (new) — parsing, wrapper
    prefixes, truncated envelopes, non-string fields, neighbouring families
  - `packages/webapp/tests/ui/wc/wc-message-view.test.ts` — both CTA branches,
    the failing-provider exclusion, store-throws degradation, read-only parity
  - `packages/webapp/tests/providers/cross-provider-model-resolution.test.ts` —
    `getAlternativeModelProviders`, including the policy-denied case
  - `packages/webcomponents/tests/chat/slicc-error-card.test.ts` — secondary
    button render order, glyphs, event isolation, unknown-value inertness,
    `no-action` suppression, disconnect unbinding
  - `packages/ios-app/…/QuotaExceededDetailTests.swift` (new) — the Swift mirror
- [ ] Manual smoke against a live exhausted budget — **not done**: reproducing it
      means actually burning a weekly Adobe budget. Covered by the fixture-driven
      tests above and the Storybook stories.

**Pre-existing failures**: none observed.

## Screenshots

New Storybook stories `Chat/ErrorCard/QuotaExceeded` (+ `…Dark`) render the card
in isolation. Both show the state where another provider **is** connected, so
both CTAs appear.

**Light:**

Out of AI budget card, light mode

**Dark:**

Out of AI budget card, dark mode

With no other provider connected the card collapses to a single **Add a
provider** button — asserted in `wc-message-view.test.ts`, not just described.

## Documentation

- [x] `docs/operational-telemetry.md` — the user-fixable family list the `error`
      checkpoint drops now names `quota-exceeded`
- [x] Doc comments on `errorCardEl`, `quotaCtaAttrs`, `wireChangeModelEvent`,
      `wireOpenSettingsSurfaces`, `slicc-error-card`, and the Swift `ErrorCard`
      carry the routing story
- [x] Storybook stories are the component-level docs for the new variant

## Risk & rollback

Blast radius is one error-card branch: a message that does not carry
`quota_exceeded` takes exactly the path it took before. No flags, no persisted
state, no migration — revert cleanly.

### DIFF
--- a/docs/operational-telemetry.md
+++ b/docs/operational-telemetry.md
@@ -118,7 +118,7 @@ SLICC uses helix-rum-js's supported checkpoint types with SLICC-specific semanti
 | `convert`      | Cone fed a scoop   | scoop folder name                         | `'scoop-feed'`                      | `emitScoopLifecycle('feed', ...)`, called from `scoop-message-router.ts`                                                                                       |
 | `leave`        | Scoop completed    | scoop folder name                         | `'scoop-complete'`                  | `emitScoopLifecycle('complete', ...)`, called from `scoop-completion-service.ts`                                                                               |
 
-A scoop failure reuses the `error` checkpoint row above rather than getting its own row: `trackScoopLifecycle()` sets `source: 'scoop:<name>'` (not the bare scoop name used by `enter`/`convert`/`leave`) and runs the message through the same `sanitizeError` used for `trackError`, dropping known user-fixable error families (no-api-key, invalid-model, auth-expired) before emitting.
+A scoop failure reuses the `error` checkpoint row above rather than getting its own row: `trackScoopLifecycle()` sets `source: 'scoop:<name>'` (not the bare scoop name used by `enter`/`convert`/`leave`) and runs the message through the same `sanitizeError` used for `trackError`, dropping known user-fixable error families (no-api-key, invalid-model, auth-expired, quota-exceeded) before emitting.
 
 ### Auto-instrumented (from enhancer, CLI/Electron only)
 
--- a/packages/ios-app/SliccFollower/Models/ErrorFamilies.swift
+++ b/packages/ios-app/SliccFollower/Models/ErrorFamilies.swift
@@ -1,0 +1,89 @@
+import Foundation
+
+/// The pieces an exhausted-provider-budget failure is rendered from.
+///
+/// Swift mirror of `parseQuotaExceededError` in
+/// `packages/webapp/src/core/error-families.ts`. The leader's error card shows
+/// the provider's own prose under an "Out of AI budget" header; without this
+/// the follower would keep printing the raw
+/// `429 {"error":{"type":"quota_exceeded",…}}` envelope under a generic
+/// "Something went wrong", which is the same failure told twice as badly.
+///
+/// Only the TEXT is mirrored. The leader's card also offers "Switch provider
+/// and try again" / "Add a provider"; the follower renders no CTA at all
+/// (see `ErrorCard`), because both act on leader-side state the
+/// follower→leader protocol carries no message for.
+struct QuotaExceededDetail: Equatable {
+    /// The provider's explanation, with the trailing "connect your own LLM
+    /// provider" sentence dropped — on the follower there is nothing to click.
+    let message: String
+    /// ISO-8601 instant the budget refills, when the provider sent one.
+    let resetsAt: String?
+
+    /// Header the error card shows instead of "Something went wrong".
+    static let label = "Out of AI budget"
+
+    /// Body for an envelope that carried no usable message of its own.
+    static let fallbackMessage = "The usage budget for this provider has been fully used."
+
+    /// The `error.type` the Adobe proxy stamps on an exhausted-budget refusal.
+    /// Matched instead of the prose, so re-worded proxy copy — or a
+    /// `Scoop "…" failed with unrecoverable error: ` wrapper — never drops the
+    /// failure out of detection.
+    private static let typeToken = "quota_exceeded"
+
+    /// Trailing self-service sentence the leader's CTAs replace.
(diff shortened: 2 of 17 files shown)


trieloff:
Storybook renders of the new `Chat/ErrorCard/QuotaExceeded` story — the state where another provider **is** connected, so both CTAs show.

**Light:**

Out of AI budget card, light mode

**Dark:**

Out of AI budget card, dark mode

With no other provider connected the card collapses to a single **Add a provider** button (asserted in `wc-message-view.test.ts`).

PR #2970, opened 2026-09-08
Fix Adobe login SW race; default dev harness to hosted origin

catalan-adobe:
## What

Three related fixes found during a fresh Canary smoke test.

### Adobe provider — SW timing race on first login (`adobe.ts`)

`fetchProxyConfig` failed on the very first login attempt in a fresh Chrome profile with `Failed to fetch`. The service worker intercepts the cross-origin fetch to the adobe proxy before its bridge config is settled (Chrome SW timing race on a fresh profile). In that window the SW routes to same-origin wrangler, which returns 404 (`Fetch proxy not available in worker mode`). `fetchProxyConfig` catches the rejection, returns `{}`, and `resolveClientId({})` throws:

> Login failed: Could not determine IMS client ID — proxy /v1/config did not return one and adobe-config.json is empty

**Fix:** split `fetchProxyConfig` into `attemptFetchProxyConfig` (single attempt) and a wrapper that retries once after 600 ms. The SW settles within that window so the retry goes through the bridge correctly.

Also cleaned up all pre-existing `as unknown as T` casts with `SAFETY:` comments (required by `require-safety-comment-for-as-unknown-as` which was triggered after touching the file), and documented three intentional empty catch blocks.

### Dev harness — wrong default origin (`dev-standalone-fresh.sh`)

The script always started wrangler and opened Chrome at `localhost:8787`. OAuth (IMS/Adobe, GitHub) does not work from localhost origins — IMS only allowlists `www.sliccy.ai`. The node-server already defaults to `SLICC_HOSTED_ORIGIN` when `WORKER_BASE_URL` is unset; the script was overriding that unconditionally.

**Fix:** default to the production hosted origin (`https://www.sliccy.ai`). Chrome opens `https://www.sliccy.ai/?bridge=ws://localhost:5715/cdp&bridgeToken=…` — the hosted UI connects back to the local node-server via the bridge parameter. Wrangler is only started when `WORKER_BASE_URL` is explicitly set to a localhost URL. `BRIDGE_DEV_ALLOWED_ORIGINS` is only set for the local-worker path; `sliccy.ai` is already in the hard-coded `BRIDGE_ALLOWED_ORIGINS` list.

### Skill documentation (`cdp-smoke-test/SKILL.md`)

Documented the hosted-origin mode as the **recommended default** with the correct launch command. Added explicit note that OAuth/IMS will not work from localhost origins. Wrangler / local-worker mode demoted to an opt-in alternate section.

### DIFF
--- a/.agents/skills/cdp-smoke-test/SKILL.md
+++ b/.agents/skills/cdp-smoke-test/SKILL.md
@@ -27,21 +27,28 @@ belong to the developer's production SLICC. Never bind, reap, or kill either
 port. The smoke-test lane uses bridge `:5715`; its Chrome CDP port is
 auto-resolved.
 
+### Default mode — hosted origin (recommended)
+
+Chrome opens `https://www.sliccy.ai` with a `?bridge=ws://localhost:5715/cdp`
+parameter that routes it back to the local node-server. No wrangler needed.
+OAuth, IMS (Adobe), and all provider relays work out of the box because they
+are hosted on the production origin.
+
 ```bash
 # 1. Record production listeners before doing anything else.
 PROD_BRIDGE_PIDS=$(lsof -nP -tiTCP:5710 -sTCP:LISTEN 2>/dev/null | sort -n | paste -sd, -)
 PROD_CDP_PIDS=$(lsof -nP -tiTCP:9222 -sTCP:LISTEN 2>/dev/null | sort -n | paste -sd, -)
 printf 'production preflight: :5710=%s :9222=%s\n' "${PROD_BRIDGE_PIDS:-none}" "${PROD_CDP_PIDS:-none}"
 
-# 2. Build the latest code (cherry regenerates worker bridge assets).
+# 2. Build only node-server (no webapp build required — UI loads from sliccy.ai).
 npm install
-npm run build -w @ai-ecoverse/cherry -w @slicc/webapp -w @slicc/node-server
+npm run build -w @slicc/node-server
 
-# 3. Launch — wrangler :8787, isolated bridge :5715, ephemeral profile.
+# 3. Launch — hosted origin, isolated bridge :5715, ephemeral profile.
 #    CHROME_PATH is optional; default is a labeled Chrome for Testing clone.
 export SLICC_HARNESS_LOG=/tmp/slicc-dev-harness-5715.log
 CHROME_PATH="/Applications/Google Chrome Canary.app" \
-  PORT=5715 WRANGLER_PORT=8787 \
+  PORT=5715 \
   nohup npm run dev:standalone:fresh > "$SLICC_HARNESS_LOG" 2>&1 &
 
 # 4. Wait for boot and export the auto-resolved CDP port from this lane's log.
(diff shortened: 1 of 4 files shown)


catalan-adobe:
### Strengths

- **[packages/webapp/providers/adobe.ts:149-191]** — The single-attempt helper is cleanly separated from retry and caching policy. Only successful responses are cached, preserving recovery from transient failures, and both HTTP and network failures remain observable.
- **[packages/dev-tools/tools/dev-standalone-fresh.sh:242-249]** — The no-override path now correctly launches the production hosted origin without building or starting wrangler.
- **[packages/dev-tools/tools/dev-standalone-fresh.sh:322-335]** — `BRIDGE_DEV_ALLOWED_ORIGINS` is empty in the default hosted path, avoiding an unnecessary expansion of the bridge allowlist.
- **[.agents/skills/cdp-smoke-test/SKILL.md:76-93]** — The documentation clearly marks local wrangler as optional and warns that OAuth/IMS cannot be tested from localhost.
- **[packages/webapp/providers/adobe.ts:1028-1338]** — The added `SAFETY:` comments explain the structural assumptions behind the existing double casts rather than merely restating the cast.

### Issues

*(none found)*

- **[packages/dev-tools/tools/dev-standalone-fresh.sh:139-143]** — Every non-empty `WORKER_BASE_URL` is classified as a local-worker invocation.
  - The plan says wrangler should start only for an explicitly configured localhost URL. As implemented, an inherited `WORKER_BASE_URL=https://www.sliccy.ai` or staging URL unnecessarily builds the UI and starts wrangler at lines 256-298; an unrelated listener on `:8787` can therefore prevent an otherwise valid hosted launch.
  - Local URLs also need port consistency: `WORKER_BASE_URL=http://localhost:9000` still starts and probes wrangler on the default `WRANGLER_PORT=8787`, then opens a URL where nothing was started.
  - Parse and validate the URL. Start wrangler only for loopback hosts, deriving its port from the URL or rejecting a mismatch with `WRANGLER_PORT`; explicit remote origins should launch directly without wrangler.

- **[packages/webapp/tests/providers/adobe-provider.test.ts:761-788]** — The existing test helper no longer mirrors the production retry implementation.
  - For example, the recovery test at lines 835-849 expects the first `fetchProxyConfig` call to fail and a second caller invocation to recover. The real implementation now retries within the first invocation after 600 ms, so it should succeed immediately from the caller’s perspective. Likewise, two always-failing production invocations now make four fetches, while this test still expects two.
  - Consequently, the PR’s central race fix has no automated regression test even though this suite appears to cover it.
  - Extract the retry/cache logic into a DOM-independent module and test the actual helper with fake timers: first-attempt rejection followed by success, non-OK followed by success, both attempts failing without caching, and successful cache reuse.

- **[.agents/skills/cdp-smoke-test/SKILL.md:43-45]** — The recommended smoke-test mode cannot exercise local webapp changes, including this PR’s Adobe fix.
  - The hosted page loads its JavaScript and service worker from the deployed `https://www.sliccy.ai`; building only `@slicc/node-server` means the local `packages/webapp/providers/adobe.ts` retry code is never executed. This contradicts the skill’s claim at line 19 that it tests the “latest local build” and leaves no documented pre-merge end-to-end path for OAuth changes.
  - This is primarily a limitation in the plan itself rather than a coding deviation. Describe this mode as “deployed webapp + local node-server,” and provide an OAuth-allowlisted preview/development origin that serves the current webapp commit if provider changes must be smoke-tested before deployment.

#### 💡 Minor (Nice to Have)

- **[packages/dev-tools/tools/dev-standalone-fresh.sh:2-16]** — The script header still says the harness always launches “wrangler UI + node-server” and requires a webapp build.
  - That is no longer true for the default hosted mode and conflicts with the updated skill instructions.
  - Update the header and prerequisites to distinguish hosted-default and local-wrangler modes.

- **[packages/webapp/providers/adobe.ts:1204-1212]** — `buildPiAiModelMap` still catches every `getModels` failure without logging it.
  - A provider initialization or registry regression would silently remove metadata and cost inheritance from Adobe models. A comment documents the fallback but does not make unexpected failures observable.
  - Log a deduplicated warning containing the provider name, or catch only the specific expected “not ready/unregistered” condition.

### Recommendations

- Prefer tests that import production retry/origin-selection helpers rather than maintaining copied implementations.
- Add a small origin-mode test matrix covering unset, production, staging, localhost default port, localhost custom port, and malformed `WORKER_BASE_URL`.
- Clearly distinguish smoke tests of deployed UI code from smoke tests of the current checkout.

### Assessment

**Reasoning:** The main runtime changes are directionally correct, but origin classification does not match the stated plan, and the Adobe retry is currently covered by a stale test that asserts the old behavior. The documented hosted workflow also cannot validate the local webapp fix before deployment.

---

PR #2973, opened 2026-09-08
feat: headline budget usage, not $, when the provider bills against a rolling allowance

trieloff:
Closes nothing yet — pairs with the Adobe LLM proxy's `GET /v1/usage` (aemcoder/vibemigration#200).

## Why

A provider that bills against a **rolling allowance** rather than per token makes session dollars the wrong headline. A family-priced model can bill `$0.00` while the shared weekly window burns down, so `$1.40/hour` says "nothing is happening" during the hour that ends the week's work.

Every cost surface now headlines **percent USED** of the window on such a provider. Providers that meter per token are untouched — no `getBudgetUsage()` hook means today's `$` headline, byte-identical.

## Two commits

**1. `feat(webcomponents)` — stories + presentational model** (approved before wiring)

- `primitives/budget-usage.ts`: one shared vocabulary (`BudgetUsage`, thresholds, formatting, copy fragments) so a pill and its hover card can never disagree about what `9.5%` means.
- `<slicc-floatbar>`: `budget-percent` / `-status` / `-window` / `-resets` (or the `budget` property) swap the `$N/h` segment for a gauge + `9.5%`, escalating ink → amber (80%) → rose (95%, or rate-limited, which also swaps the gauge for an alert octagon).
- `<slicc-cost-overlay>`: a `budget` property leads the card with the window and a meter; the total row becomes "This session".
- The monitor needed **no new type** — a budget hero is an ordinary `MonitorVital` — so its stories are model-only.
- Stories cover mid-window, near-limit, rate-limited, overrun, and a side-by-side `$`-vs-budget contrast on all three surfaces.

**2. `feat(webapp)` — the live wiring**

- `ProviderConfig.getBudgetUsage?()` is the seam. Its three-way contract is load-bearing: `null` = no budget concept (re-probed in 30 min), **throw** = the call failed (retried in 5 min, last reading stays on screen), a window = a reading. Adobe implements it in `adobe-usage.ts` — pure and `fetch`-injectable, since `adobe.ts` itself cannot be imported under vitest.
- `budget-window-cache.ts`: success 60 s, one shared in-flight probe, keyed by provider id so a switched account never inherits the previous one's allowance.
- `SessionStatsMsg.budget` carries the window with its ISO `resetsAt`. The facade attaches a **cached snapshot** and fires the refresh without awaiting it — the handler is on the request loop. Only a session's first pull waits (2.5 s, under the panel's 5 s stats timeout) so a budget provider's pill leads with its window instead of showing `$/h` and then flipping.
- Monitor: budget hero, burn rate demoted to an ordinary tile (kept — a runaway scoop is still readable), Cost group named by the window, and warn/refusal alerts.

## Conventions, and why

- **Percent USED, never remaining** — the provider's own convention. One direction on every surface: a rising number always means worse.
- **The bar clamps at 100%; the figure does not.** A provider reporting 104% has said something true.
- **`rate-limited` is critical whatever the percent says** — the reported percent lags the refusal, so a refused window at 40% still stops work while a quiet one at 92% does not. Carried by a word *and* a glyph, never tint alone (`--waffle` is 2.09:1 on these surfaces).
- **No clock in a component.** Reset copy is relative and locale-free (`resets in 18h`), formatted at the reading edge from the ISO instant on the wire — a string computed in the worker would be stale by the time a pill rendered it.
- The webapp deliberately **duplicates** the webcomponents formatter and thresholds so the shell keeps a DOM-free copy; `provider-budget.test.ts` pins the two together.

## ⚠️ One contract change

`cost --json` is now the envelope `{ "budget": <window|null>, "scoops": [ … ] }` — read rows from `.scoops`. It is also parseable in the no-data case now (it used to answer that path with prose). `docs/shell-reference.md` and the delegation skill are updated in the same commit.

## Two design calls worth a look

1. **The monitor keeps four vitals tiles.** The grid is four fixed columns, so a fifth wraps at hero width. Live processes is the tile that yields — the only vital the process table below repeats verbatim.
2. **The floatbar still hides its cost segment below a 560px nav.** In budget mode that is now the headline being hidden; the collapse ladder was left as-is rather than re-ranked as a drive-by.

## Verification

lint ✅ · typecheck (all 9 projects) ✅ · `npm run test` 20 837 passed ✅ · webapp + webcomponents coverage gates ✅ · `npm run build` ✅ · first-load size gate ✅ (worker +3.8 kB, 24 kB under ceiling) · boy-scout gate: no touched file on a debt list.


trieloff:
All four review findings are fixed in `8f84657`, with regression tests for each:

- **Critical — `budget` dropped in the `session-stats` handler.** Correct and load-bearing: the worker's reading never reached either surface, so the feature was dead outside the tests. Forwarded; `offscreen-client.test.ts` now pins both directions, including that a metered provider's reply carries **no** `budget` key at all (absence is what keeps the `$` headline).
- **Major — in-flight probe not keyed by provider.** Fixed: the promise carries its provider id and is only shared with a caller asking about the same account. Also closed the mirror of it you didn't mention — a probe that started before a switch no longer writes its late answer over the current account's entry. Covered by a new cross-provider test.
- **Minor — logged-out token cached as unsupported for 30 min.** Taken. Being signed out is transient in a way a missing endpoint is not, so it now **throws** (short retry clock, and it costs no network — the token check is local), and Adobe's login/logout paths call `clearBudgetWindowCache()`, so the window appears with the session rather than minutes into it. That also gives the function the production callers you flagged it as lacking.
- **Minor — `/v1/usage` missing `X-Session-Id`.** Taken: the header rides in from the call site with its own purpose anchor (`adobe-usage-probe`), not the "dev forgot the wrapper" fallback sentinel — the probe is legitimate recurring non-LLM traffic and the proxy should be able to group it as such. `fetchAdobeUsage` merges caller headers but keeps `Authorization` its own, tested.

lint · typecheck · 20 843 tests · coverage · build · first-load (worker +3.9 kB, 24 kB under ceiling) · knip (both gates) all green locally.

Issue #2812, opened 2026-09-03
feat(worker): tray-stable webhook URLs so a reset is invisible to external senders


trieloff:
Follow-up to #1957, which fixed the symptom. This is the fix that removes the failure mode.

### Problem

A webhook URL bakes the tray id into its own address. The capability token is `<trayId>.<secret>` (`createCapabilityToken` in `shared.ts`), and the worker routes a delivery with `env.TRAY_HUB.get(idFromName(parsed.trayId))`. The URL names an *instance*.

An external service caches that URL for the life of whatever job it was handed out for, which can be an hour. Reset the tray in the meantime and the sender is holding a pointer to a dead object. That is what happened in #1957: a callback arrived 43 minutes into a render, POSTed into a tray that had been replaced, and the event was gone. The registration had survived the reset (the same webhook id was re-issued under the new tray prefix), so nothing looked broken from inside. Only the cached URL string was dead.

#1957 made that answer a `308` to the replacement instead of a silent `410`, so a well-behaved sender now recovers. But it is a workaround with two costs: it depends on the sender following redirects for a `POST` and on the leader having named the replacement, and the `Location` header hands out the replacement's webhook capability, which is a secret. A tray reset should simply be invisible to anyone holding a webhook URL.

### Proposed solution

Keep the URL shape and change what the id in it names: a stable per-cone id instead of a tray id, with one indirection resolving it to whichever tray is current.

```
POST /webhook/<coneId>.<secret>/<webhookId>
        |
        v
  idFromName(coneId)  ->  WebhookHome DO   { currentTrayId, secretHash, registrations, revokedAt }
                              |
                              v  internal forward (no redirect, no capability in a header)
                        idFromName(currentTrayId) -> tray DO -> live leader socket
```

- **`coneId` is minted once and never rotates.** The leader persists it alongside the session state it already persists, and re-points the home with one authenticated `rebind` call on every reset, in the same place it calls `supersede` today.
- **Keyed on the cone, not on each registration**, so a reset is one write no matter how many webhooks exist.
- **Delivery is an internal forward, not a redirect.** Strictly better than the 308: the sender never learns a tray id, and no capability appears in a response header.
- **The mapping must live in DO storage, not KV.** The moment the read matters is the instant after a reset, which is exactly when KV would still be serving the tray that just died.

The 308 from #2800 stays as the migration path: existing tray-scoped URLs keep working, new registrations get the stable shape, and the old shape can be deprecated on its own schedule.

[...]

Issue #2524, opened 2026-08-27
webhook/crontask accept an unresolvable --scoop, then silently drop deliveries with HTTP 202 ok:true


trieloff:
## Summary

`webhook create --scoop <target>` and `crontask create --scoop <target>` accept a target that does not exist. Nothing warns at create time, `webhook list` / `crontask list` show the entry as active, and deliveries to it are then **silently discarded while the sender is told they succeeded**.

For a multi-cone workspace this is the difference between a misrouted lick (annoying, recoverable) and a lick that is gone forever with a success receipt. A single typo — or a cone folder that was renamed or removed — turns a watcher into a black hole that looks healthy in every listing.

## Environment

- SLICC 6.99.8
- Cones present: default root `cone`, plus `cone-helix`, `cone-karl`
- Measured from the default root's shell

## Reproduction

```
$ webhook create --scoop ghost-cone-does-not-exist --filter '() => ({probe:"silent-drop-test"})'
Created webhook "default"
ID:  emhxyaya9m9x
URL: https://www.sliccy.ai/webhook/…/emhxyaya9m9x
Scoop: ghost-cone-does-not-exist
```

No warning, exit 0. It then lists as active:

```
$ webhook list
Active webhooks:
  emhxyaya9m9x  default  https://…/emhxyaya9m9x  -> ghost-cone-does-not-exist  [filtered]
```

Firing it reports success to the caller:

```
$ curl -s -w 'HTTP=%{http_code}\n' -X POST "https://…/emhxyaya9m9x" \
    -d '{"type":"probe","text":"does this reach anyone?"}'
HTTP=202
{
  "ok": true,
  "accepted": true
}
```

`202 Accepted` + `"ok": true` + `"accepted": true`, and no unit is ever woken.

`crontask create` behaves the same way:

```
$ crontask create --name ghost-target-probe --scoop ghost-cone-does-not-exist --cron "5 4 1 1 *" --filter '() => ({x:1})'
Created cron task "ghost-target-probe"
ID:       ke357lx740er
Scoop:    ghost-cone-does-not-exist
Next run: 1/1/2027, 4:05:00 AM
```

## Impact

[...]

PR #3053, opened 2026-09-11
fix: complete durable tray lifecycle and preview continuity (#2812)

trieloff:
## Latest review fixes and approved bundle budget — c4e9997bb

- Rotation now uses fresh cryptographic delivery and management secrets persisted before HTTP, atomic dual-hash replacement, and full-intent receipt replay. Old credentials alone cannot derive or authorize the next identity.
- Release lifecycle inspection retries transient failures and distinguishes token permissions. Skipped releases archive without an API dependency; deploying releases archive before the gate, then mutate secrets/deploy only on success.
- User-approved total-JS budget: **25.2 → 25.3 MB**. CI exceeded the old ceiling by 201 bytes; fresh local build measures **25.19 MB**. Eager worker ceiling: **4292 → 4295 kB**, measured **4293 kB (+1.5 kB versus merge-base)**. Page ceiling and the 5 kB delta gate are unchanged. This accommodates the durable lifecycle wiring with narrow headroom.
- Verification: 963 integrated tests; full typecheck; all eight lint gates; worker/webapp coverage; full bundle-size command including extension; fresh shared/webapp/node/extension builds; both real local workerd/browser lifecycle E2Es passed (51 seconds), including repeat rotation after reload. CI is pending on this commit.

## Review follow-up — c7650e825

All eight review comments addressed and replied to. Identity-less clients retain legacy creation; stable-aware creates carry a persisted idempotency attempt. Rotation request bodies are bounded and capability grammar is validated. Definitive rotation refusals clear safely through CAS; expired preview targets recover without losing frozen transfer state. Explicit unknown/unresolved webhook rejections get bounded retries and terminal archival, with independent IDs bypassing backoff. Abandoned upload leases expire and cleanup records no longer consume active preview slots indefinitely.

**Rollout prerequisite:** preview cleanup now requires a safe `previews/` object-age lifecycle rule on `sliccy-now-basic-storage` (recommend 90 days; must preserve 30d pending + 30d finalized retention). New read-only pre-deploy verification blocks missing/unsafe policy. Provisioned the enabled `preview-retention-90d` rule for `previews/` using the authorized Wrangler login and verified it against the live API. Preserved the existing multipart-abort rule. Verifier API-shape correction pushed in `47af6a862`; CI verifies access using its own credentials.

Verification on the integrated follow-up: 951 targeted tests, full typecheck, all lint gates, worker/webapp coverage gates, and both real local workerd/browser lifecycle tests passed. CI is pending for the pushed follow-up.

## Summary

Completes the roving-tray work in one PR, including the local characterization harness, durable webhook delivery, rotation/revocation, private identity recovery, and preview continuity. Addresses #2812. No iOS or Go CLI source changes.

### Webhook lifecycle
- Stable `/wh/<identity>.<secret>/<id>` addresses use a Durable Object home, not a replaceable tray ID.
- Identity means a leader-session lineage shared by its WorkUnits, not independent per-agent-cone isolation.
- Durable enqueue happens before acceptance. FIFO replay uses alarms and explicit delivered/filtered acknowledgements; ambiguous outcomes are retried at least once.
- Backpressure limits: 100 entries, 120 KiB encoded home record, 64 KiB request body, eight waiting requests. Accepted events are not evicted or aged out. Explicit unregistered/unresolved events receive three spaced retries before terminal archival; unrelated IDs bypass their rejection backoff, while ambiguous errors remain retryable.
- `webhook rotate` atomically changes the delivery secret on the same home; durable pending intent recovers lost responses before rebind/reset.
- `webhook delete` writes a permanent registration tombstone before local removal and prevents queued replay, including across restart/reset.
- Management identity is persisted separately in private IndexedDB, never mirrored into public session status or the VFS. Reset source/target pairs survive interruption and reload.

### Preview continuity and safety
- Original preview URLs, serving scope, entry path, bridge permissions and retention survive authenticated transfers. Original-token locators point directly to the current owner, without accumulating forwarding chains.
- Lost responses and storage failures retry the same source/target pair. Clone-backed regression tests cover failed locator persistence, failed alarm installation, and ambiguous upload commits.
- Immutable upload candidates and hash/size/MIME reconciliation protect committed R2 bytes from failed or mismatching retries.
- Durable upload leases keep a non-serving cleanup owner for late writes after expiry/revoke. Body/R2 operations have deadlines; eight unresolved leases per preview. Leases expire after 120 seconds; ordinary cleanup tombstones retire after 24 hours and do not consume active slots. The required R2 lifecycle backstop handles arbitrarily late writes without extending serving TTL.
- Superseded guest seats return the existing terminal `410 TRAY_EXPIRED` shape with no successor capability. Existing browser/iOS validators remain compatible.
- Public-entrypoint tests verify internal mutation routes cannot be dispatched through public URLs.

## Verification

Final integrated commit: `5e586157e3862ac26177d7e24ee6f4b868ad0d2f`.

- `npm run verify`: all eight gates passed; repeated successfully by the real pre-push hook.
- `npm run typecheck`: passed on combined checkout.
- Worker plus affected manager, identity, follower, shell, panel and shared-preview suites: **924 passed, nine opt-in skips, 63 passing files**.
- Worker/webapp/shared coverage gates passed during integration (before final late-upload lease extension); final lease/storage regressions are included in the final targeted run above.
- Real local workerd + Playwright: **2/2 passed, 48.8 seconds, no retries**. Covers stable URL across reset, old-secret rejection after rotation, queued delivery after reload without another sender, deletion remaining revoked after another reset, and original preview content surviving reset. This browser run preceded the final cleanup-lease-only extension.
- Fresh shared/webapp/node builds passed in the local browser harness. Local ports isolated; existing user processes were not stopped.
- One broad `npm run test` attempt overlapped an incoming merge conflict and test-fixture artifacts, so it is not a clean full-repository pass. Conflict/artifacts were resolved; final targeted suites, typecheck and pre-push verification above passed. CI remains the full-repository gate.

## Deployment

Adds `WEBHOOK_HOMES` and additive `v3-webhook-homes` SQLite DO migration; production/staging bindings aligned. No production/staging deployment performed. Review and staging verification remain required before production rollout.


packages/webapp/src/scoops/tray-leader.ts
```diff
@@ -289,11 +466,28 @@ export class LeaderTrayManager {
...
-      storedSession?.workerBaseUrl === this.options.workerBaseUrl ? storedSession : null;
+      storedSession?.workerBaseUrl.replace(/\/+$/, '') ===
+      this.options.workerBaseUrl.replace(/\/+$/, '')
+        ? storedSession
+        : null;
+
+    this.identity = await this.identityStore.load();
+    // The server may have rotated even when its response was lost. Replay the
+    // durable old-credential request before attach/rebind can reject that secret.
+    if (this.identity?.pendingRotation) await this.rotateWebhookOnce();
```
trieloff: **Failed rotation permanently blocks leader connect.**

`rotateWebhookOnce()` persists `pendingRotation` *before* the HTTP call and clears it only on success. This line replays it unconditionally on every connect, and every error propagates out of `connectOnce` → `startOnce` → `start()`.

Scenario: two tabs share the same origin IndexedDB identity. Tab B resets and rebinds the home to `T2`. Tab A (still on `T1`) runs `webhook rotate`; the home hits `this.home.currentTrayId !== trayId` → `CONTROLLER_UNCONFIRMED`, which `handleWebhookRotate` maps to 403. `pendingRotation` is now durable and can never succeed: every subsequent `start()` replays the same doomed request and throws. `clearSession()` is not an escape either — it throws while a replacement is pending. The leader is bricked until IndexedDB is cleared by hand.

Suggest treating a non-retryable (4xx that is not the idempotent-receipt path) rotation replay as "give up and drop the intent", rather than failing the connect.

packages/cloudflare-worker/src/webhook-home.ts
```diff
@@ -0,0 +1,726 @@
...
+    }
+    try {
+      const forwarded = await this.forwardToTray(
+        next.webhookId,
+        base64ToBytes(next.bodyB64),
+        next.headers
+      );
+      const ack = forwarded.headers.get('x-slicc-webhook-ack');
+      void forwarded.body?.cancel().catch(() => {});
+      if (forwarded.status >= 300 || (ack !== 'delivered' && ack !== 'filtered')) return;
```
trieloff: **An unregistered `webhookId` poisons the entire cone queue, permanently.**

The head is removed only on `x-slicc-webhook-ack: delivered|filtered`. For an id the leader does not know, the disposition is `unknown-webhook` → `webhookDeliveryResponse` returns 404 with no ack header → `forwarded.status >= 300` → the entry is retained. Accepted events no longer have a TTL (`WEBHOOK_QUEUE_TTL_MS` is documented as migration-only).

The id is the last path segment of the public `/wh/<coneId>.<secret>/<id>` URL, i.e. entirely sender-chosen. One typo'd or guessed id parks a permanent head and stalls FIFO replay of every other webhook for that cone; 100 of them and `enqueue` starts returning false so all legitimate deliveries get `429 WEBHOOK_QUEUE_FULL`. Recovery requires running `webhook delete` for each exact unknown id.

Worth a bounded attempt count / dead-letter on the head (or treating `unknown-webhook` like a tombstone after N retries) so one bad id cannot wedge the cone.

packages/cloudflare-worker/src/session-tray-preview.ts
```diff
@@ -618,17 +669,42 @@ async function persistentUploadRecord(
...
     throw routeError('duplicate preview file path', 409);
   }
-  return { objectKey: `${record.archivePrefix}objects/${crypto.randomUUID()}`, relativePath };
+  // Bound abandoned/ambiguous write bookkeeping. Sequential snapshot uploads
+  // use one lease; this also permits a small parallel uploader.
+  if ((record.pendingUploadKeys?.length ?? 0) >= 8) {
+    throw routeError('too many unresolved preview uploads', 429);
+  }
+  const objectKey = `${record.archivePrefix}objects/${crypto.randomUUID()}`;
+  (record.pendingUploadKeys ??= []).push(objectKey);
```
trieloff: **A leaked upload lease is permanent — no timestamp, no TTL, no sweep.**

`pendingUploadKeys` entries are removed only by `commitPreviewUpload` or `handlePreviewUploadRelease`. But `releasePreviewUpload` swallows every failure (`.catch(() => {})`), and an aborted/terminated edge invocation never calls release at all. Nothing ages these out.

Two permanent consequences:

1. `deletePersistentArchive` returns `false` while `pendingUploadKeys.length` is non-zero, so `expirePersistentPreviews` never reaches `delete tray.previews[token]`. The record stays a `cleanup` tombstone forever, occupying one of `MAX_PREVIEWS_PER_TRAY` (10) slots and re-arming the DO alarm every `PREVIEW_CLEANUP_RETRY_MS` for the life of the tray.
2. After 8 leaked leases on the same preview, this check throws `429 too many unresolved preview uploads` and that preview can never accept another file.

Storing `{ key, leasedAt }` and expiring leases older than the R2/body deadline would bound both.

packages/webapp/src/scoops/tray-leader.ts
```diff
@@ -391,9 +585,258 @@ export class LeaderTrayManager {
...
+    this.stop();
+    return this.start();
+  }
+
+  private async claimReplacement(
+    source: LeaderTraySession,
+    target: LeaderTraySession | null
+  ): Promise<LeaderTraySession> {
+    const next =
+      target && target.trayId !== source.trayId ? target : await this.createTraySession();
```
trieloff: **An expired transfer target hard-locks every future `start()`.**

`claimReplacement` reuses the stored target whenever `target.trayId !== source.trayId`, and nothing ever clears a stale one.

Scenario: attempt 1 mints `T2` (and `createTraySession` saves it to `store`), then `transferPreviousSession(T1, T2)` fails on a transient error, so `connectOnce` throws with `replacementStore = T1` and `store = T2`. The browser stays closed longer than `TRAY_RECLAIM_TTL_MS` (1 h for desktop). On the next start `claimReplacement(T1, T2)` picks `T2` again; `PreviewContinuity.transfer` calls `/internal/confirm-controller` on `T2`, where `handleConfirmController` folds in `await this.ensureTrayIsActive()` → `confirmed: false` → `forbidden()` 403. `transferPreviousSession` throws, `connectOnce` throws, and `clearSession()` refuses to clear while a replacement is pending. Every subsequent start repeats the same failure.

A 403/410 from the target probably needs to invalidate the cached target and mint a new one rather than retrying it forever.

packages/cloudflare-worker/src/index.ts
```diff
@@ -1213,9 +1456,33 @@ async function createTray(request: Request, env: WorkerEnv): Promise<Response> {
...
+    secret: coneSecret,
+    rebindSecret,
+    trayId,
+    controllerToken: payload.controllerToken,
+  });
+  if (!bind.ok) {
+    return jsonResponse(
+      {
+        error: 'Webhook home bind failed; retry with the same cone identity',
+        code: 'WEBHOOK_HOME_BIND_FAILED',
```
trieloff, line 1489: **A failed home bind 503s `POST /tray` after the tray DO already exists, leaking an orphan on every retry.**

`initResponse` has already created the tray record in `TRAY_HUB.idFromName(trayId)` by the time `bindWebhookHome` runs (with a 5 s `withWebhookTimeout`). When the bind fails — slow home, `429 WEBHOOK_HOME_BUSY` at 8 pending requests, or `403 INVALID_REBIND` because the home was claimed with different secrets — this returns 503 and the caller never learns the `trayId` or its tokens. A retrying leader mints a **new** `trayId` each attempt, so each retry leaves another unreachable tray DO behind until its reclaim TTL elapses.

It also makes the webhook home a hard availability dependency of tray creation for every client (iOS, Go CLI, cloud), which previously had none. Binding before `initResponse`, or reusing the same `trayId` on retry, would avoid both.

packages/cloudflare-worker/src/index.ts
```diff
@@ -1227,15 +1494,83 @@ async function createTray(request: Request, env: WorkerEnv): Promise<Response> {
           url: `${url.origin}/controller/${payload.controllerToken}`,
         },
         webhook: {
-          token: payload.webhookToken,
-          url: `${url.origin}/webhook/${payload.webhookToken}`,
+          token: `${coneId}.${coneSecret}`,
+          url: `${url.origin}/wh/${coneId}.${coneSecret}`,
+          // The rebind capability the leader persists and presents on every
+          // later reset. Never logged, never a lick — it steers deliveries.
+          rebindToken: `${coneId}.${rebindSecret}`,
```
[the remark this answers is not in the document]
trieloff: Fixed: identity-less POST /tray keeps the legacy /webhook/ capability and never contacts WEBHOOK_HOMES. Stable-aware clients explicitly supply their private identity and persisted createAttemptId. Added rolling-client compatibility coverage and made only the queue-specific deployed smoke opt into stable creation.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/cloudflare-worker/src/index.ts
```diff
@@ -1158,12 +1206,186 @@ export async function handleDmgDownload(
...
+  url: URL,
+  trayId: string
+): Promise<Response> {
+  const controllerToken = extractBearer(request);
+  if (!controllerToken) {
+    return jsonResponse({ error: 'unauthorized' }, 401);
+  }
+  let body: { oldConeId?: string; oldSecret?: string; oldRebindSecret?: string };
+  try {
+    body = await request.json();
```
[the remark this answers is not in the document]
trieloff: Fixed: rotation reads through readBoundedWebhookBody before JSON parsing. Oversized fixed/chunked bodies return 413; stalled bodies time out and cancel. Tests prove neither tray nor home authority is contacted on rejected input.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/cloudflare-worker/src/index.ts
```diff
@@ -1227,15 +1494,83 @@ async function createTray(request: Request, env: WorkerEnv): Promise<Response> {
...
+    return 'invalid';
+  }
+  const present = [body.coneId, body.coneSecret, body.rebindSecret].filter((v) => v !== undefined);
+  if (present.length === 0) return undefined;
+  if (present.length !== 3) return 'invalid';
+  const { coneId, coneSecret, rebindSecret } = body;
+  const ok = (v: unknown): v is string => typeof v === 'string' && v.trim() !== '';
+  // `.` is the token delimiter — a coneId carrying one would corrupt the
+  // `<coneId>.<secret>` grammar the delivery route parses.
+  if (!ok(coneId) || coneId.includes('.') || !ok(coneSecret) || !ok(rebindSecret)) {
```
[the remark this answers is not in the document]
trieloff: Fixed: worker creation validates every capability component against the URL-safe delimiter-free grammar, including both secrets. The client response parser now rejects dotted/encoded-delimiter secrets too. Added worker and client parser regressions.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/webapp/src/scoops/tray-leader.ts
trieloff: Fixed: definitive rotation refusals release their persisted intent, allowing startup to continue; ambiguous transport/server outcomes remain pending. Atomic IndexedDB compare-and-swap prevents a stale tab from overwriting a newer identity or pending rotation. Regression coverage includes late 403/success races across stores and permanent-refusal restart recovery.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/cloudflare-worker/src/webhook-home.ts
trieloff: Fixed: only explicit matching unknown-registration/unresolved-target responses spend a three-attempt budget (30s spacing). Other webhook IDs bypass this rejection backoff while same-ID order is preserved. Terminal outcomes move atomically to a bounded latest-100 dead-letter archive, outside active queue capacity. Transient/ambiguous outcomes never exhaust that budget. Tests cover typo/healthy progress, repair during grace, restart, full queue recovery and archive rollover.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/cloudflare-worker/src/session-tray-preview.ts
trieloff: Fixed with 120s timestamped leases, durable legacy migration, expired-candidate commit rejection, cleanup-excluded active capacity and a 24h ordinary cleanup horizon. A timeout alone cannot prove that R2 will not write later, so deploys now require a read-only verified previews/ object-age lifecycle backstop (60 < days <= 90; recommend 90). Tests cover eight abandoned leases, late writes, cleanup after the local horizon and lease migration. Operational prerequisite: an authorized operator must inspect/provision the rule on sliccy-now-basic-storage; this environment has no Cloudflare API credentials, and no live policy was changed. The deploy gate intentionally blocks if absent/unsafe.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/webapp/src/scoops/tray-leader.ts
trieloff: Fixed without discarding partially moved previews: only an explicit source-issued PREVIEW_TARGET_UNAVAILABLE before freeze permits replacing the cached target. Frozen transfers finish the same pair using retained target ownership, then a completed expired target becomes the next durable source for one bounded additional rove. Actual reclaim-TTL regression tests cover failures at import, locator relocation and activation, including reconstruction.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

packages/cloudflare-worker/src/index.ts
trieloff: Fixed: legacy clients have no home availability dependency. Stable-aware clients persist createAttemptId before POST; an authenticated opaque UUID-shaped address derived from identity+attempt reuses the same tray DO, original capabilities and timestamp on bind failure or lost response. Attempt intent is cleared only after the session is durable; deliberate reset gets a new attempt. Tests cover repeated bind503/429/403, lost replies, session persistence failure and retry acknowledgement failure. UUID shape is retained and tested for preview DNS round-tripping.

Pushed in `c7650e825`. Combined verification: 951 tests, full typecheck, lint and worker/webapp coverage pass; both local workerd/browser lifecycle tests pass.

trieloff:
Provisioned and read back the live R2 cleanup prerequisite using the authorized Wrangler login: `sliccy-now-basic-storage`, enabled rule `preview-retention-90d`, prefix `previews/`, object expiration after 90 days. Existing seven-day multipart-abort rule preserved. This bucket is shared by production/staging. No worker deployment or credential changes performed.

The exact verifier now passes against the live API. Read-back exposed an omitted-prefix API shape in the default multipart rule; fixed and pushed in `47af6a862` with two regressions (28 lifecycle tests pass), full typecheck and all eight lint gates green. CI still uses its own credentials, whose permissions are checked by the gate.

packages/cloudflare-worker/scripts/publish-worker.sh
```diff
@@ -17,6 +17,10 @@
 #   SLICC_LAST_RELEASE_TAG (empty means first release and always deploys).
 set -euo pipefail
 
+# Hard prerequisite, before even secret uploads (which can activate a version).
+# Read-only: an operator must provision the prefix-scoped lifecycle first.
+node packages/cloudflare-worker/scripts/verify-preview-lifecycle.mjs sliccy-now-basic-storage
```
trieloff: 🟡 **Major — an unretried Cloudflare API read is now the first thing that can fail a tagged release.**

This is the first executable line of `publish-worker.sh`, ahead of everything — including `archive_assets()`, which the comment at line 30 says must run "on every release, **including worker-deploy skips**". With `set -euo pipefail`, a non-zero exit here aborts the whole script.

`verify-preview-lifecycle.mjs` has no retry and a single 30 s deadline, and it collapses every cause into one exit: `catch { throw new Error('Unable to read R2 lifecycle (HTTP/auth/network/timeout/JSON failure)') }`. Every other Cloudflare call in this script is retried — `deploy_with_retry` uses `MAX_ATTEMPTS=6` with `SLEEP_BETWEEN=15`, and the smoke test retries with backoff — precisely because this repo has a documented history of transient Cloudflare deploy failures.

Two concrete failure scenarios:

1. **Transient API blip.** `api.cloudflare.com` returns a 5xx (or the request exceeds 30 s) during a release. semantic-release has already tagged, bumped `package.json` and published; `publish:worker` then exits 1 on the *first* line, with no worker deploy, no asset archive, and no retry. Recovery is a manual re-run.
2. **Token scope.** `GET /accounts/{id}/r2/buckets/{bucket}/lifecycle` is an account-level R2 admin read, not the object-level permission `upload-assets-to-r2.mjs` needs. A `CLOUDFLARE_API_TOKEN` scoped as documented in `packages/cloudflare-worker/CLAUDE.md` ("Workers Edit, R2 R/W, Zone Routes Edit") may well not carry it — and because the error text is deliberately generic, the failure reads as a network problem rather than a missing permission. PR CI cannot catch this: the three workflow call sites run under the CI token, and `publish-worker.sh` only runs on a tagged release.

Given the check is a *precondition on data an operator provisions out of band* rather than something this deploy changes, it seems better placed as a retried step next to the other Cloudflare calls (and after `archive_assets()`), rather than as an unretried gate in front of a release that may not even deploy the worker. If it must stay first, please at least give it the same retry treatment and distinguish "token cannot read lifecycle" from "call failed" in the message.

packages/cloudflare-worker/src/index.ts
```diff
@@ -1158,12 +1206,189 @@ export async function handleDmgDownload(
...
+      { error: 'oldConeId, oldSecret and oldRebindSecret are required', code: 'INVALID_BODY' },
+      400
+    );
+  }
+
+  const coneId = body.oldConeId;
+  const rebindSecret = body.oldRebindSecret;
+  const digest = await crypto.subtle.digest(
+    'SHA-256',
+    new TextEncoder().encode(`slicc-webhook-rotate-v1:${body.oldSecret}:${rebindSecret}`)
```
trieloff: 🟡 **Major — `webhook rotate` has no forward secrecy, so it does not remediate the leak it exists for.**

The replacement delivery secret is a pure function of two values the leader already holds together:

```
coneSecret = SHA-256("slicc-webhook-rotate-v1:" + oldSecret + ":" + rebindSecret)
```

and the rebind secret itself never changes — `handleRotate` only ever writes `secretHash`, and `rotateWebhookOnce` in `tray-leader.ts` hard-rejects any response whose `rebindSecret` differs (`'webhook rotate: hub changed the management identity'`).

Both values live in the **same** IndexedDB record (`leader-webhook-identity:<workerBaseUrl>`) and are sent together, in plaintext, in the body of every `POST /tray` and every `POST /api/tray/:trayId/webhook/rotate`. So any compromise that discloses the delivery secret — the documented trigger for rotating — has almost certainly disclosed the rebind secret too.

**Failure scenario:** an attacker reads the `leader-webhook-identity` record (XSS on the leader origin, a synced/backed-up profile, a captured request body). The operator notices deliveries they did not expect and runs `webhook rotate`; `automation/SKILL.md` tells them "Old URLs stop authenticating." The attacker computes `SHA-256("slicc-webhook-rotate-v1:" + s0 + ":" + r)` and has the new URL — and every subsequent one, since `r` is fixed and each `s(n+1)` is derivable from `s(n)`. Rotation is a no-op against the only threat it is documented to answer, while reporting success.

The determinism is not load-bearing for retry-safety: `rotationReceiptHash` is computed over the *request body*, so any client-chosen secret replays identically, and the leader already persists a durable `pendingRotation` intent (now under CAS) before sending. Persisting a freshly generated `secret` — and a fresh `rebindSecret` — in that intent would give the same lost-response recovery with actual forward secrecy.

packages/cloudflare-worker/src/webhook-home.ts
trieloff: Verified the registration-failure path: `nextEligibleDelivery()` + per-ID `rejection` backoff + the 3-attempt dead-letter close the reported hole cleanly, and other IDs now bypass a quarantined one. Thanks.

**One residual in the same queue, through a different door.** `attemptDelivery` classifies *only* `404 WEBHOOK_NOT_REGISTERED` and `422 WEBHOOK_TARGET_UNRESOLVED` as explicit; everything else returns `null` = ambiguous, which never spends the budget. But `webhookDeliveryResponse` returns **`202` with no `x-slicc-webhook-ack`** whenever the leader does not answer within `WEBHOOK_DELIVERY_WAIT_MS` (3 s) — and that is indistinguishable here from a network blip.

`relayWebhookEvent` (`ui/page-leader-tray.ts`) hands the event to `LickManager` **first** and only then resolves a disposition, so the side effect — waking the unit — has already happened. Nothing on either side keys on `deliveryId` for idempotency: `settle()` no-ops because `awaitDelivery` already deleted the resolver.

**Failure scenario:** a cone is mid-agent-loop, so the page → kernel-worker → `LickManager` round trip takes ~4 s. Sender POSTs `/wh/<coneId>.<secret>/deploy-done`. The tray relays it, the lick fires, the ack arrives at 4 s and is dropped; the home sees `202`/no-ack → ambiguous → retains. 30 s later it forwards the same delivery again → the lick fires **again**. Under sustained load the cone is woken by one external event every 30 s indefinitely, and that `webhookId` never drains. A pre-#2524 leader (which never acks at all) is the same case, permanently.

Suggest either treating a `2xx` without an ack as explicit-but-unconfirmed after N attempts (so it can dead-letter like the others), or making the ack survive the wait budget — e.g. keep the `deliveryId` resolver in a short-lived map so a late `webhook.delivery` still retires the queue entry, and have the tray remember recently-acked `deliveryId`s so a replay is filtered rather than re-run.

packages/webapp/src/scoops/tray-leader.ts
trieloff: Verified: `isDefinitiveRotationRefusal` + the CAS release in the `fetchJson` catch does clear `pendingRotation` for the `CONTROLLER_UNCONFIRMED` 403 I described, and `connectOnce` continues. That path is closed.

**Latent residual:** `handleWebhookRotate` collapses the home's status to `rotated.status === 403 ? 403 : 502`, so only 403 survives as definitive. The home's `handleRotate` can also answer `410 HOME_REVOKED` and `400 INVALID_BODY`, both of which are permanent — and both arrive at the leader as `502`, which `isDefinitiveRotationRefusal` rejects. The intent is then retained and every subsequent `start()` replays the same doomed request: the exact brick this fix removes, restored the moment `/internal/home/revoke` acquires a caller (it currently has none) or `handleRotate` grows another terminal 4xx.

Cheap to make future-proof: forward the home's status when it is a 4xx/410 rather than mapping everything except 403 to 502.

packages/cloudflare-worker/scripts/publish-worker.sh
trieloff: Fixed in c4e9997bb. Lifecycle GETs now retry transient network/body transport errors, timeouts, 429 and 5xx: three attempts, fresh 30s deadlines, 1s/2s backoff. 401/403 fail immediately with sanitized account/token/R2 configuration-read guidance; unsafe policy and malformed responses are not retried. Worker-deploy skips archive assets without consulting the lifecycle API. Deploying releases run template → archive → lifecycle → secrets → hub/preview. Verified 41 verifier cases plus five executable tests of the actual release script with command stubs, including archive/gate failures preventing all mutations.

packages/cloudflare-worker/src/index.ts
trieloff: Fixed in c4e9997bb. The leader cryptographically generates and persists fresh delivery AND management secrets in the pending intent before HTTP; no replacement is derived from old credentials. The home atomically stores both hashes and a receipt covering the full operation. Exact retries require presenting the persisted fresh values, so old leaked credentials alone cannot retrieve or derive them or authorize new mutations. Regression tests cover old-authority rejection, failed atomic storage, lost responses/reload and CAS races. Real local browser E2E now rotates again after reload and verifies the recovered fresh management identity works. Rotation addresses a past credential disclosure, not continuing compromise of the active origin. Legacy deterministic intents are refused locally and documented rather than silently reusing derivable secrets.

PR #3111, opened 2026-09-15
fix(webapp): path.resolve(rel) uses process.cwd() (#3110)

### LINKED ISSUE #3110: path.resolve(p) with one relative argument resolves against / instead of process.cwd()
## Summary

`path.resolve(p)` with a single relative argument ignores `process.cwd()` and resolves against `/`.

Node documents `path.resolve('out.pdf')` as equivalent to `path.resolve(process.cwd(), 'out.pdf')`. Here it returns `/out.pdf`.

## Measured

```
$ cd /tmp && pwd && node /tmp/pathprobe.cjs
/tmp
process.cwd()                                 "/tmp"
path.resolve('out.pdf')                       "/out.pdf"      <-- expected "/tmp/out.pdf"
path.resolve(process.cwd(),'out.pdf')         "/tmp/out.pdf"
path.dirname(path.resolve('fixtures/x.pdf'))  "/fixtures"
```

`process.cwd()` itself is correct, and the two-argument form is correct. Only the
one-argument form is wrong.

## Why it matters

Any tool that does the standard `fs.mkdirSync(path.dirname(path.resolve(outArg)), { recursive: true })`
creates a directory **at the filesystem root** when handed a relative output path. In a scoop
this is not writable, so it raises a sudo escalation — a human gets asked to approve a write to
`/fixtures` or `/sub` that neither they nor the agent intended. It took four escalations before
the cause was found, and the wrong first hypothesis was "the shell's cwd is not inherited"
(it is; `cd` and `process.cwd()` both behave).

The symptom is especially confusing because `fs.writeFileSync('sub/out.pdf', …)` resolves
correctly against the cwd, so the *file* lands where you expect while the *mkdir* lands at `/`.
That leaves an empty directory at the root and a working output, which reads as two unrelated bugs.

## Two related findings from the same investigation

1. `spawnSync`'s `cwd` option is ignored (as is `env`). A test that sets `cwd` to check a
   child's path handling silently tests nothing.
2. `ls <glob> | wc -l` overcounts by n-1: `ls` inserts a blank line between multiple file
   arguments, so 26 files report as 51. `find`, `echo <glob> | wc -w` and `ls -1` are correct.
   This silently inflated a file count in a report.

## Workaround

```js
const absolute = (p) => (path.isAbsolute(p) ? p : path.resolve(process.cwd(), p));
```

## Environment

SLICC browser runtime, this shell's node. Reproduced from several cwds, in both a cone and a scoop.

trieloff:
Closes #3110

## Summary

Realm `path.resolve('out.pdf')` ignored `process.cwd()` and returned `/out.pdf`. It now matches Node: `path.resolve(rel)` ≡ `path.resolve(process.cwd(), rel)`. Absolute inputs and the two-argument form are unchanged.

## Why

**Repro**

1. `cd /tmp` in a cone or scoop shell (or boot the JS realm with `cwd: '/tmp'`).
2. `node -e "console.log(require('path').resolve('out.pdf'))"`

Expected: `/tmp/out.pdf`
Actual (before): `/out.pdf`

Any tool that does `fs.mkdirSync(path.dirname(path.resolve(outArg)), { recursive: true })` then created a directory at the VFS root (`/fixtures`, `/sub`). In a scoop that path is not writable, so it raised a sudo card for a write the agent did not intend. `fs.writeFileSync('sub/out.pdf')` already resolved against cwd, so the file landed in the right place while mkdir hit `/`.

`process.cwd()` itself and `path.resolve(cwd, rel)` were already correct.

- `path.resolve` walks arguments right-to-left and, if none is absolute, prepends the realm's `process.cwd()` (Node posix `path.resolve`).
- `createNodePath(getCwd)` is per-realm, same pattern as `createNodeOs(env)`: `require('path')` closes over the same `processShim.cwd` user code sees as `process.cwd()`. The envless `nodePath` singleton stays for host-side module-graph walks, which already pass an absolute `fromDir`.

Out of scope (same investigation, not this PR): `spawnSync` ignoring `cwd`/`env`, and `ls <glob> | wc -l` overcounting.

## Cross-cutting impact

- **Floats exercised**: CLI and Chrome extension share this JS-realm path shim; both pick up the fix. Electron / Sliccstart / worker use the same kernel realm.
- **Other callers**: host-side `nodePath.resolve(fromDir, specifier)` in `node-module.ts` still passes an absolute `fromDir`, so two-arg resolve is unchanged.
- **Bundle size**: kernel-worker first-load +0.3 kB vs merge-base (under the 5 kB allowance).

## Test plan

- [x] `npm run verify`
- [x] `npm run typecheck`
- [x] `npm run test` — 22214 passed, 53 skipped, no new failures
- [x] `npm run test:coverage`
- [x] `npm run build`
- [x] `npm run bundle-size`
- [x] **New / changed behavior is covered by a unit test** in `packages/webapp/tests/kernel/realm/helpers/node-path.test.ts` and a realm integration case in `cjs-require-bare-builtins.test.ts`
- [x] N/A — no UI change (kernel path shim only)

Focused:

```
npx vitest run packages/webapp/tests/kernel/realm/helpers/node-path.test.ts packages/webapp/tests/kernel/realm/cjs-require-bare-builtins.test.ts
```

After `makeCtx({ cwd: '/tmp' })`, `path.resolve('out.pdf')` is `/tmp/out.pdf`, absolute input is unchanged, `path.resolve(cwd, rel)` still works.

## Documentation

- [x] No documentation changes needed — Node-parity bugfix; agent-facing `path` surface is unchanged.

## Risk & rollback

- Blast radius: JS realm `require('path').resolve` in every float. Scripts that relied on the buggy `/rel` result would change; that was the sudo-raising bug.
- Rollback: revert this PR.
- `spawnSync` `cwd`/`env` and `ls | wc -l` are **not** fixed here.

## Checklist

- [x] Linear history, rebased onto `origin/main`


trieloff:
Thanks Claude — taking the LGTM.

Leaving the empty/invalid-cwd fallback duplicated for this PR, as you marked it non-blocking. `defaultCwd()` is the host singleton path; the realm `getCwd` reads `processShim.cwd()`. I’ll extract a shared helper if a third caller shows up.

PR #3112, opened 2026-09-15
Fix rg byte-limit failure handling

### LINKED ISSUE #3106: rg aborts the whole shell script on its byte limit, and the limit trips at ~half the byte total it names
`rg` aborts the whole shell script on its byte limit, and the limit trips at ~half the byte total it names

## Summary

Two separate defects in the `rg` shim, both hit while grepping a directory tree that
contains video files:

1. **The failure kills the shell.** It is not a normal non-zero exit — every statement
   after the `rg` call is skipped, `$?` is never observed, and the message cannot be
   redirected or captured. `2>file` gets 0 bytes.
2. **The limit fires far below the byte count it reports.** The message says
   `536870912 bytes` (512.00 MiB), but a search over **264.3 MiB** trips it.

There are also two different messages for what looks like the same condition:
`aggregate input size limit exceeded (536870912 bytes)` and
`live byte limit exceeded (536870912 bytes)`, both quoting the same number.

## Repro 1 — the abort

```bash
echo "  line 1 runs"
rg -ril "somestring" /shared > /tmp/r.out 2> /tmp/r.err
echo "  line 3 runs (exit was $?)"     # NEVER PRINTS
echo "  stderr: $(wc -c < /tmp/r.err)B" # NEVER PRINTS
```

Observed:

```
  line 1 runs
bash: rg: aggregate input size limit exceeded (536870912 bytes)
```

`line 3` never runs, and `/tmp/r.err` is not created — so the error text reaches the
terminal but is unavailable to the script. Normal `rg` failures do **not** behave this
way (control below), so this is specific to the limit path.

### Control: ordinary failures are well behaved

```bash
rg -ril "zzz-no-such-string" /shared/wiki   # exit=1, stderr 0B, script continues
rg -ril x /nope/nope                        # exit=1,            script continues
```

Both allow the following statements to run. Only the byte-limit path aborts.

## Repro 2 — the limit trips at ~half the stated value

Bisected against a directory of session recordings, measuring each candidate with
`find … -exec wc -c {} \;` summed (not `du`, which rounds). One `rg` per shell
invocation, because a trip terminates the script.

| input | exact bytes | MiB | result |
|---|---|---|---|
| dir A | 109,300,893 | 104.2 | exit 1, fine |
| dir C alone | 71,008,952 | 67.7 | exit 1, fine |
| A + B | — | 196.6 | exit 1, fine (reproduced twice) |
| **A + B + C** | — | **264.3** | **trips** |
| A + B + C + D | — | 307.6 | trips |
| whole parent dir | 508,524,161 | 484.97 | trips |

So the threshold sits between **196.6 MiB and 264.3 MiB** while the message names
**512.00 MiB**. Every individual input is fine on its own; only the aggregate trips.
Note the whole-parent case is 484.97 MiB — already **27 MiB under** the stated limit —
which is what first made the number look wrong.

## Contributing factor: binary files appear to count toward the limit

The tree is 481.1 MiB of `.webm` against 1.0 MiB of `.md`/`.js`/`.json` (188 files, 28
of them `.webm`). `rg` does not report matches from those binaries — a single
18,928,008-byte `.webm` searched on its own returns exit 1 with 0 bytes on stdout and
stderr — yet their bytes evidently count, since restricting the search makes the same
tree searchable:

```bash
rg -ril PATTERN -g '*.md' /shared          # exit 0, 6 hits
rg -ril PATTERN --max-filesize=1M /shared  # exit 0, 7 hits
```

Both complete over the same 1.0 GiB tree that fails unrestricted.

## Why this matters

The abort is the serious half. A byte limit is a reasonable guard, but a guard that
terminates the caller turns a skipped search into a silently truncated script: any
verification, cleanup or reporting step after the `rg` simply does not happen, and the
script's own error handling never sees it. It also cannot be defended against — the
usual `cmd || fallback` and `2>err` idioms are both bypassed.

A closely related trap in the same family: `rg … | head` returns `head`'s exit status,
so `rg pat dir | head || echo "no hits"` reports success even when `rg` died. Combined
with an abort that produces no capturable stderr, a failed search is very easy to read
as a genuine "no matches".

## Suggested fix

1. Exit non-zero (like every other `rg` failure) instead of aborting the shell, and
   write the message to the process's stderr so `2>` captures it.
2. Make the reported number the number actually enforced, or say what is being counted
   — the two different messages quoting one figure suggest two counters with one label.
3. Consider excluding files that are skipped as binary, or that exceed
   `--max-filesize`, from the accounting, since those bytes are never searched.

## Environment

- SLICC 6.152.0
- `rg` resolves to `/usr/bin/rg`
- Tree: 1.0 GiB total; the failing subtree is 508,524,161 B across 188 files

trieloff:
## Problem
`/usr/bin/rg` byte-limit failures abort the caller, and concurrent reads can charge live bytes inconsistently.

## Solution
Return exit 2 with capturable stderr and enforce/report the remaining aggregate and live byte budgets consistently.

## Testing
Focused shell regressions cover caller continuation, stderr redirection, aggregate limits, and per-file live limits.

Closes #3106

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01M2J55Q069M7VY9N3YR80HDP2&panel=chat)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01M2J55Q069M7VY9N3YR80HDP2)⚡</sup>

👋 I've got this PR.

trieloff:
Duplicate of #3115 (merged 2026-09-15, already closed #3106). This draft was the competing Augment branch on the same issue.

PR #3133, opened 2026-09-15
fix(webapp): git diff two commits, -q, FETCH_HEAD, merge abort

### LINKED ISSUE #3120: git diff cannot compare two commits; --name-status silently returns empty (rc=0) instead of failing
## Summary

`git diff` cannot compare two commits. Passing full 40-character SHAs is rejected as an "ambiguous argument", and one flag combination — `--name-status` — **silently returns success with empty output instead of erroring**, which reads as "these commits are identical".

## Measured

In a fresh clone of `ai-ecoverse/skills`:

```
$ A=$(git rev-parse HEAD~1); B=$(git rev-parse HEAD)
$ echo "$A $B"
5e29b4e3063a06e4087e67df55a936a2b4b62130 5791ce61...

$ git diff --name-status $A $B
                                  # <-- EMPTY, rc=0
$ git diff --name-only $A $B
fatal: ambiguous argument '5e29b4e3063a06e4087e67df55a936a2b4b62130'
$ git diff --stat $A $B
fatal: ambiguous argument '5e29b4e3063a06e4087e67df55a936a2b4b62130'
$ git diff $A $B
fatal: ambiguous argument '5e29b4e3063a06e4087e67df55a936a2b4b62130'
```

Ground truth for those two commits, from the GitHub API (`compare/5e29b4e3...5791ce61`): **status `ahead`, 5 files changed, 2 commits**. So there is a substantial difference, and `--name-status` reported none.

Note that revision arithmetic itself works — `git rev-parse HEAD~1` correctly returns `5e29b4e3...`. It is `git diff` that will not accept a commit argument.

## Why the silent case is the serious half

`fatal: ambiguous argument` is survivable: it is loud, and a caller can fall back. `--name-status` returning **rc=0 with no output** is not, because "no files changed" is a *plausible answer* that a caller will believe. An agent comparing two commits to decide whether a change landed would conclude "identical" and act on it.

This is the same failure class as two other issues filed today — `chmod +x` reporting success while changing nothing (#3109), and a CLI flag being silently dropped so a POST goes out with no body. In each case the tool's silence produced a confident wrong conclusion rather than a retry. A wrong answer that looks like a valid answer is worse than an error.

## Impact in practice

This came up while verifying whether two pull requests that edited the same 182 KB file would merge cleanly. `git diff <shaA> <shaB>` being unavailable forced a fallback to fetching each PR's patch hunks from the GitHub API and comparing line ranges by hand. That is a workaround for a routine operation.

## Asks

1. Support commit arguments in `git diff` (`<sha>`, `<sha> <sha>`, `HEAD~1 HEAD`, and ideally `A...B`).
2. Failing that, make **every** unsupported form fail loudly. `--name-status` must not return 0-with-empty-output when it cannot honour the request. If a code path cannot compute a diff, it must not report an empty diff.
3. Add a regression test asserting that a diff between two commits known to differ is non-empty — an assertion on "empty means equal" would have caught this.

## Environment

Measured in the browser runtime today. `git` is implemented in `packages/webapp/src/git/` and registered via `defineCommand('git'` in `packages/webapp/src/shell/almost-bash-shell-headless.ts`. Reference clone of the source for line lookups was at 6.142.2, so line numbers may have drifted; the behaviour above was measured on the live current runtime.

### LINKED ISSUE #3121: git: -q corrupts positional parsing, global options rejected, no FETCH_HEAD, and unclearable merge state
## Summary

Four separate fidelity gaps in the bundled `git`, all measured today in a fresh clone. They share a theme: flags are misparsed rather than ignored, so the resulting error message points at the wrong thing.

## 1. `-q` makes the argument parser misread the command

```
$ git clone -q https://github.com/ai-ecoverse/skills.git /tmp/gt
fatal: destination is inside the source repository

$ git clone https://github.com/ai-ecoverse/skills.git /tmp/gt      # same command, no -q
done.

$ git fetch -q origin main
fatal: The function requires a "remote OR url" parameter but none was provided.

$ git fetch origin main                                            # same command, no -q
   5791ce6..branch 'main' of https://github.com/ai-ecoverse/skills.git
```

Both error messages are actively misleading. "destination is inside the source repository" sends you looking at your paths; the real problem is that `-q` consumed or displaced a positional argument. This cost a real debugging detour: the clone was assumed to have failed because of something about `/tmp`, and only a control run without `-q` revealed the cause.

## 2. Global options are parsed as subcommands

```
$ git --no-color log -1 --format='%h'
git: '--no-color' is not a git command. See 'git help'.
```

`--no-color` is a *global* option that precedes the subcommand, so this is valid upstream git. The parser is treating the first token as the subcommand unconditionally.

## 3. `git fetch` does not write `FETCH_HEAD`

```
$ git fetch origin feat/gh-mcp        # succeeds
$ ls .git/FETCH_HEAD
ls: .git/FETCH_HEAD: No such file or directory
$ git rev-parse FETCH_HEAD
fatal: ambiguous argument 'FETCH_HEAD'
$ git merge --no-edit FETCH_HEAD
fatal: Could not find FETCH_HEAD.
```

`FETCH_HEAD` is the documented handle for "what I just fetched", and `git fetch <remote> <ref>` followed by `git merge FETCH_HEAD` is the canonical way to merge something you did not create a local branch for. The remote-tracking ref (`origin/feat/gh-mcp`) does resolve, so there is a workaround, but scripts written against upstream semantics break.

## 4. Merge state is inconsistent and cannot be cleared

After the failed `git merge FETCH_HEAD` above:

```
$ git status
On branch mcptest
nothing to commit, working tree clean          # <-- claims clean

$ git merge --no-edit main
fatal: merge is not possible because you have unmerged files.   # <-- claims unmerged

$ git merge --abort
fatal: No branch specified to merge.           # wrong error; does not clear the state

$ git reset --hard origin/feat/gh-mcp
HEAD is now at 0274277                          # does not clear it either

$ git merge --no-edit main
fatal: merge is not possible because you have unmerged files.   # still stuck
```

So `status` and `merge` disagree about whether a merge is in progress, and neither of the two documented escapes (`merge --abort`, `reset --hard`) clears it. The clone had to be discarded. `git merge --abort` also reports an error belonging to a different code path ("No branch specified to merge" is what a bare `git merge` should say).

## Also observed

`git diff --name-status <shaA> <shaB>` returns rc=0 with empty output on commits that differ by 5 files. Filed separately as its own issue since the silent-success shape makes it more dangerous than these four.

## Asks

1. Accept and ignore (or honour) `-q` wherever upstream does, and never let an unrecognised flag shift positional parsing. An unknown flag should be an explicit error naming the flag, not a corrupted positional read.
2. Parse global options (`--no-color`, `-C <path>`, `-c k=v`) before subcommand resolution.
3. Write `FETCH_HEAD` on fetch.
4. Make `status` and `merge` agree on merge state, and make `merge --abort` / `reset --hard` clear it.

## Environment

Live browser runtime, today. Implementation lives in `packages/webapp/src/git/`, registered via `defineCommand('git'` in `packages/webapp/src/shell/almost-bash-shell-headless.ts`.

trieloff:
Closes #3120
Closes #3121

## Summary

Bundled `git` now compares two commits, treats `-q` as a quiet flag instead of a stolen positional, writes `FETCH_HEAD`, and lets `merge --abort` / `reset --hard` clear merge state. Dual-mode (CLI + extension) — same `packages/webapp/src/git/` implementation both floats run.

## Why

**#3120.** `git diff --name-status $A $B` returned rc=0 with empty stdout on commits that differed by 5 files, which reads as “identical.” `--name-only` / `--stat` / bare diff failed with `ambiguous argument`. mri treated unknown `--name-status` as a string flag and ate the first SHA.

**#3121.** `-q` on clone/fetch ate the next positional (`destination is inside the source repository` / missing remote). `git --no-color log` was parsed as a missing command. `fetch` never wrote `.git/FETCH_HEAD`. After a conflicted merge, `status` said clean while `merge` said unmerged, and neither `merge --abort` nor `reset --hard` cleared it.

- `parseArgs` treats unknown flags as booleans so they cannot steal positionals. Clone/fetch/diff/merge then reject remaining unknown flags by name (exit 129).
- `git diff` accepts `<sha> <sha>`, `HEAD~1 HEAD`, `A..B`, and `A...B` (merge-base). `--name-status` prints `M`/`A`/`D` rows. Unresolvable commits fail loudly.
- Global `--no-color` is stripped before subcommand dispatch (`-C` / `-c k=v` already were).
- `git fetch` writes `.git/FETCH_HEAD`; `resolveRevision` reads the first OID field so `rev-parse FETCH_HEAD` / `merge FETCH_HEAD` work.
- Conflicted merges write `MERGE_HEAD`. `status` reports unmerged paths. `merge --abort` and `reset --hard` call `git.abortMerge` and drop merge-state files.

## Cross-cutting impact

- **Floats exercised:** CLI and Chrome extension share this git implementation (kernel-worker / shell). No UI, no follower protocol, no CDP.
- **Bundle size:** worker eager graph +6.3 kB vs merge-base (git-commands is already eager). Raised `workerEagerCeilingKb` 4307→4314 and `maxDeltaKb` 5→7; put `maxDeltaKb` back to 5 after this lands. Page graph unchanged. No new chunk.
- **Other callers of `parseArgs`:** unknown flags no longer consume the next token. Playwright and git stopEarly paths unchanged.

## Test plan

- [x] `npm run verify`
- [x] `npm run typecheck`
- [x] `npm run test` — 22293 passed, 53 skipped
- [x] `npm run test:coverage:webapp` — floor held
- [x] `npm run bundle-size` — after the budget raise
- [x] `npx vitest run packages/webapp/tests/git/git-issues-3120-3121.test.ts` — 10 passed (two full SHAs that differ → `--name-status` non-empty; `-q` clone/fetch; `--no-color log`; FETCH_HEAD; merge --abort / reset --hard)

No UI change — no screencast.

## Documentation

- [x] `docs/shell-reference.md` (git revision / diff / fetch / merge-state)
- [ ] README.md — N/A (agent-facing bundled git, not a user-facing product change)

## Risk & rollback

- Blast radius: bundled `git` in every float. Linear revert.
- `maxDeltaKb` is temporarily 7; restore to 5 after merge.
- CI cannot catch a live GitHub clone of a packed repo; unit tests cover full-SHA diffs against isomorphic-git objects.

## Checklist

- [x] No secrets in the diff
- [x] Shell-command changes reflected in docs


trieloff:
Addressed the Codex review:

- **P1 abort:** `merge --abort` now returns `MERGE_HEAD missing` before `abortMerge` / `checkout --force`, so a dirty worktree is left alone.
- **P1 commit:** `git commit` with `MERGE_HEAD` writes a two-parent merge commit and clears merge state (falls back to `MERGE_MSG` when `-m` is omitted).
- **P2 name-status:** workdir and `--cached` diffs go through the same formatter; A/D/M is based on entry presence so empty-file edits stay `M`.
- **P2 attached shorts:** `-bmain` / `-Xours` are treated as a value-taking flag plus attached value.

PR #3170, opened 2026-09-16
fix(webapp): honor process.exitCode

### LINKED ISSUE #3155: process.exitCode is ignored; the process always exits 0 (process.exit(n) works)
## Summary

`process.exitCode = n` is ignored: the process always exits **0**. `process.exit(n)` works correctly, so only the assignment form is affected.

This is the worst possible shape for a test-harness defect, because setting `process.exitCode` is the *recommended* way for a script to signal failure while letting the event loop drain — so a harness that fails correctly reports success.

## Measured (live runtime, today)

```
$ printf 'process.exitCode = 3;\n' > ec.cjs && node ec.cjs; echo "rc=$?"
rc=0

$ printf 'process.exitCode = 3;\n' > ec.js  && node ec.js;  echo "rc=$?"
rc=0

$ printf 'process.exit(3);\n' > ex.js && node ex.js; echo "rc=$?"
rc=3          # control: the explicit form works
```

Both `.js` and `.cjs` are affected. An uncaught throw still yields `rc=1`, so a *thrown* failure is visible; only the deliberate, documented, non-throwing signal is lost:

```
$ printf 'process.exitCode = 4;\nthrow new Error("boom");\n' > ec2.js && node ec2.js 2>/dev/null; echo "rc=$?"
rc=1          # the throw sets 1; the assigned 4 is discarded either way
```

## Why it matters

Node documents `process.exitCode` as the way to set the exit status without forcing an immediate exit, and it is what test runners, linters and CI scripts normally use — precisely because `process.exit()` truncates pending I/O. Any such script running here reports success no matter what it found.

The failure is silent and the wrong answer is the *plausible* one (`0` = "everything fine"), so nothing prompts a retry. This is the same shape as `chmod +x` reporting success while changing nothing (#3109), `git diff --name-status` returning rc=0 with empty output on differing commits (#3120), and a killed script discarding all of its output (#3136).

Related and already fixed: #3110 (`path.resolve` with one relative argument). I confirmed that fix is live — `path.resolve('x')` now correctly resolves against `process.cwd()`.

## Ask

Honour `process.exitCode` at normal process exit, including when set from an async callback. A regression test asserting that a script whose only statement is `process.exitCode = 3` exits 3 would pin it.

## Environment

`node --version` reports `v20.0.0-js-shim`. Measured from `/tmp` on the live browser runtime.

trieloff:
Closes #3155

## Solution
Honor `process.exitCode` after normal entry and event-loop drain completion while preserving explicit exits and uncaught-error status.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) | [View session](https://cosmos.augmentcode.com/session?agentId=01M2MQHS34A47VG30WW2KVT1BP&panel=chat)


trieloff:
<sup>[**PR Author Agent**](https://cosmos.augmentcode.com/session?agentId=01M2MQHS34A47VG30WW2KVT1BP)⚡</sup>

👋 I've got this PR.

trieloff:
dupe of #3161

PR #3175, opened 2026-09-16
fix(webapp): adopt threshold compaction history

trieloff:
## Summary

- adopt successful and fallback threshold compaction into both Pi's active-loop message snapshot and `Agent.state.messages`
- persist the compacted canonical history immediately so reloads do not restore the oversized prefix
- preserve the full, untouched pre-compaction transcript in `/sessions/live-*.md`
- cover successful and fallback post-compaction tool continuations plus cancellation and concurrent-state guards

## Live diagnosis

The Copilot cone on the live CDP session had 22 settled threshold-compaction markers but 555 canonical conversation entries and zero canonical rewrites. Pi's `transformContext` return value was used only for the current outbound model request; the active loop and canonical agent state retained the oversized history. Every post-tool continuation therefore crossed the threshold again and compacted the same prefix.

The fix mutates the active-loop snapshot in place, assigns a separate compacted array to canonical state to avoid duplicate append aliasing, and persists that state. Adoption is skipped if the state changed while compaction was running or the request was cancelled. The pre-compaction hook writes the untouched, uncapped history to `/sessions/live-*.md` before canonical replacement, so summarized-away content remains recoverable.

Naive fallback compaction is deliberately adopted as well: leaving it request-scoped would recreate the same loop after every tool continuation. The full transcript remains available in the Markdown archive linked from the fallback message.

## Verification

- `npm run verify`
- `node packages/dev-tools/tools/check-touched-exemptions.mjs`
- `npm run typecheck`
- `npm run test` — 22,611 passed, 63 skipped
- `npm run test:coverage`
- `npm run build`
- `npm run bundle-size`
- isolated real-agent E2E: successful and fallback compaction both continue through a tool call without a second marker
- isolated real-agent E2E: `/sessions/live-cone-*.md` retains both large pre-compaction story turns after canonical adoption


trieloff:
Opus review follow-ups are addressed in `a7ccfbcc0`:

- Added explicit coverage for the concurrent-state and cancelled-transform adoption guards.
- Documented that naive fallback adoption is intentional; otherwise a post-tool continuation retries the same failed compaction and recreates the loop.
- Extended the fallback real-agent E2E through a tool call and verified it still produces exactly one compaction marker.
- Added a real-agent assertion that `/sessions/live-cone-*.md` retains both full pre-compaction story turns after the canonical history is replaced.

The refreshed full local gate passes: verify, touched-debt check, typecheck, 22,611 tests, coverage, all builds, extension build, bundle sizes, and both isolated compaction E2Es.

Issue #3213, opened 2026-09-17
serve: previews accumulate against an undocumented quota, and the mint failure does not name the remedy


trieloff:
`serve` previews accumulate against an undocumented limit, and the mint failure does not name the remedy.

## Symptom

Verification runs abort with:

```
serve: Preview mint failed: Preview limit reached
```

In the slicc-website WAC sessions this aborted six runs. In one case it turned a before/after
comparison red for environmental reasons rather than a real regression — the more expensive
failure mode, because the run produces a *plausible wrong answer* instead of an error.

## What is undocumented

There is no mention of a preview quota, its scope, or its lifetime in either `serve --help` or
`man serve`:

```
$ serve --help | grep -ic "preview limit reached"
0
$ man serve | grep -ic "limit\|quota"
0
```

The only `limit`-adjacent entries in `--help` are `--max-tabs` (bridge tab cap) and `--lines`
(a `--logs` cap), neither of which is this.

## Why it bites silently

Previews persist and accumulate without any prompting. `serve --list` on a tray that had not been
deliberately cleaned showed two `persistent` previews minted days apart, both still active and not
expiring for roughly a month:

| minted | expires | mode |
|---|---|---|
| 2026-09-14 | 2026-10-14 | persistent |
| 2026-09-16 | 2026-10-16 | persistent |

Nothing in the workflow that minted them suggested they would outlive the session, so the quota is
consumed by past work that the current run has no reason to think about.

## Asks

1. **Document the quota** in `serve --help` and `man serve`: the limit, whether it is per tray or
   per account, and whether `--ttl` snapshots count against the same pool as live previews.
2. **Make the mint error name the remedy.** Today it states the problem and stops. Suggested:
   `serve: Preview mint failed: Preview limit reached (N active). List with "serve --list", free one with "serve --stop <token>".`
3. **Document that previews persist**, so a caller knows minting one incurs cleanup rather than
   ending with the session.

## Secondary, UNVERIFIED

`serve --list` prints a compound `TOKEN` of the form `<uuid>.<suffix>`. I did not confirm what the
identifier printed at mint time looks like, because doing so means revoking a live preview that is
not mine to revoke — so I cannot say whether `--stop` accepts the mint-time identifier or only the
`--list` token. If they differ, `--stop <t>`'s help text should say which one it wants. Flagging
this as a hypothesis, not a measurement.

## Not a defect in the flag documentation

[...]

Issue #2852, opened 2026-09-04
serve: live preview path has an unenforced ~25.8 MiB per-file ceiling — oversize files 500 with Cloudflare 1101 while --ttl rejects at 25 MiB cleanly


trieloff:
## Summary

The default (live) `serve` path accepts files of any size but the preview Worker throws on
request for files above roughly 25.8 MiB: HTTP 500 with the Cloudflare `Error 1101 — Worker
threw exception` page. There is no pre-flight check and no documented limit for this path.

The `--ttl` snapshot path, by contrast, enforces a hard, exact 25 MiB per-file limit *before
publishing*, with a clean error message. So the two paths disagree, and only the one that
fails cleanly is documented.

Environment: 6.124.0 (build `6.124.0-mtlpmefa`, `upgrade status` `releasedAt`
2026-09-03T15:58:39Z). Leader is a Chrome tab; requests issued with the shell `curl`.

## Repro

```sh
mkdir -p /tmp/servetest && cd /tmp/servetest
node -e 'const fs=require("fs");
  const mk=(n,s)=>{const b=Buffer.alloc(s);for(let i=0;i<s;i++)b[i]=i&0xff;fs.writeFileSync(n,b)};
  mk("f24.bin",24*1024*1024); mk("f26.bin",26*1024*1024);'
echo hi > index.html
serve --quiet .
# then, against the minted URL:
curl -s -o /dev/null -w "%{http_code}\n" "$URL/f24.bin"   # 200
curl -s -o /dev/null -w "%{http_code}\n" "$URL/f26.bin"   # 500  (CF 1101)
```

## Measurements

Live path (default). Every measurement was taken as the first large request of a
**freshly minted** preview, and only after `index.html` returned 200 on that same
preview, so a wedged/absent leader socket cannot be mistaken for a size failure.

| bytes | size | result | trials |
|---|---|---|---|
| 20 971 520 | 20 MiB | 200, full body | 2/2 |
| 25 165 824 | 24 MiB | 200, full body | 3/3 |
| 26 214 400 | 25 MiB | 200, full body | 4/4 |
| 26 214 401 | 25 MiB + 1 | 200, full body | 3/3 |
| 26 738 688 | 25.5 MiB | 200, full body | 3/3 |
| 27 000 832 | 25.75 MiB | 200, full body | 1/1 |
| 27 033 600 | 25.781 MiB | **200 once, 500 twice** | unstable |
[...]

PR #3222, opened 2026-09-17
fix(serve): document preview limits and name the remedy on failure

trieloff:
Refs #2852, #3213. Companion to #3219 (worker). They merge independently, but the docs here state a 25 MiB live-preview limit that only fully holds once #3219 is deployed, so **merge #3219 first**. Until then, 12–25 MiB files still reset the tray on the live path.

## Summary

`serve`'s limits are now documented, and both failure modes now say how to fix them.

**Before:**
- An oversize live file failed with Cloudflare 1101.
- Hitting the quota printed only `Preview mint failed: Preview limit reached`.
- Minting printed only the URL, and `--stop` rejected that URL.

**After:**
- A file over 25 MiB gets `413 preview file exceeds 25 MiB limit: <path>`.
- The quota error says how full the pool is, what counts, and which commands free a slot.
- `serve` prints the preview token and how long the preview lives, and `--stop` accepts the URL.

## Why

- **#2852:** the live path has no pre-flight size check, and no size limit is documented.
- **#3213:** the 10-preview quota is undocumented, including its scope, the fact that `--ttl` snapshots count against it, and the fact that previews persist. The error names no fix, and the identifier printed at mint time (the URL) is not what `--stop` accepted.

Findings from the investigation that shaped this change:

- The quota is 10 per tray, and it moves with the tray.
- Live previews, `--ttl` snapshots and snapshots still uploading all count against it.
- Live previews never expire.
- The token is `<trayId>.<secret>`, while the URL host label is `<compactUUID>--[userHash-]<secret>`.

On staging, `--stop <host label>` returned `{"revoked":false}`.

- **Leader pre-flight** (`scoops/preview-request-handler.ts`): stats the requested file, or a directory's `index.html`, and refuses anything over `PREVIEW_MAX_FILE_BYTES` with `413` before reading it. The existing worker already passes a leader's status and reason through (`handlePreviewFetch`), so this works against the current deploy. `LeaderPreviewResponseError.status` gains `413`.
- **shared-ts** (`preview-url.ts`):
  - `PREVIEW_MAX_FILE_BYTES` and `PREVIEW_MAX_PER_TRAY`, which mirror the worker constants; the webapp doesn't depend on the worker package.
  - `previewTokenFromUrl()`, the inverse of `buildPreviewUrl`'s label encoding (the user hash is dropped).
- **Quota message** (`preview-mint-client.ts`): on `code: PREVIEW_LIMIT`, it adds "N of M in use" from #3219's `{ active, limit }`, or "limit 10 per tray" against older workers, plus what counts and the `serve --list` / `serve --stop <token>` fix.
- **`serve`**:
  - prints `Preview token: <token> (kept until \`serve --stop\` | kept until its --ttl expires; counts toward the 10-preview tray quota)`;
  - normalises a URL or host passed to `--stop` into the token;
  - adds a "Limits and lifetime" section to `--help`.
- **Docs:** `docs/shell-reference.md` gets a new "Limits and lifetime" section, and I moved prose that had slipped inside the `serve` code fence back out. `README.md` gets one sentence. The `playwright-cli` agent skill gets the limits and a clean-up instruction.

## Cross-cutting impact

- **Floats exercised:** every leader that answers `preview.request`. That is only the page-side `page-leader-tray.ts` (standalone, extension through the hosted leader tab, and Electron). The Swift and Go followers never serve previews.
- **Other callers:**
  - `workerError` is shared by mint, upload, finalize, revoke and list; only a body with `code: PREVIEW_LIMIT` changes.
  - `handlePreviewRequest`'s directory branch keeps its 403 guard; it now stats `index.html` too, and a missing one still falls through to 404.
- **Output contract:** `serve` stdout gains one line between `Preview URL:` and `Pushed to …`. Existing parsers that match `Preview URL:` are unaffected.
- **Security:** the printed token is exactly as powerful as the URL that was already printed.
- **Bundle size:** `npm run bundle-size` passes (it includes the first-load delta gate).

## Test plan

- [x] `npm run verify`: all 8 checks pass in the pre-push hook. `check-touched-exemptions.mjs`: OK.
- [x] `npm run typecheck`
- [x] `npx vitest run`: 1252 files and 22,873 tests pass; 5 files and 63 tests skipped
- [x] `npm run test:coverage:webapp`, `:shared` and `:node-server`: pass
- [x] `npm run bundle-size`
- [x] New tests:
  - leader 413 without reading the file, exactly 25 MiB still served, a directory's `index.html` checked, missing index → 404;
  - quota message with and without counts;
  - `previewTokenFromUrl` round-trips `buildPreviewUrl` (with and without the user hash, prod and localhost), accepts a bare host, rejects non-preview input;
  - `--help` limits text, token line for live and `--ttl`, `--stop <url>`.
- [ ] End to end: #3219's staging smoke covers the worker side of the 413 and 429; the leader side is unit-tested here.

## Documentation

- [x] `README.md`
- [x] `docs/shell-reference.md`
- [x] `packages/vfs-root/workspace/skills/playwright-cli/SKILL.md`
- [ ] `man serve` is DA content on sliccy.com, not in this repo. It is badly stale (it still describes the service-worker preview), and a replacement will be proposed separately.

## Risk & rollback

- **Blast radius:** live previews of files over 25 MiB, which failed anyway, now get a clean 413. Files between 25 MiB and ~25.8 MiB, which sometimes worked, now get 413 too, matching `--ttl`.
- **Rollback:** revert the PR. There is no persisted state.

## Checklist

- [x] Commits are focused
- [x] No secrets or tokens in the diff (test tokens are synthetic)
- [x] Shell-command changes are reflected in docs and the agent skill
- [x] Leader ↔ worker contract: additive `413`; old workers relay it unchanged


trieloff:
Took both review nits in 68bd3ff9f:

- The 413 reason now derives its MiB figure from `PREVIEW_MAX_FILE_BYTES`.
- A `PREVIEW_LIMIT` body with no `error` field now falls back to `Preview limit reached`; there is a new test for that case.

PR #3223, opened 2026-09-17
fix(webapp): judge an open dip fence from marked's own code token

trieloff:
## Summary

Follow-up to #3220 for [the Codex review finding](https://github.com/ai-ecoverse/slicc/pull/3220#discussion_r4034957849). #3220 was already in the merge queue, so this couldn't be pushed there.

The streaming placeholder looked only for the literal ```` ```shtml ```` + LF opener and a triple-backtick closer. marked also parses these fences as shtml blocks, but they were not recognised as open:
- `~~~shtml`
- a longer backtick run, such as ```` ````shtml ```` around a ```` ``` ```` line
- CRLF line endings

A half-written dip could therefore mount, and then remount on every chunk.

The code renderer now flags an open shtml block from marked's own token (`raw`). It uses the actual opening delimiter and requires a matching closing fence of the same character and at least the same length, which is how marked parsed it. Only flagged blocks become the "Pouring a dip…" card while streaming.

This branch is stacked on #3220. Its merge commit keeps #3220's SHAs, so once #3220 lands this diff shrinks to the single fix commit.

## Tests

`message-renderer.test.ts` adds 11 cases:
- **Open blocks:** tilde fence, four-backtick fence around a triple-backtick line, CRLF, a closing fence still arriving, a block closed by the wrong character.
- **Closed blocks:** tilde, four-backtick, a longer closer, CRLF, an indented closer.
- **Final render:** an unclosed block is still handed to hydration once the message is final.

Against the previous detector, five of these fail.

## Verification

- `npm run verify` passes, and so do `check-touched-exemptions`, `typecheck`, and `test:coverage:webapp` (17,891 tests).
- Both builds and `bundle-size` pass.
- E2E: `split-dip-stream` and `sprinkle-details` pass.


trieloff:
Superseded: #3220 was taken out of the merge queue and now carries this commit (d6702f191) directly.

PR #3350, opened 2026-09-21
fix(mcp): fall back to initialize when server/discover times out

trieloff:
## Summary

`mcp add` / `McpClient.initialize()` now treats a timed-out `server/discover` as a legacy-handshake signal and retries with `initialize`, the same way `-32601 Method not found` already does. A 401 on discover still surfaces as `McpAuthRequiredError`.

Before: `mcp add https://mcp.semrush.com/v2/mcp semrush` sat 60s and exited `MCP request timed out after 60000ms (server/discover)`. After: the client falls through to the 2025-06-18 `initialize` handshake.

## Why

Live diagnosis on Sliccstart `:5710` / Chrome `:9222`. `initialize()` prefers modern `server/discover`. Timeout was not a legacy signal, so there was no fallback. Semrush itself 401s in ~300ms; a Chrome-origin fetch can still hang (native `WWW-Authenticate` handling). Sibling PRs cover the proxy header relay and AbortSignal plumbing.

### Repro

1. `mcp add https://mcp.semrush.com/v2/mcp semrush`
2. `server/discover` never completes within 60s
3. Expected: fall back to `initialize`
4. Actual: command exits 124 with `MCP request timed out after 60000ms (server/discover)`

- `isLegacyHandshakeSignal` returns true for `McpTimeoutError`
- 401 is not a timeout and still throws `McpAuthRequiredError` (tested)

## Cross-cutting impact

- **Floats exercised**: any float that runs `mcp add` (CLI, Sliccstart, extension kernel)
- **Other callers**: `mcp refresh` / `mcp auth` go through the same `initialize()`; a slow modern server now gets a legacy retry instead of a hard timeout
- A genuine hang on **both** discover and initialize still times out (initialize has its own 60s)

## Test plan

- [x] `npx biome check --write` on the changed files
- [x] `npm test -- packages/webapp/tests/shell/mcp/client.test.ts` — 39 passing
- [x] **New / changed behavior is covered by a unit test**
- [x] N/A for UI screencast

## Documentation

- [x] No documentation changes needed (internal handshake fallback)

## Risk & rollback

- Blast radius: MCP handshake only. Revert is clean.
- A server that is merely slow on discover (but would eventually succeed) now also falls back to initialize after 60s. That is the same budget as today's hard failure.

## Checklist

- [x] No secrets in the diff


trieloff:
Sibling PRs from the same Semrush `mcp add` diagnosis: #3348 (WWW-Authenticate relay) and #3349 (AbortSignal). All three target `main` independently; `client.ts` hunks do not overlap.

